PR #113, opened 2024-07-25
Validate Templates with external source

eromanova:
Also, consider template type from spec as well.

HMC-108

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ tidy:
 	go mod tidy
 
 .PHONY: test
-test: generate-all fmt vet envtest tidy ## Run tests.
+test: generate-all fmt vet envtest tidy external-crd ## Run tests.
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -105,8 +105,15 @@ vet: ## Run go vet against code.
 tidy:
 	go mod tidy
 
+EXTERNAL_CRDS_DIR := external-crds
+
+.PHONY: external-crds
+external-crds:
+	@mkdir -p $(EXTERNAL_CRDS_DIR)
+	@curl -sL https://github.com/fluxcd/source-controller/releases/download/v1.3.0/source-controller.crds.yaml > $(EXTERNAL_CRDS_DIR)/source-controller.crds.yaml
```
Kshatrix, line 113: We would also need helm-controller CRDs. can we bring back changes from 244280acb394384d6a133a732e14ea8150ce8900 that got removed for some reason?

Kshatrix (changes requested):

Kshatrix (approved):

PR #117, opened 2024-07-27
Disable Flux network policies creation

a13x5:
HMC-109

Flux chart installs network policies by default. It breaks networking connectivity, so they should be disabled.

### DIFF
--- a/templates/hmc/values.yaml
+++ b/templates/hmc/values.yaml
@@ -41,6 +41,8 @@ cert-manager:
       namespace: hmc-system
 flux2:
   enabled: true
+  policies:
+    create: false
   imageAutomationController:
     create: false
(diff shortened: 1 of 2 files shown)


templates/hmc/values.yaml
```diff
@@ -41,6 +41,8 @@ cert-manager:
       namespace: hmc-system
 flux2:
   enabled: true
+  policies:
```
Kshatrix, line 44: please change it in `config/default/hmc_values.yaml` instead

Kshatrix (changes requested):

Kshatrix:
please run `make generate-all` and amend the changes

Kshatrix (approved):

PR #127, opened 2024-08-06
K0smotron bump

Kshatrix:
#136

### DIFF
--- a/templates/k0smotron/Chart.yaml
+++ b/templates/k0smotron/Chart.yaml
@@ -13,12 +13,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 0.1.1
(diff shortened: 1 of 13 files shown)


templates/k0smotron/Chart.yaml
```diff
@@ -13,12 +13,12 @@ type: application
...
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 0.1.1
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "0.1.0"
+appVersion: "1.2.2"
```
eromanova, line 21: Should it be `1.0.2`?

Kshatrix: yep, thanks! let's wait for the fix in 1.0.3 though before updating

eromanova (approved):

PR #167, opened 2024-08-08
Validate Deployment objects

### LINKED ISSUE #141: Validate Deployment objects
Admission controller should validate that the Deployment spec is correct:
- Template from _spec.template_ exists, is valid and has deployment type
- Required providers are enabled and were successfully deployed (check _Management_ status)

eromanova:
Closes #141

### DIFF
--- a/internal/webhook/deployment_webhook.go
+++ b/internal/webhook/deployment_webhook.go
@@ -16,9 +16,11 @@ package webhook // nolint:dupl
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/Mirantis/hmc/api/v1alpha1"
+	"github.com/Mirantis/hmc/internal/utils"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


internal/webhook/deployment_webhook.go
```diff
@@ -48,12 +53,28 @@ var (
...
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
-func (*DeploymentValidator) ValidateUpdate(_ context.Context, _ runtime.Object, _ runtime.Object) (admission.Warnings, error) {
+func (v *DeploymentValidator) ValidateUpdate(ctx context.Context, _ runtime.Object, new runtime.Object) (admission.Warnings, error) {
+	deployment, ok := new.(*v1alpha1.Deployment)
+	if !ok {
+		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected Deployment but got a %T", new))
+	}
+	err := v.isValidDeployment(ctx, deployment)
+	if err != nil {
+		return admission.Warnings{fmt.Sprintf("The Deployment is invalid: %s", err.Error())}, InvalidDeployment
```
squizzi, line 76: `admission.Warnings` accepts a list of warnings to present to the end user, instead of printing a sentinel error with `InvalidDeployment` it might make sense to just make an `InvalidDeploymentError` type with some more useful information for the user for that particular error.

```go
        type InvalidDeploymentError struct {
	        TemplateName string
	        Reason       string
        }

	err := v.isValidDeployment(ctx, deployment)
	if err != nil {
		return admission.Warnings{
			fmt.Sprintf("The Deployment is invalid: %s", InvalidDeploymentError{Reason: err.Error(), TemplateName: deployment.Spec.Template}),
		}, nil
	}
	return nil, nil
```

squizzi: Oh derp, I see what's going on here, you're returning `InvalidDeployment` as error too, I think one or the other should be used here, and I think the better choice is error with the wrapping as this `Deployment` won't work if the template is invalid.  Warnings are displayed to the user in a bulleted list and then Kube allows the `create/update` to pass through.

Also, I know it's draft ... and well take or leave my comments, I'm just trying to look around this repo and provide some thoughts on things I'm familiar with.

eromanova: @squizzi Thank you for the review. I got your idea and changed this part accordingly.

Kshatrix (approved):

PR #179, opened 2024-08-09
Add bastion configuration for AWS templates

### LINKED ISSUE #173: Make bastion host option available for AWS templates
We need to make available bastion host creation through templates `.spec.bastion` option in `AWSCluster`.

It's needed, since there is no other way to make SSH available for instances automatically. We can only create separate SG and attach it separately after cluster creation.

The second reason is that in most cases creating clusters with nodes with public IPs is considered bad practice and generally avoided. in this case bastion host remains the only way to access nodes using SSH

a13x5:
Resolves #173

### DIFF
--- a/docs/aws/cluster-parameters.md
+++ b/docs/aws/cluster-parameters.md
@@ -15,3 +15,38 @@ clusterawsadm ami list
 ```
 
 For details, see [Pre-built Kubernetes AMIs](https://cluster-api-aws.sigs.k8s.io/topics/images/built-amis.html).
+
+## SSH access to cluster nodes
+
+To access the nodes using the SSH protocol, several things should be configured:
+
+- An SSH key added in the region where you want to deploy the cluster
+- Bastion host is enabled
+
+### SSH keys
+
+Only one SSH key is supported and it should be added in AWS prior to creating
+the `Deployment` object. The name of the key should then be placed under `.spec.config.sshKeyName`.
+
+The same SSH key will be used for all machines and a bastion host.
+
+To enable bastion you should add `.spec.config.bastion.enabled` option in the
+`Deployment` object to `true`.
+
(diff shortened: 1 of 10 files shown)


docs/aws/cluster-parameters.md
```diff
@@ -15,3 +15,39 @@ clusterawsadm ami list
...
+
+### SSH keys
+
+Only one SSH key is supported and it should be added in AWS preor creating
+the `Deployment` object name of the key then can be placed under `.spec.config.sshKeyName`
+
+For the bastion host key named **default** must be created, since setting key name
+for bastion host is not possible.
+
+To enable bastion you should add `.spec.config.bastion.enabled` option in the
```
Kshatrix, line 34: We aim to provide default values for all configurable options (it's acceptable if only keys are present). It also would be nice to adjust JSON schemas with the newly added params.

Kshatrix (changes requested):

docs/aws/cluster-parameters.md
```diff
@@ -15,3 +15,39 @@ clusterawsadm ami list
...
+
+- Apropriate SSH keys are added in the region where you want to deploy the cluster
+- Bastion host enabled
+
+### SSH keys
+
+Only one SSH key is supported and it should be added in AWS preor creating
+the `Deployment` object name of the key then can be placed under `.spec.config.sshKeyName`
+
+For the bastion host key named **default** must be created, since setting key name
```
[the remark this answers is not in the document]
a13x5: I overlooked this parameter. Then it makes sense to just use it and drop `sshKeyName` parameters in `AWSMachineTemplate` as I did it the latest push.

docs/aws/cluster-parameters.md
a13x5: I expanded all available bastion options and did some basic tests. values.schema.json updated as well

Kshatrix (approved):

PR #180, opened 2024-08-09
Make root volume size configurable

### LINKED ISSUE #168: Expose configurations options for EBS volumes in the templates
## Description

Right now we don't pass any volume related configuration to machine templates, which results creating instances with a single root volume with size of 8GB (minimal AWS volume size)

It makes cluster pretty much unusable, since overtime this volume get filled up completely, making nodes to experience disk pressure condition.

## Task

We need to make `.spec.rootVolume` and possibly `.spec.nonRootVolumes` configurable through templates (standalone and hosted)

I propose making single value `rootVolumeSize` for root volume and additional list with `Volume` type named `nonRootVolumes` which will passed as is like in `AWSMachine` specification. 

`/var/lib/k0s` should reside on a separate EBS volume by default

a13x5:
Resolves #168


templates/aws-hosted-cp/templates/awsmachinetemplate.yaml
```diff
@@ -22,3 +22,5 @@ spec:
       sshKeyName: {{ .Values.sshKeyName | quote }}
       {{- end }}
       publicIP: {{ .Values.publicIP }}
+      rootVolume:
```
Kshatrix, line 25: could you please update jsonschemas with the new params?

Kshatrix (changes requested):

Kshatrix (approved):

PR #184, opened 2024-08-12
Automatically create Management object only once

### LINKED ISSUE #175: Management object should be automatically created only once
If initially requested, a management object should be automatically created only once. Object removal should not lead to its recreation.

eromanova:
Closes #175

### DIFF
--- a/internal/controller/release_controller.go
+++ b/internal/controller/release_controller.go
@@ -25,6 +25,7 @@ import (
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	"github.com/pkg/errors"
 	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/storage/driver"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -131,6 +132,8 @@ func (p *Poller) ensureManagement(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+
+		hmcConfig := make(chartutil.Values)
 		release, err := actionConfig.Releases.Last("hmc")
 		if err != nil {
 			if !errors.Is(err, driver.ErrReleaseNotFound) {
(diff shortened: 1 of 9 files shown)


templates/hmc/values.yaml
```diff
@@ -6,32 +6,38 @@ admissionWebhook:
   port: 9443
   certDir: "/tmp/k8s-webhook-server/serving-certs/"
 
-controllerManager:
-  manager:
-    args:
-      - --leader-elect
```
Kshatrix, line 12: we should make it true by default and not expose at all I think

internal/helm/values.go
```diff
@@ -0,0 +1,45 @@
...
+	case chartutil.Values:
+		return c, true
+	case map[string]interface{}:
+		return c, true
+	default:
+		return nil, false
+	}
+}
+
+func OverrideValuesWithValues(dst chartutil.Values, src chartutil.Values) {
```
Kshatrix, line 32: can we use `chartutil.CoalesceTables` instead?

internal/controller/release_controller.go
```diff
@@ -131,28 +132,37 @@ func (p *Poller) ensureManagement(ctx context.Context) error {
...
-				values, err := json.Marshal(release.Config)
-				if err != nil {
-					return err
-				}
-				_ = applyDefaultCoreConfiguration(mgmtObj)
-				mgmtObj.Spec.Core = &hmc.DefaultCoreConfiguration
-				mgmtObj.Spec.Core.HMC.Config = &apiextensionsv1.JSON{
-					Raw: values,
-				}
+				helm.OverrideValuesWithValues(hmcConfig, release.Config)
```
Kshatrix, line 144:
```suggestion
				chartutil.CoalesceTables(hmcConfig, release.Config)
```

internal/controller/release_controller.go
```diff
@@ -131,28 +132,37 @@ func (p *Poller) ensureManagement(ctx context.Context) error {
...
 			}
 		}
 
+		// Initially set createManagement:false to automatically create Management object only once
+		helm.OverrideValuesWithValues(hmcConfig, map[string]interface{}{
```
Kshatrix, line 149:
```suggestion
		chartutil.CoalesceTables(hmcConfig, map[string]interface{}{
```

Kshatrix (changes requested):

internal/helm/values.go
eromanova: Sure. Thank you.

Kshatrix (approved):

PR #196, opened 2024-08-15
Track deployment status

### LINKED ISSUE #138: Heartbeat event to track Deployment status

### LINKED ISSUE #231: Make telemetry disableable
Userstory:
```As a Platform Lead I would like to disable the Telemetry sending of 2A as in some specific situations Telemetry cannot legaly be sent or the networking (airgapped) will prevent it anyway```

Acceptance criteria:
- Telemetry is by default enabled
- Telemetry can be disabled through a controller parameter or a `Management` Object config (whatever feels better)

### PR DESCRIPTION
**Pull request by eromanova**
Added a new poller to track the deployments status in the heartbeat event according to the tracking plan every 10 minutes.

### Performance testing

To analyse and decrease memory and CPU usage the following approaches were tested on 1000 test deployments:

Testing discovered that the first approach performed best on the management cluster. It was tested using the metrics server (`kubectl top pod`):

1. List all deployments using the default (cached) manager client
```
NAMESPACE            NAME                                            CPU           Memory
hmc-system           hmc-controller-manager-58bbbc8d7c-wkpfs         6m            67Mi
kube-system          kube-apiserver-hmc-dev-control-plane            82m           562Mi
```
2. List all deployments using uncached manager client
```
NAMESPACE            NAME                                            CPU           Memory
hmc-system           hmc-controller-manager-d8c546f54-667gh          11m           92Mi
kube-system          kube-apiserver-hmc-dev-control-plane            189m          589Mi
```
3. Use pagination to list limited blocks of deployments with cached client. Rejected since pagination does not work using cached manager client https://github.com/kubernetes-sigs/controller-runtime/issues/1960

4. Use pagination to list limited blocks of deployments with uncached client
```
NAMESPACE            NAME                                            CPU           Memory
hmc-system           hmc-controller-manager-6b598b4745-c9tp8         10m           61Mi
kube-system          kube-apiserver-hmc-dev-control-plane            120m          666Mi
```

Closes #138 
Closes #231

### DIFF
--- a/internal/telemetry/event.go
+++ b/internal/telemetry/event.go
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	deploymentCreateEvent = "deployment-create"
+	deploymentCreateEvent    = "deployment-create"
+	deploymentHeartbeatEvent = "deployment-heartbeat"
 )
 
 func TrackDeploymentCreate(id, deploymentID, template string, dryRun bool) error {
@@ -32,6 +33,20 @@ func TrackDeploymentCreate(id, deploymentID, template string, dryRun bool) error
 		"dryRun":       dryRun,
 	}
 	return TrackEvent(deploymentCreateEvent, id, props)
+}
+
+func TrackDeploymentHeartbeat(id, deploymentID, clusterID, template, templateHelmChartVersion, infrastructureProvider, bootstrapProvider, controlPlaneProvider string) error {
+	props := map[string]interface{}{
+		"hmcVersion":               build.Version,
+		"deploymentID":             deploymentID,
+		"clusterID":                clusterID,
+		"template":                 template,
+		"templateHelmChartVersion": templateHelmChartVersion,
+		"infrastructureProvider":   infrastructureProvider,
+		"bootstrapProvider":        bootstrapProvider,
+		"controlPlaneProvider":     controlPlaneProvider,
+	}
+	return TrackEvent(deploymentHeartbeatEvent, id, props)
 }
 
 func TrackEvent(name, id string, properties map[string]interface{}) error {
--- a/internal/telemetry/tracker.go
+++ b/internal/telemetry/tracker.go
@@ -1,0 +1,107 @@
+// Copyright 2024
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: internal/telemetry/event.go [hunk 1]
```diff
@@ -34,6 +35,18 @@ func TrackDeploymentCreate(id, deploymentID, template string, dryRun bool) error
 	return TrackEvent(deploymentCreateEvent, id, props)
 }
 
+func TrackDeploymentHeartbeat(id, deploymentID, clusterID, template, templateVersion, targetInfrastructure string) error {
+	props := map[string]interface{}{
+		"hmcVersion":           build.Version,
+		"deploymentID":         deploymentID,
+		"clusterID":            clusterID,
+		"targetInfrastructure": targetInfrastructure,
```
Kshatrix, line 43: I think this should be infrastructureProvider in CAPI terms (and we can also add other providers - bootstrap and controlplane just in case) @Schnitzel

internal/telemetry/tracker.go
```diff
@@ -0,0 +1,100 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/Mirantis/hmc/api/v1alpha1"
+)
+
+type Tracker struct {
+	crclient.Client
+}
+
+const interval = 10 * time.Minute
```
Kshatrix, line 35: @Schnitzel any preferences for the interval?

Kshatrix (changes requested):

internal/telemetry/tracker.go
squizzi: This should be configurable, even if it's not a CRD option, but just a hidden configuration like an env var.  Also shouldn't there be an option to disable telemetry outright?  A CRD bool or something?

internal/telemetry/tracker.go
```diff
@@ -0,0 +1,108 @@
...
+	var errs error
+	deployments := &v1alpha1.DeploymentList{}
+	err = t.List(ctx, deployments, &crclient.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, deployment := range deployments.Items {
+		template := templates[deployment.Spec.Template]
+		templateVersion := template.Spec.Helm.ChartVersion
```
Schnitzel, line 89: as this is actually the version of the helmchart, we might want to change `templateVersion` to `templateHelmChartVersion`? to not suggest that templates have versions (as we decided that templates don't have versions)

internal/telemetry/tracker.go
Schnitzel: @Kshatrix @eromanova 
I think 10mins is great, and I don't think we need to make it configurable interval, I can't see any other tools being able to configure the telemetry interval

@squizzi 
Yea correct we should probably do that, I created a new issue for this: https://github.com/Mirantis/hmc/issues/231

internal/telemetry/event.go
Schnitzel: I like the updated ones, good for me to go!

Kshatrix (approved):

PR #197, opened 2024-08-15
Specify minLength validation for deployment's template reference

eromanova:
We should ensure the `spec.template` field is set and is not empty. `kubebuilder:validation:Required` marker is not enough since an empty template bypasses the validation: https://github.com/kubernetes-sigs/kubebuilder/issues/3675

### DIFF
--- a/api/v1alpha1/deployment_types.go
+++ b/api/v1alpha1/deployment_types.go
@@ -60,6 +60,7 @@ type DeploymentSpec struct {
 	DryRun bool `json:"dryRun,omitempty"`
 	// Template is a reference to a Template object located in the same namespace.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
 	Template string `json:"template"`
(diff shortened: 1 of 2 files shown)


internal/webhook/deployment_webhook.go
```diff
@@ -59,6 +59,9 @@ func (v *DeploymentValidator) ValidateCreate(ctx context.Context, obj runtime.Ob
 	if !ok {
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected Deployment but got a %T", obj))
 	}
+	if deployment.Spec.Template == "" {
```
Kshatrix, line 62: Template is already set as "required" in the openapi schema, so this one is unnecessary

eromanova: You're right. However, I found that it's no longer working when the admission webhooks are enabled. I'll double-check.

Kshatrix (approved):

Issue #211, opened 2024-08-19
Common k8s cluster verification wf/action


Kshatrix:
* Should be reusable and provider-agnostic
* Should rely on k0s status (no overlapping)
* Should verify ccm/csi integrations

Issue #212, opened 2024-08-19
On-demand AWS verification wf


Kshatrix:
* Deployment of both aws-standalone-cp and aws-hosted-cp templates
* Only the default configuration of templates (will be extended later)
* Pre-setup of aws infra for hosted-cp 
* Verification check from #211 
* Post-cleanup with aws-nuke or a similar tool

Issue #201, opened 2024-08-18
Automated Testing of AWS Templates


DinaBelova:
**Goals**
- Ensure that we deliver functioning templates to customers which cannot cause any outages or other issues when they use them to deploy clusters

**Major deliverables**
- All Templates that are provided by project are automatically tested and verified with actually deploying them to the infrastructure providers. This happens every time we create a new version of a template or a new version of project itself. Per-commit testing is not required for all tests, test that take a long time can be done on a release/tag level.

**Who it benefits**
- Business: Less outages and downtimes of applications
- Platform: Reduction of urgent issues
- Mirantis: Less angry customers

**Acceptance criteria**
- All existing AWS templates are fully automated tested against real world infrastructure
- All testing happens in Github
- Verification of cluster successfully being created: we can rely on CAPI/Infra Providers returning to us that the clsuter has been successfully created
    - Also verify working of additional helm charts installed into the clusters

**Assumptions**
- It is actually possible to test against real world infrastructure without massive amount of costs or engineering time involved to do so. 

**Limitations**
- Budget for test infra

**Out of scope**
- Infrastructures covered by CAPI but not provided with Templates
- Upgrade testing, which will be covered in the relevant implementation epic

Related issues:
- #211 
- #212

DinaBelova:
Note: fell out from sprint-6 to sprint-7 due to couple of delays around certificates / permissions mgmt for CI as well as hosted vs standalone tests specifics. Is planned to be closed early sprint-7.

PR #220, opened 2024-08-22
Add automated testing of AWS templates, rework existing e2e testing to use kubeclient

squizzi:
This PR:

* Reworks the existing e2e tests so that we no longer use iterations of `exec.Command` and instead use a new `KubeClient` type that implements `kubernetes.Interface`
* Adds new Makefile targets `make dev-aws-nuke` and `make cloudnuke` to purge AWS resources affiliated with a desired deployment name, whether that's the test created one or the one created by `config/dev/deployment.yaml`
* Utilizes this new Makefile target to purge deployed AWS resources during teardown.
* Adds a series of provider-agnostic test validators to verify creation and deletion; which can be used to validate a provider deployed cluster, such as the newly added AWS tests.

Closes: #211
See: #201


squizzi:
Any thoughts on what the timeout should be when validating resources coming online from a provider?  I've seen AWS take up to 20 min sometimes to deploy resources, I currently have it set to 30 min with 10 second polling.

squizzi:
:wave: this can be reviewed, but do note the `aws-hosted-cp` template is still not functional so for now I'm skipping that test while I continue to debug it for inclusion.  This way the provider-agnostic verification code can be potentially committed and this is no longer blocking other provider testing initiatives.

squizzi:
I don't know what's going on with the AWS key/secret that IT added, for whatever reason `clusterawsadm` is barfing when it tries to encode the credentials but `make test-e2e` works locally for me.

config/dev/cloud_nuke.yaml.tpl
a13x5, line 1: This probably will not be enough if resources weren't created through CAPI. For example: volumes or LBs that were created by CCM and CSI plugins. I didn't checked it, but AFAIK they don't have a clue about cluster name.

test/deployment/deployment.go
```diff
@@ -0,0 +1,133 @@
...
+func GetUnstructuredDeployment(provider ProviderType, templateName Template) *unstructured.Unstructured {
+	GinkgoHelper()
+
+	generatedName := uuid.New().String()[:8] + "-e2e-test"
+	_, _ = fmt.Fprintf(GinkgoWriter, "Generated cluster name: %q\n", generatedName)
+
+	switch provider {
+	case ProviderAWS:
+		// XXX: Maybe we should just use automatic AMI selection here.
+		amiID := getAWSAMI()
```
a13x5, line 71: AMI is getting selected automatically using image lookup see our docs [here](https://github.com/Mirantis/hmc/blob/main/docs/aws/cluster-parameters.md#image-lookup). 
So this should be removed, so we can test that lookup works as intended.

test/deployment/resources/aws-hosted-cp.yaml.tpl
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  template: aws-hosted-cp
+  config:
+    vpcID: ${AWS_VPC_ID}
+    region: ${AWS_REGION}
+    publicIP: ${PUBLIC_IP:=true}
+    subnets:
+      - id: ${AWS_SUBNET_ID}
+        availabilityZone: ${AWS_SUBNET_AVAILABILITY_ZONE}
+    amiID: ${AWS_AMI_ID}
```
a13x5, line 14: Should be removed

test/deployment/resources/aws-standalone-cp.yaml.tpl
```diff
@@ -0,0 +1,19 @@
...
+  name: ${DEPLOYMENT_NAME}
+spec:
+  template: aws-standalone-cp
+  config:
+    region: ${AWS_REGION}
+    publicIP: ${PUBLIC_IP:=true}
+    controlPlaneNumber: ${CONTROL_PLANE_NUMBER:=1}
+    workersNumber: ${WORKERS_NUMBER:=1}
+    controlPlane:
+      amiID: ${AWS_AMI_ID}
```
a13x5, line 13: Here as well

a13x5 (changes requested):

config/dev/cloud_nuke.yaml.tpl
squizzi: :thinking: ya, that's a good point, I wanted to make sure if concurrent tests are running we don't just nuke all of AWS so I'll need to figure out the best way to filter resources.

squizzi:
Action plan: I'm going to wrap up the `make test-e2e` stuff by finalizing the cloud-nuke filters, break the GitHub e2e test out again so it's separate and make it run on `pull_request_target` with a labeler workflow action and this should be ready for merge.  It will no longer close #212 but should close #211 and allow @a13x5 to work on his e2e tests for Azure. 

In a follow-up I'll add in the `aws-hosted-cp` template testing and any additional fixes for the GitHub actions running.

config/dev/cloud_nuke.yaml.tpl
squizzi: Looks like AWS tags the resources with: `kubernetes.io/cluster/<cluster name>` which is nice, I'll add that to the filter list.

squizzi: anddddd classic `cloud-nuke` does not support tag filtering of ELB's, 😆 , guess I'll be writing a script to deal with these with the `aws` cli _sigh_

squizzi: @a13x5 alright, I've added a custom script that cleans up CCM resources since `cloud-nuke` wasn't up to the task, we may want to just continue building on this custom script instead of using cloud-nuke because it seems like an easier approach and it rids us of yet another dependency but we can do that in the long run.

test/kubeclient/kubeclient.go
```diff
@@ -0,0 +1,268 @@
...
+	Namespace string
+
+	Client         kubernetes.Interface
+	ExtendedClient apiextensionsclientset.Interface
+	Config         *rest.Config
+}
+
+// NewFromLocal creates a new instance of KubeClient from a given namespace
+// using the locally found kubeconfig.
+func NewFromLocal(namespace string) (*KubeClient, error) {
```
a13x5, line 52: Isn't it easier to use [controller runtime client](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client)?

test/e2e/e2e_test.go
```diff
@@ -15,106 +15,240 @@
...
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(func() error {
+					return deployment.VerifyProviderDeleted(context.Background(), kc, clusterName)
+				}).WithTimeout(10 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
+			})
+		}
 	})
 })
+
+func verifyControllerUp(kc *kubeclient.KubeClient, labelSelector string, name string) error {
```
a13x5, line 165: I would use higher level resources (like deployments) + some controllers can have more than one pod (Azure).
Also I believe CAPI operator objects could be verified as well (or at least gathered for further analysis).

a13x5 (changes requested):

scripts/aws-nuke-ccm.sh
```diff
@@ -0,0 +1,52 @@
...
+# This script will remove all resources affiliated with the AWS CCM, such as
+# ELB or CSI driver resources that can not be filtered by cloud-nuke.
+# It should be ran after running cloud-nuke to remove any remaining resources.
+if [ -z $CLUSTER_NAME ]; then
+    echo "CLUSTER_NAME must be set"
+    exit 1
+fi
+
+if [ -z $YQ ]; then
+    YQ=$(which yq)
```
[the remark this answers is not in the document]
squizzi: whyyyy

test/e2e/e2e_test.go
squizzi: This was a direct copy of the existing test, but I'm happy to fix it.

Makefile
```diff
@@ -413,6 +452,6 @@ set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
 GOBIN=$(LOCALBIN) go install $${package} ;\
-mv "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1) ;\
+if [ ! -f $(1) ]; then mv -f "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1); fi ;\
```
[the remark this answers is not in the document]
squizzi: GitHub actions, there's a commit back in the history that describes why:

```
When GitHub Actions runs the 'make cli-install' target it fails
because the GOBIN is already set to ./bin.  GitHub appears to
use a 'mv' that by default does not wish to clobber the file
that now exists (even with -f specified), so place a check in
front of the 'mv' to prevent it from ever running if the file
exists there.  This does not change the existing behavior of
the make target as it does not run if a file exists anyways.
```

Makefile
```diff
@@ -275,23 +291,38 @@ dev-azure-creds: envsubst
...
+
+.PHONY: dev-aws-apply
+dev-aws-apply: ## Apply the AWS deployment
+	$(KUBECTL) -n $(NAMESPACE) apply -f config/dev/deployment.yaml
+
+.PHONY: dev-aws-destroy
+dev-aws-destroy: ## Delete the AWS deployment
+	$(KUBECTL) -n $(NAMESPACE) delete -f config/dev/deployment.yaml
 
-.PHONY: dev-creds-apply
```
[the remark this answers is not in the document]
squizzi: Rebase issues, fixing it.

Makefile
```diff
@@ -382,6 +415,12 @@ yq: $(YQ) ## Download yq locally if necessary.
 $(YQ): | $(LOCALBIN)
 	$(call go-install-tool,$(YQ),github.com/mikefarah/yq/v4,${YQ_VERSION})
 
+.PHONY: cloud-nuke
```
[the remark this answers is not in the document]
squizzi: Not necessarily, we could write our own but it wouldn't be trivial.  I'd like to just add this to our script, but this will do for now; it was specifically requested in [#222 ](https://github.com/Mirantis/hmc/issues/212#:~:text=Post%2Dcleanup%20with%20aws%2Dnuke%20or%20a%20similar%20tool)

test/kubeclient/kubeclient.go
squizzi: Matter of opinion I guess, I felt for writing tests client-go's lack of abstraction made it easier.  I've written operators and worked with both clients and find it easier to work with client-go.

Makefile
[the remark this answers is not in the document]
a13x5: I meant here that I didn't see any use of cloud-nuke (maybe I overlooked it). I do see that you have a script and some aws cli stuff, so probably there is no need to install it.

Makefile
```diff
@@ -275,23 +291,38 @@ dev-azure-creds: envsubst
...
+.PHONY: test-apply
+test-apply: kind-deploy registry-deploy dev-push dev-deploy dev-templates
 
-.PHONY: dev-provider-delete
-dev-provider-delete: envsubst
-	@NAMESPACE=$(NAMESPACE) $(ENVSUBST) -no-unset -i config/dev/$(DEV_PROVIDER)-deployment.yaml | $(KUBECTL) delete -f -
+.PHONY: test-destroy
+test-destroy: kind-undeploy registry-undeploy
+
+.PHONY: get-local-bin
```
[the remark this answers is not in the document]
squizzi: Originally I was going to use it in some scripts to determine the bin locations of various tools, but I decided against it.  I'll remove it.

test/e2e/e2e_test.go
a13x5: Ah this is from a part which was generated. I guess we should fix it, but probably not in this PR I will not insist on that

squizzi: > Also I believe CAPI operator objects could be verified as well (or at least gathered for further analysis).

I agree, but it's a bit of scope creep for the original issue, I fixed this test because I was trying to get away from the use of `exec.Command` and some changes that came with the rebase when `capi-operator` was added, it might make sense to target some improvements for this in a follow-up that improves the testing of the operator stuff in general.

Makefile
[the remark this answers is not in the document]
a13x5: Ok got it Actions quirks

test/kubeclient/kubeclient.go
a13x5: Fair enough. I just have a feeling certain stuff (like getting config) already exists in controller runtime client. 
But I guess it's really a matter of taste here.
@Kshatrix WDYT?

squizzi:
It might make sense to break out the testing into infrastructure suites that require more expensive CI testing (like using an actual AWS or Azure cloud) and then stuff that runs locally in `kind`.  Then the `test-e2e` stuff can be separate from that as well.  We can probably make the suites depend on each other but be skipped in certain circumstances.

kylewuolle (approved):

slysunkin (dismissed):

a13x5:
@squizzi Yes the choose of client is not _that_ important. The rest looks good.

a13x5 (approved):

a13x5 (approved):

PR #224, opened 2024-08-22
Use CAPI operator to manage providers

### LINKED ISSUE #214: Use CAPI Operator to manage providers
* Install CAPI Operator as part of the HMC chart
* Use CAPI Operator resources in provider templates

### LINKED ISSUE #215: Install CAPI Operator as part of the HMC chart

eromanova:
1. Add CAPI operator sub chart to HMC. The CAPI operator requires the cert-manager to be installed, thus, this subchart is disabled by default and will be enabled by the HMC controller once the cert-manager is ready. It's also allowed to disable the CAPI operator shipped with HMC to support an external CAPI operator installation.
2. Rework cluster API, cluster API provider AWS, and k0smotron templates to use CAPI operator resources.
3. Change AWS credentials secret data to follow the format required by the CAPI operator.
4. Allow passing providers configuration options in the secret.
6. Change docs accordingly.
7. Added CAPZ template (enabled by default).

### Providers configuration
The CAPI operator allows passing `configSecret` with all the configuration variables: https://doc.crds.dev/github.com/kubernetes-sigs/cluster-api-operator/operator.cluster.x-k8s.io/InfrastructureProvider/v1alpha2@v0.11.0#spec-configSecret

HMC provider templates were adapted to either pass a pre-created configuration secret name and namespace or receive the configuration parameters as helm chart values (the secret with the provided configuration will be automatically created). This PR introduces the common structure for passing providers' additional parameters.

Examples:
1. To pass the pre-created configuration secret, set:
```
configSecret:
  name: variables-secret-name
  namespace: variables-secret-namespace // optional. Defaults to `hmc-system`
```

2. To pass the configuration as a helm value and create a secret with the configuration automatically, set:
```
configSecret:
  create: true
  name: variables-secret-name // required 
  namespace: variables-secret-namespace // optional. Defaults to `hmc-system`
config:
  CAPI_INSECURE_DIAGNOSTICS: "true"
```

3. If no overrides for the provider are required and the `configSecret` configuration is unset, the provider will be deployed with the defaults.

Closes #214 
Closes #215


internal/controller/management_controller.go
```diff
@@ -252,6 +258,15 @@ func (r *ManagementReconciler) enableAdmissionWebhook(ctx context.Context, mgmt
 
 	admissionWebhookValues["enabled"] = true
 	config["admissionWebhook"] = admissionWebhookValues
+
+	// Enable HMC capi operator only if it was not explicitly disabled in the config to
+	// support installation with existing cluster api operator
+	if capiOperatorValues["enabled"] != false {
```
Kshatrix, line 264: This looks a bit weird to me. I think if we want to preserve the current API (keep CAPI as a separate component of the Management object), we should make the CAPI field optional (now it's required in the CRD)

templates/cluster-api/values.yaml
```diff
@@ -1,11 +1 @@
...
-    repository: registry.k8s.io/cluster-api
-    name: cluster-api-controller
-    tag: v1.7.2
-  imagePullPolicy: IfNotPresent
-
-clusterTopology: false
-
-expClusterResourceSet: false
-expMachinePool: false
+version: v1.7.2
```
Kshatrix, line 1: I'm not sure whether we want to make it configurable here or make it strictly dependent on the HMC version @Schnitzel

Kshatrix (changes requested):

internal/controller/management_controller.go
eromanova: @Kshatrix I'm not sure I fully understood. This condition only affects the `CAPI Operator` subchart, not the CAPI itself. To support an external installation of the CAPI operator (in case it already exists on the management cluster). However, CAPI remains a separate component of the management object (no changes here). Could you please elaborate?

Kshatrix: Nvm, I overlooked that operator is installed as part of HMC itself. If it requires certmanager to be running, it makes sense. Thanks!

templates/cluster-api-provider-aws/values.yaml
```diff
@@ -1,31 +1,3 @@
...
-  image:
-    repository: registry.k8s.io/cluster-api-aws
-    name: cluster-api-aws-controller
-    tag: v2.6.1
-  imagePullPolicy: IfNotPresent
+version: v2.6.1
 
-credentialsSecretName: ""
-
-autoControllerIdentityCreator: true
```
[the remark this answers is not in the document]
eromanova: I doubted it but yes, it makes sense. Thanks, I'll update the PR.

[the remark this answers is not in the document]
eromanova: @Kshatrix Done. Please, check the 1st comment on the PR. I've described the approach of using the CAPI operator `configSecret`.

templates/cluster-api/values.yaml
Schnitzel: I'm ok to have the CAPI version bound to a specific 2A version and NOT have it configurable, as in the end there is so many ways that CAPI and 2A could be out of sync if we have too much version skew.

Kshatrix (approved):

PR #230, opened 2024-08-26
Add CAPI status/conditions to deployment status

### LINKED ISSUE #219: Update Deployment status based on CAPI objects condition

kylewuolle:
Resolves #219. This PR adds a new function that fetches the CAPI status / conditions of the cluster and places them in the deployment status/conditions. To ensure the conditions are updated the controller now requeues until the CAPI cluster conditions are all true.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -27,6 +27,7 @@ import (
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/dynamic"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
@@ -152,6 +153,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	dc, err := dynamic.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		setupLog.Error(err, "failed to create dynamic client")
+		os.Exit(1)
+	}
+
 	if err = (&controller.TemplateReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
@@ -160,9 +167,10 @@ func main() {
 		os.Exit(1)
 	}
 	if err = (&controller.DeploymentReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
-		Config: mgr.GetConfig(),
+		Client:        mgr.GetClient(),
+		Scheme:        mgr.GetScheme(),
+		Config:        mgr.GetConfig(),
+		DynamicClient: dc,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Deployment")
 		os.Exit(1)
--- a/internal/controller/deployment_controller.go
+++ b/internal/controller/deployment_controller.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/labels"
+
 	hcv2 "github.com/fluxcd/helm-controller/api/v2"
 	fluxmeta "github.com/fluxcd/pkg/apis/meta"
 	fluxconditions "github.com/fluxcd/pkg/runtime/conditions"
@@ -30,8 +32,11 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -48,8 +53,9 @@ import (
 // DeploymentReconciler reconciles a Deployment object
 type DeploymentReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
-	Config *rest.Config
+	Scheme        *runtime.Scheme
+	Config        *rest.Config
+	DynamicClient *dynamic.DynamicClient
 }
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
@@ -57,7 +63,6 @@ type DeploymentReconciler struct {
 func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	l := log.FromContext(ctx).WithValues("DeploymentController", req.NamespacedName)
 	l.Info("Reconciling Deployment")
-
 	deployment := &hmc.Deployment{}
 	if err := r.Get(ctx, req.NamespacedName, deployment); err != nil {
 		if apierrors.IsNotFound(err) {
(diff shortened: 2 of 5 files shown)


internal/controller/deployment_controller.go
```diff
@@ -87,6 +89,63 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		return true, fmt.Errorf("failed to create client configuration: %v", err)
+	}
+
+	resourceId := schema.GroupVersionResource{
+		Group:    "cluster.x-k8s.io",
+		Version:  "v1beta1",
+		Resource: "clusters",
+	}
+
+	list, err := dc.Resource(resourceId).Namespace(deployment.Namespace).Get(ctx, deployment.Name, metav1.GetOptions{})
```
Kshatrix, line 104: I think it's more reliable to use helm labels instead of a specific name as naming patterns are not enforced. I think it's fine to assume that we expect exactly one cluster object for now.

Kshatrix (dismissed):

internal/controller/deployment_controller.go
```diff
@@ -87,6 +89,63 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if apierrors.IsNotFound(err) {
+		l.Info("Cluster not found, ignoring since object must be deleted or not yet created")
+		return true, nil
+	}
+
+	if err != nil {
+		return true, fmt.Errorf("failed to get cluster information for deployment: %s : %v", deployment.Name, err)
```
squizzi, line 111:
```suggestion
		return true, fmt.Errorf("failed to get cluster information for deployment: %s: %w", deployment.Name, err)
```

internal/controller/deployment_controller.go
```diff
@@ -87,6 +89,63 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	}
+
+	requeue := true
+	if statusField, ok := list.Object["status"]; ok {
+		if statusMap, ok := statusField.(map[string]interface{}); ok {
+			if conditionsField, ok := statusMap["conditions"]; ok {
+				bytes, err := json.Marshal(conditionsField)
+				if err != nil {
+					return true, fmt.Errorf("failed to get deserialize cluster information for deployment: %s : %v",
```
squizzi, line 120:
```suggestion
					return true, fmt.Errorf("failed to get deserialize cluster information for deployment: %s: %w",
```

internal/controller/deployment_controller.go
```diff
@@ -87,6 +89,63 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		l.Info("Cluster not found, ignoring since object must be deleted or not yet created")
+		return true, nil
+	}
+
+	if err != nil {
+		return true, fmt.Errorf("failed to get cluster information for deployment: %s : %v", deployment.Name, err)
+	}
+
+	requeue := true
+	if statusField, ok := list.Object["status"]; ok {
```
squizzi, line 115: You don't need to do this, the API returns `unstructured.Unstructured` types and there's an entire library to help you make these conversions and get this content much much easier:

```go
	conditions, found, err := unstructured.NestedMap(list.Object, "status", "conditions")
	if err != nil {
		return true, fmt.Errorf("failed to get cluster information for deployment: %s: %w", deployment.Name, err)
	}
	if !found {
		return true, fmt.Errorf("failed to get cluster information for deployment: %s: status.conditions not found", deployment.Name)
	}

	var conditionsList []metav1.Condition

	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(conditions, &conditionsList); err != nil {
		return true, fmt.Errorf("failed to convert unstructured conditions to []metav1.Condition for deployment: %s: %w", deployment.Name, err)
	}

	allConditionsComplete := true
	for _, condition := range conditionsList {
		if condition.Status != "True" {
			allConditionsComplete = false
		}

		if condition.Reason == "" && condition.Status == "True" {
			condition.Reason = "Succeeded"
		}
		apimeta.SetStatusCondition(deployment.GetConditions(), condition)
	}
	requeue := !allConditionsComplete

	return requeue, nil
```

kylewuolle: Kind of works thanks! Turns out the conditions field is a list of interfaces and not maps though so I had to change it a bit.

internal/controller/deployment_controller.go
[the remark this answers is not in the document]
kylewuolle: Good point I've added that in to all of them.

internal/controller/deployment_controller.go
kylewuolle: Is the concern around uniqueness? I believe the cluster name must be unique within a namespace.

config/dev/deployment.yaml
```diff
@@ -1,17 +1,17 @@
 apiVersion: hmc.mirantis.com/v1alpha1
```
squizzi, line 1: Looks like you checked your `deployment.yaml` in, run a `git checkout upstream/main -- config/dev/deployment.yaml` and grab the original.

internal/controller/deployment_controller.go
```diff
@@ -87,7 +89,66 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+func (r *DeploymentReconciler) setStatusFromClusterStatus(ctx context.Context, dc *dynamic.DynamicClient, l logr.Logger, deployment *hmc.Deployment) (bool, error) {
+	resourceId := schema.GroupVersionResource{
+		Group:    "cluster.x-k8s.io",
+		Version:  "v1beta1",
+		Resource: "clusters",
+	}
+
+	list, err := dc.Resource(resourceId).Namespace(deployment.Namespace).Get(ctx, deployment.Name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		l.Info("Cluster not found, ignoring since object must be deleted or not yet created")
```
squizzi, line 101: This is a list object but the info returned implies a single cluster, perhaps `No clusters found` is a better log message here.

internal/controller/deployment_controller.go
```diff
@@ -224,6 +285,20 @@ func (r *DeploymentReconciler) Update(ctx context.Context, l logr.Logger, deploy
...
+		if err != nil {
+			if requeue {
+				return ctrl.Result{RequeueAfter: 10 * time.Second}, err
+			} else {
+				return ctrl.Result{}, err
+			}
+		}
+
+		if requeue {
+			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
squizzi, line 299: Instead of requeuing every 10 seconds, perhaps its better to add a `Watches` for the cluster object (once it's added to scheme) within `SetupWithManager`, that way when the status of the cluster object changes we would requeue rather then relying on polling.

internal/controller/deployment_controller.go
Kshatrix: No guarantee that cluster.Name == deployment.Name

kylewuolle: Ah ok I've changed it to use the label

internal/controller/deployment_controller.go
kylewuolle: Looks like there's only one watch allowed per controller unfortunately.

internal/controller/deployment_controller.go
```diff
@@ -87,7 +91,69 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	return r.Update(ctx, l, deployment)
 }
 
+func (r *DeploymentReconciler) setStatusFromClusterStatus(ctx context.Context, dc *dynamic.DynamicClient, l logr.Logger, deployment *hmc.Deployment) (bool, error) {
+	resourceId := schema.GroupVersionResource{
+		Group:    "cluster.x-k8s.io",
+		Version:  "v1beta1",
+		Resource: "clusters",
```
squizzi, line 98: I don't want to scope creep you, perhaps it's in a different issue, but Is `clusters` the only resource that this status should be referencing?  What about `Workers` and specifically `MachineDeployment`'s, the cluster could be `ready: True` when the workers aren't yet up.

kylewuolle: Michael said to just go with the clusters for now.

squizzi (changes requested):

squizzi:
@kylewuolle just need to rebase to get rid of this merge commit: https://github.com/Mirantis/hmc/pull/230/commits/0635d504fd94874418e1cf426e00c9e425a0dae3 and it lgtm

cmd/main.go
```diff
@@ -163,6 +170,7 @@ func main() {
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Config: mgr.GetConfig(),
+		Dc:     dc,
```
squizzi, line 173: nit: Any specific reason for leaving this `Dc`, if I was in my IDE and typing out `r.` and looking at completions `DynamicClient` would be a much more descriptive "oh I need this" then `Dc`.

squizzi (approved):

PR #234, opened 2024-08-27
Improve hosted docs with 'gotemplate' example to make configuring the deployment.yaml less tedious

squizzi:


slysunkin (dismissed):
Looks good

kylewuolle (dismissed):

a13x5 (dismissed):
IMO this makes docs more obscure. My intention was to actually show an example of the object without variables.

If you want to make fast/smart example I suggest to just add gotemplate and add it below the full object example.

Go template example:

```yaml
kind: Deployment
metadata:
  name: aws-hosted-cp
spec:
  template: aws-hosted-cp
  config:
    vpcID: "{{.spec.network.vpc.id}}"
    region: us-west-1
    subnets:
      - id: "{{(index .spec.network.subnets 0).resourceID}}"
        availabilityZone: "{{(index .spec.network.subnets 0).availabilityZone}}"
    amiID: ami-0bf2d31c356e4cb25
    instanceType: t3.medium
    securityGroupIDs:
      - "{{.status.networkStatus.securityGroups.node.id}}"
```

Then it can just be evaluated with:

```bash
kubectl get awscluster cluster -o go-template="$(cat template)" > deployment.yaml
```
I believe it looks cleaner than bunch exports and envsubst.

squizzi:
> IMO this makes docs more obscure. My intention was to actually show an example of the object without variables.

@a13x5 We can go with go template, that's fine, but you're comment makes me think perhaps their should just be a hosted subsection in the dev doc which includes the fast way of doing this.  I think most instances of hosted setup for someone that isn't a dev aren't going to reuse standalone resources anyways.  Thoughts?

a13x5:
@squizzi I think it's better to add convenient shortcuts in this doc and not in the dev doc, since it's provider specific. We will have other quirks for other providers described in their respective docs. For example I will add the same section for Azure as well.

slysunkin (approved):

PR #239, opened 2024-08-28
Add defaultRegistryURL and defaultRepoType for non-OCI HelmRepository support

squizzi:
While working on figuring out what registry to use for CI in #220  I thought why not use Mirantis Secure Registry (MSR)?welp.. that was a journey.

Currently we hardcode the `HelmRepostory` type to `oci` when reconciling the default `hmc-templates` repository.  The problem with this is, for whatever reason, MSR does not like OCI (even though, I'm pretty sure it implements the spec) and it results in `insufficient_scope` errors when we try to pull the charts.  After some digging through `helm-controller` CRDs and docs I discovered this setting: https://fluxcd.io/flux/components/source/helmrepositories/#type and changed our `HelmRepository` we reconcile to use `'default'` which fixed the issue and pulled the chart with no problems.

---

This patch renames the existing 'defaultOCIRegistry' flag and associated Helm values to 'defaultRegistryURL' and introduces a new type: 'defaultRepoType' which allows for modification of the previously hardcoded 'oci' type for chart repositories.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -56,20 +56,23 @@ func init() {
 }
 
 func main() {
-	var metricsAddr string
-	var probeAddr string
-	var secureMetrics bool
-	var enableHTTP2 bool
-	var defaultOCIRegistry string
-	var insecureRegistry bool
-	var registryCredentialsSecret string
-	var createManagement bool
-	var createTemplates bool
-	var hmcTemplatesChartName string
-	var enableTelemetry bool
-	var enableWebhook bool
-	var webhookPort int
-	var webhookCertDir string
+	var (
+		metricsAddr               string
+		probeAddr                 string
+		secureMetrics             bool
+		enableHTTP2               bool
+		defaultRegistryURL        string
+		defaultRepoType           string
+		insecureRegistry          bool
+		registryCredentialsSecret string
+		createManagement          bool
+		createTemplates           bool
+		hmcTemplatesChartName     string
+		enableTelemetry           bool
+		enableWebhook             bool
+		webhookPort               int
+		webhookCertDir            string
+	)
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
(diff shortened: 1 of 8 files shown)


templates/hmc/values.yaml
```diff
@@ -7,7 +7,8 @@ admissionWebhook:
   certDir: "/tmp/k8s-webhook-server/serving-certs/"
 
 controller:
-  defaultOCIRegistry: "oci://ghcr.io/mirantis/hmc/charts"
+  defaultRegistryURL: "oci://ghcr.io/mirantis/hmc/charts"
+  defaultRepoType: "oci"
```
Kshatrix, line 11: It would be nice to automatically discover the type based on URL scheme. what do you think?

squizzi: I need to make sure the OCI spec does not allow for URLs to start with `http/https` too, but I'd be game, but if that's the case -- it makes me wonder why Flux didn't adopt URL scheme and introduced type in the first place.

squizzi: There's nothing in the OCI spec that says `oci://` should be the URI scheme, it simply notes that OCI operates over `http://`, my only concern with changing this is the what if some customer comes along and uses an OCI type repo and they enter their registry URL as `https://` because that's what makes sense to them.

cmd/main.go
```diff
@@ -56,29 +56,34 @@ func init() {
...
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&secureMetrics, "metrics-secure", false,
 		"If set the metrics endpoint is served securely")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
-	flag.StringVar(&defaultOCIRegistry, "default-oci-registry", "oci://ghcr.io/mirantis/hmc/charts",
+	flag.StringVar(&defaultRegistryURL, "default-registry-url", "oci://ghcr.io/mirantis/hmc/charts",
 		"The default OCI registry to download Helm charts from.")
```
eromanova, line 84: Let's remove `OCI` from the variable description since now it's not the only registry type supported.

squizzi: whoops, thought I updated that, thanks.

templates/hmc/values.yaml
squizzi: It seems Flux makes an assumption based on URL scheme and provides useful errors if an OCI typed repository is used with an `http/https` URL scheme so I think it's probably fine for us to assume as well.

Kshatrix (approved):

PR #242, opened 2024-08-30
Improved HelmRelease deletion

slysunkin:
## Description

Fixing https://github.com/Mirantis/hmc/issues/217

This fix is intended to resolve an issue when deletion of HelmRelease resulted to non-deleted (orphaned) objects.

The root cause of the issue was with the removal AWSCluster while Machine objects have not been deleted.

### DIFF
--- a/api/v1alpha1/managedcluster_types.go
+++ b/api/v1alpha1/managedcluster_types.go
@@ -22,11 +22,14 @@ import (
 )
 
 const (
+	BlockingFinalizer       = "hmc.mirantis.com/cleanup"
 	ManagedClusterFinalizer = "hmc.mirantis.com/managed-cluster"
 
 	FluxHelmChartNameKey = "helm.toolkit.fluxcd.io/name"
 	HMCManagedLabelKey   = "hmc.mirantis.com/managed"
 	HMCManagedLabelValue = "true"
+
+	ClusterNameLabelKey = "cluster.x-k8s.io/cluster-name"
 )
 
 const (
--- a/internal/controller/managedcluster_controller.go
+++ b/internal/controller/managedcluster_controller.go
@@ -20,8 +20,6 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/labels"
-
 	hcv2 "github.com/fluxcd/helm-controller/api/v2"
 	fluxmeta "github.com/fluxcd/pkg/apis/meta"
 	fluxconditions "github.com/fluxcd/pkg/runtime/conditions"
@@ -33,6 +31,7 @@ import (
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
@@ -58,6 +57,30 @@ type ManagedClusterReconciler struct {
 	DynamicClient   *dynamic.DynamicClient
 	SystemNamespace string
 }
+
+type providerSchema struct {
+	machine, cluster schema.GroupVersionKind
+}
+
+var (
+	gvkAWSCluster = schema.GroupVersionKind{
+		Group:   "infrastructure.cluster.x-k8s.io",
+		Version: "v1beta2",
+		Kind:    "awscluster",
+	}
+
+	gvkAzureCluster = schema.GroupVersionKind{
+		Group:   "infrastructure.cluster.x-k8s.io",
+		Version: "v1beta1",
+		Kind:    "azurecluster",
+	}
+
+	gvkMachine = schema.GroupVersionKind{
+		Group:   "cluster.x-k8s.io",
+		Version: "v1beta1",
+		Kind:    "machine",
+	}
+)
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
@@ -385,12 +408,109 @@ func (r *ManagedClusterReconciler) Delete(ctx context.Context, l logr.Logger, ma
 		}
 		return ctrl.Result{}, err
 	}
+
 	err = helm.DeleteHelmRelease(ctx, r.Client, managedCluster.Name, managedCluster.Namespace)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	err = r.releaseCluster(ctx, managedCluster.Namespace, managedCluster.Name, managedCluster.Spec.Template)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	l.Info("HelmRelease still exists, retrying")
 	return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+}
+
+func (r *ManagedClusterReconciler) releaseCluster(ctx context.Context, namespace, name, templateName string) error {
+	providers, err := r.getProviders(ctx, templateName)
+	if err != nil {
+		return err
+	}
+
+	providerGVKs := map[string]providerSchema{
+		"aws":   {machine: gvkMachine, cluster: gvkAWSCluster},
+		"azure": {machine: gvkMachine, cluster: gvkAzureCluster},
+	}
+
+	// Associate the provider with it's GVK
+	for _, provider := range providers {
+		gvk, ok := providerGVKs[provider]
+		if !ok {
+			continue
+		}
+
+		cluster, err := r.getCluster(ctx, namespace, name, gvk.cluster)
+		if err != nil {
+			return err
+		}
+
+		found, err := r.machinesAvailable(ctx, namespace, cluster.Name, gvk.machine)
+		if err != nil {
+			return err
+		}
+
+		if !found {
+			return r.removeClusterFinalizer(ctx, cluster)
+		}
+	}
+
+	return nil
+}
+
+func (r *ManagedClusterReconciler) getProviders(ctx context.Context, templateName string) ([]string, error) {
+	template := &hmc.ClusterTemplate{}
+	templateRef := types.NamespacedName{Name: templateName, Namespace: r.SystemNamespace}
+	if err := r.Get(ctx, templateRef, template); err != nil {
+		log.FromContext(ctx).Error(err, "Failed to get Template", "templateName", templateName)
+		return nil, err
+	}
+	return template.Status.Providers.InfrastructureProviders, nil
+}
+
+func (r *ManagedClusterReconciler) getCluster(ctx context.Context, namespace, name string, gvk schema.GroupVersionKind) (*metav1.PartialObjectMetadata, error) {
+	opts := &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{hmc.FluxHelmChartNameKey: name}),
+		Namespace:     namespace,
+	}
+	itemsList := &metav1.PartialObjectMetadataList{}
+	itemsList.SetGroupVersionKind(gvk)
+	if err := r.Client.List(ctx, itemsList, opts); err != nil {
+		return nil, err
+	}
+	if len(itemsList.Items) == 0 {
+		return nil, fmt.Errorf("%s with name %s was not found", gvk.Kind, name)
+	}
+
+	return &itemsList.Items[0], nil
+}
+
+func (r *ManagedClusterReconciler) removeClusterFinalizer(ctx context.Context, cluster *metav1.PartialObjectMetadata) error {
+	originalCluster := *cluster
+	finalizersUpdated := controllerutil.RemoveFinalizer(cluster, hmc.BlockingFinalizer)
+	if finalizersUpdated {
+		log.FromContext(ctx).Info("Allow to stop cluster", "finalizer", hmc.BlockingFinalizer)
+		if err := r.Client.Patch(ctx, cluster, client.MergeFrom(&originalCluster)); err != nil {
+			return fmt.Errorf("failed to patch cluster %s/%s: %w", cluster.Namespace, cluster.Name, err)
+		}
+	}
+
+	return nil
+}
+
+func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namespace, clusterName string, gvk schema.GroupVersionKind) (bool, error) {
+	opts := &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{hmc.ClusterNameLabelKey: clusterName}),
+		Namespace:     namespace,
+		Limit:         1,
+	}
+	itemsList := &metav1.PartialObjectMetadataList{}
+	itemsList.SetGroupVersionKind(gvk)
+	if err := r.Client.List(ctx, itemsList, opts); err != nil {
+		return false, err
+	}
+	return len(itemsList.Items) != 0, nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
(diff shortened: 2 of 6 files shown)


internal/controller/deployment_controller.go
```diff
@@ -395,6 +395,12 @@ func (r *DeploymentReconciler) Delete(ctx context.Context, l logr.Logger, deploy
 		}
 		return ctrl.Result{}, err
 	}
+
+	err = r.keepOrReleaseAWSCluster(ctx, l, deployment.Namespace, deployment.Name)
```
eromanova, line 399: If we want this logic for AWS provider only, let's check the deployment type first (that infrastructure provider equals AWS). If we're going to keep this behavior for all providers, please adapt the name of the function and error messages to make it more common.

internal/controller/deployment_controller.go
```diff
@@ -403,6 +409,60 @@ func (r *DeploymentReconciler) Delete(ctx context.Context, l logr.Logger, deploy
...
+func (r *DeploymentReconciler) keepOrReleaseAWSCluster(ctx context.Context, l logr.Logger, namespace, clusterName string) error {
+	found, err := r.machinesAvailable(ctx, namespace, clusterName)
+	if err != nil {
+		return err
+	}
+
+	return r.updateAWSCluster(ctx, l, namespace, clusterName, found)
+}
+
+func (r *DeploymentReconciler) updateAWSCluster(ctx context.Context, l logr.Logger, namespace, clusterName string, disableDeletion bool) error {
```
eromanova, line 421: You can get logger from context instead of passing it as a parameter:
```
l := log.FromContext(ctx)
```

internal/controller/deployment_controller.go
```diff
@@ -403,6 +409,60 @@ func (r *DeploymentReconciler) Delete(ctx context.Context, l logr.Logger, deploy
...
+	cluster := &capiaws.AWSCluster{}
+	err := r.Get(ctx, types.NamespacedName{
+		Name:      clusterName,
+		Namespace: namespace,
+	}, cluster)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
```
eromanova, line 429: Let's handle the `NotFound` error (if the case the AWSCluster is already gone).

internal/controller/deployment_controller.go
```diff
@@ -403,6 +409,60 @@ func (r *DeploymentReconciler) Delete(ctx context.Context, l logr.Logger, deploy
...
+
+	finalizersUpdated := false
+	if disableDeletion {
+		l.Info("Ensure we have cannot stop AWSCluster", "finalizer", hmc.BlockingFinalizer)
+		finalizersUpdated = controllerutil.AddFinalizer(cluster, hmc.BlockingFinalizer)
```
eromanova, line 434: Can we consider the option to set the finalizer initially, when creating the `AWSCluster` object? I mean, add the finalizer to the `AWSCluster` manifest in the templates?

internal/controller/deployment_controller.go
```diff
@@ -409,33 +409,36 @@ func (r *DeploymentReconciler) Delete(ctx context.Context, l logr.Logger, deploy
...
-	return r.updateAWSCluster(ctx, l, namespace, clusterName, found)
+	if !found {
+		return r.removeAWSClusterFinilizer(ctx, namespace, clusterName)
+	}
+
+	return nil
 }
 
-func (r *DeploymentReconciler) updateAWSCluster(ctx context.Context, l logr.Logger, namespace, clusterName string, disableDeletion bool) error {
+func (r *DeploymentReconciler) removeAWSClusterFinilizer(ctx context.Context, namespace, clusterName string) error {
```
eromanova, line 425:
```suggestion
func (r *DeploymentReconciler) removeAWSClusterFinalizer(ctx context.Context, namespace, clusterName string) error {
```

templates/aws-hosted-cp/templates/awscluster.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   name: {{ include "cluster.name" . }}
   annotations:
     cluster.x-k8s.io/managed-by: k0smotron
+  finalizers:
+  - hmc.mirantis.com/do-not-delete
```
eromanova, line 8: Shouldn't we add it to `aws-standalone-cp` as well?

slysunkin: I don't think so. The issue was with AWSCluster removed before underlying objects. Or do you see any other cases?

Kshatrix: It's especially important for the hosted template. please add and verify if anything else is needed.

a13x5: @slysunkin @Kshatrix This probably won't work. In my tests on initial apply the `finalizers` field was emptied (thus the need for patch in the docs). This should be checked thoroughly. 
`AzureCluster` is also affected, so we should add fix for it as well,

a13x5 (changes requested):
Initial set of `finalizer` should be rechecked
`AzureCluster` is affected, so it should be handled here as well

templates/aws-hosted-cp/templates/awscluster.yaml
slysunkin: I'm working on Azure now

templates/aws-standalone-cp/templates/awscluster.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
 kind: AWSCluster
 metadata:
   name: {{ include "cluster.name" . }}
+  finalizers:
```
a13x5, line 5: It makes no sense to add it for standalone. Remove.

internal/controller/deployment_controller.go
```diff
@@ -409,55 +435,58 @@ func (r *DeploymentReconciler) Delete(ctx context.Context, l logr.Logger, deploy
 	return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 }
 
-func (r *DeploymentReconciler) releaseAWSCluster(ctx context.Context, namespace, clusterName, templateName string) error {
-	aws, err := r.hasAWSProvider(ctx, templateName)
+func (r *DeploymentReconciler) releaseCluster(ctx context.Context, namespace, clusterName, templateName string) error {
+	prividers, err := r.getProviders(ctx, templateName)
```
squizzi, line 439:
```suggestion
	providers, err := r.getProviders(ctx, templateName)
```

internal/controller/deployment_controller.go
```diff
@@ -409,55 +435,58 @@ func (r *DeploymentReconciler) Delete(ctx context.Context, l logr.Logger, deploy
...
 	if err != nil {
 		return err
 	}
-	if !aws {
-		// nothing to do - we can skip this step
-		return nil
-	}
+	if slices.Contains(prividers, "aws") {
```
squizzi, line 443: The only thing that differs between these two blocks in the provider name string and the `gvk` being used by `removeClusterFinalizer`, the more providers we add the longer this will grow.

:golf: 

```
type providerSchema struct {
	machine, cluster schema.GroupVersionKind
}

func (r *DeploymentReconciler) releaseCluster(ctx context.Context, namespace, clusterName, templateName string) error {
	providers, err := r.getProviders(ctx, templateName)
	if err != nil {
		return err
	}

	providerGVKs := map[string]providerSchema{
		"aws":   providerSchema{machine: gvkAWSMachine, cluster: gvkAWSCluster},
		"azure": providerSchema{machine: gvkAzureMachine, cluster: gvkAzureCluster},
	}

	// Associate the provider with it's GVK
	for _, provider := range providers {
		gvk := providerGVKs[provider]

		found, err := r.machinesAvailable(ctx, namespace, clusterName, gvk.machine)
		if err != nil {
			return err
		}

		if !found {
			return r.removeClusterFinalizer(ctx, namespace, clusterName, gvk.cluster)
		}

	}

	return nil
}
```

slysunkin: Thanks for the improvement

internal/controller/managedcluster_controller.go
```diff
@@ -404,6 +440,90 @@ func (r *ManagedClusterReconciler) Delete(ctx context.Context, l logr.Logger, ma
...
+	}
+
+	// Associate the provider with it's GVK
+	for _, provider := range providers {
+		gvk, ok := providerGVKs[provider]
+		if !ok {
+			continue
+		}
+
+		found, err := r.machinesAvailable(ctx, namespace, clusterName, gvk.machine)
```
Kshatrix, line 461: Same situation here, managedCluster.Name != cluster.Name

internal/controller/managedcluster_controller.go
```diff
@@ -404,6 +440,90 @@ func (r *ManagedClusterReconciler) Delete(ctx context.Context, l logr.Logger, ma
...
+			return r.removeClusterFinalizer(ctx, namespace, clusterName, gvk.cluster)
+		}
+	}
+
+	return nil
+}
+
+func (r *ManagedClusterReconciler) getProviders(ctx context.Context, templateName string) ([]string, error) {
+	template := &hmc.Template{}
+	templateRef := types.NamespacedName{Name: templateName, Namespace: utils.DefaultSystemNamespace}
```
Kshatrix, line 476:
```suggestion
	templateRef := types.NamespacedName{Name: templateName, Namespace: r.SystemNamespace}
```

api/v1alpha1/managedcluster_types.go
```diff
@@ -22,11 +22,14 @@ import (
 )
 
 const (
+	BlockingFinalizer       = "hmc.mirantis.com/do-not-delete"
```
Kshatrix, line 25:
```suggestion
	BlockingFinalizer       = "hmc.mirantis.com/cleanup"
```

templates/aws-hosted-cp/templates/awscluster.yaml
Kshatrix, line 8:
```suggestion
  - hmc.mirantis.com/cleanup
```

templates/azure-hosted-cp/templates/azurecluster.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   name: {{ include "cluster.name" . }}
   annotations:
     cluster.x-k8s.io/managed-by: k0smotron
+  finalizers:
+  - hmc.mirantis.com/do-not-delete
```
Kshatrix, line 8:
```suggestion
  - hmc.mirantis.com/cleanup
```

internal/controller/managedcluster_controller.go
```diff
@@ -396,6 +426,12 @@ func (r *ManagedClusterReconciler) Delete(ctx context.Context, l logr.Logger, ma
 		}
 		return ctrl.Result{}, err
 	}
+
+	err = r.releaseCluster(ctx, managedCluster.Namespace, managedCluster.Name, managedCluster.Spec.Template)
```
Kshatrix, line 430: Should probably be moved after helm.DeleteHelmRelease

Kshatrix (changes requested):

templates/aws-standalone-cp/templates/awscluster.yaml
slysunkin: I've retained this finalizer: I encountered several times when Machines were left behind after AWSCluster is removed.

internal/controller/managedcluster_controller.go
slysunkin: I did some experiments and found the following:
Machine, AWSMachine or AzureMachine objects don't have labels like "helm.toolkit.fluxcd.io/name" (which can be used to filter the objects by managed cluster name), e.g. they have these labels only:
```labels:
    cluster.x-k8s.io/cluster-name: aws-dev-XXXXX
    cluster.x-k8s.io/control-plane: ""
    cluster.x-k8s.io/control-plane-name: aws-dev-XXXXX-cp
```
MachineDeployment object has "helm.toolkit.fluxcd.io/name", but MachineDeployment is removed immediately when Helm Release started deletion process (but Machines are still in the process of removal).
Should we apply labels on Machines during reconciliation and then use such labels watch them?

Kshatrix: You can get the name of the cluster using the flux label. Then use this name to query all associated machines.

Kshatrix (dismissed):

templates/aws-standalone-cp/templates/awscluster.yaml
squizzi: I just encountered this behavior in test, the `AWSMachine` objects still exist even with the `AWSCluster` gone:

```
k get awsmachines.infrastructure.cluster.x-k8s.io  -n hmc-system
NAME                              CLUSTER            STATE     READY   INSTANCEID                              MACHINE
squizzi-e2e-test-cp-0             squizzi-e2e-test   running   true    aws:///us-west-1a/i-003dbb531c5686cad   squizzi-e2e-test-cp-0
squizzi-e2e-test-md-xjskf-bg9hf   squizzi-e2e-test   running   true    aws:///us-west-1a/i-0c12931c54ef9d1db   squizzi-e2e-test-md-xjskf-bg9hf
```

```
k get -n hmc-system awscluster
No resources found in hmc-system namespace.
```

squizzi (approved):

a13x5 (approved):
I have only minor remarks. 
I think we should put all this to test and fix if any of my suspicions will prove to be right.

internal/controller/managedcluster_controller.go
```diff
@@ -385,14 +408,111 @@ func (r *ManagedClusterReconciler) Delete(ctx context.Context, l logr.Logger, ma
...
+
+	return &itemsList.Items[0], nil
+}
+
+func (r *ManagedClusterReconciler) removeClusterFinalizer(ctx context.Context, cluster *metav1.PartialObjectMetadata) error {
+	originalCluster := *cluster
+	finalizersUpdated := controllerutil.RemoveFinalizer(cluster, hmc.BlockingFinalizer)
+	if finalizersUpdated {
+		log.FromContext(ctx).Info("Allow to stop cluster", "finalizer", hmc.BlockingFinalizer)
+		if err := r.Client.Patch(ctx, cluster, client.MergeFrom(&originalCluster)); err != nil {
```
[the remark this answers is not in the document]
slysunkin: Patch was recommended here: https://github.com/Mirantis/hmc/pull/242#discussion_r1741054646

templates/cluster/aws-hosted-cp/templates/awscluster.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   name: {{ include "cluster.name" . }}
   annotations:
     cluster.x-k8s.io/managed-by: k0smotron
+  finalizers:
+  - hmc.mirantis.com/cleanup
```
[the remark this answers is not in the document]
slysunkin: @squizzi tested this functionality as well

internal/controller/managedcluster_controller.go
```diff
@@ -59,6 +58,30 @@ type ManagedClusterReconciler struct {
 	SystemNamespace string
 }
 
+type providerSchema struct {
+	machine, cluster schema.GroupVersionKind
+}
+
+var (
```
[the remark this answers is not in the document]
slysunkin: ​In Go, constants can only be fixed, literal values.

templates/cluster/aws-hosted-cp/templates/awscluster.yaml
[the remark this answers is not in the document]
squizzi: I didn't test it, I just thought this would fix the behavior since I encountered it on standalone in test.

internal/controller/managedcluster_controller.go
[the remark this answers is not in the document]
Kshatrix: you can't use Update with PartialObjectMetadata

templates/cluster/aws-hosted-cp/templates/awscluster.yaml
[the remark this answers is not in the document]
Kshatrix: Do you suspect some controller to remove it accidentally? I briefly checked capa, and they seem to add/remove only `awscluster.infrastructure.cluster.x-k8s.io` without affecting anything else

Kshatrix (approved):

PR #256, opened 2024-09-04
Add vSphere provider template

### LINKED ISSUE #252: VMware provider template for HMC

a13x5:
Resolves #252


a13x5:
Note: I added default config with empty values since they're all mandatory, but they're only used for `clusterctl cluster generate` commands, so they're irrelevant for the controller which will be using `VSphereClusterIdentity` anyway.
This is just how capi-operator works

Kshatrix (approved):

Issue #249, opened 2024-09-03
Do not attempt Helm install until CAP provider operators are installed/running/ready


squizzi:
Currently HMC will try to install a Helm template for a product that relies on a CRD that does not yet exist, we should perform some  status checks to verify the provider templates we support are ready/running and provide that status rather than the Helm installation failure that may be ambiguous to an end user.

For example, the following indicates the `AWSCluster` Kind does not yet exist because CAPA has not yet been installed to provide those CRDs.

```
k get deployments.hmc.mirantis.com -n hmc-system
NAME      READY   STATUS
aws-dev   False   Helm install failed for release hmc-system/aws-dev with chart aws-standalone-cp@0.1.2: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "aws-dev" namespace: "" from "": no matches for kind "AWSCluster" in version "infrastructure.cluster.x-k8s.io/v1beta2"...
```

eromanova:
Already been implemented in https://github.com/Mirantis/hmc/pull/487. Now, in case the required provider is not deployed yet, or failed to be deployed, the cluster template will be marked as invalid. The admission webhook will block the `ManagedCluster` creation since the template is invalid.

PR #257, opened 2024-09-04
Improve ManagedCluster validation

### LINKED ISSUE #236: [Admission] Verify requested providers readiness before admitting Deployment creation
Verification base on the Management object status

### LINKED ISSUE #244: [Admission] Improve ManagedCluster validation
We should check the referenced template validity on ManagedCluster Update only if the template was changed in ManagedCluster's spec (compare the old `spec.template` with the new `spec.template` value). 

Currently, the admission webhook will block any update to the ManagedCluster object if the referenced template became invalid for some reason. But we should still allow changing other fields except for the `spec.template` (e.g. `spec.config`).

eromanova:
Closes #236 #249

1. Verify core components' and required providers' readiness on ManagedCluster creation according to the Management object status
3. Adapt testing cases accordingly

### DIFF
--- a/api/v1alpha1/management_types.go
+++ b/api/v1alpha1/management_types.go
@@ -86,6 +86,13 @@ func (m *ManagementSpec) SetProvidersDefaults() error {
 	return nil
 }
 
+func (m *ManagementSpec) GetCoreTemplates() map[string]bool {
+	return map[string]bool{
+		m.Core.HMC.Template:  true,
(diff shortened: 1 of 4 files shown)


squizzi:
This looks like it will also address #249

Kshatrix:
We have some updates to the Management CR planned. Let's get back to this issue after we land them.

eromanova:
Closing as redundant

PR #284, opened 2024-09-06
Remove dependency on hmc-system namespace

### LINKED ISSUE #283: Remove dependency on hmc-system namespace

Kshatrix:
Closes #283

### DIFF
--- a/api/v1alpha1/management_types.go
+++ b/api/v1alpha1/management_types.go
@@ -30,8 +30,7 @@ const (
         }
 	}`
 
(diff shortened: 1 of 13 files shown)


eromanova (approved):
LGTM. But we still depend on the `hmc-system` namespace in the cert-manager (`global.leaderElection.namespace`). If the HMC will be installed in another namespace this parameter should also be overwritten. Or should be reworked if possible.

eromanova (approved):

PR #285, opened 2024-09-06
Vet doc files

zerospiel:
Tidy up all the *.md files according to common lint rules

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ### TLDR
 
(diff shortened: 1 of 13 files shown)


a13x5:
What linter was used? Lazy ordered lists are atrocious when you want to actually refer step numbers. I don't see it in our case, but if it will be enforced I would prefer normal ordering.

zerospiel:
@a13x5 the [rules](https://github.com/DavidAnson/vscode-markdownlint?tab=readme-ov-file#rules) provided by the VSCode extension have been followed

PR #287, opened 2024-09-09
Split Templates in multiple types

### LINKED ISSUE #267: Split templates in multiple types
Having a single Template resource appears to be confusing for many people. Therefore, we need to split the existing Template resource into multiple resources representing each template type respectively:
* ClusterTemplate (ns scope)
* ServiceTemplate (ns scope)
* ProviderTemplate (provider and core templates together for now. We will potentially migrate core into a separate Release CR) (cluster scope)

eromanova:
1. Split templates into 3 types:
* `ClusterTemplates` (ns scoped) - Templates for `ManagedCluster` deployments
* `ServiceTemplates` (ns scoped) - Templates for services
* `ProviderTemplates` (cluster scoped) - Templates for Core components and Provider deployments

2. All Templates are reconciled in the same way
3. Deprecated Template `type` parameter from spec and status
4. Templates are now located in different folders (`template/cluster`, `templates/provider`, `templates/service`) depending on the template kind
5. Adapted testing accordingly

Closes #267


slysunkin (approved):

api/v1alpha1/template_types.go
```diff
@@ -37,28 +39,32 @@ const (
...
-	// TemplateTypeProvider is the type used for adding CAPI providers in the HMC Management object.
-	TemplateTypeProvider TemplateType = "provider"
-	// TemplateTypeCore is the type used for HMC and CAPI core components
-	TemplateTypeCore TemplateType = "core"
-)
+// +kubebuilder:object:generate=false
+// +k8s:deepcopy-gen=false
+
+// Template is the interface defining a list of methods to interact with templates
+type Template interface {
```
Kshatrix, line 48: Let's keep this interface on the consumer side

api/v1alpha1/template_types.go
```diff
@@ -37,28 +39,32 @@ const (
...
+	Get() TemplateState
+	UpdateStatus(context.Context, crclient.Client, TemplateStatusMixin) error
```
Kshatrix, line 50: I don't think we need it. We could rather use crclient.Object interface here.

api/v1alpha1/template_types.go
```diff
@@ -37,28 +39,32 @@ const (
...
+}
+
+// +kubebuilder:object:generate=false
+// +k8s:deepcopy-gen=false
 
-// TemplateSpec defines the desired state of Template
-type TemplateSpec struct {
+type TemplateState struct {
```
Kshatrix, line 56: I don't think we need a dedicated struct. Let's add getters that return pointers to status/spec instead.

api/v1alpha1/template_types.go
```diff
@@ -37,28 +39,32 @@ const (
...
+	metav1.ObjectMeta `json:",inline"`
+	Kind              TemplateKind        `json:"kind"`
```
Kshatrix, line 58: GetObjectKind() will be accessible, so there's no need to have it separately

internal/controller/template_controller.go
```diff
@@ -119,98 +171,90 @@ func (r *TemplateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	if templateState.Spec.Helm.ChartRef != nil {
 		// HelmChart is not managed by the controller
 		return nil, nil
 	}
 	helmChart := &sourcev1.HelmChart{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      template.Name,
-			Namespace: template.Namespace,
+			Name:      templateState.ObjectMeta.Name,
+			Namespace: r.SystemNamespace,
```
Kshatrix, line 257: I think we want to use the namespace of associated Cluster/ServiceTemplate object here. SystemNamespace is only for ProviderTemplate (when Namespace is empty)

Kshatrix (changes requested):

Kshatrix (approved):

PR #291, opened 2024-09-10
Disable installBeachHeadServices by default on `aws-standalone-cp` template

wahabmk:
## Description
Disabling installBeachHeadServices by default on `aws-standalone-cp` template to address https://github.com/Mirantis/hmc/issues/286.

### DIFF
--- a/config/dev/aws-managedcluster.yaml
+++ b/config/dev/aws-managedcluster.yaml
@@ -13,4 +13,5 @@ spec:
     worker:
       instanceType: t3.small
     workersNumber: 1
+    installBeachHeadServices: false
   template: aws-standalone-cp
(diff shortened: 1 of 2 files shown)


eromanova (commented):
LGTM. But can we also consider making `HelmRepository` (for cert-manager) and `OCIRepository` (for nginx) names dependent on the cluster name to avoid this situation if more than one cluster should be deployed with `installBeachHeadServices` enabled?

a13x5 (approved):
LGTM. There is some additional work required on the feature, but I guess this will suffice for now to remove the issue.

Kshatrix (approved):
Let's no focus on the current approach too much as it's going to be reimplemented.

PR #280, opened 2024-09-05
Add support for testing `aws-hosted-cp`

squizzi:
Note: This PR was intentionally created on a repository branch instead of my fork so we can iterate on the e2e testing workflow and then change it to `pull_request_target` for merge.

---

* Break KubeClient helpers into provider specific file
* Finish aws-hosted-cp test and add comments through test to make it
  easier to understand.
* Use GinkgoHelper across e2e tests, populate hosted vars from
  AWSCluster.
* No longer rely on local registry for images in test/e2e.
* Support OS for awscli install.
* Prepend hostname to collected log artifacts.
* Support no cleanup of provider specs, differentiate ci
  cluster names.
* Add docs on running tests, do not wait for all providers
  if configured.
* Reinstantiate resource validation map on each instance of
  validation.
* Enable the external-gc feature via annotation, featureGate
  bool. (Closes: https://github.com/Mirantis/hmc/issues/152)
* Bump aws-*-cp templates to 0.1.3
* Bump cluster-api-provider-aws template to 0.1.2
* Improve test logging to log template name and validation
  phase.
* Bump k0s version to v1.30.4+k0s.0, set CCM nodeSelector to
  null for aws-hosted-cp. (Closes: https://github.com/Mirantis/hmc/issues/290)
* Break cleanup into seperate job so that it is unaffected by
  concurrency group cancellations.

Closes: #212

PR #300, opened 2024-09-12
Implement Azure e2e template tests

### LINKED ISSUE #210: E2e verification of Azure templates
Should be based on #201

kylewuolle:
Implement Azure template e2e tests. Closes #210.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -143,7 +143,7 @@ var _ = Describe("controller", Ordered, func() {
 				}
 
 				By("creating a Deployment")
-				d := managedcluster.GetUnstructured(managedcluster.ProviderAWS, template)
+				d := managedcluster.GetUnstructured(managedcluster.ProviderAWS, template, namespace)
 				clusterName = d.GetName()
 
 				deleteFunc, err = kc.CreateManagedCluster(context.Background(), d)
@@ -151,7 +151,7 @@ var _ = Describe("controller", Ordered, func() {
 
 				By("waiting for infrastructure providers to deploy successfully")
 				Eventually(func() error {
-					return managedcluster.VerifyProviderDeployed(context.Background(), kc, clusterName)
+					return managedcluster.VerifyProviderDeployed(context.Background(), kc, clusterName, managedcluster.ProviderAWS)
 				}).WithTimeout(30 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
 
 				By("verify the deployment deletes successfully")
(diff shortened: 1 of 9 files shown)


squizzi:
You'll need to reopen this with a repository branch and modify the test to run as `pull_request` or just rebase off my PR, #280 since there's a bunch of test related fixes in there.  Then you'll be able to see if the tests pass and iterate on them before we merge it in as `pull_request_target`.  This is the best way I can think of to do this until we have more time to figure out the secrets stuff in actions.

test/kubeclient/kubeclient.go
```diff
@@ -123,6 +131,75 @@ func new(configBytes []byte, namespace string) (*KubeClient, error) {
...
+			return fmt.Errorf("failed to create dynamic client: %w", err)
+		}
+
+		exists, err := dc.Get(ctx, credentialResource.GetName(), metav1.GetOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to check for existing credential: %w", err)
+		}
+
+		if exists == nil {
+			if _, err = dc.Create(ctx, credentialResource, metav1.CreateOptions{}); err != nil {
```
squizzi, line 194: You don't even need to call `.Get` prior to `.Create` you can just call `.Create` and then check for `apierrors.IsAlreadyExists(err)` and return `nil` in that case.

test/kubeclient/kubeclient.go
```diff
@@ -123,6 +131,75 @@ func new(configBytes []byte, namespace string) (*KubeClient, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to fetch group resources: %w", err)
+	}
+
+	yamlReader := yamlutil.NewYAMLReader(bufio.NewReader(bytes.NewReader(yamlFile)))
+	for {
+		yamlDoc, err := yamlReader.Read()
+
+		if err != nil {
+			if err == io.EOF {
```
squizzi, line 158:
```suggestion
			if errors.Is(err, io.EOF) {
```

test/kubeclient/kubeclient.go
```diff
@@ -123,6 +131,75 @@ func new(configBytes []byte, namespace string) (*KubeClient, error) {
...
 
+func (kc *KubeClient) CreateAzureCredentialsKubeSecret(ctx context.Context) error {
+	serializer := yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
+	yamlFile, err := os.ReadFile("./config/dev/azure-credentials.yaml")
+
+	if err != nil {
+		return fmt.Errorf("failed to read azure credential file: %w", err)
+	}
+
+	yamlFile, err = envsubst.Bytes(yamlFile)
```
squizzi, line 142: If you're going to perform substitutions on the file it should probably be a `.tpl` file.  Perhaps create an `resources/azure-credentials.yaml.tpl` and embed it above and then call `envsubst` on it.

test/kubeclient/kubeclient.go
```diff
@@ -123,6 +131,75 @@ func new(configBytes []byte, namespace string) (*KubeClient, error) {
 	}, nil
 }
 
+func (kc *KubeClient) CreateAzureCredentialsKubeSecret(ctx context.Context) error {
+	serializer := yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
+	yamlFile, err := os.ReadFile("./config/dev/azure-credentials.yaml")
```
squizzi, line 136: Instead of calling `os.ReadFile` on a path that might change use `go:embed` to embed the file at runtime, see: https://gobyexample.com/embed-directive you can embed directly into `[]byte` and then manipulate it as needed below.

test/managedcluster/managedcluster.go
```diff
@@ -50,16 +52,22 @@ var awsStandaloneCPManagedClusterTemplateBytes []byte
...
 }
 
 // GetUnstructured returns an unstructured ManagedCluster object based on the
 // provider and template.
-func GetUnstructured(provider ProviderType, templateName Template) *unstructured.Unstructured {
+func GetUnstructured(provider ProviderType, templateName Template, namespace string) *unstructured.Unstructured {
 	GinkgoHelper()
 
-	generatedName := uuid.New().String()[:8] + "-e2e-test"
+	generatedName := "e2etest-" + uuid.New().String()[:8]
```
squizzi, line 70: Note that this name doesn't align with the one ran in CI, if you want to change that name to make sure you modify `MANAGED_CLUSTER_NAME` in the actions workflow when you rebase.

test/managedcluster/validate_deleted.go
```diff
@@ -41,7 +43,7 @@ func VerifyProviderDeleted(ctx context.Context, kc *kubeclient.KubeClient, clust
 func validateClusterDeleted(ctx context.Context, kc *kubeclient.KubeClient, clusterName string) error {
 	// Validate that the Cluster resource has been deleted
 	cluster, err := kc.GetCluster(ctx, clusterName)
-	if err != nil {
+	if err != nil && !apierrors.IsNotFound(err) {
```
squizzi, line 46: 👍

squizzi (changes requested):

test/e2e/e2e_test.go
```diff
@@ -163,6 +163,68 @@ var _ = Describe("controller", Ordered, func() {
...
+				By("deleting the deployment")
+				err = deleteFunc()
+				Expect(err).NotTo(HaveOccurred())
+			}
+
+		})
+
+		for _, template := range []managedcluster.Template{
+			managedcluster.TemplateAzureStandaloneCP,
+			managedcluster.TemplateAzureHostedCP,
```
a13x5, line 203: You're creating hosted cluster template in the local kind cluster. It should fail. And if it's not - the test should be reworked

test/kubeclient/kubeclient.go
```diff
@@ -123,6 +131,75 @@ func new(configBytes []byte, namespace string) (*KubeClient, error) {
 	}, nil
 }
 
+func (kc *KubeClient) CreateAzureCredentialsKubeSecret(ctx context.Context) error {
```
a13x5, line 134: It looks more complex than it should be. @squizzi already proposed several fixes and I tend to agree with him

a13x5 (changes requested):
The hosted test looks incorrect, should be rechecked and reworked. 
Since hosted cluster should be create on cluster which is deployed on Azure. Also artifacts should be passed there somehow. I don't see any of that. 
Probably you should wait for #280 to be merged and then build hosted CP tests on top of that.

Issue #246, opened 2024-09-03
[Admission] Forbid to remove the Template object if one or more ManagedCluster references it


eromanova:

slysunkin:
I'd like to clarify a few details.

We have the following hierarchy:
for AWS
```
MachineDeployment
  - K0sWorkerConfigTemplate
  - AWSMachineTemplate
```

for Azure
```
MachineDeployment
  - K0sWorkerConfigTemplate
  - AzureMachineTemplate
```

So, we need to wait for MachineDeployment termination before K0sWorkerConfigTemplate, AWSMachineTemplate/AzureMachineTemplate (respectively to provider type).

eromanova:
@slysunkin It will be enough to check the `ManagedCluster` existence that references (`spec.template`) the Template that was triggered for the removal and forbid it if at least one `ManagedCluster` is found. Similarly, it's currently done for Management object removal: https://github.com/Mirantis/hmc/blob/main/internal/webhook/management_webhook.go#L65 No need to check internal cluster objects.

PR #301, opened 2024-09-12
Prevent ClusterTemplate of being removed if ManagedCluster exists

slysunkin:
## Description

Added code for blocking the removal of ClusterTemplate if ManagedCluster exists.

Implements #246


internal/webhook/template_webhook.go
```diff
@@ -55,7 +62,27 @@ func (*ClusterTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
...
-func (*ClusterTemplateValidator) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
+func (v *ClusterTemplateValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	template, ok := obj.(*v1alpha1.ClusterTemplate)
+	if !ok {
+		return admission.Warnings{"Wrong object"}, apierrors.NewBadRequest(fmt.Sprintf("expected ClusterTemplate but got a %T", obj))
+	}
+
+	managedClusters := &v1alpha1.ManagedClusterList{}
+	opts := &client.ListOptions{
+		Namespace: template.Namespace,
```
eromanova, line 73: You should list ManagedClusters across all namespaces, not just the template namespace. ManagedClusters can be created in any namespace but should reference the templates from the system namespace.

internal/webhook/template_webhook_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/Mirantis/hmc/test/objects/template"
+	"github.com/Mirantis/hmc/test/scheme"
+)
+
+func TestClusterTemplateValidateDelete(t *testing.T) {
+	g := NewWithT(t)
+
+	ctx := context.Background()
+
+	tpl := template.NewClusterTemplate()
```
eromanova, line 37: You also can pass the template parameters like this:
```
tpl := template.NewClusterTemplate(template.WithName("testTemplateFail"), template.WithNamespace("testNamespaceFail"))
```
But it's up to you.

internal/webhook/template_webhook.go
```diff
@@ -55,7 +62,24 @@ func (*ClusterTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
...
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
-func (*ClusterTemplateValidator) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
+func (v *ClusterTemplateValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	template, ok := obj.(*v1alpha1.ClusterTemplate)
+	if !ok {
+		return admission.Warnings{"Wrong object"}, apierrors.NewBadRequest(fmt.Sprintf("expected ClusterTemplate but got a %T", obj))
+	}
+
+	managedClusters := &v1alpha1.ManagedClusterList{}
```
Kshatrix, line 71: we should add a FieldIndexer to index over the `.spec.template` field instead of iterating over all existing clusters ( `mgr.GetFieldIndexer().IndexField()`)

Kshatrix (changes requested):

internal/webhook/template_webhook.go
slysunkin: During the fixing of unit tests I've found a case where indexing will not work:
1. cluster template exists (because it is created first)
2. managed cluster doesn't exist (for any reason, i.e. ProviderTemplateReconciler fails or as in internal/controller/template_controller_test.go where only templates are created)
3. cluster template deletion will fail due to "Index with name field..." error

internal/webhook/template_webhook.go
```diff
@@ -55,7 +64,31 @@ func (*ClusterTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
...
+	listOptions := client.ListOptions{
```
eromanova, line 74: Currently, `ManagedCluster` may refer the `ClusterTemplate` from the system namespace only. It will be changed soon but until then, let's validate the removal of the `ClusterTemplates` in the system namespace only.

internal/webhook/template_webhook.go
```diff
@@ -55,7 +64,31 @@ func (*ClusterTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
...
+		FieldSelector: fields.SelectorFromSet(fields.Set{v1alpha1.TemplateKey: template.Name}),
+		Limit:         1,
+	}
+	err := v.Client.List(ctx, managedClusters, &listOptions)
+	if err != nil {
+		if err.Error() == "Index with name field:.spec.template does not exist" {
+			// ManagedCluster may not be indexed because of the incomplete reconciliation
+			log.FromContext(ctx).Error(err, "ClusterTemplateValidator")
+			return nil, nil
+		}
```
eromanova, line 84: Is it correct that we allow the `ClusterTemplate` deletion in this case? What if this error occurred but there are still `ManagedClusters` that should prevent the template from being removed? I think it should be handled in a different way.

eromanova (changes requested):

internal/webhook/template_webhook.go
slysunkin: We may have a workaround to remove the templates iterating over all ManagedClusters.
In fact, I'l not sure what we can do if we have a half-initialized system (where the problem occurred somewhere between template creation and proper initialization of a Manager).

internal/webhook/template_webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+			existingObjects: []runtime.Object{managedcluster.NewManagedCluster()},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := fake.NewClientBuilder().
+				WithScheme(scheme.Scheme).
+				WithRuntimeObjects(tt.existingObjects...).
+				WithIndex(tt.existingObjects[0], v1alpha1.TemplateKey, controller.ExtractTemplateName).
```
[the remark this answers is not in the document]
slysunkin: If we pass a wrong object then line 519 of [internal/controller/managedcluster_controller.go](https://github.com/Mirantis/hmc/pull/301/files#diff-f79d441347dc2389fb8d9662097d55099d962aee056e150a45bcf767a9d3445c) will handle this. It will be a test for that part of code.

internal/webhook/template_webhook.go
slysunkin: I can, but previously it was requested to search across all namespaces https://github.com/Mirantis/hmc/pull/301#discussion_r1760023794

internal/webhook/template_webhook.go
Kshatrix: You have this "Index with name field:.spec.template does not exist" error because you haven't added this indexer to the manager used in the tests.

internal/controller/managedcluster_controller.go
```diff
@@ -514,8 +514,24 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
...
+	}
+	if cluster.Spec.Template == "" {
+		return []string{}
+	}
+	return []string{cluster.Spec.Template}
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *ManagedClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetFieldIndexer().
```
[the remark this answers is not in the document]
eromanova: But just to confirm since I'm still in doubt. We should register the FieldIndexer not only for the ManagedCluster reconciler but also for the Webhook. In case, we register it here will it work in the admission webhook code?

internal/webhook/template_webhook.go
eromanova: Sorry, I probably wasn't clear enough. These suggestions are a little bit different. In https://github.com/Mirantis/hmc/pull/301#discussion_r1760023794 I meant to list `ManagedClusters` across all namespaces while trying to find clusters that refer to the `ClusterTemplate` planned to be removed. But in this comment, I'm suggesting validating the removal of `ClusterTemplate` objects in the system namespace only. Since now, all templates that are not in the system namespace can't be referenced and thus, can be removed without any extra validation. But it's not that important since this logic will be changed soon (after #317  `ManagedCluster` will be allowed to refer to the `ClusterTemplate` from the same namespace only), so, you can ignore it.

internal/webhook/template_webhook.go
Kshatrix: it's in `suite_test.go`
If you add the indexer here, you will no longer have this error in tests.

internal/controller/managedcluster_controller.go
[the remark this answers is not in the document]
Kshatrix: The indexer is registered for the manager, not for the reconciler, so it doesn't really matter where we do that. Logically it makes sense to keep it near the component that is supposed to use it (webhook in our case) or right after we set up the manager.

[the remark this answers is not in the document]
slysunkin: Thanks for the suggestions. "Reuse" is a great argument. I reworked this functionality.

internal/webhook/template_webhook.go
```diff
@@ -55,7 +65,26 @@ func (*ClusterTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
...
+		FieldSelector: fields.SelectorFromSet(fields.Set{TemplateKey: template.Name}),
+		Limit:         1,
+	}
+	err := v.Client.List(ctx, managedClusters, &listOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(managedClusters.Items) > 0 {
+		return admission.Warnings{"The ClusterTemplate object can't be removed if ManagedCluster objects still exist"}, ErrTemplateDeletionForbidden
```
Kshatrix, line 85:
```suggestion
		return admission.Warnings{"The ClusterTemplate object can't be removed if ManagedCluster objects referencing it still exist"}, ErrTemplateDeletionForbidden
```

Kshatrix (changes requested):

Kshatrix (approved):

eromanova (approved):

PR #303, opened 2024-09-15
ClusterTemplate and ServiceTemplate distribution system

### LINKED ISSUE #272: ClusterTemplate distribution

### LINKED ISSUE #273: ServiceTemplate distribution

### LINKED ISSUE #315: [admission] Validate TemplateManagement Update
Restrict the `TemplateManagement` update if any `ManagedCluster` references the `ClusterTemplate` which is not present in the access rules of the updated `TemplateManagement` specification.

eromanova:
This PR introduces the ClusterTemplate and ServiceTemplate distribution system.

The TemplateManagement controller:
1. Watches for the TemplateManagement object
2. Gets all existing ClusterTemplates and ServiceTemplates with the `hmc.mirantis.com/managed: true` label across all namespaces (gets actual state)
3. Parses the access rules defined in the TemplateManagement spec, reads the template chains, and gets the expected Templates state
4. Creates missing Cluster and Service Templates that reference helm charts in the system namespace
5. Deletes Cluster and Service Templates from the namespaces if it's no longer present in the access rule

Also:
1. Reconcile default HelmRepository for each Template not managed by the templates distribution system
2. Added `TemplateManagement` update validation (checks that new access rules do not affect in-used `ClusterTemplates`

Closes #272
Closes #273
Closes #315


internal/controller/templatemanagement_controller.go
```diff
@@ -48,7 +62,282 @@ func (r *TemplateManagementReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			return err
+		}
+	} else {
+		if target.GetLabels()[hmc.HMCManagedLabelKey] == hmc.HMCManagedLabelValue {
+			l.Info("Template is not managed by HMC, skipping an update", "namespace", targetNamespace, "name", name)
+			return nil
+		}
+	}
+
+	_, err = ctrl.CreateOrUpdate(ctx, r.Client, target, func() error {
```
Kshatrix: templates are immutable, so we are limited to create/delete operations

internal/controller/templatemanagement_controller.go
```diff
@@ -48,7 +63,306 @@ func (r *TemplateManagementReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		return []string{}, err
+	}
+	result := make([]string, len(namespaces.Items))
+	for i, ns := range namespaces.Items {
+		result[i] = ns.Name
+	}
+	return result, nil
+}
+
+func (r *TemplateManagementReconciler) createTemplates(ctx context.Context, state map[string]map[string]bool, kind string) error {
```
Kshatrix, line 255: we might need to change the state structure here. if we need to distribute one template across x namespaces, we have to make x copyTemplate calls. I think it makes sense to inverse it and keep Templates as keys and two lists/a map of namespaces as values.

internal/controller/templatemanagement_controller.go
```diff
@@ -48,9 +54,198 @@ func (r *TemplateManagementReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		helmRepo := &sourcev1.HelmRepository{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: ns,
+				Name:      defaultRepoName,
+				Labels: map[string]string{
+					hmc.HMCManagedLabelKey: hmc.HMCManagedLabelValue,
+				},
+			},
+		}
+		err := r.Get(ctx, types.NamespacedName{
```
Kshatrix, line 128: for helmrepository in contrary to templates we can use createourupdate :D but let's decide if we want to distribute these resources or keep them all in the system namespace first

internal/controller/templatemanagement_controller.go
```diff
@@ -48,9 +54,156 @@ func (r *TemplateManagementReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 		return ctrl.Result{}, err
 	}
+
+	defer func() {
+		statusErr := ""
+		if err != nil {
+			statusErr = err.Error()
+		}
+		templateMgmt.Status.Error = statusErr
+		err = errors.Join(err, r.updateStatus(ctx, templateMgmt))
```
Kshatrix, line 64: I think ObservedGeneration should be moved here

templates/provider/hmc-templates/files/templates/azure-hosted-cp.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: hmc.mirantis.com/v1alpha1
 kind: ClusterTemplate
 metadata:
   name: azure-hosted-cp
+  labels:
```
Kshatrix, line 5: I don't think we want to label those in the system namespace as managed. Only distributed once.

internal/webhook/templatemanagement_webhook.go
```diff
@@ -0,0 +1,126 @@
...
+	var errs []error
+	for templateName, namespaces := range expectedState.ClusterTemplatesState {
+		for namespace, keep := range namespaces {
+			if !keep {
+				managedClusters, err := getManagedClustersForTemplate(ctx, v.Client, namespace, templateName)
+				if err != nil {
+					return nil, err
+				}
+
+				if len(managedClusters) > 0 {
```
Kshatrix, line 84: I think we can use warnings to not overpopulate errors with clusters

Kshatrix (changes requested):

Kshatrix (approved):

PR #331, opened 2024-09-17
Add user-facing RBAC roles

### LINKED ISSUE #305: Add RBAC roles associated with HMC management

Kshatrix:
Resolves #305 #306


templates/provider/hmc/templates/rbac/controller/roles.yaml
```diff
@@ -55,13 +39,6 @@ rules:
   resources:
   - managements
   verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
```
eromanova, line 64:
```
{{ include "rbac.editorVerbs" . | nindent 4 }}
```
is missing?

Kshatrix: no, editorVerbs is wider than needed in this case

eromanova: But we should still allow the controller to get/watch/update/etc. `Management` object. Now, the `verbs` are empty. Or am I missing something?

Kshatrix: ah, this one is definitely missing, thanks

eromanova (approved):

PR #340, opened 2024-09-17
Added reenabling of the createManagement parameter check

slysunkin:
## Description
Added code for rejecting of reenabling the createManagement parameter in Management HMC spec.

Implements https://github.com/Mirantis/hmc/issues/245

### DIFF
--- a/internal/webhook/management_webhook.go
+++ b/internal/webhook/management_webhook.go
@@ -34,7 +34,8 @@ type ManagementValidator struct {
 }
 
 var (
-	ErrManagementDeletionForbidden = errors.New("management deletion is forbidden")
+	ErrManagementDeletionForbidden         = errors.New("management deletion is forbidden")
+	ErrCreateManagementReenablingForbidden = errors.New("reenabling of the createManagement parameter is forbidden")
 )
 
 func (in *ManagementValidator) SetupWebhookWithManager(mgr ctrl.Manager) error {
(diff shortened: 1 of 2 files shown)


internal/webhook/management_webhook.go
```diff
@@ -57,7 +58,35 @@ func (*ManagementValidator) ValidateCreate(_ context.Context, _ runtime.Object)
...
+
+	// Check if the 'controller' field exists and is a map.
+	controller, ok := vals["controller"].(map[string]interface{})
+	if !ok {
+		return nil, nil
+	}
+
+	// Check if 'createManagement' exists and is true.
+	if createManagement, ok := controller["createManagement"].(bool); ok && createManagement {
+		return admission.Warnings{"The Management object can't reenable createManagement"}, ErrCreateManagementReenablingForbidden
```
eromanova, line 87: I don't think we need a warning here since this is already covered in an error message.

internal/webhook/management_webhook_test.go
```diff
@@ -75,3 +77,89 @@ func TestManagementValidateDelete(t *testing.T) {
...
+
+	tests := []struct {
+		name            string
+		management      *v1alpha1.Management
+		existingObjects []runtime.Object
+		err             string
+		warnings        admission.Warnings
+	}{
+		{
+			name:            "should fail if ManagedCluster object has controller.createManagement=true",
```
eromanova, line 127:
```suggestion
			name:            "should fail if Management object has controller.createManagement=true",
```

internal/webhook/management_webhook_test.go
```diff
@@ -75,3 +77,89 @@ func TestManagementValidateDelete(t *testing.T) {
...
+			management:      objToFail,
+			existingObjects: []runtime.Object{managedcluster.NewManagedCluster()},
```
eromanova, line 129: It's not necessary to create ManagedClusters here

eromanova (changes requested):

internal/webhook/management_webhook_test.go
```diff
@@ -75,3 +77,80 @@ func TestManagementValidateDelete(t *testing.T) {
...
+	tests := []struct {
+		name       string
+		management *v1alpha1.Management
+		err        string
+		warnings   admission.Warnings
+	}{
+		{
+			name:       "should fail if Management object has controller.createManagement=true",
+			management: objToFail,
+			warnings:   admission.Warnings{"The Management object can't reenable createManagement"},
```
eromanova, line 128: Let's remove warnings from the test spec if we don't check them eventually.

eromanova (approved):
LGTM. One small comment left.

eromanova (approved):

PR #341, opened 2024-09-17
Get status.conditions from CAPI operator during `updateComponentsStatus`

### LINKED ISSUE #235: Wait for Infrastructure/ControlPlane/BootstrapProvider readiness before reporting success
```components:
      cluster-api:
        success: true
      cluster-api-provider-aws:
        success: true
      cluster-api-provider-azure:
        success: true
      hmc:
        success: true
      k0smotron:
        success: true```

### PR DESCRIPTION
**Pull request by squizzi**
In this PR we now update our component statuses by iterating the provider CRDs affiliated with CAPI operator and returning a status error if any of the `status.conditions` are `False` or if we cannot obtain any information pertaining to status condition.  For example:

In this case `cluster-api-provider-aws` and `-azure` are not yet installed:

```
  status:
    availableProviders: {}
    components:
      cluster-api:
        success: true
      cluster-api-provider-aws:
        error: 'failed to find any provider resources affiliated with template: cluster-api-provider-aws'
      cluster-api-provider-azure:
        error: 'failed to find any provider resources affiliated with template: cluster-api-provider-azure'
      hmc:
        success: true
      k0smotron:
        success: true
```

In this case `cluster-api-provider-aws` is reporting some status conditions false:

```
  status:
    availableProviders: {}
    components:
      cluster-api:
        success: true
      cluster-api-provider-aws:
        error: 'InfrastructureProvider: aws is not yet ready, has false conditions: MinimumReplicasAvailable'
      cluster-api-provider-azure:
        success: true
      hmc:
        success: true
      k0smotron:
        success: true
```

In this case `cluster-api-provider-aws` has no `status.conditions` because something is going wrong.  Perhaps we could add some additional statuses for special circumstances; here it's because `aws-variables` does not yet exist yet, but I couldn't find anyway to assume this type of circumstance, capi-operator doesn't produce any event in this case, it only generates a log.

```
    components:
      cluster-api:
        success: true
      cluster-api-provider-aws:
        error: 'failed to get status for template: cluster-api-provider-aws: failed
          to get conditions: no status conditions found for InfrastructureProvider:
          aws'
      cluster-api-provider-azure:
        success: true
      hmc:
        success: true
      k0smotron:
        success: true
```

Closes #235

### DIFF
--- a/internal/controller/managedcluster_controller.go
+++ b/internal/controller/managedcluster_controller.go
@@ -109,7 +109,7 @@ func (r *ManagedClusterReconciler) setStatusFromClusterStatus(
 ) (bool, error) {
 	l := ctrl.LoggerFrom(ctx)
 
-	resourceConditions, err := status.GetResourceConditions(ctx, r.DynamicClient, schema.GroupVersionResource{
+	resourceConditions, err := status.GetResourceConditions(ctx, managedCluster.Namespace, r.DynamicClient, schema.GroupVersionResource{
 		Group:    "cluster.x-k8s.io",
 		Version:  "v1beta1",
 		Resource: "clusters",
--- a/internal/controller/managedcluster_controller.go
+++ b/internal/controller/managedcluster_controller.go
@@ -109,7 +109,7 @@ func (r *ManagedClusterReconciler) setStatusFromClusterStatus(
 ) (bool, error) {
 	l := ctrl.LoggerFrom(ctx)
 
-	_, _, conditions, err := status.ConditionsFromResource(ctx, r.DynamicClient, schema.GroupVersionResource{
+	resourceConditions, err := status.GetResourceConditions(ctx, r.DynamicClient, schema.GroupVersionResource{
 		Group:    "cluster.x-k8s.io",
 		Version:  "v1beta1",
 		Resource: "clusters",
@@ -124,7 +124,7 @@ func (r *ManagedClusterReconciler) setStatusFromClusterStatus(
 	}
 
 	allConditionsComplete := true
-	for _, metaCondition := range conditions {
+	for _, metaCondition := range resourceConditions.Conditions {
 		if metaCondition.Status != "True" {
 			allConditionsComplete = false
 		}
--- a/internal/controller/managedcluster_controller.go
+++ b/internal/controller/managedcluster_controller.go
@@ -33,9 +33,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
@@ -53,6 +51,7 @@ import (
 	"github.com/Mirantis/hmc/internal/helm"
 	"github.com/Mirantis/hmc/internal/sveltos"
 	"github.com/Mirantis/hmc/internal/telemetry"
+	"github.com/Mirantis/hmc/internal/utils/status"
 )
 
 const (
@@ -96,7 +95,8 @@ func (r *ManagedClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			l.Error(err, "Failed to get Management object")
 			return ctrl.Result{}, err
 		}
-		if err := telemetry.TrackManagedClusterCreate(string(mgmt.UID), string(managedCluster.UID), managedCluster.Spec.Template, managedCluster.Spec.DryRun); err != nil {
+		if err := telemetry.TrackManagedClusterCreate(
+			string(mgmt.UID), string(managedCluster.UID), managedCluster.Spec.Template, managedCluster.Spec.DryRun); err != nil {
 			l.Error(err, "Failed to track ManagedCluster creation")
 		}
 	}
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: internal/controller/management_controller.go [hunk 1]
```diff
@@ -142,6 +158,54 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
 	return ctrl.Result{}, nil
 }
 
+// checkProviderStatus checks the status of a provider associated with a given
+// ProviderTemplate name.  Since there's no way to determine resource Kind from
```
squizzi, line 162: Is there a way we could add `ProviderTypes` to our templates (providers like `k0smotron` are both infra and bootstrap providers) or extract the `Kind`'s from the `provider.yaml`'s somehow?  If so, we don't need to do this weird iteration thing.

squizzi:
I can probably add some unit tests to this using `envtest`, I just wanted to get some thoughts/eyes on before I do that.

internal/utils/conditions.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2024
```
wahabmk, line 1: Not super important but it would be nice if this was placed in package other than `internal/utils/` as it can easily become a mishmash of different functions.

wahabmk (approved):

internal/controller/management_controller.go
```diff
@@ -147,6 +163,54 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
+
+		if len(falseConditionTypes) > 0 {
+			return fmt.Errorf("%s: %s is not yet ready, has false conditions: %s", name, kind, strings.Join(falseConditionTypes, ", "))
+		}
+
+		return nil
+	}
+
+	// We've exhausted all possible provider types and found no items.
+	return fmt.Errorf("failed to find any provider resources affiliated with template: %s", providerTemplateName)
```
Kshatrix, line 211: some providers might not have any affiliated capi operator resources (currently sveltos), so we shouldn't fail here.

internal/controller/management_controller.go
```diff
@@ -147,6 +163,54 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
+			if condition.Status != v1.ConditionTrue {
+				falseConditionTypes = append(falseConditionTypes, condition.Type)
+			}
+		}
+
+		if len(falseConditionTypes) > 0 {
+			return fmt.Errorf("%s: %s is not yet ready, has false conditions: %s", name, kind, strings.Join(falseConditionTypes, ", "))
+		}
+
+		return nil
```
Kshatrix, line 207: some providettemplates have multiple providers defined (e.g. k0smotron), we should collect statuses from all

Kshatrix (changes requested):

squizzi:
I'll plan to add an e2e test in a follow-up once the controller tests have been refactored as part of #360

internal/utils/status/status.go
```diff
@@ -0,0 +1,118 @@
...
+// the name and kind of the resource and a slice of metav1.Condition.  If the
+// resource is not found, returns a ResourceNotFoundError which can be checked
+// by the caller to prevent reconciliation loops.
+//
+//nolint:revive
+func ConditionsFromResource(
+	ctx context.Context, dynamicClient dynamic.Interface,
+	gvr schema.GroupVersionResource, labelSelector string,
+) (string, string, []metav1.Condition, error) {
+	list, err := dynamicClient.Resource(gvr).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
```
zerospiel, line 86: should we limit the `list` call to `2`? just to avoid unnecessarily machinery during the runtime

internal/controller/management_controller.go
```diff
@@ -128,6 +134,16 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
 			errs = errors.Join(errs, errors.New(errMsg))
 			continue
 		}
+
+		if component.Template != hmc.CoreHMCName {
+			if err := r.checkProviderStatus(ctx, component.Template); err != nil {
+				errMsg := err.Error()
+				updateComponentsStatus(detectedComponents, &detectedProviders, component.Template, template.Status, errMsg)
+				errs = errors.Join(errs, errors.New(errMsg))
```
zerospiel, line 142:
```suggestion
				errs = errors.Join(errs, err)
```

internal/utils/status/status.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return kind, name, conditions, nil
+}
+
+func ObjKindName(unstrObj *unstructured.Unstructured) (kind string, name string) {
+	kind = unstrObj.GetKind()
+	name = unstrObj.GetName()
+
+	return kind, name
```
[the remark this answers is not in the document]
squizzi: The linter didn't like that so I changed it to this, it was that before.

internal/controller/management_controller.go
```diff
@@ -145,6 +161,57 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
+		"controlplaneproviders",
+		"bootstrapproviders",
+	} {
+		gvr := schema.GroupVersionResource{
+			Group:    "operator.cluster.x-k8s.io",
+			Version:  "v1alpha2",
+			Resource: resourceType,
+		}
+
+		name, kind, conditions, err := status.ConditionsFromResource(ctx, r.DynamicClient, gvr,
```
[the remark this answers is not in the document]
squizzi: doh

internal/utils/status/status.go
```diff
@@ -0,0 +1,118 @@
...
+// the provided GroupVersionResource and labelSelector.  The function returns
+// the name and kind of the resource and a slice of metav1.Condition.  If the
+// resource is not found, returns a ResourceNotFoundError which can be checked
+// by the caller to prevent reconciliation loops.
+//
+//nolint:revive
+func ConditionsFromResource(
+	ctx context.Context, dynamicClient dynamic.Interface,
+	gvr schema.GroupVersionResource, labelSelector string,
+) (string, string, []metav1.Condition, error) {
```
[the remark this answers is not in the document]
squizzi: I've always been told named returns are a no-no, but I'm happy to add.

zerospiel (approved):

squizzi:
@zerospiel I had to rebase and it dismissed the review... mind reviewing again?

zerospiel:
@squizzi it seems that the last revision downloaded has erased the changes you made after the review, please bring the right revision back

squizzi:
@zerospiel good catch and sorry about that, I've fixed the rebase issues.

zerospiel (commented):
LGTM, please rebase if required

squizzi:
Ugh I used GitHub's update branch button, that was a mistake, I'll get rid of that merge commit.

slysunkin (approved):

internal/utils/status/status.go
```diff
@@ -0,0 +1,117 @@
...
```
a13x5, line 85: Function signature looks confusing to me tbh. Structs in the arguments and in the return statement would be much clearer. 
As an option the list op could be done at `checkProviderStatus` and here you can do only basic checks on `metav1.List`

a13x5 (dismissed):
Some resolved conversations aren't resolved for some reason.
Several other notes, mostly minor

internal/utils/status/status.go
```diff
@@ -0,0 +1,117 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+)
+
+// ConditionsFromUnstructured fetches all of the status.conditions from an
+// unstructured object and returns them as a slice of metav1.Condition.  The
+// status.conditions field is expected to be a slice of map[string]any
+// which can be cast into a metav1.Condition.
+func ConditionsFromUnstructured(unstrObj *unstructured.Unstructured) ([]metav1.Condition, error) {
```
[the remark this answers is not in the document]
squizzi: We already use this function elsewhere to check status for other types so it made sense to just reuse it, it's seems odd to rewrite the function specifically for the CAPI operator `Status` type when it can be used on a bunch of different `Status` types.

internal/utils/status/status.go
squizzi: @zerospiel had some ideas of how to make this easier but I've rebased this damn thing so many times I rebased my changes away apparently.  I don't mind doing a struct.

internal/utils/status/status.go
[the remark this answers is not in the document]
a13x5: Ok, let's keep it. I still don't like it though

internal/utils/status/status.go
squizzi: I did a struct for the return fields which helped alleviate some of the confusion with types that match.  

>> As an option the list op could be done at checkProviderStatus and here you can do only basic checks on metav1.List

A lot of the choices made were so that we could use similar functions in multiple places instead of something specific to `checkProviderStatus`.

The `ConditionsFromResource` function is performing the list, the condition parsing is within `ConditionsFromUnstructured`, so perhaps the name here could change to make it less confusing.  A modified `ConditionsFromResource` to `GetResourceConditions`

internal/controller/managedcluster_controller.go
```diff
@@ -118,60 +118,36 @@ func (r *ManagedClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		l.Info("Clusters not found, ignoring since object must be deleted or not yet created")
-		return true, nil
-	}
-
-	if err != nil {
-		return true, fmt.Errorf("failed to get cluster information for managedCluster %s in namespace: %s: %w",
-			managedCluster.Namespace, managedCluster.Name, err)
-	}
-	conditions, found, err := unstructured.NestedSlice(list.Items[0].Object, "status", "conditions")
+	}, labels.SelectorFromSet(map[string]string{hmc.FluxHelmChartNameKey: managedCluster.Name}).String())
```
eromanova, line 139: We should also filter it by the namespace of the `ManagedCluster`.

eromanova (changes requested):

eromanova (approved):

squizzi:
@eromanova another rebase cleared your review, if you don't mind approving again

eromanova (approved):

PR #342, opened 2024-09-18
Add Credential CR and controller

### LINKED ISSUE #321: Credentials API and CRDs
# CRD

**Simple cluster identity reference**

```yaml
kind: Credential
metadata:
  name: azure-cred
spec:
  description: |
    Main Azure cluster credentials
  identityRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    kind: AzureClusterIdentity
    name: azure-cred-identity
    namespace: hmc-system
status:
  state: Ready
```

User must maintain the cluster identity object by itself (i.e. create cluster identity and corresponding secret). The `Credential` object just references the existing cluster identity object.

`.spec.identityRef.namespace` is an optional field which currently only applies to the Azure provider, since all other providers have cluster scoped cluster identity resources.

`.spec.description` (optional) is a free form description of the credentials. Will be also added to print columns, so it will be easier for the user to choose what `Credential` to use.

`.status.state` field signify if any problem appeared with the resource referenced (e.g. it doesn’t exist).

Controller is not responsible for the cluster identity resource and its secret.

### LINKED ISSUE #328: ManagedCluster controller credentials reference and checks
Several fields must be added to the `ManagedCluster` object:

`.spec.credential` to reference `Credential` object which will be used to provision the cluster.
`.status.credentialState` state of the referenced credential (same as the .status.state of the `Credential`)
`.status.credentialProvisioned` state of the provisioning of the CCM/CSI credentials (`bool`)

a13x5:
Resolves #321 
Resolves #328


api/v1alpha1/credential_types.go
```diff
@@ -0,0 +1,75 @@
...
+// CredentialSpec defines the desired state of Credential
+type CredentialSpec struct {
+	IdentityRef *corev1.ObjectReference `json:"identityRef"`
+	// +optional
+	Description string `json:"description,omitempty"`
+}
+
+// CredentialStatus defines the observed state of Credential
+type CredentialStatus struct {
+	State CredentialState `json:"state,omitempty"`
```
Kshatrix, line 46: Let's use conditions instead (but not necessary to change it now)

internal/controller/credential_controller.go
```diff
@@ -0,0 +1,95 @@
...
+	clIdty.SetKind(cred.Spec.IdentityRef.Kind)
+	clIdty.SetName(cred.Spec.IdentityRef.Name)
+	clIdty.SetNamespace(cred.Spec.IdentityRef.Namespace)
+
+	err = r.Client.Get(ctx, client.ObjectKey{
+		Name:      cred.Spec.IdentityRef.Name,
+		Namespace: cred.Spec.IdentityRef.Namespace,
+	}, clIdty)
+	if err != nil {
+		stateErr := r.setState(ctx, cred, hmc.CredentialNotFound)
```
Kshatrix, line 62: please verify that the returned error is indeed NotFound before reporting it in the status

internal/controller/managedcluster_controller.go
```diff
@@ -514,6 +555,36 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
...
+	err := json.Unmarshal(values.Raw, &valuesJSON)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling values: %s", err)
+	}
+
+	identityVals := map[string]any{
+		"name": idRef.Name,
+	}
+
+	switch idRef.Kind {
```
Kshatrix, line 569: We should not be provider-specific here. whatever is defined in the `idRef` should be passed as is. It should be up to the template owner to decide which fields to use.

Kshatrix (changes requested):

api/v1alpha1/managedcluster_types.go
```diff
@@ -72,7 +74,8 @@ type ManagedClusterSpec struct {
 	// Template is a reference to a Template object located in the same namespace.
 	Template string `json:"template"`
 	// DryRun specifies whether the template should be applied after validation or only validated.
-	DryRun bool `json:"dryRun,omitempty"`
+	DryRun     bool                `json:"dryRun,omitempty"`
+	Credential CredentialReference `json:"credential,omitempty"`
```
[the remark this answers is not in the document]
a13x5: @Kshatrix Sorry, but I don't understand what you mean here by the same template. Maybe from the same *namespace*?

api/v1alpha1/credential_types.go
a13x5: Ok, I'll leave the state field for now and create a separate task for that

internal/controller/managedcluster_controller.go
```diff
@@ -269,9 +272,47 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+		apimeta.SetStatusCondition(managedCluster.GetConditions(), metav1.Condition{
+			Type:    hmc.CredentialReadyCondition,
+			Status:  metav1.ConditionFalse,
+			Reason:  hmc.FailedReason,
+			Message: fmt.Sprintf("Failed to get Credential: %s", err),
+		})
+		return ctrl.Result{}, err
+	}
+
+	if cred.Status.State != hmc.CredentialReady {
```
Kshatrix, line 290: we will have this check in admission for the initial managedcluster creation. but after it passed validation initially (e.g. the credential was ready when managedcluster was created, but got invalid some time later), I don't think it makes sense to fail the reconcile anymore

Kshatrix (changes requested):

Kshatrix (approved):

PR #352, opened 2024-09-18
Implemented Azure e2e tests

kylewuolle:
Implement Azure template e2e tests. Closes https://github.com/Mirantis/hmc/issues/210.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -34,6 +34,7 @@ import (
 	"github.com/Mirantis/hmc/test/kubeclient"
 	"github.com/Mirantis/hmc/test/managedcluster"
 	"github.com/Mirantis/hmc/test/managedcluster/aws"
+	"github.com/Mirantis/hmc/test/managedcluster/azure"
 	"github.com/Mirantis/hmc/test/managedcluster/vsphere"
 	"github.com/Mirantis/hmc/test/utils"
 )
@@ -62,7 +63,6 @@ var _ = Describe("controller", Ordered, func() {
 	Context("Operator", func() {
 		It("should run successfully", func() {
 			kc := kubeclient.NewFromLocal(namespace)
-			aws.CreateCredentialSecret(context.Background(), kc)
 
 			By("validating that the hmc-controller and capi provider controllers are running")
 			Eventually(func() error {
@@ -73,6 +73,11 @@ var _ = Describe("controller", Ordered, func() {
 				}
 				return nil
 			}).WithTimeout(15 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
+			GinkgoT().Setenv("NAMESPACE", namespace)
+			cmd := exec.Command("make", "DEV_PROVIDER=aws", "dev-creds-apply")
+			_, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred())
+			// aws.CreateCredentialSecret(context.Background(), kc)
 		})
 	})
 
@@ -89,7 +94,11 @@ var _ = Describe("controller", Ordered, func() {
 		BeforeAll(func() {
 			By("ensuring AWS credentials are set")
 			kc = kubeclient.NewFromLocal(namespace)
-			aws.CreateCredentialSecret(context.Background(), kc)
+			// aws.CreateCredentialSecret(context.Background(), kc)
+			GinkgoT().Setenv("NAMESPACE", namespace)
+			cmd := exec.Command("make", "DEV_PROVIDER=aws", "dev-creds-apply")
+			_, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred())
 		})
 
 		AfterEach(func() {
@@ -159,11 +168,15 @@ var _ = Describe("controller", Ordered, func() {
 			cmd = exec.Command("make", "dev-templates")
 			_, err = utils.Run(cmd)
 			Expect(err).NotTo(HaveOccurred())
+			GinkgoT().Setenv("NAMESPACE", namespace)
+			cmd = exec.Command("make", "DEV_PROVIDER=aws", "dev-creds-apply")
+			_, err = utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred())
 			Expect(os.Unsetenv("KUBECONFIG")).To(Succeed())
 
 			// Ensure AWS credentials are set in the standalone cluster.
 			standaloneClient = kc.NewFromCluster(context.Background(), namespace, clusterName)
-			aws.CreateCredentialSecret(context.Background(), standaloneClient)
+			// aws.CreateCredentialSecret(context.Background(), standaloneClient)
 
 			templateBy(managedcluster.TemplateAWSHostedCP, "validating that the controller is ready")
 			Eventually(func() error {
@@ -179,7 +192,7 @@ var _ = Describe("controller", Ordered, func() {
 
 			// Populate the environment variables required for the hosted
 			// cluster.
-			aws.PopulateHostedTemplateVars(context.Background(), kc)
+			aws.PopulateHostedTemplateVars(context.Background(), kc, clusterName)
 
 			templateBy(managedcluster.TemplateAWSHostedCP, "creating a ManagedCluster")
 			hd := managedcluster.GetUnstructured(managedcluster.TemplateAWSHostedCP)
@@ -211,7 +224,7 @@ var _ = Describe("controller", Ordered, func() {
 			)
 			Eventually(func() error {
 				return deploymentValidator.Validate(context.Background(), standaloneClient)
-			}).WithTimeout(30 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
+			}).WithTimeout(60 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
 
 			// Delete the hosted ManagedCluster and verify it is removed.
 			templateBy(managedcluster.TemplateAWSHostedCP, "deleting the ManagedCluster")
@@ -308,7 +321,7 @@ var _ = Describe("controller", Ordered, func() {
 			)
 			Eventually(func() error {
 				return deploymentValidator.Validate(context.Background(), kc)
-			}).WithTimeout(30 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
+			}).WithTimeout(60 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
 
 			deletionValidator := managedcluster.NewProviderValidator(
 				managedcluster.TemplateVSphereStandaloneCP,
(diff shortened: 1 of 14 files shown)


test/e2e/e2e_test.go
```diff
@@ -118,215 +371,166 @@ var _ = Describe("controller", Ordered, func() {
 			}
 		})
 
-		It("should work with an AWS provider", func() {
-			// Deploy a standalone cluster and verify it is running/ready.
-			// Deploy standalone with an xlarge instance since it will also be
-			// hosting the hosted cluster.
-			GinkgoT().Setenv(managedcluster.EnvVarAWSInstanceType, "t3.xlarge")
-			GinkgoT().Setenv(managedcluster.EnvVarInstallBeachHeadServices, "false")
+		It("should work with an Azure provider", func() {
```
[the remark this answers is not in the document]
kylewuolle: It'd be a good idea. Maybe in a follow on PR?

test/e2e/e2e_test.go
```diff
@@ -325,8 +327,205 @@ var _ = Describe("controller", Ordered, func() {
...
+						err := deleteFunc()
+						Expect(err).NotTo(HaveOccurred())
+					}
+				}
+			}
+		})
+
+		It("should work with an Azure provider", func() {
+			templateBy(managedcluster.TemplateAzureStandaloneCP,
+				fmt.Sprintf("creating a Deployment using template %s", managedcluster.TemplateAzureStandaloneCP))
```
squizzi, line 374:
```suggestion
				templateBy("creating a ManagedCluster")
```
We've changed `Deployment` types to `ManagedCluster` now.

The `templateBy` is already setting `[templateName] log description` so it's kind of assumed that the `ManagedCluster` is being created with this template.

test/e2e/e2e_test.go
```diff
@@ -325,8 +327,205 @@ var _ = Describe("controller", Ordered, func() {
...
+			standaloneDeleteFunc := kc.CreateManagedCluster(context.Background(), sd)
+
+			// verify the standalone cluster is deployed correctly
+			deploymentValidator := managedcluster.NewProviderValidator(
+				managedcluster.TemplateAzureStandaloneCP,
+				sdName,
+				managedcluster.ValidationActionDeploy,
+			)
+
+			By("waiting for infrastructure providers to deploy successfully")
```
squizzi, line 387:
```suggestion
			templateBy(TemplateAzureStandaloneCP, "waiting for infrastructure provider to deploy successfully")
```

test/e2e/e2e_test.go
```diff
@@ -325,8 +327,205 @@ var _ = Describe("controller", Ordered, func() {
...
+			Eventually(func() error {
+				return deploymentValidator.Validate(context.Background(), kc)
+			}).WithTimeout(90 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
+
+			// setup environment variables for deploying the hosted template (subnet name, etc)
+			setAzureEnvironmentVariables(sdName, kc)
```
squizzi, line 393: Both vsphere and aws providers have an environment variable set function as well; it might be worth just DRYing these all out into a single helper that can be used across providers.  I tried to do this in my PR a bit with `vsphere.CheckEnv`.

Either way this should probably live in either the `azure` package since it's specific to Azure OR if you do a common one create a `common` package or place it in test utils.

test/e2e/e2e_test.go
```diff
@@ -325,8 +327,205 @@ var _ = Describe("controller", Ordered, func() {
...
+func deployOnAzureCluster(kubeCfgPath string) {
+	GinkgoT().Helper()
+	GinkgoT().Setenv("KUBECONFIG", kubeCfgPath)
+	cmd := exec.Command("kubectl", "create", "-f",
+		"https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/master/deploy/example/"+
+			"storageclass-azuredisk-csi.yaml")
+	_, err := utils.Run(cmd)
+	Expect(err).NotTo(HaveOccurred())
+
+	cmd = exec.Command("kubectl", "patch", "storageclass", "managed-csi", "-p",
```
squizzi, line 475: @kylewuolle @a13x5  Is this necessary for the provider?  It was my understanding at least in AWS that we shouldn't have to configure the CSI driver and it should be present  as the default storageclass from get-go, is that not the case on Azure?

test/e2e/e2e_test.go
```diff
@@ -325,8 +327,205 @@ var _ = Describe("controller", Ordered, func() {
...
+
+	networkSpec, found, err := unstructured.NestedMap(spec, "networkSpec")
+	Expect(err).NotTo(HaveOccurred())
+	Expect(found).To(BeTrue())
+
+	vnet, found, err := unstructured.NestedMap(networkSpec, "vnet")
+	Expect(err).NotTo(HaveOccurred())
+	Expect(found).To(BeTrue())
+	vnetName := vnet["name"]
+	GinkgoT().Setenv("AZURE_VM_NET_NAME", vnetName.(string))
```
squizzi, line 503: It's a test so it's not as big of a deal if it panics but you could do a check on each of the type casts here with:

```suggestion
	vnetName, ok := vnet["name"].(string)
	Expect(ok).To(BeTrue()
```

Prior to setting the environment var.

test/managedcluster/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/restmapper"
+
+	hmc "github.com/Mirantis/hmc/api/v1alpha1"
+	"github.com/Mirantis/hmc/test/kubeclient"
+)
+
+func GetAzureInfo(ctx context.Context, name string, kc *kubeclient.KubeClient) (map[string]interface{}, error) {
```
squizzi, line 40: nit: The package name is `azure`, so this could just be `GetInfo` since it'll be `azure.GetInfo` when it's imported.

test/e2e/e2e_test.go
```diff
@@ -325,8 +327,205 @@ var _ = Describe("controller", Ordered, func() {
...
+	Expect(err).NotTo(HaveOccurred())
+
+	cmd = exec.Command("make", "dev-templates")
+	_, err = utils.Run(cmd)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(os.Unsetenv("KUBECONFIG")).To(Succeed())
+}
+
+func setAzureEnvironmentVariables(clusterName string, kc *kubeclient.KubeClient) {
+	GinkgoT().Helper()
```
squizzi, line 491:
```suggestion
	GinkgoHelper()
```

test/managedcluster/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+	list, err := dc.List(ctx, metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{hmc.FluxHelmChartNameKey: name}).String(),
+	})
+
+	if apierrors.IsNotFound(err) || len(list.Items) == 0 {
+		return nil, fmt.Errorf("clusters %s not found", name)
+	}
+
+	spec, found, err := unstructured.NestedMap(list.Items[0].Object, "spec")
+	if !found || err != nil {
```
squizzi, line 57: Careful with the `!found || err != nil` check here, there's instances where that or might be true and you'll lose the error, it's safer to just check for `!found` and `error` separately like:

```go
spec, found, err := unstructured.NestedMap(list.Items[0].Object, "spec")
Expect(found).To(BeTrue())
Expect(err).NotTo(HaveOccurred(), "cluster %s not found", name)
```

test/managedcluster/azure/azure.go
squizzi, line 40: `GetAzureInfo` isn't running inside the `validate_*` code so it doesn't need to return error (since it's not running within `Eventually`), this means this can just be a `GinkgoHelper()` and you can use Gomega assertions inside for your error and `!found` handling.

test/managedcluster/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to fetch group resources: %w", err)
+	}
+
+	yamlReader := yamlutil.NewYAMLReader(bufio.NewReader(bytes.NewReader(yamlFile)))
+	for {
+		yamlDoc, err := yamlReader.Read()
+
+		if err != nil {
+			if err == io.EOF {
```
squizzi, line 87:
```suggestion
			if errors.Is(err, io.EOF) {
```

test/managedcluster/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	spec, found, err := unstructured.NestedMap(list.Items[0].Object, "spec")
+	if !found || err != nil {
+		return nil, fmt.Errorf("clusters spec %s not found", name)
+	}
+	return spec, nil
+}
+
+func CreateCredentialSecret(ctx context.Context, kc *kubeclient.KubeClient) error {
```
squizzi, line 63: This seems really complex, you're doing a lot of work to just read over the yaml document to create the credential secret, it'd probably just be easier to create some `map[string]interface{}` object for `AzureClusterIdentity` and then use `corev1` to create a `Secret`, but whatever.

squizzi, line 63: This can be a `GinkgoHelper()` too and then you don't need to return `error` and can use Gomega assertions.

test/managedcluster/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		dc := kc.GetDynamicClient(schema.GroupVersionResource{
+			Group:    credentialResource.GroupVersionKind().Group,
+			Version:  credentialResource.GroupVersionKind().Version,
+			Resource: mapping.Resource.Resource,
+		})
+
+		exists, err := dc.Get(ctx, credentialResource.GetName(), metav1.GetOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
```
squizzi, line 114: the `exists == nil` check is redundant because you've already checked `apierrors.IsNotFound` so you know that `exists == nil`, by putting them together with the `&&` you've added the necessity of the `exists == nil` check.

Something like this would be easier:

```go
	if !apierrors.IsNotFound(err) {
		Expect(err).NotTo(HaveOccurred(), "failed to get Azure credential")
                return
	}

        dc.Create(ctx, credentialResource, metav1.CreateOptions{})
        Expect(err).NotTo(HaveOccurred(), "failed to create Azure credentials"
```

test/managedcluster/providervalidator.go
```diff
@@ -64,6 +64,8 @@ func NewProviderValidator(template Template, clusterName string, action Validati
 		case TemplateAWSStandaloneCP, TemplateAWSHostedCP:
 			resourcesToValidate["ccm"] = validateCCM
 			resourceOrder = append(resourceOrder, "ccm")
+		case TemplateAzureStandaloneCP, TemplateVSphereHostedCP:
+			delete(resourcesToValidate, "csi-driver")
```
squizzi, line 68: I see earlier you were setting up a `csi-driver` to be used with Azure, should this test be enabled for Azure then?  Or is the `csi-driver` intentionally only for `hosted`?

test/e2e/e2e_test.go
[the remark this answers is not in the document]
squizzi: > It'd be a good idea. Maybe in a follow on PR?

Ya, probably, we probably need to revisit the tests and streamline them in general once they're all in.

I'd also like to add some spec focusing using the `provider` labels and what not.

squizzi:
Looking at the latest test failure, my suspicion is the names aren't unique for the cluster names for each provider test and they're running over each other.  It might be worth appending a provider name to the end of the `ManagedCluster` that gets created for each test so they don't clobber each other and race.

test/managedcluster/resources/vsphere-standalone-cp.yaml.tpl
```diff
@@ -3,7 +3,7 @@ kind: ManagedCluster
 metadata:
   name: ${MANAGED_CLUSTER_NAME}
 spec:
-  template: vsphere-standalone-cp
+  template: vsphere-standalone-cp-vsphere
```
squizzi, line 6: This looks like a find and replace mistake:

```suggestion
  template: vsphere-standalone-cp
```

test/e2e/e2e_test.go
kylewuolle: Yes unfortunately it's not there by default

squizzi: ya, but is that _intended_

test/managedcluster/managedcluster.go
```diff
@@ -66,17 +76,18 @@ func GetProviderLabel(provider ProviderType) string {
 
 // GetUnstructured returns an unstructured ManagedCluster object based on the
 // provider and template.
-func GetUnstructured(templateName Template) *unstructured.Unstructured {
+func GetUnstructured(templateName Template, provider ProviderType) *unstructured.Unstructured {
```
squizzi, line 79: As specified above, we do not need `ProviderType` here anymore, its unused below anyways.  Rebase might have brought it back accidentally or something.  I'm surprised the linter didn't catch it's unused.

squizzi:
@kylewuolle In the future can use try not to use force-pushes?  At least not after someone has initially reviewed your code.  It makes reviewing your code changes after a big initial review harder, you can always just squash the ending result into a single commit.

test/managedcluster/azure/azure.go
```diff
@@ -0,0 +1,153 @@
...
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/restmapper"
+
+	hmc "github.com/Mirantis/hmc/api/v1alpha1"
+	"github.com/Mirantis/hmc/test/kubeclient"
+)
+
+func getAzureInfo(ctx context.Context, name string, kc *kubeclient.KubeClient) (map[string]interface{}, error) {
```
squizzi, line 42:
```suggestion
func getAzureInfo(ctx context.Context, name string, kc *kubeclient.KubeClient) map[string]interface{}
```

test/managedcluster/azure/azure.go
```diff
@@ -0,0 +1,153 @@
...
+		LabelSelector: labels.SelectorFromSet(map[string]string{hmc.FluxHelmChartNameKey: name}).String(),
+	})
+
+	Expect(err).NotTo(HaveOccurred())
+	Expect(len(list.Items)).NotTo(BeEquivalentTo(0))
+
+	spec, found, err := unstructured.NestedMap(list.Items[0].Object, "spec")
+	Expect(err).NotTo(HaveOccurred())
+	Expect(found).To(BeTrue())
+	return spec, nil
```
squizzi, line 61:
```suggestion
	return spec
```

squizzi (approved):

Kshatrix:
@kylewuolle please squash commits like "fix linting error" and let's verify azure part is green before we land it

squizzi:
@kylewuolle if you cannot reproduce the latest failure locally -- you can use try to use the tmate action: https://github.com/mxschmitt/action-tmate to live debug what's going on with the `infrastructure-aws` provider, it'll host a tmate session you can ssh into so you can get into the actions container and figure out potentially what's going on.

squizzi (approved):
As discussed, go ahead and merge what you currently have @kylewuolle and I'm going to work to get the E2E tests all green in #360

PR #355, opened 2024-09-19
Improve Templates deletion validation

### LINKED ISSUE #329: [admission] Restrict manual removal of Templates managed by HMC
The admission webhook should:
1. Forbid the removal of `ClusterTemplates` and `ServiceTemplates`, managed by the HMC distribution system (with `hmc.mirantis.com/managed: true` label) or the template is in use by `ManagedCluster` or template chain.
3.  Forbid the removal of `ProviderTemplates` enabled in the `Management` object.

eromanova:
The following template deletion rules are validated:
1. `ClusterTemplate` or `ServiceTemplate` can't be removed if it is in use by `ManagedCluster` or owned by `TemplateChain`
2. `ProviderTemplate` can't be removed if it's a core provider or enabled in Management `spec.providers`

Closes #329

### DIFF
--- a/internal/webhook/template_webhook.go
+++ b/internal/webhook/template_webhook.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
+	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -29,14 +31,23 @@ import (
 	"github.com/Mirantis/hmc/api/v1alpha1"
 )
 
-type ClusterTemplateValidator struct {
+var errTemplateDeletionForbidden = errors.New("template deletion is forbidden")
+
+type TemplateValidator struct {
 	client.Client
+	SystemNamespace   string
+	templateKind      string
+	templateChainKind string
 }
 
-var errTemplateDeletionForbidden = errors.New("template deletion is forbidden")
+type ClusterTemplateValidator struct {
+	TemplateValidator
+}
 
 func (v *ClusterTemplateValidator) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	v.Client = mgr.GetClient()
+	v.templateKind = v1alpha1.ClusterTemplateKind
+	v.templateChainKind = v1alpha1.ClusterTemplateChainKind
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(&v1alpha1.ClusterTemplate{}).
 		WithValidator(v).
@@ -66,16 +77,18 @@ func (v *ClusterTemplateValidator) ValidateDelete(ctx context.Context, obj runti
 		return admission.Warnings{"Wrong object"}, apierrors.NewBadRequest(fmt.Sprintf("expected ClusterTemplate but got a %T", obj))
 	}
 
-	managedClusters := &v1alpha1.ManagedClusterList{}
-	if err := v.Client.List(ctx, managedClusters,
-		client.InNamespace(template.Namespace),
-		client.MatchingFields{v1alpha1.ManagedClusterTemplateIndexKey: template.Name},
-		client.Limit(1)); err != nil {
+	inUseByCluster, err := v.templateIsInUseByCluster(ctx, template)
+	if err != nil {
 		return nil, err
 	}
+	if inUseByCluster {
+		return admission.Warnings{fmt.Sprintf("The %s object can't be removed if ManagedCluster objects referencing it still exist", v.templateKind)}, errTemplateDeletionForbidden
+	}
 
-	if len(managedClusters.Items) > 0 {
-		return admission.Warnings{"The ClusterTemplate object can't be removed if ManagedCluster objects referencing it still exist"}, errTemplateDeletionForbidden
+	owners := getOwnersWithKind(template, v.templateChainKind)
+	if len(owners) > 0 {
+		return admission.Warnings{fmt.Sprintf("The %s object can't be removed if it is managed by %s: %s",
+			v.templateKind, v.templateChainKind, strings.Join(owners, ", "))}, errTemplateDeletionForbidden
 	}
 
 	return nil, nil
@@ -87,12 +100,13 @@ func (*ClusterTemplateValidator) Default(context.Context, runtime.Object) error 
 }
 
 type ServiceTemplateValidator struct {
-	client.Client
-	SystemNamespace string
+	TemplateValidator
 }
 
 func (v *ServiceTemplateValidator) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	v.Client = mgr.GetClient()
+	v.templateKind = v1alpha1.ServiceTemplateKind
+	v.templateChainKind = v1alpha1.ServiceTemplateChainKind
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(&v1alpha1.ServiceTemplate{}).
 		WithValidator(v).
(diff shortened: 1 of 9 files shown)


internal/webhook/template_webhook.go
```diff
@@ -157,11 +184,45 @@ func (*ProviderTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Ob
...
+		return false, err
+	}
+	if v.InjectUserInfo != nil {
+		v.InjectUserInfo(&req)
+	}
+	// Allow all templates' deletion for the HMC controller
+	if serviceAccountIsEqual(req, os.Getenv(ServiceAccountEnvName)) {
+		return true, nil
+	}
+	// Cluster-scoped ProviderTemplates and Templates from the system namespace are not allowed to be deleted
```
Kshatrix, line 219: It should be ok to delete templates from the system namespace as long as they are not in-use in any templatechains

internal/webhook/template_webhook.go
```diff
@@ -157,11 +184,45 @@ func (*ProviderTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Ob
...
 func (*ProviderTemplateValidator) Default(_ context.Context, _ runtime.Object) error {
 	return nil
 }
+
+func (v TemplateValidator) isTemplateDeletionAllowed(ctx context.Context, template templateutil.Template) (bool, error) {
+	req, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		return false, err
+	}
+	if v.InjectUserInfo != nil {
```
Kshatrix, line 212: is this one only for unit tests?

Kshatrix (changes requested):

internal/webhook/template_webhook.go
eromanova: Yes, it's needed for tests only.

internal/webhook/userinfo.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	ServiceAccountEnvName = "SERVICE_ACCOUNT"
+)
+
+func serviceAccountIsEqual(req admission.Request, serviceAccount string) bool {
+	return strings.HasPrefix(req.UserInfo.Username, "system:serviceaccount:") &&
```
Kshatrix, line 28: I think you can use `SplitUsername` from "k8s.io/apiserver/pkg/authentication/serviceaccount" here

Kshatrix (changes requested):

eromanova:
Tests are failing in the CI env due to https://github.com/Mirantis/hmc/issues/624 but succeeded when run locally.

internal/controller/templatechain_controller_test.go
```diff
@@ -231,9 +231,6 @@ var _ = Describe("Template Chain Controller", func() {
 
 				Eventually(k8sClient.Get, 1*time.Minute, 5*time.Second).WithArguments(ctx, chain, clusterTemplateChainResource).Should(HaveOccurred())
 			}
-			for _, template := range []*hmcmirantiscomv1alpha1.ClusterTemplate{ctTemplates["test"], ctTemplates["ct0"], ctTemplates["ct2"]} {
```
eromanova, line 234: Removing the templates' cleanup since the webhook now denies these requests. Templates have owner reference set and should be automatically deleted by the garbage collector. The owner reference existence is checked below.

eromanova: Redundant since the admission validation webhook was disabled in tests.

internal/webhook/template_webhook.go
```diff
@@ -185,11 +190,134 @@ func (*ProviderTemplateValidator) ValidateUpdate(_ context.Context, _, _ runtime
...
+			return false, err
+		}
+		if len(chainList.Items) > 0 {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (v TemplateValidator) providerTemplateIsInUse(ctx context.Context, templateName string) (bool, error) {
```
Kshatrix, line 284: I think in our terms "provider template is in use" == it's part of some `Release` or it's explicitly set in the `Management` object

internal/webhook/template_webhook.go
```diff
@@ -185,11 +190,134 @@ func (*ProviderTemplateValidator) ValidateUpdate(_ context.Context, _, _ runtime
...
+		return true, nil
+	}
+	return false, nil
+}
+
+func (v TemplateValidator) templateIsInUseByTemplateChain(ctx context.Context, template client.Object) (bool, error) {
+	listOpts := []client.ListOption{
+		client.InNamespace(template.GetNamespace()),
+		client.MatchingFields{v1alpha1.TemplateChainSupportedTemplatesIndexKey: template.GetName()},
+		client.Limit(1),
```
Kshatrix, line 261: I think there will always be a reasonable amount of template chains, so we don't have to limit the number. Also it would be nice to show which templatechains are preventing the operation

internal/webhook/template_webhook.go
```diff
@@ -185,11 +190,134 @@ func (*ProviderTemplateValidator) ValidateUpdate(_ context.Context, _, _ runtime
...
+	if len(managedClusters.Items) > 0 {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (v TemplateValidator) templateIsInUseByTemplateChain(ctx context.Context, template client.Object) (bool, error) {
+	listOpts := []client.ListOption{
+		client.InNamespace(template.GetNamespace()),
+		client.MatchingFields{v1alpha1.TemplateChainSupportedTemplatesIndexKey: template.GetName()},
```
Kshatrix, line 260: AFAIR we only wanted to prevent removal of "distributed" templates (based on ownership). Because otherwise they would've been recreated anyways.

Kshatrix (changes requested):

zerospiel (approved):

internal/webhook/template_webhook.go
```diff
@@ -185,11 +202,72 @@ func (*ProviderTemplateValidator) ValidateUpdate(_ context.Context, _, _ runtime
...
+		return admission.Warnings{"Wrong object"}, apierrors.NewBadRequest(fmt.Sprintf("expected ProviderTemplate but got a %T", obj))
+	}
+
+	owners := getOwnersWithKind(template, v1alpha1.ReleaseKind)
+	if len(owners) > 0 {
+		return admission.Warnings{fmt.Sprintf("The ProviderTemplate %s cannot be removed while it is part of existing Releases: %s",
+			template.GetName(), strings.Join(owners, ", "))}, errTemplateDeletionForbidden
+	}
+
+	mgmtList := &v1alpha1.ManagementList{}
```
Kshatrix, line 217: nit: we already have this logic in the release webhook, let's make a common function out of it

internal/webhook/template_webhook.go
```diff
@@ -185,11 +202,72 @@ func (*ProviderTemplateValidator) ValidateUpdate(_ context.Context, _, _ runtime
...
+	}
+	if len(mgmtList.Items) == 0 {
+		return nil, nil
+	}
+	if len(mgmtList.Items) > 1 {
+		return nil, fmt.Errorf("expected 1 Management object, got %d", len(mgmtList.Items))
+	}
+	mgmt := mgmtList.Items[0]
+	if slices.Contains(mgmt.Templates(), template.Name) {
+		return admission.Warnings{fmt.Sprintf("The ProviderTemplate %s cannot be removed while it is enabled in the Management spec",
```
Kshatrix, line 229:
```suggestion
		return admission.Warnings{fmt.Sprintf("The ProviderTemplate %s cannot be removed while it is used in the Management spec",
```

Kshatrix (changes requested):

Kshatrix (approved):

PR #362, opened 2024-09-20
Add Services (ServiceTemplates) to ManagedCluster to deploy on target cluster

wahabmk:
# Description

To be filled....

# Testing

To be filled....


internal/sveltos/clusterprofile.go
```diff
@@ -0,0 +1,131 @@
...
+	})
+	if err != nil {
+		return nil, operation, err
+	}
+
+	return cp, operation, nil
+}
+
+// DeleteClusterProfile issues delete on ClusterProfile object.
+func DeleteClusterProfile(ctx context.Context, cl client.Client, namespace string, name string) error {
```
slysunkin, line 119: Can we create some generic "remover"? Currently the difference between DeleteHelmRelease and DeleteClusterProfile is only in client.Object for Delete method, but the method parameters and logic is the same.

api/v1alpha1/managedcluster_types.go
```diff
@@ -60,6 +62,45 @@ const (
...
 
+// ManagedClusterServiceSpec represents a Service within ManagedCluster
+type ManagedClusterServiceSpec struct {
+	// Template is a reference to a Template object located in the same namespace.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Template string `json:"template"`
+	// Install tells if this service should be installed.
+	// +kubebuilder:default:=true
+	Install bool `json:"install"`
```
Kshatrix, line 73: I'd suggest `Disable` instead. As it not only controls installation, but also removes the service from the cluster. It's also inverse, so will be false by default and hidden with ommitempty.

internal/controller/managedcluster_controller.go
```diff
@@ -270,55 +275,162 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+	if !fluxconditions.IsReady(hr) {
+		return ctrl.Result{RequeueAfter: DefaultRequeueInterval}, nil
+	}
+
+	return r.updateServices(ctx, l, managedCluster)
+}
+
+// updateServices reconciles services provided in ManagedCluster.Spec.Services.
+// TODO(https://github.com/Mirantis/hmc/issues/361): Set status to ManagedCluster object at appropriate places.
+func (r *ManagedClusterReconciler) updateServices(ctx context.Context, l logr.Logger, mc *hmc.ManagedCluster) (ctrl.Result, error) {
```
Kshatrix, line 335: can we please use logger from context instead

internal/controller/managedcluster_controller.go
```diff
@@ -270,55 +275,162 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+		})
+	}
+
+	cp, _, err := sveltos.ReconcileClusterProfile(ctx, r.Client, mc.Namespace, mc.Name,
+		map[string]string{
+			hmc.FluxHelmChartNamespaceKey: mc.Namespace,
+			hmc.FluxHelmChartNameKey:      mc.Name,
+		},
+		sveltos.ReconcileClusterProfileOpts{
+			OwnerReference: &metav1.OwnerReference{
```
Kshatrix, line 384: Cluster-scoped dependents can only specify cluster-scoped owners.
ClusterProfile is cluster-scoped, ManagedCluster in namespace-scoped.

internal/sveltos/clusterprofile.go
```diff
@@ -0,0 +1,131 @@
...
+// ReconcileClusterProfile reconciles a Sveltos ClusterProfile object.
+func ReconcileClusterProfile(ctx context.Context,
+	cl client.Client,
+	namespace string,
+	name string,
+	matchLabels map[string]string,
+	opts ReconcileClusterProfileOpts,
+) (*sveltosv1beta1.ClusterProfile, controllerutil.OperationResult, error) {
+
+	cp := &sveltosv1beta1.ClusterProfile{
```
Kshatrix, line 63: Can we use Profile instead? It seems to be just a namespace-scoped version of ClusterProfile if I'm not missing something

internal/sveltos/clusterprofile.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, operation, err
+	}
+
+	return cp, operation, nil
+}
+
+// DeleteClusterProfile issues delete on ClusterProfile object.
+func DeleteClusterProfile(ctx context.Context, cl client.Client, namespace string, name string) error {
+	err := cl.Delete(ctx, &sveltosv1beta1.ClusterProfile{
```
Kshatrix, line 120: If sveltos applies finalizers, we can remove them to force-delete the object (we don't really need to wait for services cleanup when we remove the whole cluster)

templates/service/ingress-nginx/Chart.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: A Helm chart to refer the official ingress-nginx helm chart
+type: application
+version: 4.11.0
+appVersion: "1.11.0"
+dependencies:
+  - name: ingress-nginx
+    version: 4.11.0
+    repository: https://kubernetes.github.io/ingress-nginx
+annotations:
+  hmc.mirantis.com/type: service
```
Kshatrix, line 12: this one seems to be redundant as we have a separate ServiceTemplate kind now

templates/service/kyverno/Chart.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: A Helm chart to refer the official kyverno helm chart
+type: application
+version: 3.2.6
+appVersion: "v1.12.5"
+dependencies:
+  - name: kyverno
+    version: 3.2.6
+    repository: https://kyverno.github.io/kyverno/
+annotations:
+  hmc.mirantis.com/type: service
```
Kshatrix, line 12: same

api/v1alpha1/managedcluster_types.go
```diff
@@ -60,6 +62,45 @@ const (
 	ProgressingReason string = "Progressing"
 )
 
+// ManagedClusterServiceSpec represents a Service within ManagedCluster
+type ManagedClusterServiceSpec struct {
+	// Template is a reference to a Template object located in the same namespace.
+	// +kubebuilder:validation:Required
```
eromanova, line 68: The `Required` kubebuilder flag is redundant now and can be removed.

internal/controller/managedcluster_controller.go
```diff
@@ -270,55 +275,162 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+				UID:        mc.UID,
+			},
+			HelmChartOpts: opts,
+		})
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to create ClusterProfile: %w", err)
 	}
-	return ctrl.Result{}, nil
+
+	l.Info(fmt.Sprintf("Successfully created ClusterProfile (%s)", cp.Name))
```
eromanova, line 396: This log will be shown every time the ClusterProfile is created/updated or not changed. I'd move it to the `ReconcileClusterProfile` function instead and handle the logging message depending on the `operation` result.

internal/controller/managedcluster_controller.go
```diff
@@ -270,55 +275,162 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+			OwnerReference: &metav1.OwnerReference{
+				APIVersion: hmc.GroupVersion.String(),
+				Kind:       hmc.ManagedClusterKind,
+				Name:       mc.Name,
+				UID:        mc.UID,
+			},
+			HelmChartOpts: opts,
+		})
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to create ClusterProfile: %w", err)
```
eromanova, line 393:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to reconcile ClusterProfile: %w", err)
```

templates/provider/hmc/templates/rbac/controller/roles.yaml
```diff
@@ -131,12 +131,22 @@ rules:
...
-  verbs: {{ include "rbac.viewerVerbs" . | nindent 4 }}
+  - clusterprofiles
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
eromanova, line 149:
```suggestion
  verbs: {{ include "rbac.editorVerbs" . | nindent 4 }}
```

eromanova (changes requested):

api/v1alpha1/managedcluster_types.go
```diff
@@ -60,6 +62,45 @@ const (
...
+	// +optional
+	ReleaseNamespace string `json:"releaseNamespace"`
+	// CreateNamespace create the release namespace if not present.
+	// It will defaults to true if not provided.
+	// +kubebuilder:default:=true
+	// +optional
+	CreateNamespace bool `json:"createNamespace"`
+	// RegistryConfig to provide options to interact with registry.
+	// +optional
+	RegistryConfig *ManagedClusterServiceRegistryConfig `json:"registryConfig,omitempty"`
```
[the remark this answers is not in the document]
wahabmk: Retrieved URL and Insecure from Flux `HelmRepository` in `getServiceTemplateSource()` func

internal/controller/managedcluster_controller.go
wahabmk: Used Sveltos `Profile` instead.

internal/sveltos/clusterprofile.go
wahabmk: Used Sveltos `Profile` instead.

internal/controller/managedcluster_controller.go
```diff
@@ -420,8 +530,13 @@ func (r *ManagedClusterReconciler) Delete(ctx context.Context, l logr.Logger, ma
 		return ctrl.Result{}, err
 	}
 
+	err = sveltos.DeleteProfile(ctx, r.Client, managedCluster.Namespace, managedCluster.Name)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
wahabmk, line 536: Concerning your [suggestion](https://github.com/Mirantis/hmc/pull/362#discussion_r1773348198) to remove Sveltos finalizers. I think it will be better if we don't remove Sveltos finalizer because otherwise the objects being cleaned up in [`reconcileDeleteCommon()`](https://github.com/projectsveltos/addon-controller/blob/29d37c8a90fe47c4518967a4c041a2511fe596c7/controllers/profile_utils.go#L976-L1014) func may be left hanging in the `hmc-system` namespace after the `Profile` object has been deleted.

internal/controller/managedcluster_controller.go
```diff
@@ -303,24 +308,129 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+	for _, svc := range mc.Spec.Services {
+		if svc.Disable {
+			l.Info(fmt.Sprintf("Skip adding Template (%s) to Profile (%s) because Disable=true", svc.Template, mc.Name))
+			continue
+		}
+
+		tmpl := &hmc.ServiceTemplate{}
+		tmplRef := types.NamespacedName{Name: svc.Template, Namespace: mc.Namespace}
+		if err := r.Get(ctx, tmplRef, tmpl); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get Template (%s)", tmplRef.String())
```
eromanova, line 350:
```suggestion
			return ctrl.Result{}, fmt.Errorf("failed to get Template (%s): %w", tmplRef.String(), err)
```

internal/controller/managedcluster_controller.go
```diff
@@ -303,24 +308,129 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+	tmplRef := types.NamespacedName{Namespace: tmpl.Namespace, Name: tmpl.Name}
+
+	if tmpl.Status.ChartRef == nil {
+		return nil, fmt.Errorf("status for ServiceTemplate (%s) has not been updated yet", tmplRef.String())
+	}
+
+	hc := &sourcev1.HelmChart{}
+	if err := r.Get(ctx, types.NamespacedName{
+		Namespace: tmpl.Status.ChartRef.Namespace,
+		Name:      tmpl.Spec.Helm.ChartName,
```
eromanova, line 416: Can we also get the chart name from the `Status.ChartRef`? Template spec API supports configuring either `chartName` or `chartRef`: https://github.com/Mirantis/hmc/blob/main/api/v1alpha1/templates_common.go#L43
So, `chartName` may be unset.

internal/controller/managedcluster_controller.go
```diff
@@ -303,24 +308,129 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+	}
+
+	repo := &sourcev1.HelmRepository{}
+	if err := r.Get(ctx, types.NamespacedName{
+		// Using chart's namespace because it's source
+		// (helm repository in this case) should be within the same namespace.
+		Namespace: hc.Namespace,
+		Name:      hc.Spec.SourceRef.Name,
+	}, repo); err != nil {
+		return nil, fmt.Errorf("failed to get HelmRepository (%s)", tmplRef.String())
```
eromanova, line 428:
```suggestion
		return nil, fmt.Errorf("failed to get HelmRepository (%s): %w", tmplRef.String(), err)
```

internal/sveltos/profile.go
```diff
@@ -0,0 +1,142 @@
...
+
+// ReconcileProfile reconciles a Sveltos Profile object.
+func ReconcileProfile(ctx context.Context,
+	cl client.Client,
+	l logr.Logger,
+	namespace string,
+	name string,
+	matchLabels map[string]string,
+	opts ReconcileProfileOpts,
+) (*sveltosv1beta1.Profile, controllerutil.OperationResult, error) {
```
eromanova, line 58: It looks like the `OperationResult` is unused and not needed here as a result now

internal/sveltos/profile.go
```diff
@@ -0,0 +1,142 @@
...
+			cp.Spec.HelmCharts = append(cp.Spec.HelmCharts, helmChart)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, operation, err
+	}
+
+	if operation == controllerutil.OperationResultCreated || operation == controllerutil.OperationResultUpdated {
+		l.Info(fmt.Sprintf("Successfully %s Profile (%s)", string(operation), cp.Name))
```
eromanova, line 127: Let's log the profile `Namespace` here as well

internal/controller/managedcluster_controller.go
```diff
@@ -303,24 +308,129 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+	if tmpl.Status.ChartRef == nil {
+		return nil, fmt.Errorf("status for ServiceTemplate (%s) has not been updated yet", tmplRef.String())
+	}
+
+	hc := &sourcev1.HelmChart{}
+	if err := r.Get(ctx, types.NamespacedName{
+		Namespace: tmpl.Status.ChartRef.Namespace,
+		Name:      tmpl.Spec.Helm.ChartName,
+	}, hc); err != nil {
+		return nil, fmt.Errorf("failed to get HelmChart (%s)", tmplRef.String())
```
eromanova, line 418:
```suggestion
		return nil, fmt.Errorf("failed to get HelmChart (%s): %w", tmplRef.String(), err)
```

internal/controller/managedcluster_controller.go
```diff
@@ -303,24 +308,129 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+		}
+
+		source, err := r.getServiceTemplateSource(ctx, tmpl)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not get repository url: %w", err)
+		}
+
+		opts = append(opts, sveltos.HelmChartOpts{
+			RepositoryURL:  source.Spec.URL,
+			RepositoryName: tmpl.Spec.Helm.ChartName,
```
eromanova, line 360: Is it correct that the `RepositoryName` is set to the `ChartName`?

eromanova (changes requested):

internal/controller/managedcluster_controller.go
wahabmk: The Sveltos docs use this format for OCI registries (scroll to the bottom): https://projectsveltos.github.io/sveltos/addons/helm_charts/

We are using OCI for now so it is ok. But good point, as I can check to see which type the HelmRepository is. If it is not OCI, then looking at Sveltos docs, the chartName needs to be in the format of `repository/chart` which in our case will become `ChartName/ChartName`, because we don't have a "RepositoryName" available in our spec.

wahabmk:
The CI failure is an auth related one:
```
ERROR: failed to solve: golang:1.22: failed to resolve source metadata for docker.io/library/golang:1.22: failed to authorize: failed to fetch oauth token: unexpected status from GET request to https://auth.docker.io/token?scope=repository%3Alibrary%2Fgolang%3Apull&service=registry.docker.io: 401 Unauthorized
```

internal/controller/managedcluster_controller.go
wahabmk: Pushed new commit to set these fields based on whether the source repository is of default or oci type.

internal/controller/managedcluster_controller.go
```diff
@@ -303,24 +309,138 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
...
+		if err := r.Get(ctx, tmplRef, tmpl); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get Template (%s): %w", tmplRef.String(), err)
+		}
+
+		source, err := r.getServiceTemplateSource(ctx, tmpl)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not get repository url: %w", err)
+		}
+
+		opts = append(opts, sveltos.HelmChartOpts{
```
Kshatrix, line 359: although this works well with templates from our registry, it won't be the case for other flux sources (e.g. git repo source or registries under auth).

it would be nice to use chart urls hosted by source-controller itself. but only if sveltos downloads charts on the mothership cluster. could you please check?

internal/controller/managedcluster_controller.go
Kshatrix: ack. then we should wait for Profile removal before we remove finalizers from the managedcluster object (to not leave hanging resources behind unnoticed)

Kshatrix (changes requested):

internal/controller/managedcluster_controller.go
wahabmk: I have created the following ticket to implement this in future: https://github.com/Mirantis/hmc/issues/413. This PR as is will work for dev and demo purposes.

wahabmk:
> ack. then we should wait for Profile removal before we remove finalizers from the managedcluster object (to not leave hanging resources behind unnoticed)

done

Kshatrix (approved):

PR #371, opened 2024-09-23
Support to reference ClusterTemplate from the ManagedClusters' namespace only

### LINKED ISSUE #346: Reference ClusterTemplates in the same namespace in ManagedClusters

eromanova:
Closes #346


Kshatrix (changes requested):
in the template webhook we should be limited to the current namespace now when we validate templates in-use

Kshatrix (approved):

PR #377, opened 2024-09-23
support for AWS_SESSION_TOKEN in development mode

Schnitzel:


Kshatrix:
@Schnitzel it seems it was already fixed in #378

Schnitzel:
@Kshatrix oh good find, documentation is missing though, will create a PR just with the doc entry :)

PR #378, opened 2024-09-23
Restore missing session token variable for aws credentials

kylewuolle:
Simple fix to restore the AWS_SESSION_TOKEN functionality for aws credentials.


slysunkin (approved):

a13x5:
@kylewuolle This probably won't work with normal credentials - when you have `AWS_SESSION_TOKEN` empty, since controller will assume that you have `AWS_SESSION_TOKEN` and it's empty - which is incorrect credentials.
Please recheck

PR #380, opened 2024-09-24
Handle TemplateManagement creation and removal

### LINKED ISSUE #373: Properly handle TemplateManagement creation and removal
`TemplateManagement` should be a singleton, created by the release controller together with the `Management` object (with empty spec) and removed only when HMC is removed.

eromanova:
1. Support automatically creating the `TemplateManagement` with no access rules (enabled by default).
2. Forbid the creation of more than one `TemplateManagement` object.
3. Forbid the removal of the `TemplateManagement` object if the Management object exists and was not triggered for the deletion.

Closes #373

### DIFF
--- a/internal/webhook/templatemanagement_webhook.go
+++ b/internal/webhook/templatemanagement_webhook.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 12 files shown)


internal/webhook/templatemanagement_webhook.go
```diff
@@ -113,8 +126,8 @@ func getManagedClustersForTemplate(ctx context.Context, cl client.Client, namesp
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
-func (v *TemplateManagementValidator) ValidateDelete(ctx context.Context, _ runtime.Object) (admission.Warnings, error) {
-	return nil, nil
+func (v *TemplateManagementValidator) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
+	return nil, ErrTemplateManagementDeletionForbidden
```
Kshatrix, line 130: I think we should add ownerreference to the management object instead. And reject TemplateManagement deletion only when Management deletion timestamp is unset

Kshatrix (changes requested):

internal/webhook/templatemanagement_webhook.go
eromanova: I found it excessive since the `TemplateManagement` will be automatically removed after the `Management` object deletion once all the CRDs are removed. But in case we change the deletion logic sometimes your approach makes sense.

Kshatrix: Yes, I understand, but I think it's better not to rely on the side effect of CRD removal when we have an explicit dependency

Kshatrix (approved):

PR #381, opened 2024-09-24
Kubebuilder API project chores

zerospiel:
* separated CRDs per files, automatically generated with the `kubebuilder` tool
* updated the `PROJECT` according to the current state with the `kubebuilder` tool
* removed unnecessary kubebuilder pragmas as of `+optional` and `kubebuilder:validation:Required`
* renamed templates_types.go to templates_common.go
* renamed `templatechain_types.go` to `templates_common.go`
* renamed `*Mixin` types to `*Common` to avoid python-style patterns
* bump golangci-lint to `v1.61.0`
* minor style nits


api/v1alpha1/managedcluster_types.go
```diff
@@ -62,27 +62,25 @@ const (
 
 // ManagedClusterSpec defines the desired state of ManagedCluster
 type ManagedClusterSpec struct {
-	// DryRun specifies whether the template should be applied after validation or only validated.
-	// +optional
-	DryRun bool `json:"dryRun,omitempty"`
-	// Template is a reference to a Template object located in the same namespace.
-	// +kubebuilder:validation:Required
```
eromanova, line 69: We should not remove the `Required` flag here to restrict `ManagedClusters` creation with an empty `spec.template` before the request is passed to the webhook. `MinLength` validation is not enough since an empty string bypasses this check.

api/v1alpha1/templates_common.go
```diff
@@ -28,10 +28,9 @@ const (
 	ChartAnnotationControlPlaneProviders = "hmc.mirantis.com/control-plane-providers"
 )
 
-// TemplateSpecMixin is a Template configuration common for all Template types
-type TemplateSpecMixin struct {
+// TemplateSpecCommon is a Template configuration common for all Template types
+type TemplateSpecCommon struct {
 	// Helm holds a reference to a Helm chart representing the HMC template
-	// +kubebuilder:validation:Required
```
eromanova, line 34: ditto

Kshatrix (approved):

eromanova (changes requested):

api/v1alpha1/managedcluster_types.go
Kshatrix: there were no real changes to the generated CRDs, so I assume those are irrelevant

api/v1alpha1/templates_common.go
zerospiel: https://github.com/Mirantis/hmc/pull/382/files#diff-4ed5924873ee60026c42ac4878d7b17dc575fb0ec9bc623d0e0b90d12205eaedR33

api/v1alpha1/managedcluster_types.go
zerospiel: https://github.com/Mirantis/hmc/pull/382/files#diff-98472f9e89833cf1be765a0d990eeb34a6ba3523f073cf7f5c315039149cf680R70

zerospiel: @Kshatrix it is so, but suddenly the schema-validation validates even the empty values and the requests proceeds further to the admission, which is really odd

zerospiel: due to the `admission.Defaulter` implementation, having no `spec.config` leads to further proceeding, and in particular: schema and then the validationwebhook, so in reality, that chang is actually noop

eromanova: Sorry, nevermind. You're right, we can safely remove it. The mutation happens before the schema validation.

PR #391, opened 2024-09-25
Validate available upgrades for managed clusters

### LINKED ISSUE #372: Allow ManagedCluster's upgrade to supported templates only

eromanova:
1. Added new field: `status.availableUpdates` to the `ManagedCluster`.
2. Validate that the new template is in the list of `availableUpdates`.

Closes #372


internal/templateutil/interface.go
```diff
@@ -0,0 +1,36 @@
...
+
+	hmc "github.com/Mirantis/hmc/api/v1alpha1"
+)
+
+// Template is the interface defining a list of methods to interact with templates
+type Template interface {
+	client.Object
+	GetSpec() *hmc.TemplateSpecCommon
+	GetStatus() *hmc.TemplateStatusCommon
+}
```
zerospiel, line 28: nit+optional: used only within the `internal/controller` pkg, so I assume might be reverted back to the packaged and be unexported

internal/templateutil/interface.go
```diff
@@ -0,0 +1,36 @@
...
+
+func IsManagedByHMC(template Template) bool {
+	labels := template.GetLabels()
+	if labels == nil {
+		return false
+	}
+	return labels[hmc.HMCManagedLabelKey] == hmc.HMCManagedLabelValue
```
zerospiel, line 35: I think just

```go
return template.GetLabels()[hmc.HMCManagedLabelKey] == hmc.HMCManagedLabelValue
```

would be enough

internal/templateutil/state.go
```diff
@@ -105,34 +100,25 @@ func ParseAccessRules(ctx context.Context, cl client.Client, rules []hmc.AccessR
...
+			errs = errors.Join(errs, err)
+			continue
+		}
+		supportedServiceTemplates, err := getSupportedTemplates(ctx, cl, rule, hmc.ServiceTemplateKind)
+		if err != nil {
+			errs = errors.Join(errs, err)
+			continue
+		}
 		var clusterTemplates []string
 		var serviceTemplates []string
```
zerospiel, line 114: nit: i suggest preallocating both the slices with the corresponding lengths

internal/templateutil/state.go
```diff
@@ -193,3 +206,35 @@ func getTargetNamespaces(ctx context.Context, cl client.Client, targetNamespaces
...
+
+func getSupportedTemplates(ctx context.Context, cl client.Client, accessRule hmc.AccessRule, templateKind hmc.TemplateKind) ([]hmc.SupportedTemplate, error) {
+	switch templateKind {
+	case hmc.ClusterTemplateKind:
+		var result []hmc.SupportedTemplate
+		for _, chainName := range accessRule.ClusterTemplateChains {
+			chain := &hmc.ClusterTemplateChain{}
+			err := cl.Get(ctx, types.NamespacedName{
+				Name: chainName,
+			}, chain)
```
zerospiel, line 218: I suggest calling the `list` once, and then filtering required names rather than fetching the API `N` times

Besides, I think the `list` could be called just once before looping over the `rules`, in total that'd save `NxM-1` calls for each of the kinds, which might require creating 1/2 (depending on generics usage) extra small functions to list but I believe it worths it

internal/templateutil/state.go
```diff
@@ -193,3 +206,35 @@ func getTargetNamespaces(ctx context.Context, cl client.Client, targetNamespaces
...
+			result = append(result, chain.Spec.SupportedTemplates...)
+		}
+		return result, nil
+	case hmc.ServiceTemplateKind:
+		var result []hmc.SupportedTemplate
+		for _, chainName := range accessRule.ServiceTemplateChains {
+			chain := &hmc.ServiceTemplateChain{}
+			err := cl.Get(ctx, types.NamespacedName{
+				Name: chainName,
+			}, chain)
```
zerospiel, line 231: ditto

internal/templateutil/state.go
```diff
@@ -193,3 +206,35 @@ func getTargetNamespaces(ctx context.Context, cl client.Client, targetNamespaces
...
+			if err != nil {
+				return nil, err
+			}
+			result = append(result, chain.Spec.SupportedTemplates...)
+		}
+		return result, nil
+	}
+	return nil, fmt.Errorf("invalid template kind. Supported values: %s and %s", hmc.ClusterTemplateKind, hmc.ServiceTemplateKind)
```
zerospiel, line 239: nit: i'd personally put under the `default` case but in general no difference

internal/templateutil/state.go
```diff
@@ -160,6 +146,33 @@ func ParseAccessRules(ctx context.Context, cl client.Client, rules []hmc.AccessR
...
+	if len(tmList.Items) == 0 {
+		return false, fmt.Errorf("TemplateManagement is not found")
+	}
+	for _, accessRule := range tmList.Items[0].Spec.AccessRules {
+		supportedTemplates, err := getSupportedTemplates(ctx, cl, accessRule, templateKind)
+		if err != nil {
+			return false, err
+		}
+		for _, supportedTemplate := range supportedTemplates {
+			if supportedTemplate.Name == source {
```
zerospiel, line 164: style: the statement could be reverted in order to reduce levels of nesting

internal/templateutil/state.go
```diff
@@ -160,6 +146,33 @@ func ParseAccessRules(ctx context.Context, cl client.Client, rules []hmc.AccessR
...
+				for _, availableUpgrade := range supportedTemplate.AvailableUpgrades {
+					if availableUpgrade.Name == target {
+						return true, nil
+					}
+				}
+			}
```
zerospiel, line 170: it may be replaced with the `slices` package function `Contains` call

internal/webhook/managedcluster_webhook.go
```diff
@@ -71,18 +72,35 @@ func (v *ManagedClusterValidator) ValidateCreate(ctx context.Context, obj runtim
...
-	template, err := v.getManagedClusterTemplate(ctx, newManagedCluster.Namespace, newManagedCluster.Spec.Template)
-	if err != nil {
-		return nil, fmt.Errorf("%s: %v", errInvalidManagedCluster, err)
-	}
-	err = v.isTemplateValid(ctx, template)
-	if err != nil {
-		return nil, fmt.Errorf("%s: %v", errInvalidManagedCluster, err)
+	oldTemplateName := oldManagedCluster.Spec.Template
+	newTemplateName := newManagedCluster.Spec.Template
+	if oldTemplateName != newTemplateName {
```
zerospiel, line 86: style:

```go
if oldTemplateName == newTemplateName { return nil, nil };
// the logic
```

internal/templateutil/state.go
```diff
@@ -160,6 +146,33 @@ func ParseAccessRules(ctx context.Context, cl client.Client, rules []hmc.AccessR
...
+func IsAvailableForUpgrade(ctx context.Context, cl client.Client, templateKind hmc.TemplateKind, source, target string) (bool, error) {
+	tmList := &hmc.TemplateManagementList{}
+	err := cl.List(ctx, tmList)
+	if err != nil {
+		return false, err
+	}
+	if len(tmList.Items) == 0 {
+		return false, fmt.Errorf("TemplateManagement is not found")
+	}
+	for _, accessRule := range tmList.Items[0].Spec.AccessRules {
```
zerospiel, line 158: I am not in the context, why only the first entry is being looped over? Would it be a good idea to comment on the idea behind it?

internal/webhook/managedcluster_webhook.go
```diff
@@ -71,18 +72,35 @@ func (v *ManagedClusterValidator) ValidateCreate(ctx context.Context, obj runtim
...
+		if !newManagedCluster.Spec.DryRun && templateutil.IsManagedByHMC(template) {
+			availableForUpgrade, err := templateutil.IsAvailableForUpgrade(ctx, v.Client, v1alpha1.ClusterTemplateKind, oldTemplateName, newTemplateName)
+			if err != nil {
+				return nil, fmt.Errorf("failed to check if the ClusterTemplate %s is available for the upgrade from %s: %w", newTemplateName, oldTemplateName, err)
+			}
+			if !availableForUpgrade {
+				return nil, fmt.Errorf("ClusterTemplate %s is not available for the upgrade from %s", newTemplateName, oldTemplateName)
+			}
+		}
+		err = v.isTemplateValid(ctx, template)
```
zerospiel, line 100: validation might be executed before the check on the eligibility for the upgrade (technically it might save a couple of network calls)

internal/webhook/managedcluster_webhook.go
```diff
@@ -71,18 +72,35 @@ func (v *ManagedClusterValidator) ValidateCreate(ctx context.Context, obj runtim
...
+		if err != nil {
+			return nil, fmt.Errorf("%s: %v", errInvalidManagedCluster, err)
+		}
+		if !newManagedCluster.Spec.DryRun && templateutil.IsManagedByHMC(template) {
+			availableForUpgrade, err := templateutil.IsAvailableForUpgrade(ctx, v.Client, v1alpha1.ClusterTemplateKind, oldTemplateName, newTemplateName)
+			if err != nil {
+				return nil, fmt.Errorf("failed to check if the ClusterTemplate %s is available for the upgrade from %s: %w", newTemplateName, oldTemplateName, err)
+			}
+			if !availableForUpgrade {
+				return nil, fmt.Errorf("ClusterTemplate %s is not available for the upgrade from %s", newTemplateName, oldTemplateName)
```
zerospiel, line 97: nit+optional: i see the API wrappers over errors above `NewBadRequest` and I think the errors might also be wrapped, e.g. in this particular line if it is not available for the upgrade, then forbidden or something. It is more or less about the UX

internal/webhook/managedcluster_webhook_test.go
```diff
@@ -28,12 +28,28 @@ import (
...
 var (
-	testTemplateName = "template-test"
-	testNamespace    = "test"
+	testNamespace = "test"
+
+	testTemplateName      = "template-test"
+	templateUpgradeSource = "template-1-0-0"
+	templateUpgradeTarget = "template-1-0-1"
+
+	ctChainName = "test"
```
zerospiel, line 52: i think those may be separated into the `const` section to avoid any accidental data changes

zerospiel (changes requested):

internal/templateutil/state.go
eromanova: Oh, no. You're right. I looked at the old code.

internal/templateutil/state.go
zerospiel: no comparable types or extra code is required:

```go
				if slices.ContainsFunc(supportedTemplate.AvailableUpgrades, func(au hmc.AvailableUpgrade) bool {
					return target == au.Name
				}) {
					return true, nil
				}
```

api/v1alpha1/templates_common.go
```diff
@@ -26,6 +26,9 @@ const (
 	ChartAnnotationBootstrapProviders = "hmc.mirantis.com/bootstrap-providers"
 	// ChartAnnotationControlPlaneProviders is an annotation containing the CAPI control plane providers associated with Template.
 	ChartAnnotationControlPlaneProviders = "hmc.mirantis.com/control-plane-providers"
+
+	ClusterTemplateKind string = "ClusterTemplate"
+	ServiceTemplateKind string = "ServiceTemplate"
```
zerospiel, line 31:
```suggestion
	ClusterTemplateKind = "ClusterTemplate"
	ServiceTemplateKind = "ServiceTemplate"
```

zerospiel (approved):
lgtm

internal/webhook/managedcluster_webhook.go
```diff
@@ -84,6 +97,15 @@ func (v *ManagedClusterValidator) ValidateUpdate(ctx context.Context, _ runtime.
 	if err != nil {
 		return nil, fmt.Errorf("%s: %v", errInvalidManagedCluster, err)
 	}
+	if !newManagedCluster.Spec.DryRun && templateutil.IsManagedByHMC(template) {
```
Kshatrix, line 100: We wanted to have a list of available upgrades in the managedcluster status and only check that the requested template is in the list in admission

Kshatrix (changes requested):

internal/controller/managedcluster_controller.go
```diff
@@ -993,6 +1000,33 @@ func setIdentityHelmValues(values *apiextensionsv1.JSON, idRef *corev1.ObjectRef
 	}, nil
 }
 
+func (r *ManagedClusterReconciler) setAvailableUpgrades(ctx context.Context, managedCluster *hmc.ManagedCluster, template *hmc.ClusterTemplate) error {
+	if template == nil || template.Labels[hmc.HMCManagedLabelKey] != hmc.HMCManagedLabelValue {
```
Kshatrix, line 1004: I don't think we need to rely on these labels. And therefore I don't think we need the distinction between an empty list and nil here. managecluster-controller should not be aware of any subsequent validations and by default we should only allow upgrades within the chains. We might want to disable this validation on global or per-namespace basis though

Kshatrix (changes requested):

internal/controller/managedcluster_controller.go
```diff
@@ -1000,6 +1007,29 @@ func setIdentityHelmValues(values *apiextensionsv1.JSON, idRef *corev1.ObjectRef
 	}, nil
 }
 
+func (r *ManagedClusterReconciler) setAvailableUpgrades(ctx context.Context, managedCluster *hmc.ManagedCluster, template *hmc.ClusterTemplate) error {
+	if template == nil {
+		return nil
+	}
+	chainName := template.Labels[hmc.HMCManagedByChainLabelKey]
```
Kshatrix, line 1014: we might have multiple chains containing the same template

api/v1alpha1/managedcluster_types.go
```diff
@@ -95,8 +95,13 @@ type ManagedClusterStatus struct {
...
 	// provided by the corresponding ClusterTemplate.
 	KubernetesVersion string `json:"k8sVersion,omitempty"`
-	// Conditions contains details for the current state of the ManagedCluster
+	// Conditions contains details for the current state of the ManagedCluster.
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// AvailableUpgrades is the list of ClusterTemplate names to which
+	// this cluster can be upgraded. It can be an empty array, which means no upgrades are
+	// available.
+	AvailableUpgrades []AvailableUpgrade `json:"availableUpgrades,omitempty"`
```
zerospiel, line 104: do we expect to extend in some way the `availableUpgrade` struct? at the moment it seems it the field might be a simple slice of strings. wdyt?

internal/controller/managedcluster_controller.go
```diff
@@ -1019,5 +1058,34 @@ func (r *ManagedClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+							NamespacedName: client.ObjectKey{
+								Namespace: cluster.Namespace,
+								Name:      cluster.Name,
+							},
+						})
+					}
+				}
+				return req
+			}),
+		).
```
zerospiel, line 1089: are there limited event types that we should watch or do we need all of them? at first glance, it seems that the `generic` and `delete` events are redundant in the watch

nits: `types.` -> `client.`; `reconcile.` -> `ctrl.`; instead of reflection just a simple `chain.Name` (instead of the `o.GetName()`) may be called (and other fields if any are required) where applicable

zerospiel (commented):
I left a couple questions, if they are not applicable, then pls resolve them and then LGTM

zerospiel (approved):

Kshatrix (approved):

PR #395, opened 2024-09-25
Modify Telemetry Payload Interval

briantd:
Objective
* Increase the interval between heartbeats from every 10 minutes to daily

### DIFF
--- a/internal/telemetry/tracker.go
+++ b/internal/telemetry/tracker.go
@@ -34,7 +34,8 @@ type Tracker struct {
 	SystemNamespace string
 }
 
-const interval = 10 * time.Minute
+// Daily Heartbeat
+const interval = 1440 * time.Minute
 
 func (t *Tracker) Start(ctx context.Context) error {
 	timer := time.NewTimer(0)
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -18,7 +18,7 @@ import (
 	"github.com/segmentio/analytics-go"
 )
 
-var segmentToken = ""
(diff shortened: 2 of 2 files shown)


internal/telemetry/tracker.go
```diff
@@ -34,7 +34,8 @@ type Tracker struct {
 	SystemNamespace string
 }
 
-const interval = 10 * time.Minute
+// Daily Heartbeat
+const interval = 1440 * time.Minute
```
a13x5, line 38: Use `24 * time.Hour`

a13x5 (changes requested):
Hardcode of any tokens is bad practice. 
Please add token to the GH secrets and reference it as env variable (`SEGMENT_TOKEN`) in the build workflow

internal/telemetry/tracker.go
briantd: @a13x5 -- Thanks for the guidance.

@Schnitzel requested adding a default token so that if another party cloned the repo and built the binary that we'd [potentially] receive telemetry. If I understand your suggestion to use GH secrets, image builds done outside of at GH workflow would not be able to emit telemetry.

Schnitzel:
> Hardcode of any tokens is bad practice. Please add token to the GH secrets and reference it as env variable (`SEGMENT_TOKEN`) in the build workflow

as discussed in slack: this is intentional and will allow us to track telemetry for locally built 2A installations. the token itself is not a secret and can be easily distributed (we already distribute it to built versions of 2A, so if somebody knows how segment works they could figure out the token also from the built version)

PR #405, opened 2024-09-27
CCM/CSI credentials propagation

### LINKED ISSUE #324: CCM/CSI credentials propagation
Credentials that are required for CCM/CSI on a managed cluster must be collected and passed via Kubernetes API by creating a required secrets.
Highly provider dependent.

### LINKED ISSUE #325: Adjust cluster templates according to new CCM/CSI secret propagation
- No cloud-init credentials
- CCM/CSI accept pre-defined secret as a source for credentials

a13x5:
Resolves #324
Resolves #325

Notes:

1. This must be merged **after** #342. When #342 merged `Add Credential CR and controller` will be removed from this PR. Now it's cherry-picked here for clarity.
2. Azure CCM Helm chart has issue (kubernetes-sigs/cloud-provider-azure#7188) with controller flags, so it isn't working with secret. CSI works fine. **Please don't merge this PR before it's fixed!**
3. I'm not sure if we should fail on unsupported infrastructure provider. Now the `ManagedCluster` will not transition to Ready when some other (than AWS, Azure or vSphere) infrastructure provider will be used. Probably this isn't right.


internal/controller/managedcluster_controller.go
```diff
@@ -683,6 +695,357 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
...
+func (r *ManagedClusterReconciler) reconcileCredentialPropagation(ctx context.Context, l logr.Logger, managedCluster *hmc.ManagedCluster) (ctrl.Result, error) {
+	l.Info("Reconciling CCM credentials propagation")
+	var err error
+	defer func() {
+		err = errors.Join(err, r.updateStatus(ctx, managedCluster))
+	}()
+
+	providers, err := r.getInfraProviders(ctx, managedCluster.Namespace, managedCluster.Spec.Template)
+	if err != nil {
+		managedCluster.Status.CCMSecretReconciled = false
```
Kshatrix, line 707: you can initially set it to False in the beginning of the reconcile and only change it to True once it succeed

internal/controller/managedcluster_controller.go
```diff
@@ -365,6 +373,10 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
 			return ctrl.Result{RequeueAfter: DefaultRequeueInterval}, nil
 		}
 
+		if !managedCluster.Status.CCMSecretReconciled {
+			return r.reconcileCredentialPropagation(ctx, l, managedCluster)
```
Kshatrix, line 377: let's not return here to not skip services update

internal/controller/managedcluster_controller.go
```diff
@@ -683,6 +695,357 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
 	return len(itemsList.Items) != 0, nil
 }
 
+func (r *ManagedClusterReconciler) reconcileCredentialPropagation(ctx context.Context, l logr.Logger, managedCluster *hmc.ManagedCluster) (ctrl.Result, error) {
+	l.Info("Reconciling CCM credentials propagation")
+	var err error
+	defer func() {
```
Kshatrix, line 701: it's deffered already, you don't need another one

internal/controller/managedcluster_controller.go
```diff
@@ -365,6 +373,10 @@ func (r *ManagedClusterReconciler) Update(ctx context.Context, l logr.Logger, ma
 			return ctrl.Result{RequeueAfter: DefaultRequeueInterval}, nil
 		}
 
+		if !managedCluster.Status.CCMSecretReconciled {
```
Kshatrix, line 376: if we update credentials, new secret won't be propagated. if you want to skip this part when it was reconciled successfully, you should track both `name` and `generation` of the credential object (+ underlying ClusterIdentity reference could be updated with no sign on the credential object right now)

Kshatrix (changes requested):

internal/controller/managedcluster_controller.go
a13x5: I had an issue with conditions not populating properly without it in this scope. I will recheck

internal/controller/managedcluster_controller.go
a13x5: Yep, this is a problem. I will rewrite that part, and probably get rid of the `CCMSecretReconciled` field completely

internal/controller/managedcluster_controller.go
zerospiel: i think it is not required at all regarding [this comment](https://github.com/Mirantis/hmc/pull/405/files#r1801166440), having detailed conditions along in the status

internal/controller/managedcluster_controller.go
```diff
@@ -683,6 +695,357 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
 	return len(itemsList.Items) != 0, nil
 }
 
+func (r *ManagedClusterReconciler) reconcileCredentialPropagation(ctx context.Context, l logr.Logger, managedCluster *hmc.ManagedCluster) (ctrl.Result, error) {
```
zerospiel, line 698: logger is always in the context, so the argument is redundant

internal/controller/managedcluster_controller.go
```diff
@@ -683,6 +695,357 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
...
+		default:
+			managedCluster.Status.CCMSecretReconciled = true
+			errMsg := fmt.Sprintf("unsupported infrastructure provider %s", provider)
+			apimeta.SetStatusCondition(managedCluster.GetConditions(), metav1.Condition{
+				Type:    hmc.CredentialsPropagatedCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  hmc.FailedReason,
+				Message: errMsg,
+			})
+			return ctrl.Result{}, errors.New(errMsg)
```
zerospiel, line 778: i think such functions/methods returns should be given without errors because it'll just clog the event queue along with logs with unnecessary same events/messages, I think setting the conditions is good enough 

applicable elsewhere if any

internal/controller/managedcluster_controller.go
```diff
@@ -683,6 +695,357 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
...
+
+func makeConfigMap(name, namespace string, data map[string]string) *corev1.ConfigMap {
+	c := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Data: data,
+	}
+	c.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("ConfigMap"))
```
zerospiel, line 1027: why do we need it? will it mitigate some runtime checks performed by the controller-gen client, the client is not partial-metadata-scoped IINM

same for the `makeSecret` symbol

internal/controller/managedcluster_controller.go
```diff
@@ -15,10 +15,12 @@
 package controller
```
zerospiel, line 15: general suggestion and a couple of nit comments:

* I think we can separate the new code into a new package, IMO it'd be easier to manage (it does not matter whether it would be implemented on interfaces or just public fixed structs)
* `kube-system` consts and etc. -> `metav1.NamespaceSystem`
* complicated list options might be simplified with `"sigs.k8s.io/controller-runtime/pkg/client".InNamespace`, `"sigs.k8s.io/controller-runtime/pkg/client".HasLabels`, `"sigs.k8s.io/controller-runtime/pkg/client".Limit` and etc.
* some errors might be wrapped (`%w`) where it might make more sense than a plain string (`%s`)
* some variables (such as secret names) might be instantiated as constants
* some error-typed return arguments might be annotated/wrapped with an understandable message (`return err` -> `return fmt.Errorf("failed to perform some action during another event: %w", err)` to ease potential debugging 
* I think some of the configurations (e.g. azure/vsphere/etc.) might be commented on for future readers, now it can be only understandable by users with an in-depth experience, i.e. I know why `insecureFlag` is required for the vSphere CSI but one cannot
* personal comment (so skip it if it's not okay for you): I find reading code with `\n` between logical blocks much more convenient than reading uninterrupted huge blocks of code

internal/controller/managedcluster_controller.go
```diff
@@ -683,6 +695,357 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
...
+	}
+	return makeSecret(secretName, ns, secretData),
+		makeConfigMap("cloud-config", ns, cmData),
+		nil
+}
+
+func generateVSphereCSISecret(vCl *capv.VSphereCluster, vScrt *corev1.Secret, vMa *capv.VSphereMachine) (*corev1.Secret, error) {
+	csiCfg := `
+[Global]
+cluster-id = "{{ .ClusterID }}"
```
zerospiel, line 952: I'm wondering, don't we need to set an empty `thumbprint` here explicitly? I've seen such a setting here and there. And the same question goes regarding the CPI configuration in the `global` section

internal/controller/managedcluster_controller.go
```diff
@@ -683,6 +695,357 @@ func (r *ManagedClusterReconciler) machinesAvailable(ctx context.Context, namesp
...
+		},
+		"vcenter": map[string]any{
+			vCl.Spec.Server: map[string]any{
+				"server": vCl.Spec.Server,
+				"datacenters": []string{
+					vMa.Spec.Datacenter,
+				},
+			},
+		},
+		"labels": map[string]any{
```
zerospiel, line 932: I am not aware of such labels being set, why do we need them and with such values?

internal/controller/managedcluster_controller.go
a13x5: It's optional see [CSI docs](https://docs.vmware.com/en/VMware-vSphere-Container-Storage-Plug-in/2.0/vmware-vsphere-csp-getting-started/GUID-BFF39F1D-F70A-4360-ABC9-85BDAFBE8864.html).
The CAPV made it mandatory in their configs for some unknown reasons

internal/controller/managedcluster_controller.go
a13x5: These are not k8s labels, it's labels for use in multi datacenter or multi vcenter environment to mark VMs or other stuff. Arbitrary values could passed here. 
I took the default ones they put in [the chart](https://github.com/kubernetes/cloud-provider-vsphere/blob/master/charts/vsphere-cpi/values.yaml#L12)

internal/controller/managedcluster_controller.go
a13x5: @zerospiel That's what I meant when put this note above 
> I'm not sure if we should fail on unsupported infrastructure provider.

I can just skip return _here_ completely. 

If you mean that `reconcileCredentialPropagation` method shouldn't re-trigger reconcile then I don't understand it and it makes not sense to me.

internal/controller/managedcluster_controller.go
a13x5: The object will be with empty GVK without it. So you'll have something like:
`invalid object type: /, Kind=`

internal/controller/managedcluster_controller.go
a13x5: Rechecked and it looks I was wrong or missed something. So this is indeed extra, removed.

internal/controller/managedcluster_controller.go
a13x5: I will try to address these in more or less the same order:
1. I don't see benefits of a separate package: it makes some methods and data a bit more complicated and benefit is near zero. Probably we can move new code to a separate file, but leave in the same package.
2. Done
3. Chain method calls looks more complicated that passing struct as an argument. But I guess it's more like a personal preference. I will leave it as is if you not insist on that.
4. Please provide exact places where it will make more sense
5. If you're about CCM secret names I don't see why they should be in global scope. I specifically reduced their scope to one function.
6. Please give me specific example of such place
7. It will be a lot to comment (I don't like this idea). Maybe it's better to document it somewhere in the actual docs?
8. I didn't think about it and tbh I don't even see what was bothering you

internal/controller/managedcluster_controller.go
a13x5: @Kshatrix @zerospiel I removed the `CCMSecretReconciled` variable completely and now secrets will be applied on the managed cluster on every reconcile.

I would like to actually not do that, but as Andrey mentioned earlier it will require some changes in Credential controller to properly track changes. 

So if you're ok with the each reconcile thing we can leave it as is for now and I will create task for tracking credential resources properly. If not I will have to do the tracking task first and I would prefer to do it in a separate PR (this one is big enough already)

internal/controller/managedcluster_controller.go
zerospiel: Correct, skip the error return only here, nothing else has been meant since there is no more such code that'd behave in the same way

internal/controller/managedcluster_controller.go
zerospiel: 1. The following points actually:
* compilation point of view: having a gigantic package slows the compile time
* LSP point of view: having a gigantic package with lots of symbols slows the initiation and the following search
* code/developer point of view: clean code with separated responsibilities; symbols like `internal/controller.makeClientFromSecret` or `internal/controller.applyCCMConfigs` (and others that have been introduced) are used purely within the managed-cluster-ctrl, and a developer may misuse them (we have almost a dozen of controllers in the package) or spend time to read the code of the symbols whether those might be reused or it is better to implement another function/method at the moment

This point is not mandatory to implement within this PR and could be implemented as part of a follow-up PR. I also might be wrong with the aforementioned points thus we can ask for another opinion on that point

3. Right, that's mostly personal preferences, for me personally the 
```go
&client.ListOptions{
 // ...
  LabelSelector: labels.SelectorFromSet(map[string]string{
    hmc.ClusterNameLabelKey: managedCluster.Name,
  }),
  // ...
},
```
looks not just as passing a struct compared to the following:
```go
client.HasLabels{hmc.ClusterNameLabelKey: managedCluster.Name}
```
4. Sure, for example, the 
```go
fmt.Errorf("failed to get cluster providers for cluster %s/%s: %s" /* ... */
```
converts the underlying error (which might be 4xx/5xx) to a plain string and a caller won't be able to extract any details from it if it wanted to. This is mostly speculation, but if that happens and a caller wants to deconstruct the returned error, a developer should then change other pieces of code from `%s` to the `%w`. There are different places where an error converts to a plain string, most of them — API calls (`get`/`list`)
5. I did not mention the **global** scope, no one stops from converting `secretName := "vsphere-cloud-secret"` to the e.g. `const vsphereCloudSecretName = "vsphere-cloud-secret"` within the same line (applicable elsewhere **if any**)
6. That bullet point mostly relates to the `makeClientFromSecret` function, each error could be wrapped with a detailed message but you may skip this if `failed to create k8s client: <underlying message>` is more suitable for you than e.g. `failed to create k8s client: failed to get REST config from the <name> secret: <underlying message>`.
7. Sounds good and useful, especially for potential readers of the docs, they will know the defaults and why we have such defaults 
8. Nvm, I see now more new lines between logical blocks have been added, it's okay now

internal/controller/managedcluster_controller.go
zerospiel: Sounds ok to me 👌

internal/controller/managedcluster_controller.go
zerospiel: Agreed with @Kshatrix upon a follow-up PR

Kshatrix (approved):

internal/controller/managedcluster_controller.go
a13x5: Follow up task on that #529

internal/controller/managedcluster_controller.go
a13x5: Separate task on that #530
@zerospiel If you want to add something there feel free to do so. Thanks

PR #416, opened 2024-10-01
Add validation for ServiceTemplateChain

wahabmk:
Closes https://github.com/Mirantis/hmc/issues/316.

Originally I wrote the create validation for [Cluster/Service]TemplateChain to check if the templates being referenced in `.Spec.SupportedTemplates` actually existed in the same namespace. However, looking at how https://github.com/Mirantis/hmc/pull/407 has been implemented I dumped that so there is no validation check happening for create now.

### DIFF
--- a/internal/webhook/template_webhook.go
+++ b/internal/webhook/template_webhook.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -119,7 +120,39 @@ func (*ServiceTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
-func (*ServiceTemplateValidator) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
+func (in *ServiceTemplateValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	tmpl, ok := obj.(*v1alpha1.ServiceTemplate)
+	if !ok {
+		return admission.Warnings{"Wrong object"}, apierrors.NewBadRequest(fmt.Sprintf("expected ServiceTemplate but got a %T", obj))
+	}
+
+	var token string
+
+	for {
+		managedClusters := &v1alpha1.ManagedClusterList{}
+		if err := in.Client.List(ctx, managedClusters, &client.ListOptions{
+			Namespace: tmpl.Namespace,
+			Limit:     100,
+			Raw: &metav1.ListOptions{
+				Continue: token,
+			},
+		}); err != nil {
+			return nil, err
+		}
+
+		for _, m := range managedClusters.Items {
+			for _, s := range m.Spec.Services {
+				if s.Template == tmpl.Name {
+					return admission.Warnings{"The ServiceTemplate object can't be removed if ManagedCluster objects referencing it still exist"}, errTemplateDeletionForbidden
+				}
+			}
+		}
+
+		if token = managedClusters.Continue; token == "" {
+			break
+		}
+	}
+
 	return nil, nil
 }
 
--- a/internal/webhook/template_webhook.go
+++ b/internal/webhook/template_webhook.go
@@ -18,10 +18,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"slices"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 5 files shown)


kylewuolle (dismissed):

internal/webhook/template_webhook.go
```diff
@@ -119,7 +120,39 @@ func (*ServiceTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
...
+		return admission.Warnings{"Wrong object"}, apierrors.NewBadRequest(fmt.Sprintf("expected ServiceTemplate but got a %T", obj))
+	}
+
+	var token string
+
+	for {
+		managedClusters := &v1alpha1.ManagedClusterList{}
+		if err := in.Client.List(ctx, managedClusters, &client.ListOptions{
+			Namespace: tmpl.Namespace,
+			Limit:     100,
```
slysunkin, line 135: Why do we need to check the first 100 of clusters? What if we have more (it is estimated to 500 clusters on production) and the template is used in cluster #499?

slysunkin: Can we reuse template key, like in here:
```
	managedClusters := &v1alpha1.ManagedClusterList{}
	listOptions := client.ListOptions{
		FieldSelector: fields.SelectorFromSet(fields.Set{v1alpha1.TemplateKey: template.Name}),
		Limit:         1,
		Namespace:     template.Namespace,
	}
	err := v.Client.List(ctx, managedClusters, &listOptions)
```

wahabmk: > Why do we need to check the first 100 of clusters? What if we have more (it is estimated to 500 clusters on production) and the template is used in cluster #499?

It is actually fetching list of `ManagedCluster` objects using pagination. 100 is just an arbitrary page size that I am using to paginate.

> Can we reuse template key, like in here:

That would be nice. I need to match `template` from list of templates under `.Spec.Services`. Maybe there is a way to select fields from objects that are part of an array, but I could not find it:
```yaml
kind: ManagedCluster
. . .
spec:
  . . .
  services:
    - template: kyverno
      name: kyverno
      namespace: kyverno
    - template: ingress-nginx
      name: ingress-nginx
      namespace: ingress-nginx
```

internal/webhook/template_webhook.go
```diff
@@ -119,7 +120,39 @@ func (*ServiceTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
...
-func (*ServiceTemplateValidator) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
+func (in *ServiceTemplateValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	tmpl, ok := obj.(*v1alpha1.ServiceTemplate)
+	if !ok {
+		return admission.Warnings{"Wrong object"}, apierrors.NewBadRequest(fmt.Sprintf("expected ServiceTemplate but got a %T", obj))
+	}
+
+	var token string
+
+	for {
```
zerospiel, line 131: the pattern looks odd to me, it definitely correct but I still have a question: why don't we just list everything as a single API call? the `delete` event isn't expected to happen frequently

internal/webhook/template_webhook.go
```diff
@@ -119,7 +120,39 @@ func (*ServiceTemplateValidator) ValidateUpdate(_ context.Context, _ runtime.Obj
...
+			Raw: &metav1.ListOptions{
+				Continue: token,
+			},
+		}); err != nil {
+			return nil, err
+		}
+
+		for _, m := range managedClusters.Items {
+			for _, s := range m.Spec.Services {
+				if s.Template == tmpl.Name {
```
zerospiel, line 145: could be interchanged with a single call to `slices.Contains*`, probably it'd be easier to read

internal/webhook/template_webhook.go
zerospiel: Wahab is right, there is no way to use a selector on such a field, containing a list of items. The pattern with pagination is correct too

Kshatrix: I think it's quite possible if the extracting function returns multiple keys. and in that case equality means that one of them is matching the value.

zerospiel: do you mean trying to construct a selector with something like `[?(@.template == "kyverno")]`? i am not sure whether it's feasible with the structs/methods provided by the k8s packages (or I am not aware of them), and if so constructing a custom selector for such a simple call along with the rare event (`delete`) does not make too much sense (probably it worths to set a `TODO` comment having other pieces of code that might require optimizations)

Kshatrix: no, you need a custom [IndexerFunc](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/interfaces.go#L195) anyways. Any it can return multiple keys. Quite easy to implement and verify.

internal/webhook/template_webhook.go
wahabmk: So I want to make sure that the `ServiceTemplate` object that the user calls delete on is not used by any `ManagedCluster` objects. I don't know how many `ManagedCluster` objects there could be in the management cluster, so therefore, I thought it is safer (w.r.t memory) to use pagination to list over all `ManagedCluster` objects.

internal/webhook/template_webhook.go
wahabmk: I see that the `IndexerFunc` has been used here:
https://github.com/Mirantis/hmc/blob/9b9931d2292f0c0f133835ef3c29fe54c1f71713/api/v1alpha1/common.go#L56-L60 
I think it will be better if I use it as I can get rid of the pagination that way, but I don't understand how to specify template that is within a list of services as the key/field. Something like `.spec.services[].template`?

wahabmk: Probably something like this?
```go

func SetupManagedClusterServicesIndexer(ctx context.Context, mgr ctrl.Manager) error {
	return mgr.GetFieldIndexer().IndexField(ctx, &ManagedCluster{}, ".spec.services", func(o client.Object) []string {
		cluster, ok := o.(*ManagedCluster)
		if !ok {
			return nil
		}

		services := []string
		for _, s := range cluster.Spec.Services {
			services = append(services, s.Template)
		}

		return services
	})
}
```

Kshatrix (approved):

PR #422, opened 2024-10-02
Compatibility attributes: add API support, propagation and checks

### LINKED ISSUE #400: Compatibility attributes: implement enforcement
Part of the #353

zerospiel:
* Cluster/Service/Provider Templates now have support for compatibility attributes
* Managed/Management Clusters now have support to report the compatibility attributes
* amends to the templates controller in regards of the API changes
* amends to other parts of the code
* fill out statuses with the new attributes
  for managedclustes,managements,*templates
  resources
* fix a couple of logical errors
* check compatibility attributes in the
  admission controller
* tests
* fix typos

Closes #400 #354


api/v1alpha1/providertemplate_types.go
```diff
@@ -15,24 +15,82 @@
...
 
+// ChartAnnotationCAPIVersion is an annotation containing the CAPI exact version in the SemVer format associated with a ProviderTemplate.
+const ChartAnnotationCAPIVersion = "hmc.mirantis.com/capi-version"
+
 // ProviderTemplateSpec defines the desired state of ProviderTemplate
 type ProviderTemplateSpec struct {
-	TemplateSpecCommon `json:",inline"`
+	Helm HelmSpec `json:"helm"`
+	// Compatible CAPI provider version set in the SemVer format.
+	CAPIVersion string `json:"capiVersion,omitempty"`
```
zerospiel, line 31: At the moment this field is a noop since I don't quite understand in which entity it should be propagated and/or which entity should consume it and what validations should be implemented

api/v1alpha1/managedcluster_types.go
```diff
@@ -95,17 +94,23 @@ type ManagedClusterSpec struct {
...
+	// DryRun specifies whether the template should be applied after validation or only validated.
+	DryRun bool `json:"dryRun,omitempty"`
 }
 
 // ManagedClusterStatus defines the observed state of ManagedCluster
 type ManagedClusterStatus struct {
+	// Currently compatible K8S version of the cluster. Being set only if
+	// provided by the corresponding ClusterTemplate.
+	KubertenesVersion string `json:"k8sVersion,omitempty"`
+	// Providers represent exposed CAPI providers with constrainted compatibility versions set.
```
Kshatrix, line 111:
```suggestion
	// Providers represent required CAPI providers with constrainted compatibility versions set.
```

Do we have any benefits from duplicating them in the ManagedCluster status though? For validation we could take the referenced template instead

api/v1alpha1/providertemplate_types.go
Kshatrix: It should be used to check compatibility of the core CAPI provider (we have a dedicated field in the Release and Management objects alongside HMC itself) and other providers.

api/v1alpha1/templates_common.go
```diff
@@ -73,3 +76,66 @@ type TemplateValidationStatus struct {
...
+
+const (
+	bootstrapProvidersType providersType = iota
+	controlPlaneProvidersType
+	infrastructureProvidersType
+)
+
+func parseProviders[T any](providersGetter interface{ GetSpecProviders() ProvidersTupled }, typ providersType, annotations map[string]string, validationFn func(string) (T, error)) ([]ProviderTuple, error) {
+	pspec, anno := getProvidersSpecAnno(providersGetter, typ)
+	if len(pspec) > 0 {
```
Kshatrix, line 90: we might potentially have only part of annotations overridden in the spec(we need to inspect both annotations and spec, but prioritize values in the spec over annotations), but it's not that important for now

api/v1alpha1/templates_common.go
```diff
@@ -73,3 +76,66 @@ type TemplateValidationStatus struct {
...
+	}
+
+	var (
+		splitted = strings.Split(providers, ",")
+		pstatus  = make([]ProviderTuple, 0, len(splitted))
+		merr     error
+	)
+	for _, v := range splitted {
+		v = strings.TrimSpace(v)
+		nVerOrC := strings.SplitN(v, " ", 2)
```
Kshatrix, line 106: let's discuss the format separately

api/v1alpha1/providertemplate_types.go
```diff
@@ -15,24 +15,82 @@
...
+	// Represents required CAPI providers with exact compatibility versions set. Should be set if not present in the Helm chart metadata.
```
Kshatrix, line 32: these are providers that exposed/enabled when the template is installed

api/v1alpha1/servicetemplate_types.go
```diff
@@ -15,26 +15,104 @@
...
+	KubertenesConstraint string `json:"k8sConstraint,omitempty"`
+	// Represents required CAPI providers. Should be set if not present in the Helm chart metadata.
+	Providers Providers `json:"providers,omitempty"`
 }
 
 // ServiceTemplateStatus defines the observed state of ServiceTemplate
 type ServiceTemplateStatus struct {
+	// Constraint describing compatible K8S versions of the cluster set in the SemVer format.
+	KubertenesConstraint string `json:"k8sConstraint,omitempty"`
+	// Represents exposed CAPI providers.
```
Kshatrix, line 45:
```suggestion
	// Represents requested CAPI providers.
```

Kshatrix (changes requested):

api/v1alpha1/providertemplate_types.go
zerospiel: Should this comment be related to the `spec`? `Exposed` is already in-use in the `status` field

api/v1alpha1/templates_common.go
zerospiel: reconciled on the `;` to be used as the separator, and I will propagate it to the providers' list without constraints/exact versions, which is now only applicable for the `servicetemplates` to avoid inconsistency

api/v1alpha1/clustertemplate_types.go
```diff
@@ -15,26 +15,86 @@
...
+	ChartAnnotationKubernetesVersion = "hmc.mirantis.com/k8s-version"
+)
 
 // ClusterTemplateSpec defines the desired state of ClusterTemplate
 type ClusterTemplateSpec struct {
-	TemplateSpecCommon `json:",inline"`
+	Helm HelmSpec `json:"helm"`
+	// Compatible K8S version of the cluster set in the SemVer format.
+	KubertenesVersion string `json:"k8sVersion,omitempty"`
+	// Providers represent required CAPI providers with constrainted compatibility versions set. Should be set if not present in the Helm chart metadata.
```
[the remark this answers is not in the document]
zerospiel: Precede annotation or the spec indeed? Because from the currently implemented logic, the spec is preceding annotations

api/v1alpha1/templates_common.go
zerospiel: I have done the part collecting the both, but the same comment: current logic counts values from the spec first, and if absent then values from the annos

zerospiel:
@squizzi @Kshatrix @a13x5 requesting CR, the PR is big enough, and already awaits some time to be reviewed and with each merge, it's harder and harder to keep it up-to-date

internal/webhook/management_webhook.go
```diff
@@ -53,13 +56,91 @@ func (*ManagementValidator) ValidateCreate(_ context.Context, _ runtime.Object)
...
+	if err := v.List(ctx, ptpls); err != nil {
+		return nil, fmt.Errorf("failed to list ProviderTemplates: %v", err)
+	}
+
+	if len(ptpls.Items) == 0 {
+		return nil, nil // nothing to do
+	}
+
+	name2Tpl := make(map[string]*hmcv1alpha1.ProviderTemplate, len(ptpls.Items)-1)
+	capiRequiredVersion := new(semver.Version)
```
Kshatrix, line 95: the currently requested capi template is referenced in the management spec or (if empty) in the Release spec

internal/webhook/management_webhook.go
```diff
@@ -53,13 +56,91 @@ func (*ManagementValidator) ValidateCreate(_ context.Context, _ runtime.Object)
...
+
+		var err error
+		capiRequiredVersion, err = semver.NewVersion(v.Status.CAPIVersion)
+		if err != nil { // should never happen
+			return nil, fmt.Errorf("%s: invalid CAPI version %s in the ProviderTemplate %s to be validated against: %v", invalidMgmtMsg, v.Status.CAPIVersion, v.Name, err)
+		}
+	}
+
+	var wrongVersions error
+	for _, c := range mgmt.Status.Components {
```
Kshatrix, line 114: what do you intent to validate here? core capi update with the existing mgmt providers?

Kshatrix (changes requested):

api/v1alpha1/providertemplate_types.go
```diff
@@ -21,28 +21,35 @@ import (
...
 // ProviderTemplateSpec defines the desired state of ProviderTemplate
 type ProviderTemplateSpec struct {
-	Helm HelmSpec `json:"helm"`
-	// CAPI exact version in the SemVer format required for this ProviderTemplate.
-	// For the CAPI component it is the actual current version of the component
-	// against which other Templates will be validated.
+	Helm HelmSpec `json:"helm,omitempty"`
+	// CAPI exact version in the SemVer format.
+	// Applicable only for the cluster-api ProviderTemplate itself.
 	CAPIVersion string `json:"capiVersion,omitempty"`
```
Kshatrix, line 36: let's make mutually exclusive with CAPIVersionConstraint

internal/webhook/managedcluster_webhook.go
```diff
@@ -54,127 +56,264 @@ var (
...
-			errs = append(errs, fmt.Errorf("one or more required %s providers are not deployed yet: %v", providerType, missing))
+	var (
+		exposedProviders  = management.Status.AvailableProviders
+		requiredProviders = template.Status.Providers
+
+		missingBootstrap, missingCP, missingInfra []string
+		wrongVersionProviders                     map[string][]string
+	)
+
+	// on update we have to validate versions between exact the provider tpl and constraints from the cluster tpl
```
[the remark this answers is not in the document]
zerospiel: I've made the check to be performed independently on the request type (effectively on CREATE/UPDATE) but I am not sure that the validation should be implemented in the template-ctrl since the check requires exposed providers which are only available after management object reconciliation, making the controller dependant on another controller which is not suitable IMO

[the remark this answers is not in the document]
Kshatrix: I think it still makes sense. Templates should not be valid if they can't be used within the system and it's good to know about it before you assign them. The check itself is quite heavy  for admission and moving it to the controller would also allow us to avoid duplication - we need the same verification when we try to distribute templates through the templatemanagement object and when we validate release objects in the release controller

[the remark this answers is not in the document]
zerospiel: Will do in the follow-up PR

Kshatrix (approved):

Kshatrix (approved):

PR #433, opened 2024-10-04
Add Webhooks for ServiceTemplate to ValidatingWebhookConfiguration

wahabmk:
Missed adding this to https://github.com/Mirantis/hmc/pull/416.
Deletes on `ServiceTemplate` now blocked if it is referenced in a `ManagedCluster` object:
```sh
➜  ~ khmc delete servicetemplate ingress-nginx 
Warning: The ServiceTemplate object can't be removed if ManagedCluster objects referencing it still exist
Error from server (Forbidden): admission webhook "validation.servicetemplate.hmc.mirantis.com" denied the request: template deletion is forbidden
```

### DIFF
--- a/internal/webhook/template_webhook.go
+++ b/internal/webhook/template_webhook.go
@@ -94,12 +94,12 @@ type ServiceTemplateValidator struct {
 	client.Client
 }
 
-func (in *ServiceTemplateValidator) SetupWebhookWithManager(mgr ctrl.Manager) error {
-	in.Client = mgr.GetClient()
+func (v *ServiceTemplateValidator) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	v.Client = mgr.GetClient()
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(&v1alpha1.ServiceTemplate{}).
-		WithValidator(in).
-		WithDefaulter(in).
+		WithValidator(v).
+		WithDefaulter(v).
 		Complete()
 }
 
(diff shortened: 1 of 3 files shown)


slysunkin (dismissed):
LGTM

templates/provider/hmc/templates/webhooks.yaml
```diff
@@ -127,6 +127,29 @@ webhooks:
...
+    failurePolicy: Fail
+    matchPolicy: Equivalent
+    name: validation.servicetemplate.hmc.mirantis.com
+    rules:
+      - apiGroups:
+          - hmc.mirantis.com
+        apiVersions:
+          - v1alpha1
+        operations:
+          - CREATE
```
Kshatrix, line 147: please leave only DELETE as we don't have anything else implemented yet

internal/webhook/template_webhook.go
```diff
@@ -144,20 +142,20 @@ func (in *ServiceTemplateValidator) ValidateDelete(ctx context.Context, obj runt
 }
 
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
-func (*ServiceTemplateValidator) Default(_ context.Context, _ runtime.Object) error {
+func (*ServiceTemplateValidator) Default(context.Context, runtime.Object) error {
```
Kshatrix, line 145: probably should be reverted?

Kshatrix (changes requested):

templates/provider/hmc/templates/webhooks.yaml
wahabmk: I should probably remove CREATE and UPDATE from clustertemplate as well because that also only has validation for DELETE implemented.

Kshatrix (approved):

PR #446, opened 2024-10-07
Change ManagedCluster's short names

### LINKED ISSUE #364: Fix ManagedCluster CRD confusing points
I noticed several issues with the `ManagedCluster` CR:

1. Name `ManagedCluster` clashes with the resource `ManagedCluster` of  `containerservice.azure.com/v1api20210501` which is a part of Azure provider. This means that user can't execute `kubectl get managedcluster` and has to user the full name `kubectl get managedcluster.v1alpha1.hmc.mirantis.com`
2. Short names of the `ManagedCluster` are `hmc-deploy` and `deploy` which make confusion even more.

**The task**

Replace short names to be similar to the full name. In the docs if `ManagedCluster` resource is mentioned in scope of kubectl commands it should referenced by a full name with group and version (`managedcluster.v1alpha1.hmc.mirantis.com`).

a13x5:
Resolves #364 
Small fix to get rid of `Deployment` resource legacy.

### DIFF
--- a/api/v1alpha1/managedcluster_types.go
+++ b/api/v1alpha1/managedcluster_types.go
@@ -114,7 +114,7 @@ type ManagedClusterStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:shortName=hmc-deploy;deploy
+// +kubebuilder:resource:shortName=mcluster;mcl
 // +kubebuilder:printcolumn:name="ready",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description="Ready",priority=0
 // +kubebuilder:printcolumn:name="status",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].message",description="Status",priority=0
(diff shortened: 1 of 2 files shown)


api/v1alpha1/managedcluster_types.go
```diff
@@ -114,7 +114,7 @@ type ManagedClusterStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:shortName=hmc-deploy;deploy
+// +kubebuilder:resource:shortName=mcluster;mcl
```
zerospiel, line 117: i have already done this in the #422 but the names are different (`hmcmc;mc`), if the latter are ok maybe we could wait for the merge, otherwise LGTM

zerospiel (approved):

api/v1alpha1/managedcluster_types.go
a13x5: @zerospiel Tbh I don't have any preferences on the naming.
But I guess I will merge this to provide correct changes to the docs which I'm doing now.
Thanks!

PR #464, opened 2024-10-08
Honor the createRelease flag when reconciling

kylewuolle:
Small change to use the createRelease flag and not create the release if the flag is false.


slysunkin (approved):
LGTM

PR #475, opened 2024-10-11
Add 'make airgap-package' target and associated scripts for bundling images for air gap

### LINKED ISSUE #423: Determine best way to bundle and utilize charts and images for airgapped customers
**Acceptance Criteria**
* Documentation exists for using our airgapped bundle and loading charts/images into a registry for use with 2a
* Our release process should be updated to create an airgapped bundle.

squizzi:
This adds new scripts for creating a bundle for airgap customers.  We can easily add a step to the release CI as well, but the resulting tarball of saved images is 4.2G so I decided to not include that for now since we'll need to figure out the best place to host the tarball.

At this time this script deploys a management cluster, waits for all the providers to come up and be ready then gets a list of all the images used and pulls/saves those.

It then iterates `helm template ...` output across each of the k0s extensions we deploy and pulls/saves images affiliated with the extensions.  I've also updated `helm-package` to pull in these extensions during it's packaging process.

```
$ IMG=ghcr.io/mirantis/hmc/controller-ci:v0.0.1-287-gd9c38c3 make bundle-images
...
Done! Images bundled into /home/squizzi/go/src/github.com/Mirantis/hmc/bin/images/hmc-images-v0.0.2-14-gf02c79b.tgz

$ docker load < ./bin/images/hmc-images-v0.0.2-14-gf02c79b.tgz
...
49dc5e73a32a: Loading layer [==================================================>]  70.34MB/70.34MB
Loaded image: mcr.microsoft.com/oss/kubernetes/azure-cloud-node-manager:v1.30.4
c2453148319b: Loading layer [==================================================>]  85.39MB/85.39MB
Loaded image: quay.io/k0sproject/k0smotron:v1.0.4
4a2377d5d202: Loading layer [==================================================>]   1.69MB/1.69MB
d9486140fcec: Loading layer [==================================================>]  76.06MB/76.06MB
Loaded image: registry.k8s.io/provider-aws/cloud-controller-manager:v1.30.3
a7c367db8420: Loading layer [==================================================>]  61.15MB/61.15MB
Loaded image: registry.k8s.io/sig-storage/csi-snapshotter:v7.0.2
e4d8c7836c02: Loading layer [==================================================>]  73.34MB/73.34MB
...
```

Closes #423


scripts/bundle-images.sh
```diff
@@ -0,0 +1,245 @@
...
+        exit 1
+}
+
+echo -e "\nVerifying provider Deployments are ready...\n"
+
+# Verify each provider we support has deployed so we can get the images used
+# across the deployments.
+for TEMPLATE in $(find $TEMPLATES_DIR -name 'provider.yaml');
+do
+    RESULT=$(perl -nle 'print if m{(?=.*Provider)(?=.*kind)}' $TEMPLATE)
```
a13x5, line 99: What if you'll get?
```bash
$ perl
-bash: perl: command not found
```
It looks like it's a job for awk here

Also there is a size limit on variable content.

scripts/bundle-images.sh
```diff
@@ -0,0 +1,245 @@
...
+function ctrl_c() {
+        echo "Caught CTRL-C, exiting..."
+        exit 1
+}
+
+echo -e "\nVerifying provider Deployments are ready...\n"
+
+# Verify each provider we support has deployed so we can get the images used
+# across the deployments.
+for TEMPLATE in $(find $TEMPLATES_DIR -name 'provider.yaml');
```
a13x5, line 97: Again here, please don't use uppercase names for local variables.

scripts/bundle-images.sh
```diff
@@ -0,0 +1,245 @@
...
+        VALUES=$(grep -vw "{{" $TEMPLATE |
+            $YQ e "${EXTENSIONS_PATH}.charts[] | select(.name == \"$NAME\") | .values")
+
+        if [[ $URL == "" ]] || [[ $NAME == "" ]] || [[ $VERSION == "" ]]; then
+            echo "Error: Failed to get URL, NAME, or VERSION from $TEMPLATE"
+            exit 1
+        fi
+
+        # Check to see if a custom image is being used.
+        CUSTOM_TAG=$(echo -e "$VALUES" | $YQ e '.image.tag' -)
```
a13x5, line 198: Can we just pass values to `helm template` and omit all `custom_tag` logic?

scripts/bundle-images.sh
```diff
@@ -0,0 +1,245 @@
...
+            fi
+
+            EXTENSION_IMAGES_BUNDLED="$EXTENSION_IMAGES_BUNDLED $IMAGE"
+        done
+    done
+done
+
+echo -e "\nSaving images...\n"
+IMAGES_BUNDLED_UNIQ=$(echo "$IMAGES_BUNDLED" | tr ' ' '\n' | sort -u)
+IMAGE_IDS=
```
a13x5, line 220: Looks like it's not needed

scripts/bundle-images.sh
```diff
@@ -0,0 +1,245 @@
...
+
+bundle_images "$IMAGES_BUNDLED_UNIQ" $BUNDLE_TARBALL
+
+if [[ $EXTENSION_IMAGES_BUNDLED != "" ]]; then
+    EXTENSION_IMAGES_BUNDLED_UNIQ=$(echo "$EXTENSION_IMAGES_BUNDLED" | tr ' ' '\n' | sort -u)
+    EXTENSION_IMAGE_IDS=
```
a13x5, line 226: And this also

templates/cluster/vsphere-hosted-cp/templates/k0smotroncontrolplane.yaml
a13x5, line 1: Move these changes to #493

templates/cluster/vsphere-standalone-cp/templates/k0scontrolplane.yaml
a13x5, line 1: Same here it should be in scope of #493

a13x5 (changes requested):
Several issues with make targets
The scripts should be revisited. 

One general important note regarding bash syntax:

There is a convention that local scope variable names should be always lower snake case. Uppercase names are usually reserved only for env variables (in rare cases for global ones in scope of script)
Ditching this rule makes script harder to read and reason about. Please fix

Makefile
```diff
@@ -126,20 +126,56 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
...
+				if [[ $$url == "" ]] || [[ $$name == "" ]] || [[ $$version == "" ]]; then \
+					echo "Error: Cannot construct Helm pull command from url: $$url, name: $$name, version: $$version: one or more vars is not populated"; \
+					exit 1; \
+				fi; \
+				if [[ ! $$(find $(EXTENSION_CHARTS_PACKAGE_DIR) -name $$name-$$version*.tgz) ]]; then \
+					echo "Helm chart $$name from $$url with version $$version"; \
+					$(HELM) pull --repo $$url --version $$version $$name -d $(EXTENSION_CHARTS_PACKAGE_DIR); \
+				fi; \
+		done; \
+	done
```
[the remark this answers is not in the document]
s3rj1k: IMO: All inline-shell scripts should be moved out into some folder and bash liter added

scripts/airgap-push.sh
```diff
@@ -0,0 +1,140 @@
...
+    IMAGE_NAME=$(echo $IMAGE | grep -o '[^/]*$')
+    OLD_IMAGE=$(docker images -a | grep $IMAGE | awk '{print $1":"$2}')
+    TAG=$OLD_IMAGE | awk -F ":" '{print $2}'
+
+    docker tag $OLD_IMAGE $REPO/$IMAGE_NAME:$TAG
+    docker push $REPO/$IMAGE_NAME:$TAG
+done
+
+# Next, use Helm to push the charts to the given chart repository.
+mkdir -p hmc_charts
```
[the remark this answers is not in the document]
s3rj1k: .gitignore ?

Makefile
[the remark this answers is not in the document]
squizzi: I was following suite with other bash scripts here, if that's the feeling around this I'll move this one out.

scripts/bundle-images.sh
s3rj1k: `awk` also has different implementations, mawk, gawk,...

squizzi: Right, so it's tricky to choose the right ones, we could just suggest on mac to install `ggrep` from Brew prior to running the script and just use that, but it's the same as `perl`, it's hard to satisfy every requirement.

s3rj1k: One thing that we have on all platforms is Go, how about writing all scripts in Go?

Makefile
```diff
@@ -126,20 +126,56 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
...
+
+.PHONY: bundle-images
+bundle-images: dev-apply $(IMAGES_PACKAGE_DIR) ## Create a tarball with all images used by HMC.
+	@BUNDLE_TARBALL=$(IMAGES_PACKAGE_DIR)/hmc-images-$(VERSION).tgz EXTENSIONS_BUNDLE_TARBALL=$(IMAGES_PACKAGE_DIR)/hmc-extension-images-$(VERSION).tgz IMG=$(IMG) KUBECTL=$(KUBECTL) YQ=$(YQ) HELM=$(HELM) NAMESPACE=$(NAMESPACE) TEMPLATES_DIR=$(TEMPLATES_DIR) KIND_CLUSTER_NAME=$(KIND_CLUSTER_NAME) bash -c "./scripts/bundle-images.sh"
+
+.PHONY: airgap-package
+airgap-package: helm-package bundle-images ## Create a tarball with all images and Helm charts used by HMC, useful for deploying in air-gapped environments.
+	tar -czf hmc-charts-images-$(VERSION).tgz ./scripts/airgap-push.sh $(CHARTS_PACKAGE_DIR) $(IMAGES_PACKAGE_DIR)
```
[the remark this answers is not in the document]
squizzi: I'm confused.  It's not two commands, it's a single `tar` command, the `bundle-images` target has the logic.  I'm assuming this message was for there.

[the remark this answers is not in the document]
a13x5: Oh, wait, looks I got confused. You just adding `./scripts/airgap-push.sh` to the archive here.

scripts/bundle-images.sh
a13x5: I think conversation went a little bit to another direction.
1. I'm not aware of the requirement to run release pipeline (which this script will be part of) on MacOS (this is how you get endless scope)
2. Let's stick to GNU utilities, the MacOS quirks (if any) could be simply documented in `dev.md`. 
3. Go is an option, but I guess it's too late to make this decisions at this point. We can do followup task, but it should be evaluated separately

squizzi:
`EXPORT_ALL_VARIABLES` does not appear to play nicely with other `Makefile` targets, so I'm avoiding that.

scripts/bundle-images.sh
squizzi: Yep, good idea, I've implemented this.

templates/cluster/vsphere-hosted-cp/templates/k0smotroncontrolplane.yaml
squizzi: I can't, I get the desire to scope these to #493 but this _won't work_ without them because it tries to pull via the chart name it gets from these files and the chart doesn't exist on `charts.mirantis`

templates/cluster/vsphere-standalone-cp/templates/k0scontrolplane.yaml
squizzi: I can't, I get the desire to scope these to #493 but this _won't work_ without them because it tries to pull via the chart name it gets from these files and the chart doesn't exist on `charts.mirantis`

a13x5 (changes requested):
> EXPORT_ALL_VARIABLES does not appear to play nicely with other Makefile targets and export does not appear to actually populate the variables

`export` should work at least with GNU Make 4.4 (especially if `.EXPORT_ALL_VARIABLES` works) (see [manual](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html))

Anyway if it's became too complex I will not block this PR. Issues with Makefile will be fixed in scope of #589 

A couple of small issues should be fixed in the script however (Note that NIT comment are optional to fix)

squizzi:
> Anyway if it's became too complex I will not block this PR. Issues with Makefile will be fixed in scope of https://github.com/Mirantis/hmc/issues/589

Ya, I'm not sure what's going on, I've configured the exports just as the docs suggest with:

```
export var = value
```

and they end up empty in the script invocation, debugged with `set -x`.

squizzi:
I think I need to bring back the `custom_tag` logic, for example `kube-vip` when templated with our values produces the following error:

```
helm template --repo https://kube-vip.github.io/helm-charts kube-vip -f values.yaml
Error: execution error at (kube-vip/templates/daemonset.yaml:33:22): A valid config.address required!
```

It's because we're trying to remove the `config.address` entry prior to templatizing it.  Sure we could work around this instance with something like `--set config.address "12345"` but do we potentially want to avoid custom logic like this?  The `custom_tag` logic, although a little quirky didn't have to deal with these potential custom logic that chart maintainers may add.

squizzi:
Found a bug in the `airgap-push.sh` script, it wasn't dealing with multiple tags of images well, that's been fixed.

a13x5 (changes requested):
I tested the scripts on the airgapped ubuntu machine.
While testing I had multiple issues with airgapped k0s (some irrelevant to this PR).
I added several comments related to the issues I experienced.

Regarding the `kube-vip` issue. It's caused by the empty value we're providing via values, so it's not related to `kube-vip` chart itself. Thus I suggest adding a special case for it either adding `--debug` (so it will output some yaml) or by removing empty value from the values we're providing. 
custom_tag logic also could be considered, but IMO it looks more flaky.

scripts/bundle-images.sh
```diff
@@ -0,0 +1,226 @@
...
+        label_value=$(echo $(echo ${provider_kind_tolower} | sed -e 's/provider//g')-${provider_name})
+    fi
+
+    wait_for_deploy_exist "$LABEL_KEY=$label_value"
+
+    ${KUBECTL} wait --for condition=available --timeout=2m deploy -l ${LABEL_KEY}=${label_value} -n ${NAMESPACE}
+    if [[ $? -ne 0 ]]; then
+        echo "Error: Cannot wait for Deployment: Deployment with ${LABEL_KEY}=${label_value} label not found"
+        exit 1
+    fi
```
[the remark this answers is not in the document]
squizzi: How long does it typically take for your HMC cluster to come up and be fully ready on your machine?  It looks like our CI is fully ready in 8-9m, so 10m seems to be safe, perhaps overkill, but this target is going to run during release CI anyways.

squizzi:
>> Regarding the kube-vip issue. It's caused by the empty value we're providing via values, so it's not related to kube-vip chart itself. Thus I suggest adding a special case for it either adding --debug (so it will output some yaml) or by removing empty value from the values we're providing.

Outputting `--debug` if the `helm template` fails seems to do the trick for this instance.

scripts/bundle-images.sh
```diff
@@ -0,0 +1,240 @@
...
+do
+    if [[ $image == "" ]]; then
+        echo "Error: Failed to get image from KIND cluster, image string should not be empty"
+        exit 1
+    fi
+
+    if [[ $image == "localhost/hmc/controller:latest" ]]; then
+        # Don't try to pull and package the local controller image.
+        continue
+    fi
```
a13x5, line 144: I got this bundled for some reason
```
   "RepoTags": [
      "hmc/controller:latest"
    ],
```
Plus I don't see an actual controller image in the bundle. Since we're not pulling it here we should have a separate case for it. Maybe I'm missing something?
All other images seem to work fine

a13x5 (changes requested):
Issue with hmc-controller image
With `--insecure-skip-tls-verify` flag we're not covering http registries, but it's out of scope of this PR and should be covered in the docs.

scripts/bundle-images.sh
squizzi: The `hmc/controller` image is supposed to be added from the `IMG` set in `make docker-build`, so by default the `IMG` environment var is `hmc/controller:latest`, because that's what the image gets built as.  In CI, we'd set the IMG tag to `build.Version` like we do in the CI for e2e testing and set that variable and that image would get bundled.  We could also just try setting that from the get-go outside of IMG I guess.

a13x5: Ok, so if you do this outside of CI you have to set it to some value with proper tag. 
I believe short note on that in dev.md will suffice.

squizzi:
Looks like kube-vip still didn't want to render out, for now I've just added some logic to workaround this case, I kept the `--debug` logic in there though for other instances where that may be viable.

squizzi:
I've squashed the changes down to prepare for merge/carrying of this PR since this is my last day at Mirantis.

PR #476, opened 2024-10-11
Reconcile MultiClusterService

wahabmk:
# Description

1. Implemented Reconcile for MultiClusterService.
2. Does not implement status. That will be done in https://github.com/Mirantis/hmc/issues/460.
3. Some refactoring.

# Sanity Testing
The following demo shows some sanity testing:

https://github.com/user-attachments/assets/1c22282b-37f4-434f-92a8-9b8937991cd0


internal/sveltos/profile.go
```diff
@@ -50,96 +51,149 @@ type HelmChartOpts struct {
...
+	if operation == controllerutil.OperationResultCreated || operation == controllerutil.OperationResultUpdated {
+		l.Info(fmt.Sprintf("Successfully %s Profile %s", string(operation), p.Name))
+	}
+
+	return p, nil
+}
+
+// Spec returns a spec object to be used with
+// a Sveltos Profle or ClusterProfile object.
+func Spec(opts *ReconcileProfileOpts) (*sveltosv1beta1.Spec, error) {
```
zerospiel, line 128: wdyt renaming to the `getProfileSpec`?

internal/sveltos/profile.go
```diff
@@ -50,96 +51,149 @@ type HelmChartOpts struct {
...
+	return spec, nil
+}
 
-			cp.Spec.HelmCharts = append(cp.Spec.HelmCharts, helmChart)
-		}
-		return nil
-	})
-	if err != nil {
-		return nil, err
+func objectMeta(owner *metav1.OwnerReference) metav1.ObjectMeta {
```
zerospiel, line 182: wdyt renaming to the `getObjectMeta`?

internal/sveltos/profile.go
```diff
@@ -50,96 +51,149 @@ type HelmChartOpts struct {
...
+	if owner != nil {
+		obj.OwnerReferences = []metav1.OwnerReference{*owner}
 	}
 
-	return cp, nil
+	return obj
 }
 
+// DeleteProfile deletes a Sveltos Profile object.
 func DeleteProfile(ctx context.Context, cl client.Client, namespace string, name string) error {
```
zerospiel, line 197: i know that it's not directly related to this PR but here goes the same question about the need for explicit deletion of a `Profile` object, having owner references on it, does the object require it?

internal/sveltos/profile.go
```diff
@@ -50,96 +51,149 @@ type HelmChartOpts struct {
...
-	operation, err := ctrl.CreateOrUpdate(ctx, cl, cp, func() error {
-		if cp.Labels == nil {
-			cp.Labels = make(map[string]string)
-		}
+	spec := &sveltosv1beta1.Spec{
+		ClusterSelector: libsveltosv1beta1.Selector{
+			LabelSelector: opts.LabelSelector,
+		},
+		Tier:               tier,
+		ContinueOnConflict: !opts.StopOnConflict,
```
zerospiel, line 139: can we preallocate the slice here like `HelmCharts: make([]sveltosv1beta1.HelmChart, 0, len(opts.HelmChartOpts)),`?

internal/sveltos/profile.go
```diff
@@ -50,96 +51,149 @@ type HelmChartOpts struct {
...
-				ChartName:        hc.ChartName,
-				ChartVersion:     hc.ChartVersion,
-				ReleaseName:      hc.ReleaseName,
-				ReleaseNamespace: hc.ReleaseNamespace,
-				HelmChartAction:  sveltosv1beta1.HelmChartActionInstall,
-				RegistryCredentialsConfig: &sveltosv1beta1.RegistryCredentialsConfig{
-					PlainHTTP:             hc.PlainHTTP,
-					InsecureSkipTLSVerify: hc.InsecureSkipTLSVerify,
-				},
+		if hc.Values != nil {
```
zerospiel, line 162: we can simplify on a single call/error 

```go
		if hc.Values != nil && len(hc.Values.Raw) > 0 {
			b, err := yaml.JSONToYAML(hc.Values.Raw)
			if err != nil {
				return nil, fmt.Errorf("failed to convert values from JSON to YAML for service %s: %w", hc.RepositoryName, err)
			}

			helmChart.Values = string(b)
		}
```

internal/sveltos/profile.go
```diff
@@ -50,96 +51,149 @@ type HelmChartOpts struct {
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to convert values from JSON to YAML for service %s: %w", hc.RepositoryName, err)
 			}
 
-			if hc.Values != nil {
-				b, err := hc.Values.MarshalJSON()
-				if err != nil {
-					return fmt.Errorf("failed to marshal values to JSON for service (%s) in ManagedCluster: %w", hc.RepositoryName, err)
-				}
+			helmChart.Values = string(b)
```
zerospiel, line 173: performance nit: if we expect to have long strings then maybe converting to string without allocation might be helpful here (i.e. `unsafe.String(&b[0], b)`)

api/v1alpha1/managedcluster_types.go
```diff
@@ -91,6 +91,9 @@ type ManagedClusterSpec struct {
 	Priority int32 `json:"priority,omitempty"`
 	// DryRun specifies whether the template should be applied after validation or only validated.
 	DryRun bool `json:"dryRun,omitempty"`
+
+	// +kubebuilder:default:=false
```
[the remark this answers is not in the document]
wahabmk: It's just because without it the field does not show up in `kubectl get ... -o yaml`.

internal/controller/multiclusterservice_controller.go
```diff
@@ -29,10 +37,132 @@ type MultiClusterServiceReconciler struct {
...
+		}); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to reconcile ClusterProfile: %w", err)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// HelmChartOpts returns slice of helm chart options to use with Sveltos.
+// Namespace is the namespace of the referred templates in services slice.
+func HelmChartOpts(ctx context.Context, c client.Client, l logr.Logger, namespace string, services []hmc.ServiceSpec) ([]sveltos.HelmChartOpts, error) {
```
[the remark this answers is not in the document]
wahabmk: - It doesn't need to be public (can change that)
- I passed logger so that it will have the WithValues() already set from where this function is called e.g. [here](https://github.com/wahabmk/hmc/blob/07392894f35ba7d2a37b2140f15d0487931f0d25/internal/controller/managedcluster_controller.go#L374-L377).
- I named it without `get*` because [I had this](https://go.dev/doc/effective_go#Getters) in mind. However, maybe this doesn't qualify as a getter? In any case I am open to change it to `getHelmChartOpts` if that is preferable.

internal/controller/multiclusterservice_controller.go
```diff
@@ -29,10 +37,132 @@ type MultiClusterServiceReconciler struct {
...
+		if err != nil {
+			return nil, fmt.Errorf("could not get repository url: %w", err)
+		}
+
+		opts = append(opts, sveltos.HelmChartOpts{
+			Values:        svc.Values,
+			RepositoryURL: source.Spec.URL,
+			// We don't have repository name so chart name becomes repository name.
+			RepositoryName: tmpl.Spec.Helm.ChartName,
+			ChartName: func() string {
```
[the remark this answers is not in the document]
wahabmk: > can we use plain variables or non-anon functions instead of closures here and in the releaseNamespace field?

I think this is easier to understand no? Is computational cost the reason to remove it?

> and a tiny nit (applicable elsewhere where it makes sense): pkg-imports-wise types.NamespacedName -> client.ObjectKey since we already have imported the client pkg

ack

internal/controller/multiclusterservice_controller.go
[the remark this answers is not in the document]
zerospiel: regarding the name: thanks, haven't had it in mind, then it's ok
regarding the logger: there is `ctrl.LoggerInto(ctx)` which provides convenient way to change kind-of-global logger and put it back in the context, but having this logger there is no need, since those values are already embedded into the default logger so there is no need at all to duplicate this info (that's why I remove such an approach elsewhere)

internal/controller/multiclusterservice_controller.go
```diff
@@ -29,10 +37,132 @@ type MultiClusterServiceReconciler struct {
...
+			// over plain HTTP, which is different than what InsecureSkipTLSVerify is meant for.
+			// See: https://github.com/fluxcd/source-controller/pull/1288
+			PlainHTTP: source.Spec.Insecure,
+		})
+	}
+
+	return opts, nil
+}
+
+func (r *MultiClusterServiceReconciler) reconcileDelete(ctx context.Context, mcsvc *hmc.MultiClusterService) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
wahabmk: The reason is so that Sveltos can clean up the resources that depend on `Profile` or `ClusterProfile` object. See https://github.com/Mirantis/hmc/pull/362#discussion_r1775019326 for previous discussion.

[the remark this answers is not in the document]
zerospiel: Please correct me if I'm mistaken: **without** the current logic and relying only on the owner references, removing an `mcs` object will remove the related `ClusterProfile` object, so in this way, Sveltos **will NOT** garbage collect other objects relying on that cluster profile?

internal/controller/template_controller.go
```diff
@@ -344,3 +345,32 @@ func (r *ProviderTemplateReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&hmc.ProviderTemplate{}).
 		Complete(r)
 }
+
+// TemplateSource returns the source of the provided template.
+func TemplateSource(ctx context.Context, c client.Client, tmpl Template) (*sourcev1.HelmRepository, error) {
```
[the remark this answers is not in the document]
wahabmk: - This function is operating on the `Template` interface so it can be used by other controllers as well if needed. That is why I put it in the `template_controller.go` instead of `multiclusterservice_controller.go`.
- Yeah it can be private
- Same reason as I mentioned before about naming getters, but can change it if that's preferable.
- This way is more scalable and correct IMO. I.e. Fetching the source of a template by fetching: Template -> HelmChart -> Source. More correct because Flux's HelmChart and its source must be in the same namespace. More scalable because if we change the name of default repository or if the template exists in a different namespace than "hmc-system" this function will still work.

internal/sveltos/profile.go
wahabmk: Same reason as mentioned before about naming getters, but open to changing it to `getProfileSpec`

internal/sveltos/profile.go
wahabmk: Same reason

internal/sveltos/profile.go
wahabmk: Same reason as mentioned in the comment for `DeleteClusterProfile`.

internal/controller/multiclusterservice_controller.go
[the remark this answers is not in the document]
wahabmk: If deleting the parent will set the `DeletionTimestamp` on the child objects, then we might not need to explicitly delete the child object because Sveltos checks `DeletionTimestamp`. I don't know the answer to that. I can try and test it.

internal/controller/template_controller.go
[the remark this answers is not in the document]
zerospiel: the iface will be removed/reworked soon or later, but it's okay

I have the following concern: currently, all of the templates are being put in the only helm repo, right? if so, then I don't see the reason to make extra API calls, otherwise, surely you're right and there is no other way to properly get the `helmrepository`. we can wait for other opinions though :)

api/v1alpha1/managedcluster_types.go
```diff
@@ -91,6 +91,9 @@ type ManagedClusterSpec struct {
 	Priority int32 `json:"priority,omitempty"`
```
Schnitzel, line 91: as mentioned in Slack I expected the Priority per Service and not on the level of the cluster, as it could be that the service has individual priorities. I could also agree to have it in both places, but I feel as the Priority really only is for the Services and the cluster itself does not have a priority I would move it into each of the Service

Schnitzel (changes requested):

api/v1alpha1/managedcluster_types.go
Schnitzel: ah, so we put all the defined `services` in a `managedCluster` object into a single sveltos `Profile`? I assumed that we create a separate Profile object per defined  service.

wahabmk: Yes each Profile/ClusterProfile can have multiple services. We don't create a separate Profile/ClusterProfile per service.

Schnitzel: okay, yea I wonder what is better, but lets keep it the way we have it now.
should we then rename `Priority` to `ServicePriority` or something? as right now the word Priority suggest that this is a cluster priority, which it is not, it is only used by Services

internal/controller/multiclusterservice_controller.go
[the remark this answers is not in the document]
wahabmk: @zerospiel You are correct in that we can rely on owner references for removing dependent objects. However, in this case we need to issue a delete on `Profile`. The reason is that without this [`DeleteProfile`](https://github.com/Mirantis/hmc/blob/3f9a2addc4b5f9feda254a7d94afa594a2e9ae7b/internal/controller/managedcluster_controller.go#L591-L594) call, the Sveltos addon-controller keeps on looping with the following error:
```sh

I1016 23:21:22.030322       1 controller.go:302] "Reconciling" controller="clustersummary" controllerGroup="config.projectsveltos.io" controllerKind="ClusterSummary" ClusterSummary="hmc-system/p--wali-aws-dev-capi-wali-aws-dev" namespace="hmc-system" name="p--wali-aws-dev-capi-wali-aws-dev" reconcileID="3d98ee4c-2713-45bf-ae9b-fa7f7000fd77"
I1016 23:21:22.030397       1 clustersummary_controller.go:122] "Reconciling" controller="clustersummary" controllerGroup="config.projectsveltos.io" controllerKind="ClusterSummary" ClusterSummary="hmc-system/p--wali-aws-dev-capi-wali-aws-dev" namespace="hmc-system" name="p--wali-aws-dev-capi-wali-aws-dev" reconcileID="3d98ee4c-2713-45bf-ae9b-fa7f7000fd77"
I1016 23:21:22.030893       1 clustersummary_controller.go:225] "Reconciling ClusterSummary delete" controller="clustersummary" controllerGroup="config.projectsveltos.io" controllerKind="ClusterSummary" ClusterSummary="hmc-system/p--wali-aws-dev-capi-wali-aws-dev" namespace="hmc-system" name="p--wali-aws-dev-capi-wali-aws-dev" reconcileID="3d98ee4c-2713-45bf-ae9b-fa7f7000fd77"
E1016 23:21:22.045361       1 clustersummary_controller.go:250] "failed to remove ResourceSummary." err="failed to get API group resources: unable to retrieve the complete list of server APIs: apiextensions.k8s.io/v1: Get \"https://wali-aws-dev-apiserver-1185732421.ca-central-1.elb.amazonaws.com:6443/apis/apiextensions.k8s.io/v1\": dial tcp: lookup wali-aws-dev-apiserver-1185732421.ca-central-1.elb.amazonaws.com on 10.96.0.10:53: no such host" controller="clustersummary" controllerGroup="config.projectsveltos.io" controllerKind="ClusterSummary" ClusterSummary="hmc-system/p--wali-aws-dev-capi-wali-aws-dev" namespace="hmc-system" name="p--wali-aws-dev-capi-wali-aws-dev" reconcileID="3d98ee4c-2713-45bf-ae9b-fa7f7000fd77"
I1016 23:21:22.045893       1 controller.go:318] "Reconcile done, requeueing after 10s" controller="clustersummary" controllerGroup="config.projectsveltos.io" controllerKind="ClusterSummary" ClusterSummary="hmc-system/p--wali-aws-dev-capi-wali-aws-dev" namespace="hmc-system" name="p--wali-aws-dev-capi-wali-aws-dev" reconcileID="3d98ee4c-2713-45bf-ae9b-fa7f7000fd77"
```
So during cleanup, it tried to access the deleted target cluster and fails. This might actually be a bug in Sveltos addon-controller.

api/v1alpha1/managedcluster_types.go
wahabmk: Changed to `ServicePriority` to better reflect its purpose.

internal/controller/template_controller.go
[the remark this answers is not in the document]
wahabmk: Removed the `TemplateSource` function as it was being used only in `helmChartOpts` func anyway.

internal/controller/multiclusterservice_controller.go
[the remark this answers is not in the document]
zerospiel: Thanks for the check, the error looks unrelated though (I'm not much in the details of the sveltos). I've crawled through the addon-ctrl code and that particular error should not be related to my understanding (`*Profile` creation leads to `ClusterSummary` creation, and deletion of the `*Profile` either way leads to the deletion of the `ClusterSummary`, I've found no diff between having explicit/implicit profiles deletion regarding this behavior). Potential diff may be related to the cluster (either CAPI or Sveltos) readiness but I have no child clusters on-hands so cannot check the actual states, anyway, I do not think that is related to the problem.

In conclusion: if you are sure that the setting finalizer somehow changes the states of a cluster and depending on that state the sveltos' clustersummary controller behaves in different ways, then it's okay. 

I'd rather left comments here (`Delete(Cluster)Profile` functions) why do we even need them

zerospiel (commented):
LGTM. Regarding the integration tests, it feels like there is room to implement more checks (e.g. the actual statuses of the requested object or dependant objects) but I'd left that idea to think to the author and other reviewers

api/v1alpha1/managedcluster_types.go
mestadler: ack, change to ServicePriority seems like the right thing.  @Schnitzel all good?  this should solve the last outstanding unresolved item.

Schnitzel (dismissed):
looks good!

wahabmk:
> LGTM. Regarding the integration tests, it feels like there is room to implement more checks (e.g. the actual statuses of the requested object or dependant objects) but I'd left that idea to think to the author and other reviewers

Further tests to check status can be implemented as part of https://github.com/Mirantis/hmc/issues/460.

eromanova (approved):
LGTM

eromanova (approved):

PR #481, opened 2024-10-14
Add credential admission checks for managedcluster

### LINKED ISSUE #392: Add admission checks for Credential
Validate that Credential referenced in ManagedCluster has Ready state
Validate that Credential has the same type as Template which referenced in ManagedCluster

a13x5:
Resolves #392 
Also added basic unit tests for validations.

### DIFF
--- a/internal/controller/managedcluster_controller_test.go
+++ b/internal/controller/managedcluster_controller_test.go
@@ -38,7 +38,8 @@ var _ = Describe("ManagedCluster Controller", func() {
 			managedClusterName      = "test-managed-cluster"
 			managedClusterNamespace = "test"
 
-			templateName = "test-template"
+			templateName   = "test-template"
+			credentialName = "test-credential"
 		)
 
 		ctx := context.Background()
@@ -50,6 +51,7 @@ var _ = Describe("ManagedCluster Controller", func() {
 		managedCluster := &hmc.ManagedCluster{}
 		template := &hmc.ClusterTemplate{}
 		management := &hmc.Management{}
+		credential := &hmc.Credential{}
 		namespace := &corev1.Namespace{}
 
 		BeforeEach(func() {
@@ -92,6 +94,13 @@ var _ = Describe("ManagedCluster Controller", func() {
 							Raw: []byte(`{"foo":"bar"}`),
 						},
 					},
+					Providers: hmc.ProvidersTupled{
+						InfrastructureProviders: []hmc.ProviderTuple{
+							{
+								Name: "aws",
+							},
+						},
+					},
 				}
 				Expect(k8sClient.Status().Update(ctx, template)).To(Succeed())
 			}
@@ -106,7 +115,40 @@ var _ = Describe("ManagedCluster Controller", func() {
 					Spec: hmc.ManagementSpec{},
 				}
 				Expect(k8sClient.Create(ctx, management)).To(Succeed())
+				management.Status = hmc.ManagementStatus{
+					AvailableProviders: hmc.ProvidersTupled{
+						InfrastructureProviders: []hmc.ProviderTuple{
+							{
+								Name: "aws",
+							},
+						},
+					},
+				}
+				Expect(k8sClient.Status().Update(ctx, management)).To(Succeed())
 			}
+			By("creating the custom resource for the Kind Credential")
+			err = k8sClient.Get(ctx, typeNamespacedName, credential)
+			if err != nil && errors.IsNotFound(err) {
+				credential = &hmc.Credential{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      credentialName,
+						Namespace: managedClusterNamespace,
+					},
+					Spec: hmc.CredentialSpec{
+						IdentityRef: &corev1.ObjectReference{
+							APIVersion: "infrastructure.cluster.x-k8s.io/v1beta2",
+							Kind:       "AWSClusterStaticIdentity",
+							Name:       "foo",
+						},
+					},
+				}
+				Expect(k8sClient.Create(ctx, credential)).To(Succeed())
+				credential.Status = hmc.CredentialStatus{
+					State: hmc.CredentialReady,
+				}
+				Expect(k8sClient.Status().Update(ctx, credential)).To(Succeed())
+			}
+
 			By("creating the custom resource for the Kind ManagedCluster")
 			err = k8sClient.Get(ctx, typeNamespacedName, managedCluster)
 			if err != nil && errors.IsNotFound(err) {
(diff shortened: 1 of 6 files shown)


internal/controller/managedcluster_controller_test.go
```diff
@@ -106,7 +115,40 @@ var _ = Describe("ManagedCluster Controller", func() {
...
+							{
+								Name: "aws",
+							},
+						},
+					},
+				}
+				Expect(k8sClient.Status().Update(ctx, management)).To(Succeed())
 			}
+			By("creating the custom resource for the Kind Credential")
+			err = k8sClient.Get(ctx, typeNamespacedName, credential)
```
zerospiel, line 130: getting the deliberately wrong-named object does not make much sense (not only there but everywhere except the `mcl` resource), so maybe it'd be better to create the object ignoring the `AlreadyExist` error (which is not mandatory having removal in the `AfterEach` node)

do we actually expect integration tests for the `Credential` resource? If no **and** the tests work without such a resource's object, maybe it might be better just skip that part or leave a todo comment to implement tests on the actual business logic (like a single-node checking only a small piece of logic in the [`internal/controller/template_controller_test.go:199`](https://github.com/Mirantis/hmc/blob/b7e1f2247edf70dff71d45e0c56433b71d561063/internal/controller/template_controller_test.go#L199-L316))

internal/webhook/managedcluster_webhook.go
```diff
@@ -73,6 +73,19 @@ func (v *ManagedClusterValidator) ValidateCreate(ctx context.Context, obj runtim
...
+	}
+
+	if cred.Status.State != hmcv1alpha1.CredentialReady {
+		return nil, fmt.Errorf("%s: Credential is not Ready", invalidManagedClusterMsg)
+	}
+
+	if err := isCredMatchTemplate(cred, template); err != nil {
+		return nil, fmt.Errorf("%s: %v", invalidManagedClusterMsg, err)
+	}
+
```
zerospiel, line 88: kind of a nit: may it be a good idea to extract that block into a single method/func to reduce the duplication?

internal/webhook/managedcluster_webhook.go
```diff
@@ -298,3 +336,38 @@ func getMissingProvidersWithWrongVersions(exposed, required []hmcv1alpha1.Provid
...
+	for _, provider := range infraProviders {
+		switch provider.Name {
+		case "aws":
+			if idtyKind != "AWSClusterStaticIdentity" &&
+				idtyKind != "AWSClusterRoleIdentity" &&
+				idtyKind != "AWSClusterControllerIdentity" {
+				return errMsg(idtyKind, provider.Name)
+			}
+		case "azure":
+			if idtyKind != "AzureClusterIdentity" {
```
zerospiel, line 361: kind of nit either: what do you think if we utilize the const from [the dedicated file](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/e3e3e309b07d470fa79cb3ee172debc029c4d8f2/api/v1beta1/consts.go#L195)? Either way the dependency will be introduced by the #405 PR

If we can, then I hope there are such constants in the corresponding packages (maybe skip the `aws` ones since that dependency has not yet been brought)

side note: if you do not want to escape quotes (like in `"template \"%s\"`) you can either use the graves `` `template "%s"` `` or the [dedicated verb](https://pkg.go.dev/fmt#hdr-Printing) for double-quotes strings `template %q`

test/objects/credential/credential.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/Mirantis/hmc/api/v1alpha1"
+)
+
+const (
+	DefaultName      = "credential"
+	DefaultNamespace = "default"
+)
+
+type Opt func(managedCluster *v1alpha1.Credential)
```
zerospiel, line 29: can we drop `managedCluster` argument name? it is both redundant and misleading 

and it seems `WithDescription` function is not used, and `DefaultNamespace = "default"` is equal to the `DefaultNamespace = metav1.NamespaceDefault`

internal/webhook/managedcluster_webhook.go
```diff
@@ -298,3 +336,38 @@ func getMissingProvidersWithWrongVersions(exposed, required []hmcv1alpha1.Provid
...
+
+func isCredMatchTemplate(cred *hmcv1alpha1.Credential, template *hmcv1alpha1.ClusterTemplate) error {
+	idtyKind := cred.Spec.IdentityRef.Kind
+	infraProviders := template.Status.Providers.InfrastructureProviders
+
+	errMsg := func(idtyKind string, provider string) error {
+		return fmt.Errorf("wrong kind of the ClusterIdentity \"%s\" for provider \"%s\"", idtyKind, provider)
+	}
+
+	if len(infraProviders) == 0 {
```
zerospiel, line 348: wdyt about checking providers' absence before making the API calls (getting the `credentials` object)? i think it might save some network in theory

zerospiel (commented):
Looks good to me, left mostly nit comments, please consider them if you will make any amends/manual rebase resolution to the PR

test/objects/credential/credential.go
a13x5: Such a lame copy-paste error)

internal/webhook/managedcluster_webhook.go
zerospiel: if there are no providers available in a template, then we will fetch credentials (assuming it's ready) each time just to declare that there are no providers available, to me it looks a bit redundant, so it actually does not matter whether the creds are ready or not — the validation will fail either way

internal/webhook/managedcluster_webhook.go
a13x5: @zerospiel I know that the kind comparison by comparing strings looks kinda shitty. Unfortunately only Azure provider has these consts. VSphere has [something else](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/apis/v1beta1/vsphereclusteridentity_types.go#L66) and AWS doesn't care at all.
So I decided to go with string comparison, since it makes it uniform across all providers

internal/webhook/managedcluster_webhook.go
a13x5: tbh I still don't see much point in this, but I will fix it

zerospiel: thanks. the main point is to avoid making unnecessary network invocations, simply this :)

internal/controller/managedcluster_controller_test.go
a13x5: I don't think it's needed now for the credentials controller, since most logic for it actually in the managedcluster.

zerospiel: Correct, I meant that if the tests in the scope of the managed-ctrl are not supposed to be implemented, then it'd be great to leave a TODO comment. And the point about getting the wrong-named object: I do not understand why the initially misused code should be copied further instead of being fixed in the new pieces of code.

Either way, it is up to you, I just pointed out that the current code is being misused again and again and may be simplified. But I believe that will be resolved during the tests implementation

zerospiel (approved):

Kshatrix (approved):

PR #487, opened 2024-10-15
Compatibility attributes amends

zerospiel:
* move clustertemplates compatibility verification from the validationwebhook to the tpl ctrl
* correspinding integration tests

Related Issue: https://github.com/Mirantis/hmc/issues/400

### DIFF
--- a/internal/controller/template_controller.go
+++ b/internal/controller/template_controller.go
@@ -17,8 +17,12 @@ package controller
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"slices"
+	"time"
 
+	"github.com/Masterminds/semver/v3"
 	helmcontrollerv2 "github.com/fluxcd/helm-controller/api/v2"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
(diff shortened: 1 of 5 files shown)


internal/controller/template_controller.go
```diff
@@ -301,23 +318,114 @@ func (r *TemplateReconciler) getHelmChartFromChartRef(ctx context.Context, chart
 	return helmChart, nil
 }
 
+func (r *ClusterTemplateReconciler) validateCompatibilityAttrs(ctx context.Context, template *hmc.ClusterTemplate) error {
+	management := new(hmc.Management)
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: hmc.ManagementName}, management); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil // nothing to update in template's status
```
Kshatrix, line 325: I think we should mark it as invalid with a message of something like "waiting for Management creation to complete validation". and requeue with some static interval until it's created

Kshatrix (changes requested):

Kshatrix (approved):

PR #515, opened 2024-10-17
Fix -  Initial version of hmc-templates release gets updated concurrently 

### LINKED ISSUE #449: Initial version of hmc-templates release gets updated concurrently
Initial version of hmc-templates release gets updated concurrently multiple times with and without createRelease helm value.

kylewuolle:
Simple change to resolve #449. 

Testing performed:
1. Updated hmc_values.yaml to set the createRelease flag to true
2. Executed `make hmc-chart-release`
3. Ran `make kind-deploy registry-deploy dev-push dev-deploy dev-templates` (make dev-apply without the dev-release target)
4. Verified the release was created correctly

1. Updated hmc_values.yml to set the createRelease flag to false
2. Ran `make kind-deploy registry-deploy dev-push dev-deploy dev-templates` (make dev-apply without the dev-release target)
3. Verified the release was not created and there was no loop attempting to create it
4. ran `make dev-release` and verified the release was installed correctly

Also ran `make dev-apply` with the createRelease flag set to false (clean version of hmc_values.yaml) and verified this works as well.

### DIFF
--- a/internal/controller/release_controller.go
+++ b/internal/controller/release_controller.go
@@ -252,7 +252,7 @@ func (r *ReleaseReconciler) reconcileHMCTemplates(ctx context.Context, releaseNa
 		},
 	}
 
-	if releaseName == "" {
+	if releaseName == "" || r.CreateRelease {
 		createReleaseValues := map[string]any{
 			"createRelease": true,
 		}


slysunkin (approved):

squizzi:
Rebase to pick up the e2e controller test, it might need to be fixed up for this, it may even be useful to add a test case to the controller test.

internal/controller/release_controller.go
```diff
@@ -252,7 +252,7 @@ func (r *ReleaseReconciler) reconcileHMCTemplates(ctx context.Context, releaseNa
 		},
 	}
 
-	if releaseName == "" {
+	if releaseName == "" || r.CreateRelease {
```
Kshatrix, line 255: we only need Release to be created as part of the hcm-templates chart during the initial hmc installation. We should probably rename it to avoid confusion (along with `create-management`, `create-template-management`) (although it's mentioned in the variable description)

Kshatrix (changes requested):

kylewuolle:
Looks like this was fixed in another PR

PR #360, opened 2024-09-20
Run VSphere tests in self-hosted, refactor test suite into spec per provider, support label filtering, get all tests green

squizzi:
This PR refactors our existing e2e test spec into a more conventional e2e test run.  We now have a `BeforeSuite` that runs and installs/validates the controller.  Prior to this we had all of our e2e tests under a single `Controller` spec, but this breaks it into:

* Controller (`Label(controller)`)
* Providers (`Label(provider)`)
    * AWS (`Label(provider:aws, provider:cloud)`) - `test/e2e/provider_aws_test.go`
    * Azure (`Label(provider:azure, provider:cloud)`) - `test/e2e/provider_azure_test.go`
    * Vsphere (`Label(provider:vsphere, provider:onprem)`) - `test/e2e/provider_vsphere_test.go`
 
The labels for these tests have been documented here: https://github.com/Mirantis/hmc/pull/360/files#diff-3a6cab76ecce19612a704fe0ef89dc83f1ed2df44f7bfa28d47d2529a7949fafR139

Each of the umbrella labels correspond to a GitHub job that is dependent on the `Build and Unit Test` job which provides setup.   The `Controller` tests always run, no matter the label, right now the `Controller` test is disabled because there aren't any tests other than Before/AfterSuite, but we can enable it because that sort of is a test of the controller in itself.

The provider tests are still attached to the `test e2e` label, but I think we can break these out further with `provider` specific GitHub labels at some point.

`provider:onprem` tests run on a self-hosted GitHub runner that has access to Mirantis' network, this is to support providers that do not necessarily have cloud infrastructure we can get to from a GitHub hosted runner -- like VSphere.

This PR also adds support for creating `ClusterIdentity`, `Credential` and their associated `Secret` resources  with a new `clusteridentity` package.

And of course, with testing comes bug fixing, so this PR fixes several different issues discovered along the way.

Closes: #210, Closes: #323

---

Note: I've opened two other issues to further simplify these tests, we could make these a lot easier to write in the future with:
* https://github.com/Mirantis/hmc/issues/427
* https://github.com/Mirantis/hmc/issues/426

PR #516, opened 2024-10-17
Fix up outstanding issues with e2e tests

squizzi:
Now that #360 has been merged the tests are actually running the right way (yay :tada:) and I can open a fork and they will run (another yay :tada:), but there may be some outstanding issues with the tests we need to resolve.  This PR will serve to iterate over the e2e tests and get them green running in their native environment so that we can collectively say Y-A-Y.

### DIFF
--- a/dummy
+++ b/dummy
@@ -1,0 +1,1 @@
+So dummy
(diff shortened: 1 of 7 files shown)


squizzi:
Looks like hosted tests are broken, if I can get #536 squared away I can try to get those green.

PR #518, opened 2024-10-17
Strip leading v from VERSION in 'Get outputs' job

squizzi:
The leading `v` was removed from `VERSION` in #505, but the test outputs get generated outside of the `Makefile`, this fixes the version used within the test workflow as well.  Unfortunately, with the nature of `pull_request_target` the CI won't pass until merged because this is a workflow file change (we really need to change this behavior when things cool down around here)

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -60,7 +60,8 @@ jobs:
       - name: Get outputs
         id: vars
         run: |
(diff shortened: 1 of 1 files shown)


squizzi:
Here's the test running with the change to build confidence that this will fix the controller test: 

https://github.com/Mirantis/hmc/actions/runs/11394890001/job/31706255408?pr=519

Kshatrix (approved):

PR #536, opened 2024-10-18
Get the PR ref from the merge commit within `pull_request_target` workflows

squizzi:
With our change to `pull_request_target` for workflow runs, `actions/checkout` needs to be updated to grab the ref from the PR, not from where the job is running.  This is because `pull_request_target` runs on the target branch.

This PR uses the GitHub API to fetch the merge commit sha and use that as the ref for `actions/checkout`

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -30,10 +30,22 @@ jobs:
     outputs:
       version: ${{ steps.vars.outputs.version }}
       clustername: ${{ steps.vars.outputs.clustername }}
+      pr: ${{ steps.pr.outputs.result }}
     steps:
+      - name: Get PR ref
+        uses: actions/github-script@v6
+        id: pr
+        with:
+          script: |
+            const { data: pullRequest } = await github.rest.pulls.get({
+              ...context.repo,
+              pull_number: context.payload.pull_request.number,
+            });
+            return pullRequest
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
+          ref: ${{fromJSON(steps.pr.outputs.result).merge_commit_sha}}
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


s3rj1k:
@squizzi Is switching to `pull_request_target` safe security-wise?

How repo is protected from credentials theft by 3-rd party actors via repo fork?

squizzi:
@s3rj1k We currently have the "outside contributors require approval" setting for PRs on and only run `pull_request_target` jobs that use secrets when the `test e2e` label is added to PRs.  Labeling a PR is only permitted for users with write permissions to this repo.  So with that, I say we're as protected as we can be if code is actually reviewed to ensure secret exfiltration does not occur.

Outside of that, we can just run these as `pull_request` and then not run e2e provider tests in PRs, only running them in `main`, which doesn't allow users to iterate against provider tests unless they create the PR as a branch against `Mirantis/hmc`, that means we limit testing of provider CI to outside forks.  So if someone changes a template or some functionality that changes provider code they can't test it prior to merge.

squizzi:
@s3rj1k we could also do something like this to provide additional security: https://michaelheap.com/access-secrets-from-forks/

a13x5 (commented):
IMO we should implement security features (discussed in the Slack) in a separate PR and narrow the scope of this one.

squizzi:
> IMO we should implement security features (discussed in the Slack) in a separate PR and narrow the scope of this one.

So what do you think the scope of this should be?  Just the ref change and then I'll move the last commit and the other proposed changes from slack to the new PR?

zerospiel (approved):

slysunkin (approved):

a13x5 (approved):
LGTM

PR #538, opened 2024-10-21
Fix CLI install on ARM64 platform.

s3rj1k:


Makefile
```diff
@@ -15,6 +15,8 @@ OS := $(strip ${os})
 ARCH := $(shell uname -m)
 ifeq ($(ARCH),x86_64)
 	ARCH := amd64
+else ifeq ($(ARCH),aarch64)
+	ARCH := arm64
 endif
```
a13x5, line 20: While we fixing it, let's use `go env GOARCH` instead of `uname -m` + `if`. This starting to get too ugly.

a13x5 (changes requested):
Minor note
The fail improvements look good.

Makefile
a13x5: I've been thinking only about `ARCH := $(go env GOARCH)`, but if you think about other check to add sure add them

s3rj1k: GOHOST* seems like a better choice in this specific case, as we want to know host OS data and not cross-compile.

a13x5: Got it. Sure, let's use `GOHOSTARCH` then

s3rj1k: fixed

a13x5 (approved):
LGTM

PR #539, opened 2024-10-21
EKS e2e tests

slysunkin:
## Description

Implementing https://github.com/Mirantis/hmc/issues/387

EKS E2E tests

### DIFF
--- a/test/utils/utils.go
+++ b/test/utils/utils.go
@@ -21,11 +21,11 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/Mirantis/hmc/internal/utils/status"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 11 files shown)


test/utils/utils.go
```diff
@@ -128,6 +129,10 @@ func ValidateConditionsTrue(unstrObj *unstructured.Unstructured) error {
 			continue
 		}
 
+		if slices.Contains(excludedConditions, c.Type) {
```
squizzi, line 132: Maybe this would be a better fit for the [Go functional options pattern](https://golang.cafe/blog/golang-functional-options-pattern.html) since it's only used in this one instance, then each time `ValidateConditionsTrue` is used it wouldn't need to be passed an empty `[]string`

slysunkin:
This PR had to be a from a branch

PR #544, opened 2024-10-23
Updated EKS chart version

slysunkin:
## Description

All charts versions were updated but EKS. This PR fixes EKS chart version inconsistency.


Kshatrix:
We don't need to keep the versions aligned. If there's a change in the chart since the previous release, we'll update the version

Kshatrix (approved):

PR #548, opened 2024-10-24
Miscallenous changes improving readme/onboarding

bnallapeta:
- Update IMG variable in Makefile to use localhost as the repository for the controller image.
- Update Makefile to update the image load command for podman
- Update region value in aws-managedcluster.yaml to use the AWS_REGION environment variable.
- Update repository value in hmc_values.yaml to use localhost as the repository for the controller image.
- Add AWS_SESSION_TOKEN to the list of required environment variables in dev.md.


config/dev/aws-managedcluster.yaml
```diff
@@ -13,7 +13,7 @@ spec:
       instanceType: t3.small
     controlPlaneNumber: 1
     publicIP: true
-    region: us-west-2
+    region: ${AWS_REGION}
```
s3rj1k, line 16: I think EKS YAMLs also would benefit from the environ

bnallapeta: Done.

docs/dev.md
```diff
@@ -27,6 +27,7 @@ The following env variables must be set in order to deploy dev cluster on AWS:
 
 - `AWS_ACCESS_KEY_ID`
 - `AWS_SECRET_ACCESS_KEY`
+- `AWS_SESSION_TOKEN`
```
eromanova, line 30: The `AWS_SESSION_TOKEN` variable is recommended but not required to deploy dev cluster on AWS. Let's explicitly mention that information.

eromanova (changes requested):
LGTM. One small comment.

config/dev/aws-managedcluster.yaml
a13x5, line 16: Let's use [well-known variable names](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) instead of inventing our own.

Default value for this one should be set to `us-east-2`

config/dev/hmc_values.yaml
```diff
@@ -1,5 +1,5 @@
 image:
-  repository: hmc/controller
+  repository: localhost/hmc/controller
```
a13x5, line 2: Why this is needed?

a13x5 (changes requested):
Don't know why do we need podman support. Is there any specific reason for that?

config/dev/aws-managedcluster.yaml
s3rj1k: Making this unconfigurable is a bad practice, why not expose it via environ with some sane default value so that anyone could set custom region without manually editing _some_ YAMLs?

P.S. Any other way of changing hard-coded region is also OK.

s3rj1k:
> podman support

Podman desktop is OSS without any limitations that has Docker Desktop (licensing), as long as HMC is targeted to be OSS it should also support Podman.

config/dev/aws-managedcluster.yaml
a13x5: I'm ok with making it configurable. I'm not ok with: bad naming; absence of default value (in Makefile).

a13x5:
> Docker Desktop

We don't have any requirements on Docker Desktop. You can user Docker CE or moby

s3rj1k:
> We don't have any requirements on Docker Desktop. You can user Docker CE or moby

HMC should help community to get quickly into a project and not make it harder, on Apple Silicon the most easy way of starting with HMC is by using Podman Desktop.

a13x5:
Ok, got it. Podman is needed for MacOS users. Thanks for clarifications

bnallapeta:
> Ok, got it. Podman is needed for MacOS users. Thanks for clarifications

Its not just for MacOS users. A lot of people (irrespective of their OS) prefer to use podman for multiple reasons instead of docker. As mentioned already, we as an OSS project should support as many configs as possible (within a reasonable effort from our end of course).

config/dev/hmc_values.yaml
bnallapeta:
```
Digest: sha256:367e2184d70c7bdda48e8a444ea1bf21f03587ee0975ed96f6d2739c15007ddd
/Users/bnr/work/github/hmc/bin/kind-v0.23.0 load docker-image hmc/controller:latest -n hmc-dev
enabling experimental podman provider
ERROR: image: "hmc/controller:latest" not present locally
make: *** [dev-push] Error 1
```

The image that is created is actually tagged as

```
podman images
REPOSITORY              	TAG        	IMAGE ID  	CREATED         	SIZE
localhost/hmc/controller	latest     	b62d60df23f2  About a minute ago  61.8 MB
```

config/dev/aws-managedcluster.yaml
bnallapeta: Didn't invent a new name. Got this from clusterawsadm output: 

```
$ clusterawsadm bootstrap iam create-cloudformation-stack
AWS_REGION env not set and --region flag not provided, default configuration will be used
```

More info [here](https://docs.aws.amazon.com/sdkref/latest/guide/feature-region.html) and [here](https://stackoverflow.com/questions/59961939/what-is-the-difference-between-aws-default-region-and-aws-region-system-variable).

Set the default value to us-east-2.
(although I am against this personally; its good to ask the user to set the region that they want instead of defaulting to something.)

config/dev/hmc_values.yaml
bnallapeta: btw, tested this on both docker and podman.

a13x5: Ok, got it, it's needed for podman to function

docs/dev.md
```diff
@@ -28,6 +28,10 @@ The following env variables must be set in order to deploy dev cluster on AWS:
 - `AWS_ACCESS_KEY_ID`
 - `AWS_SECRET_ACCESS_KEY`
 
+The following env variables are recommended although not neccessary for deployment.
```
a13x5, line 31: Why it's recommended? 
I suggest using word "optional" and also don't forget to add `AWS_REGION` here and mention that it has default value

config/dev/aws-managedcluster.yaml
a13x5: I'm convinced. Thanks for going extra mile and explaining it

a13x5 (commented):
One minor note regarding the doc.
Rest looks good

docs/dev.md
bnallapeta: Done. Changed it to optional and added a small explanation too.

a13x5: When I said default, I meant the one that you actually set in the Makefile. 

Why you're mentioning `~/.aws/config`? We're using `envsubst` it doesn't know about `~/.aws/config`.

bnallapeta: Done.

a13x5 (dismissed):
LGTM

eromanova (approved):

PR #550, opened 2024-10-24
Update Services Status for ManagedCluster & MultiClusterService

wahabmk:


internal/controller/multiclusterservice_controller.go
```diff
@@ -178,6 +222,91 @@ func helmChartOpts(ctx context.Context, c client.Client, namespace string, servi
...
+			return nil, fmt.Errorf("failed to get ClusterSummary %s to fetch status: %w", summaryRef.String(), err)
+		}
+
+		idx := -1
+		for i := range servicesStatus {
+			if obj.Name == servicesStatus[i].ClusterName && obj.Namespace == servicesStatus[i].ClusterNamespace {
+				idx = i
+				break
+			}
+		}
```
zerospiel, line 294: interchangeable with the:
```go
idx := slices.IndexFunc(servicesStatus, func(o hmc.ServiceStatus) bool {
    return obj.Name == o.ClusterName && obj.Namespace == o.ClusterNamespace
})
```

internal/sveltos/profile.go
```diff
@@ -224,3 +225,50 @@ func PriorityToTier(priority int32) (int32, error) {
 
 	return 0, fmt.Errorf("invalid value %d, priority has to be between %d and %d", priority, mini, maxi)
 }
+
+// SetStatusConditions transforms status from Sveltos ClusterSummary
+// object and sets it into the provided list of conditions.
+func SetStatusConditions(summary sveltosv1beta1.ClusterSummary, conditions *[]metav1.Condition) {
```
zerospiel, line 231: can the function's signature be changed in order to have the `clustersummary` instance as a reference instead of a copy?

internal/controller/multiclusterservice_controller.go
```diff
@@ -57,37 +67,71 @@ func (r *MultiClusterServiceReconciler) Reconcile(ctx context.Context, req ctrl.
...
+
+		apimeta.SetStatusCondition(&mcs.Status.Conditions, condition)
+		err = errors.Join(err, servicesErr, r.updateStatus(ctx, mcs))
+	}()
+
+	if controllerutil.AddFinalizer(mcs, hmc.MultiClusterServiceFinalizer) {
+		if err := r.Client.Update(ctx, mcs); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update MultiClusterService %s with finalizer %s: %w", mcs.Name, hmc.MultiClusterServiceFinalizer, err)
+		}
+		return ctrl.Result{Requeue: true}, nil
```
zerospiel, line 96: why do we need an explicit `requeue` here?

internal/controller/multiclusterservice_controller.go
```diff
@@ -192,9 +321,43 @@ func (r *MultiClusterServiceReconciler) reconcileDelete(ctx context.Context, mcs
...
+		l.Error(err, msg, "Name", o.GetName(), "Namespace", o.GetNamespace())
+		return []ctrl.Request{}
+	}
+
+	// The Profile/ClusterProfile object has the same name as its
+	// owner object which is either ManagedCluster or MultiClusterService.
+	req := client.ObjectKey{Name: ownerRef.Name}
+	if ownerRef.Kind == sveltosv1beta1.ProfileKind {
+		req.Namespace = o.GetNamespace()
+	}
```
zerospiel, line 347: could you elaborate on the comment and the idea behind the code block? there is a statement that the owner of a `clustersummary` object is either an `mcl` (ns-scoped) or `mcs` (cluster-scope), so why do we contest the owner reference's kind against sveltos' `profile`? 


just in case: I am not sure whether the logger already has `name`/`namespace` values, if so, then the same keys might be confusing, so maybe something similar (not exactly those) to the `clustersummary-name`/`clustersummary-namespace` keys

internal/controller/multiclusterservice_controller.go
```diff
@@ -192,9 +321,43 @@ func (r *MultiClusterServiceReconciler) reconcileDelete(ctx context.Context, mcs
...
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *MultiClusterServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&hmc.MultiClusterService{}).
+		For(&hmc.MultiClusterService{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Watches(&sveltosv1beta1.ClusterSummary{},
+			handler.EnqueueRequestsFromMapFunc(requeueSveltosProfileForClusterSummary),
+			builder.WithPredicates(predicate.Funcs{
```
zerospiel, line 358: the same questions as of `managedcluster-ctrl`: do we need the `generic` event here and do we need to trigger the reconciliation if a `clustersummary` object gets its status update?

internal/controller/multiclusterservice_controller.go
wahabmk: I think the comment on the function I made might be confusing. It should be:
```go
// requeueSveltosProfileForClusterSummary asserts that the requested object has Sveltos ClusterSummary
// type, fetches its owner (a Sveltos Profile or ClusterProfile object), and requeues the ManagedCluster 
// or MultiClusterService object that is referenced by the same namespace/name as the Sveltos Profile/ClusterProfile.
```
The ownership of these objects goes like this:
```
ClusterSummary -> is owned by Profile/ClusterProfile -> is owned by ManagedCluster/MultiClusterService
```
Where the ManagedCluster and MultiClusterService have the same reference (namespace/name) as the Profile and ClusterProfile objects that the own respectively. So what this function is doing is:

1. Making sure that the requested reference is for a ClusterSummary object.
2. Fetching the reference for its owner (Profile or ClusterProfile) and requeuing that reference.
3. Since this reference is the same as that of an `mcl` or `mcs` object, we end up running reconcile on the `mcl` or `mcs` object, which is what we want to fetch the latest status from their respective ClusterSummary objects.

> just in case: I am not sure whether the logger already has name/namespace values

I can check that. Thanks for pointing it out.

internal/controller/multiclusterservice_controller.go
wahabmk: Same answer.

internal/controller/multiclusterservice_controller.go
zerospiel: The reason for such behavior is that adding a finalizer does not increase the generation of an object. So I would amend my initial question: do we need such a predicate for the main resource type being watched? I mean it's totally okay to have it, I just want to understand the reasoning behind this decision, since it might be tricky to debug. In general, this is a good decision to limit the number of events only to `spec` changes

internal/controller/managedcluster_controller.go
```diff
@@ -1022,5 +1021,11 @@ func (r *ManagedClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				}
 			}),
 		).
+		Watches(&sveltosv1beta1.ClusterSummary{},
```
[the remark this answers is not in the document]
zerospiel: Ok, thanks for the explanation, makes sense then not to limit it to the generation. Yes, I think the `generic` might be turned off though

internal/controller/multiclusterservice_controller.go
zerospiel: Ok, now I get it, it was my inattention to the code that I missed the ownership chain. So I think everything is fine here, and I suggest amending the func comment or adding an extra comment within the func just in case (I believe your comment here is very well described the behavior)

internal/controller/multiclusterservice_controller.go
wahabmk: Without the `GenerationChangedPredicate{}`, the reconciliation went into a dead loop indefinitely.

internal/controller/multiclusterservice_controller.go
```diff
@@ -57,37 +67,71 @@ func (r *MultiClusterServiceReconciler) Reconcile(ctx context.Context, req ctrl.
...
 	}
 
-	if controllerutil.AddFinalizer(mcsvc, hmc.MultiClusterServiceFinalizer) {
-		if err := r.Client.Update(ctx, mcsvc); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to update MultiClusterService %s with finalizer %s: %w", mcsvc.Name, hmc.MultiClusterServiceFinalizer, err)
+	return r.reconcile(ctx, mcsvc)
+}
+
+func (r *MultiClusterServiceReconciler) reconcile(ctx context.Context, mcs *hmc.MultiClusterService) (res ctrl.Result, err error) {
+	var servicesErr error
```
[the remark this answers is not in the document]
wahabmk: Yeah. To fix this, I would need to not use `err` in the return signature. So with the example you shared, it will become:
https://goplay.tools/snippet/qSblUFyHOu9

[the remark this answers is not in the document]
wahabmk: The same as above applies to the reconciliation of ManagedCluster as well with regards to fetching the status of services.

internal/controller/multiclusterservice_controller.go
wahabmk: In an attempt to reproduce the dead loop:

- I removed the requeue:
```suggestion
		return ctrl.Result{}, nil
```

- And also removed the following:
```diff
func (r *MultiClusterServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
-		For(&hmc.MultiClusterService{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&hmc.MultiClusterService{}).
		Watches(&sveltosv1beta1.ClusterSummary{},
			handler.EnqueueRequestsFromMapFunc(requeueSveltosProfileForClusterSummary),
			builder.WithPredicates(predicate.Funcs{
				DeleteFunc:  func(event.DeleteEvent) bool { return false },
				GenericFunc: func(event.GenericEvent) bool { return false },
			}),
		).
		Complete(r)
}
```

- Then I created a ManagedCluster with `servicesPriority=100`:
- Then I created a MultiClusterService with `servicesPriority=50` but did not observe the dead loop.
- So I changed the `servicesPriority=1000` for the MultiClusterService but still did not observe the dead loop as indicated by `generation: 2` below:
```sh
➜  hmc git:(services-status) khmc get multiclusterservices.hmc.mirantis.com global-ingress -o yaml
kind: MultiClusterService
metadata:
  . . .
  finalizers:
  - hmc.mirantis.com/multicluster-service
  generation: 2
  name: global-ingress
  resourceVersion: "10082"
  uid: 181fe130-95a6-493f-a9e4-dd3fcff009dc
spec:
  clusterSelector:
    matchLabels:
  services:
  - name: ingress-nginx
    namespace: ingress-nginx
    template: ingress-nginx-4-11-3
  servicesPriority: 1000
  stopOnConflict: false
status:
  conditions:
  - lastTransitionTime: "2024-10-31T07:05:35Z"
    message: ""
    reason: Succeeded
    status: "True"
    type: SveltosClusterProfileReady
  - lastTransitionTime: "2024-10-31T07:05:35Z"
    message: ""
    reason: Succeeded
    status: "True"
    type: FetchServicesStatusSuccess
  - lastTransitionTime: "2024-10-31T07:05:35Z"
    message: MultiClusterService is ready
    reason: Succeeded
    status: "True"
    type: Ready
  observedGeneration: 2
  services:
  - clusterName: wali-aws-dev
    clusterNamespace: hmc-system
    conditions:
    - lastTransitionTime: "2024-10-31T07:06:52Z"
      message: ""
      reason: Provisioned
      status: "True"
      type: Helm
    - lastTransitionTime: "2024-10-31T07:06:32Z"
      message: Release ingress-nginx/ingress-nginx
      reason: Managing
      status: "True"
      type: ingress-nginx.ingress-nginx/SveltosHelmReleaseReady
```

So this would suggest that there is no bug in the Idempotence property of the reconcile. I think I may have experienced it while I was still in the process of implementing status updates in reconcile. 

Nevertheless, I think we should keep the `GenerationChangedPredicate{}` and I added a comment explaining why the requeue is needed.

zerospiel: Ok, thanks for checking it on-hands. I'm also okay with the predicate

zerospiel (approved):

zerospiel (dismissed):

zerospiel (approved):

PR #574, opened 2024-10-29
Validate mgmt providers removal

### LINKED ISSUE #554: Forbid to remove the provider from Management if the provider is in use
The admission should validate and forbid the removal of the provider from the Management `spec.providers` if there is at least one managed cluster that uses this provider.

zerospiel:
* forbid Management .spec.providers items removal if a ManagedCluster utilizes one of the removed providers
* amends to CRDs descriptions

Closes #554


internal/webhook/management_webhook.go
```diff
@@ -55,21 +58,109 @@ func (*ManagementValidator) ValidateCreate(_ context.Context, _ runtime.Object)
...
+
+	if incompatibleContracts != "" {
+		return admission.Warnings{"The Management object has incompatible CAPI contract versions in ProviderTemplates"}, fmt.Errorf("%s: %s", invalidMgmtMsg, incompatibleContracts)
+	}
+
+	return nil, nil
+}
+
+func checkComponentsRemoval(ctx context.Context, cl client.Client, oldMgmt, newMgmt *hmcv1alpha1.Management) error {
+	removedComponentNames := []string{}
```
Kshatrix, line 94: nit: change to a map to not iterate over a slice in the loop

internal/webhook/management_webhook.go
```diff
@@ -55,21 +58,109 @@ func (*ManagementValidator) ValidateCreate(_ context.Context, _ runtime.Object)
...
+		}
+	}
+
+	if len(removedComponentNames) == 0 {
+		return nil
+	}
+
+	removedProvidersSet := make(map[string]struct{})
+	for _, m := range removedComponentNames {
+		c, ok := oldMgmt.Status.Components[m]
```
Kshatrix, line 107: the fact that it's not reflected in the status does not guarantee it wasn't installed.

internal/webhook/management_webhook.go
```diff
@@ -55,21 +58,109 @@ func (*ManagementValidator) ValidateCreate(_ context.Context, _ runtime.Object)
...
+			removedComponentNames = append(removedComponentNames, oldComp.Name)
+		}
+	}
+
+	if len(removedComponentNames) == 0 {
+		return nil
+	}
+
+	removedProvidersSet := make(map[string]struct{})
+	for _, m := range removedComponentNames {
```
Kshatrix, line 106: I don't think it makes much sense to iterate over the management object status. the diff from comparing two specs should be enough

internal/webhook/management_webhook.go
```diff
@@ -55,21 +58,109 @@ func (*ManagementValidator) ValidateCreate(_ context.Context, _ runtime.Object)
...
+			removedProvidersSet[pn] = struct{}{}
+		}
+	}
+
+	if len(removedProvidersSet) == 0 { // sanity
+		return nil
+	}
+
+	mcls := new(hmcv1alpha1.ManagedClusterList)
+	if err := cl.List(ctx, mcls); err != nil {
```
Kshatrix, line 128: we have an indexer to select managedclusters by their templates that I think we should use here

Kshatrix (changes requested):

internal/webhook/management_webhook.go
zerospiel: arrays are more performant on such small datasets even when only misses are being benchmarked 

```
goos: darwin
goarch: arm64
pkg: github.com/zerospiel/playground
cpu: Apple M1 Pro
BenchmarkStrs/arrays-10         	44406596	        28.77 ns/op	       0 B/op	       0 allocs/op
BenchmarkStrs/maps-10           	12531898	        97.03 ns/op	       0 B/op	       0 allocs/op
```

internal/webhook/management_webhook.go
zerospiel: correct but it's unclear (to me) how is it possible to extract provider names having only component names (if templates are not presented, should we fetch a `releases` object then and try to find a template reference there?)

internal/webhook/management_webhook.go
zerospiel: as of my understanding, field match works only for a single entry (if no custom selector is implemented), if so it requires multiple lists having multiple template names, so both implementations seem equal to me

Kshatrix: you should not be limited to just `MatchingFields`, but even if you were, it would still be better to have a separate api call per template with `Limit: 1`

internal/webhook/management_webhook.go
Kshatrix: yes, if not referenced directly, should be taken from the release. alternatively we'd need to ensure that the status represents the "old" version of the object by comparing generations (but in that case we'd have to fail validation if it doesn't)

internal/webhook/management_webhook.go
zerospiel: took a quick look, the existing indexers related to the `clustertemplate` resource and not the `providertemplates`, so basically the idea is to acquire the provider names, for this purpose the `providertemplates` should be fetched, and then we have to fetch all the `clustertemplates` related to a `managedclusters` object in order to determine whether the provider names are actually in use, and the latter requires to fetch all `managedclusters` objects having no `clustertemplates` preliminary. So I cannot see how I can leverage the indexer in such a case tbh

Kshatrix: yes, the idea was to identify `clustertemplates` associated with the removed `providertemplates` and then query `managedclusters` just once

Kshatrix (approved):

PR #575, opened 2024-10-29
Break workflows into separate 'E2E Test' and 'Build and Unit Test' workflows, use `pull_request` instead of `pull_request_target` across workflows

squizzi:
This PR makes the workflow changes discussed on Slack, moving away from `pull_request_target` and using `pull_request` across workflows.  The decision to move away from `pull_request_target` entirely was made here: https://github.com/Mirantis/hmc/pull/575#issuecomment-2450177121


squizzi:
The `pull_request_target` workflows are going to continue to run in this PR since they're based on `main`, so just ignore them.  `Build and Unit Tests` is the workflow being added, which is green.

squizzi:
I'm still seeing issues running tests on `pull_request_target`, I think it's time to just bite the bullet and call it what it is, with all of the discussion around `pull_request_target` and the unnecessary workarounds and additional checks we can now omit I believe the best step is to just run any work that requires running e2e tests on a repo branch and perhaps run them as nightly's to catch any outstanding issues from other PRs.

squizzi:
Closing this for now as I no longer work with Mirantis.  Feel free to reopen this and/or carry it if there's a desire to use it in the future.

PR #584, opened 2024-10-31
Cleanup component dependencies when removed

### LINKED ISSUE #553: Deinstall provider components if it was removed from the Management spec
The management controller should uninstall provider components if they are removed from the management `spec.providers`.

zerospiel:
* if component is removed from management object the related helmreleases, providertemplates and helmcharts are deleted now
* small code amends
* test coverage for the mgmt-ctrl raised from 4.4% to 45.8%

Closes #553

### DIFF
--- a/internal/controller/management_controller.go
+++ b/internal/controller/management_controller.go
@@ -45,15 +45,6 @@ import (
 	"github.com/Mirantis/hmc/internal/utils/status"
 )
 
-// Those are only needed for the initial installation
-var enforcedValues = map[string]any{
-	"controller": map[string]any{
-		"createManagement":         false,
-		"createTemplateManagement": false,
-		"createRelease":            false,
-	},
-}
-
 // ManagementReconciler reconciles a Management object
 type ManagementReconciler struct {
 	client.Client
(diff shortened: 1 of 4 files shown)


internal/controller/management_controller.go
```diff
@@ -92,101 +83,143 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
+	for compName, compSt := range management.Status.Components {
+		if compName == hmc.CoreCAPIName ||
+			compName == hmc.CoreHMCName ||
+			slices.ContainsFunc(management.Spec.Providers, func(newComp hmc.Provider) bool { return compName == newComp.Name }) {
+			continue
+		}
+
+		l.Info("Found component to remove", "component_name", compName)
+
+		// removing providertemplate will also delete its ownees, including HelmCharts
```
Kshatrix, line 199: we shouldn't remove provider templates when we disable providers

internal/controller/management_controller.go
```diff
@@ -92,101 +83,143 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
+		p.SetGroupVersionKind(hmc.GroupVersion.WithKind(hmc.ProviderTemplateKind))
+		p.SetName(compSt.Template)
+		if err := r.Client.Delete(ctx, p); client.IgnoreNotFound(err) != nil {
+			errs = errors.Join(errs, fmt.Errorf("failed to delete %s %s: %w", p.GroupVersionKind(), p.Name, err))
+			continue
+		}
+		l.Info("Removed ProviderTemplate", "reference", client.ObjectKeyFromObject(p).String())
+
+		hr := new(fluxv2.HelmRelease)
+		hr.SetGroupVersionKind(fluxv2.GroupVersion.WithKind(fluxv2.HelmReleaseKind))
```
Kshatrix, line 210: this looks redundant

internal/controller/management_controller.go
```diff
@@ -92,101 +83,143 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
 	return ctrl.Result{}, nil
 }
 
+func (r *ManagementReconciler) cleanupRemovedComponents(ctx context.Context, management *hmc.Management) error {
+	var (
+		errs error
+		l    = ctrl.LoggerFrom(ctx)
+	)
+
+	for compName, compSt := range management.Status.Components {
```
Kshatrix, line 190: the management object status cannot be relied upon as a source of information in the current implementation. I suggest determining the current state based on HelmRelease objects instead

Kshatrix (changes requested):

internal/controller/management_controller.go
```diff
@@ -92,101 +83,150 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
+
+		if componentName == hmc.CoreCAPIName ||
+			componentName == hmc.CoreHMCName ||
+			slices.ContainsFunc(management.Spec.Providers, func(newComp hmc.Provider) bool { return componentName == newComp.Name }) {
+			continue
+		}
+
+		l.Info("Found component to remove", "component_name", componentName)
+
+		if hr.Spec.ChartRef != nil {
```
Kshatrix, line 209: We should not remove helm charts here. They are associated with templates and will be garbage collected following template removal

Kshatrix (changes requested):

Kshatrix (approved):

PR #599, opened 2024-11-05
Explicit linter rules

zerospiel:
* set explicit linter rules for revive and govet
* granular exclusions for govet/revive
* removed unused exclusions
* added several extra linters

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,31 +9,39 @@ issues:
   # restore some of the defaults
   # (fill in the rest as needed)
   exclude-rules:
-    - path: "internal/*"
-      linters:
-        - dupl
-    - text: "struct-tag: unknown option 'inline' in JSON tag"
-      linters:
-        - revive
-    - text: "Unhandled error in call to function fmt.Print*"
-      linters:
-        - revive
-    - path: cmd/main.go
+    - path: (cmd\/main\.go)|(.*_test\.go)
       linters:
         - maintidx
     - path: test/
       linters:
         - perfsprint
+        - errorlint
+        - exhaustive
+    - text: 'shadow: declaration of "err" shadows declaration'
+      linters:
+        - govet
+    - text: "fieldalignment: "
+      path: (.*_test\.go|test\/|api\/?.*\/.*_types\.go) # ignore tests, and k8s API-specific files
+      linters:
+        - govet
+    - text: "max-public-structs: you have exceeded the maximum number of public struct declarations"
+      linters:
+        - revive
+      path: api/ # the api/ pkgs have lots of structs
+
 linters:
   disable-all: true
   enable:
+    - asasalint
     - asciicheck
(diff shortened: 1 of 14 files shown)


.golangci.yml
```diff
@@ -89,10 +99,35 @@ linters:
     - whitespace
 
 linters-settings:
+  depguard:
+    rules:
+      # Name of a rule.
+      main:
+        # Packages that are not allowed where the value is a suggestion.
+        deny:
+          - pkg: "github.com/pkg/errors"
```
s3rj1k, line 108: Hehe, I like this rule :)

.golangci.yml
```diff
@@ -109,37 +144,188 @@ linters-settings:
...
+  prealloc:
+    for-loops: true
   revive:
-    enable-all-rules: true
+    # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md
     rules:
       - name: add-constant
         disabled: true
+      - name: argument-limit
+        arguments: [6]
```
s3rj1k, line 204: I would suggest removing this linter, I understand that most of the time there is no sane need for 6+ args, but what if they are needed, this config will just be changed to new max value

.golangci.yml
```diff
@@ -109,37 +144,188 @@ linters-settings:
...
+          - "sayRepetitiveInsteadOfStutters"
+      - name: file-header
+        disabled: true
+      - name: flag-parameter
       - name: function-length
         disabled: true
-      - name: max-public-structs
-        disabled: true # the api/* pkgs have lots of structs
+      - name: function-result-limit
+        arguments: [3]
```
s3rj1k, line 260: same here as for argument-limit

.golangci.yml
```diff
@@ -109,37 +144,188 @@ linters-settings:
...
+          - "^[a-z][a-z0-9]{0,}$"
+      - name: import-shadowing
+      - name: imports-blocklist
+        disabled: true
+      - name: increment-decrement
+      - name: indent-error-flow
       - name: line-length-limit
         disabled: true
+      - name: max-control-nesting
+        arguments: [3]
```
s3rj1k, line 275: limit on nested ifs?

.golangci.yml
```diff
@@ -109,37 +144,188 @@ linters-settings:
...
+      - name: max-public-structs
```
s3rj1k, line 276: Also suggest to disable that, some public APIs will fail that linter

.golangci.yml
zerospiel: defautl is 5, will use it

.golangci.yml
zerospiel: public APIs are removed from this check

s3rj1k:
LGTM

Kshatrix (approved):

PR #623, opened 2024-11-11
Implemented the unmanaged cluster CRD for adopting existing k8s clusters

kylewuolle:
Resolves https://github.com/Mirantis/hmc/issues/542

This PR implements adopting existing k8s clusters into 2a.


internal/controller/unmanagedcluster_controller.go
```diff
@@ -0,0 +1,505 @@
...
+				},
+				InfrastructureRef: corev1.ObjectReference{
+					Kind:       "UnmanagedMachine",
+					Namespace:  unmanagedCluster.Namespace,
+					Name:       node.Name,
+					APIVersion: hmc.GroupVersion.Identifier(),
+				},
+				ProviderID: &node.Spec.ProviderID,
+			},
+			Status: v1beta1.MachineStatus{
```
Kshatrix, line 251: same

internal/controller/unmanagedcluster_controller.go
```diff
@@ -0,0 +1,505 @@
...
+		}
+	}
+
+	// cleanup any orphaned unmanaged machines and capi machines
+	for _, existingUnmanagedMachine := range existingMachinesByName {
+		if err := r.Delete(ctx, existingUnmanagedMachine); err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete orphaned unmanaged machine: %w", err)
+		}
+
+		if err := r.Delete(ctx, &v1beta1.Machine{
```
Kshatrix, line 282: you should rather use ownerreference instead

internal/controller/unmanagedcluster_controller.go
```diff
@@ -0,0 +1,505 @@
...
+
+func (r *UnmanagedClusterReconciler) getNodeList(ctx context.Context, unmanagedCluster *hmc.UnmanagedCluster) (*corev1.NodeList, error) {
+	l := ctrl.LoggerFrom(ctx)
+	clusterClient, err := clusterproxy.GetCAPIKubernetesClient(ctx, l, r.Client, r.Client.Scheme(), unmanagedCluster.Namespace, unmanagedCluster.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to remote cluster: %w", err)
+	}
+
+	nodelist := &corev1.NodeList{}
+	if err := clusterClient.List(ctx, nodelist); err != nil {
```
Kshatrix, line 302: I think we should process in batches here as it could be potentially a lot of nodes

internal/controller/unmanagedcluster_controller.go
```diff
@@ -0,0 +1,505 @@
...
+
+func (r *UnmanagedClusterReconciler) reconcileUnmanagedCluster(ctx context.Context, unmanagedCluster *hmc.UnmanagedCluster) (ctrl.Result, error) {
+	l := ctrl.LoggerFrom(ctx)
+	if !unmanagedCluster.DeletionTimestamp.IsZero() {
+		l.Info("Deleting UnmanagedCluster")
+		return r.reconcileDeletion(ctx, unmanagedCluster)
+	}
+
+	if err := r.createCluster(ctx, unmanagedCluster); err != nil {
+		return ctrl.Result{RequeueAfter: DefaultRequeueInterval}, err
```
Kshatrix, line 316: I suggest to rely on exponential backoff instead of a fixed one in general (unless there are reasons for that)

internal/controller/unmanagedcluster_controller.go
```diff
@@ -0,0 +1,505 @@
...
+			if len(failedCharts) > 0 {
+				helmCondition.Message = "Charts failed to deploy " + strings.Join(failedCharts, ",")
+			}
+			apimeta.SetStatusCondition(cluster.GetConditions(), helmCondition)
+		} else {
+			requeue = true
+		}
+	}
+
+	if err := r.Status().Update(ctx, cluster); err != nil {
```
Kshatrix, line 500: if anything fails before the update cal, there will be no status updated at all

internal/controller/unmanagedcluster_controller.go
```diff
@@ -0,0 +1,505 @@
...
+			}
+		}
+	}
+
+	if len(nonReadyNodes) > 0 {
+		allNodeCondition.Message = fmt.Sprintf("Nodes %s are not ready", strings.Join(nonReadyNodes, ","))
+	}
+	apimeta.SetStatusCondition(cluster.GetConditions(), allNodeCondition)
+
+	if len(cluster.Spec.Services) > 0 {
```
Kshatrix, line 462: we should unify services status updates across managed and unmanaged clusters

templates/provider/hmc/templates/rbac/controller/rolebindings.yaml
```diff
@@ -28,3 +28,20 @@ subjects:
...
+kind: RoleBinding
+metadata:
+  name: {{ include "hmc.fullname" . }}-capi-manager-rolebinding
+  namespace: {{ .Release.Namespace }}
+  labels:
+  {{- include "hmc.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: '{{ include "hmc.fullname" . }}-manager-role'
```
Kshatrix, line 43: I think assigning the whole hmc-manager-role is quite excessive here

templates/provider/hmc/templates/rbac/user-facing/unmanagedcluster-editor.yaml
```diff
@@ -0,0 +1,13 @@
+# permissions for end users to edit unmanagedclusters.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    hmc.mirantis.com/aggregate-to-global-admin: "true"
```
Kshatrix, line 6: I think rbac considerations should be similar to managedclusters, unless it was decided otherwise

Kshatrix (changes requested):

a13x5 (changes requested):
IMO The design of using unmanaged cluster as an infrastructure provider looks needlessly complicated for our purposes.

Status handling logic should be reconsidered.
kubebuilder comments must be cleared out.

templates/provider/hmc/templates/rbac/controller/rolebindings.yaml
kylewuolle: Not sure what you mean here

PR #673, opened 2024-11-21
Improve retrieval of services status from Sveltos

wahabmk:


internal/sveltos/status.go
```diff
@@ -25,13 +25,14 @@ import (
 	hmc "github.com/Mirantis/hmc/api/v1alpha1"
 )
 
-// SetStatusConditions transforms status from Sveltos ClusterSummary
-// object and sets it into the provided list of conditions.
-func SetStatusConditions(summary *sveltosv1beta1.ClusterSummary, conditions *[]metav1.Condition) error {
+// GetStatusConditions returns a list of conditions from provided ClusterSummary.
+func GetStatusConditions(summary *sveltosv1beta1.ClusterSummary) ([]metav1.Condition, error) {
```
eromanova, line 29: Minor comment: I would keep an old `SetStatusConditions` name. Since the function still focuses on setting the conditions I think this name is still more appropriate than `GetStatusConditions`. What do you think?

It would also be helpful to mention in the comment above the function that it not only returns a list of conditions but also updates them.

wahabmk: Yeah I think I agree. `GetStatusConditions` might suggest that we are retrieving status through an API call which is not the case.

eromanova (approved):

internal/sveltos/status.go
```diff
@@ -25,13 +25,14 @@ import (
...
-// object and sets it into the provided list of conditions.
-func SetStatusConditions(summary *sveltosv1beta1.ClusterSummary, conditions *[]metav1.Condition) error {
+// SetStatusConditions creates & sets a list of conditions from provided ClusterSummary and returns it.
+func SetStatusConditions(summary *sveltosv1beta1.ClusterSummary) ([]metav1.Condition, error) {
 	if summary == nil {
-		return errors.New("nil summary provided")
+		return nil, errors.New("error getting status from ClusterSummary: nil summary provided")
 	}
 
+	conditions := []metav1.Condition{}
```
zerospiel, line 34: nits: i think we can preallocate the `conditions` slice with something like `conditions := make(..., 0, len(summary.Status.FeatureSummaries)` and probably change the function's name from `Set` to `Get` or similar (because now it does not modify the given slice of conditions)

wahabmk: Preallocated `conditions`. For the name, we decided to keep it `Set`, please see convo above.

eromanova (approved):

internal/sveltos/status.go
eromanova: Ah, now I see, yes, it looks like it’s better to keep Get as Michael suggested and as you did initially. I was a little bit confused by the `SetStatusCondition`. @wahabmk  Let’s bring the initial version if you don’t mind.

eromanova: @zerospiel I'll merge this PR as is to close the task but update the function name in a separate PR.

PR #676, opened 2024-11-21
Fix - e2e test debug

kylewuolle:
This is a temporary measure to allow changes to the github workflow file on PRs so we can figure out why the e2e tests are currently failing. Once those are fixed we will set it back to pull_request_target.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -1,6 +1,6 @@
 name: CI
 on:
-  pull_request_target:
+  pull_request:
     types:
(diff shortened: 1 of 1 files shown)


slysunkin (approved):
I used the same in my experiments

a13x5:
I suggest separating e2e tests in a separate workflow and keep `pull_request_target` for build and unit tests.

kylewuolle:
That might be the only way for this to work. In the meantime I might have figured out another way to figure out what's going on.

PR #694, opened 2024-11-27
Test e2e fixes

kylewuolle:


DinaBelova:
discussed with @kylewuolle offline
this was temporary PR to verify different options (next time maybe better to mark as draft) - the actual fixing rn is happening together with https://github.com/k0rdent/kcm/pull/855

PR #697, opened 2024-11-28
Add OpenStack credential propagation support

bnallapeta:
Local Run:

```
ka get machines
NAME                           CLUSTER         NODENAME                       PROVIDERID                                          PHASE     AGE     VERSION
openstack-dev-cp-0             openstack-dev   openstack-dev-cp-0             openstack:///b52ccc71-566b-4487-a0b8-64b2c08f0e5b   Running   8m29s   v1.30.5+k0s.0
openstack-dev-cp-1             openstack-dev   openstack-dev-cp-1             openstack:///4022efcb-da46-4a99-80f3-2eaf3681238e   Running   8m29s   v1.30.5+k0s.0
openstack-dev-cp-2             openstack-dev   openstack-dev-cp-2             openstack:///87ae87e0-2ac5-4f06-b7ba-970d6dbcab3f   Running   8m29s   v1.30.5+k0s.0
openstack-dev-md-p27j6-6wfn4   openstack-dev   openstack-dev-md-p27j6-6wfn4   openstack:///5c2c51c8-f2c0-48d3-9c8f-ed28db53112a   Running   8m30s   
openstack-dev-md-p27j6-w6qvx   openstack-dev   openstack-dev-md-p27j6-w6qvx   openstack:///2df39a6d-2507-452f-9a0a-fb3e6a5d2052   Running   8m30s   
```

```
clusterctl describe cluster openstack-dev -n hmc-system
NAME                                                      READY  SEVERITY  REASON  SINCE  MESSAGE                                                        
Cluster/openstack-dev                                     True                     5m26s                                                                  
├─ClusterInfrastructure - OpenStackCluster/openstack-dev                                                                                                  
├─ControlPlane - K0sControlPlane/openstack-dev-cp                                                                                                         
│ └─3 Machines...                                         True                     7m32s  See openstack-dev-cp-0, openstack-dev-cp-1, ...                 
└─Workers                                                                                                                                                 
  └─MachineDeployment/openstack-dev-md                    True                     110s                                                                   
    └─2 Machines...                                       True                     3m43s  See openstack-dev-md-p27j6-6wfn4, openstack-dev-md-p27j6-w6qvx  
```

```
k -n hmc-system get managedclusters.hmc.mirantis.com 
NAME            READY   STATUS
openstack-dev   True    ManagedCluster is ready
```

### DIFF
--- a/config/dev/openstack-credentials.yaml
+++ b/config/dev/openstack-credentials.yaml
@@ -1,0 +1,33 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: openstack-cloud-config
+  namespace: ${NAMESPACE}
+stringData:
+  clouds.yaml: |
+    clouds:
+      openstack:
+        auth:
+          auth_url: ${OS_AUTH_URL}
+          username: ${OS_USERNAME}
+          password: ${OS_PASSWORD}
+          project_id: ${OS_PROJECT_ID}
+          project_name: ${OS_PROJECT_NAME}
+          user_domain_name: ${OS_USER_DOMAIN_NAME}
+        region_name: ${OS_REGION_NAME}
+        interface: ${OS_INTERFACE}
+        identity_api_version: ${OS_IDENTITY_API_VERSION}
+---
+apiVersion: hmc.mirantis.com/v1alpha1
+kind: Credential
+metadata:
+  name: openstack-cluster-identity-cred
+  namespace: ${NAMESPACE}
+spec:
+  description: OpenStack credentials
+  identityRef:
+    apiVersion: v1
+    kind: Secret
+    name: openstack-cloud-config
+    namespace: ${NAMESPACE}
--- a/config/dev/openstack-credentials.yaml
+++ b/config/dev/openstack-credentials.yaml
@@ -10,14 +10,12 @@ stringData:
       openstack:
         auth:
           auth_url: ${OS_AUTH_URL}
-          username: ${OS_USERNAME}
-          password: ${OS_PASSWORD}
-          project_id: ${OS_PROJECT_ID}
-          project_name: ${OS_PROJECT_NAME}
-          user_domain_name: ${OS_USER_DOMAIN_NAME}
+          application_credential_id: ${OS_APPLICATION_CREDENTIAL_ID}
+          application_credential_secret: ${OS_APPLICATION_CREDENTIAL_SECRET}
         region_name: ${OS_REGION_NAME}
         interface: ${OS_INTERFACE}
         identity_api_version: ${OS_IDENTITY_API_VERSION}
+        auth_type: ${OS_AUTH_TYPE}
 ---
 apiVersion: hmc.mirantis.com/v1alpha1
 kind: Credential
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -763,6 +763,25 @@ func (r *ClusterDeploymentReconciler) reconcileCredentialPropagation(ctx context
 				Reason:  hmc.SucceededReason,
 				Message: "vSphere CCM credentials created",
 			})
+		case "openstack":
+			l.Info("OpenStack creds propagation start")
+			if err := credspropagation.PropagateOpenStackSecrets(ctx, propnCfg); err != nil {
+				errMsg := fmt.Sprintf("failed to create OpenStack CCM credentials: %s", err)
+				apimeta.SetStatusCondition(clusterDeployment.GetConditions(), metav1.Condition{
+					Type:    hmc.CredentialsPropagatedCondition,
+					Status:  metav1.ConditionFalse,
+					Reason:  hmc.FailedReason,
+					Message: errMsg,
+				})
+				return errors.New(errMsg)
+			}
+
+			apimeta.SetStatusCondition(clusterDeployment.GetConditions(), metav1.Condition{
+				Type:    hmc.CredentialsPropagatedCondition,
+				Status:  metav1.ConditionTrue,
+				Reason:  hmc.SucceededReason,
+				Message: "OpenStack CCM credentials created",
+			})
 		default:
 			apimeta.SetStatusCondition(clusterDeployment.GetConditions(), metav1.Condition{
 				Type:    hmc.CredentialsPropagatedCondition,
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -355,7 +355,7 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, mc *hmc
 	}
 
 	if mc.Spec.PropagateCredentials {
-		if err := r.reconcileCredentialPropagation(ctx, mc); err != nil {
+		if err := r.reconcileCredentialPropagation(ctx, mc, cred); err != nil {
 			l.Error(err, "failed to reconcile credentials propagation")
 			return ctrl.Result{}, err
 		}
@@ -696,7 +696,7 @@ func (r *ClusterDeploymentReconciler) objectsAvailable(ctx context.Context, name
 	return len(itemsList.Items) != 0, nil
 }
 
-func (r *ClusterDeploymentReconciler) reconcileCredentialPropagation(ctx context.Context, clusterDeployment *hmc.ClusterDeployment) error {
+func (r *ClusterDeploymentReconciler) reconcileCredentialPropagation(ctx context.Context, clusterDeployment *hmc.ClusterDeployment, credential *hmc.Credential) error {
 	l := ctrl.LoggerFrom(ctx)
 	l.Info("Reconciling CCM credentials propagation")
 
@@ -715,8 +715,9 @@ func (r *ClusterDeploymentReconciler) reconcileCredentialPropagation(ctx context
 
 	propnCfg := &credspropagation.PropagationCfg{
 		Client:            r.Client,
-		ClusterDeployment: clusterDeployment,
+		IdentityRef:       credential.Spec.IdentityRef,
 		KubeconfSecret:    kubeconfSecret,
+		ClusterDeployment: clusterDeployment,
 		SystemNamespace:   r.SystemNamespace,
 	}
 
--- a/internal/credspropagation/common.go
+++ b/internal/credspropagation/common.go
@@ -30,8 +30,9 @@ import (
 
 type PropagationCfg struct {
 	Client            client.Client
-	ClusterDeployment *hmc.ClusterDeployment
 	KubeconfSecret    *corev1.Secret
+	IdentityRef       *corev1.ObjectReference
+	ClusterDeployment *hmc.ClusterDeployment
 	SystemNamespace   string
 }
 
@@ -53,11 +54,11 @@ func applyCCMConfigs(ctx context.Context, kubeconfSecret *corev1.Secret, objects
 	return nil
 }
 
-func makeSecret(name, namespace string, data map[string][]byte) *corev1.Secret {
+func makeSecret(name string, data map[string][]byte) *corev1.Secret {
 	s := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
-			Namespace: namespace,
+			Namespace: metav1.NamespaceSystem,
 		},
 		Data: data,
 	}
@@ -65,11 +66,11 @@ func makeSecret(name, namespace string, data map[string][]byte) *corev1.Secret {
 	return s
 }
 
-func makeConfigMap(name, namespace string, data map[string]string) *corev1.ConfigMap {
+func makeConfigMap(name string, data map[string]string) *corev1.ConfigMap {
 	c := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
-			Namespace: namespace,
+			Namespace: metav1.NamespaceSystem,
 		},
 		Data: data,
 	}
@@ -86,7 +87,11 @@ func makeClientFromSecret(kubeconfSecret *corev1.Secret) (client.Client, error) 
 	if err != nil {
 		return nil, err
 	}
-	return client.New(restConfig, client.Options{
+	cl, err := client.New(restConfig, client.Options{
 		Scheme: scheme,
 	})
+	if err != nil {
+		return nil, err
+	}
+	return cl, nil
 }
--- a/internal/credspropagation/common.go
+++ b/internal/credspropagation/common.go
@@ -87,11 +87,7 @@ func makeClientFromSecret(kubeconfSecret *corev1.Secret) (client.Client, error) 
 	if err != nil {
 		return nil, err
 	}
-	cl, err := client.New(restConfig, client.Options{
+	return client.New(restConfig, client.Options{
 		Scheme: scheme,
 	})
-	if err != nil {
-		return nil, err
-	}
-	return cl, nil
 }
--- a/internal/credspropagation/openstack.go
+++ b/internal/credspropagation/openstack.go
@@ -1,0 +1,76 @@
+// Copyright 2024
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package credspropagation
+
+import (
+	"context"
+	"fmt"
+
+	hmc "github.com/Mirantis/hmc/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func PropagateOpenStackSecrets(ctx context.Context, cfg *PropagationCfg) error {
+	openstackManagedCluster := &hmc.ManagedCluster{}
+	if err := cfg.Client.Get(ctx, client.ObjectKey{
+		Name:      cfg.ManagedCluster.Name,
+		Namespace: cfg.ManagedCluster.Namespace,
+	}, openstackManagedCluster); err != nil {
+		return fmt.Errorf("failed to get ManagedCluster %s: %w", cfg.ManagedCluster.Name, err)
+	}
+
+	openstackCredential := &hmc.Credential{}
+	if err := cfg.Client.Get(ctx, client.ObjectKey{
+		Name:      openstackManagedCluster.Spec.Credential,
+		Namespace: openstackManagedCluster.Namespace,
+	}, openstackCredential); err != nil {
+		return fmt.Errorf("failed to get OpenStackCredential %s: %w", cfg.ManagedCluster.Spec.Credential, err)
+	}
+
+	// Fetch the secret containing OpenStack credentials
+	openstackSecret := &corev1.Secret{}
+	openstackSecretName := openstackCredential.Spec.IdentityRef.Name
+	openstackSecretNamespace := openstackCredential.Spec.IdentityRef.Namespace
+	if err := cfg.Client.Get(ctx, client.ObjectKey{
+		Name:      openstackSecretName,
+		Namespace: openstackSecretNamespace,
+	}, openstackSecret); err != nil {
+		return fmt.Errorf("failed to get OpenStack secret %s: %w", openstackSecretName, err)
+	}
+
+	// Generate CCM secret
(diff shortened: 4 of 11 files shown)


internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Apply CCM config
+	if err := applyCCMConfigs(ctx, cfg.KubeconfSecret, ccmSecret); err != nil {
+		return fmt.Errorf("failed to apply OpenStack CCM secret: %w", err)
+	}
+
+	return nil
+}
+
+func generateOpenStackCCMSecret(openstackSecret *corev1.Secret) *corev1.Secret {
```
eromanova, line 66: You should create the secret with OpenStack cloud configuration following the specific format. It's not the same as the secret with credentials in the management cluster (the format: https://github.com/kubernetes/cloud-provider-openstack/blob/master/manifests/controller-manager/cloud-config).

Related thing: credentials should not be exposed, so instead of passing the credentials as chart values you should pass the name of this secret to the CCM and CSI values: https://github.com/Mirantis/hmc/pull/696/files#diff-c5afee6dd916ada2ea759da67744b5845149af9a6532b3891574865168c2c44e
See: https://github.com/kubernetes/cloud-provider-openstack/blob/master/charts/cinder-csi-plugin/values.yaml#L162

Similarly, it's done for the azure and the vsphere providers. You can use it as a reference.

eromanova (changes requested):

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,73 @@
...
+		return fmt.Errorf("failed to get ManagedCluster %s: %w", cfg.ManagedCluster.Name, err)
+	}
+
+	openstackCredential := &hmc.Credential{}
+	if err := cfg.Client.Get(ctx, client.ObjectKey{
+		Name:      openstackManagedCluster.Spec.Credential,
+		Namespace: openstackManagedCluster.Namespace,
+	}, openstackCredential); err != nil {
+		return fmt.Errorf("failed to get OpenStackCredential %s: %w", cfg.ManagedCluster.Spec.Credential, err)
+	}
```
a13x5, line 42: This call could be avoided. We already requested this object before in the managedcluster_controller.

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Fetch the secret containing OpenStack credentials
+	openstackSecret := &corev1.Secret{}
+	openstackSecretName := openstackCredential.Spec.IdentityRef.Name
+	openstackSecretNamespace := openstackCredential.Spec.IdentityRef.Namespace
```
a13x5, line 47: These aren't needed - you use this variables just once, consider passing values directly into struct below.

a13x5 (changes requested):
Several issues with overall logic.
Also I believe the scope of this PR could be broaden a little bit by adding credential passing to the cluster (`setIdentityHelmValues` function). Currently we're expecting ClusterIdentity object only, this isn't the case with OpenStack provider, so logic should be adjusted/tested. If you prepared another PR with these changes it should be merged first.

internal/credspropagation/common.go
```diff
@@ -53,11 +53,11 @@ func applyCCMConfigs(ctx context.Context, kubeconfSecret *corev1.Secret, objects
 	return nil
 }
 
-func makeSecret(name, namespace string, data map[string][]byte) *corev1.Secret {
+func makeSecret(name string, data map[string][]byte) *corev1.Secret {
```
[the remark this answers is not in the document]
bnallapeta: https://github.com/Mirantis/hmc/actions/runs/12067350380/job/33650115071

This is the reason for this change. I am not sure how it passed lint for other functions which I have updated too. Basically it would always get `metav1.NamespaceSystem` which would always evaluate to `kube-system`.

I think when we get to that point of changing the namespace, we can parameterize this.

internal/credspropagation/openstack.go
bnallapeta: Why? Its generally a good practice to use variables. What if its used just once?

internal/credspropagation/openstack.go
bnallapeta: Not that a single api call would impact api server anyways. Especially since we would have 1 Credential resource. But sure, have removed this and passed params from managedcluster_controller.go

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func PropagateOpenStackSecrets(ctx context.Context, cfg *PropagationCfg) error {
+	openstackManagedCluster := &hmc.ManagedCluster{}
+	if err := cfg.Client.Get(ctx, client.ObjectKey{
+		Name:      cfg.ManagedCluster.Name,
+		Namespace: cfg.ManagedCluster.Namespace,
+	}, openstackManagedCluster); err != nil {
+		return fmt.Errorf("failed to get ManagedCluster %s: %w", cfg.ManagedCluster.Name, err)
+	}
```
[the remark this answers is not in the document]
bnallapeta: Removed.

internal/credspropagation/openstack.go
bnallapeta: @eromanova Yes, it already follows the same format. And based on the other comments, I have removed passing values to ccm and csi and am passing just the secret. So, this should ideally work. 

@a13x5 Yes, I defined a function here with the intention of doing the INI format thing. But later realized I don't have to do that. I agree that we don't need this function. Modified.


I will anyways test this out and make sure everything works.

internal/credspropagation/common.go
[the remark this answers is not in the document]
a13x5: Well, shit...
Got it, then please adjust `makeConfigMap` as well, so both functions have the same signature. 
It would be good also to rename them to reflect that it will make secret in kube ns, but it's up to you really.

internal/credspropagation/openstack.go
a13x5: It's not just single API call here and it's not depends on the amount of `Credential` resource, but rather on amount of `ManagedCluster` resources. 
So it's more like one more call for each reconcile for each `ManagedCluster`.

internal/credspropagation/openstack.go
a13x5: Generally yes.
But in this particular example you already have variable `openstackCredential`. You copied part of it in other vars without any processing to pass them as a parameters and discard.
It would make sense to use variables here in case of:
- modifying data from `openstackCredential`
- using it in multiple places to improve readability (doubtful, but ok)
TL;DR here it just looks weird.

bnallapeta: :)

internal/credspropagation/common.go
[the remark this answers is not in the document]
bnallapeta: Have updated makeConfigMap as well. But haven't renamed it to leave it flexible for future. (if/when we add a namespace param)

bnallapeta:
@a13x5 @eromanova Would appreciate if you can take a look at the PR and merge it soon.

config/dev/openstack-credentials.yaml
```diff
@@ -0,0 +1,33 @@
...
+stringData:
+  clouds.yaml: |
+    clouds:
+      openstack:
+        auth:
+          auth_url: ${OS_AUTH_URL}
+          username: ${OS_USERNAME}
+          password: ${OS_PASSWORD}
+          project_id: ${OS_PROJECT_ID}
+          project_name: ${OS_PROJECT_NAME}
```
eromanova, line 16: AFAIR, the `project_name` is optional. Will this config be properly rendered if the `${OS_PROJECT_NAME}` is unset?

eromanova (commented):
Small comment from my side. The rest looks good.

a13x5 (approved):
Approved
Additional fixes will be made later

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,273 @@
...
+	ApplicationCredentialID     string `yaml:"application_credential_id"`
+	ApplicationCredentialName   string `yaml:"application_credential_name"`
+	ApplicationCredentialSecret string `yaml:"application_credential_secret"`
+	Username                    string `yaml:"username"`
+	Password                    string `yaml:"password"`
+	ProjectID                   string `yaml:"project_id"`
+	ProjectName                 string `yaml:"project_name"`
+	UserDomainName              string `yaml:"user_domain_name"`
+	ProjectDomainName           string `yaml:"project_domain_name"`
+}
```
zerospiel, line 72:
```suggestion
type (
	cloudConfFields struct {
		authURL                     string
		applicationCredentialID     string
		applicationCredentialName   string
		applicationCredentialSecret string
		username                    string
		password                    string
		projectID                   string
		projectName                 string
		userDomainName              string
		projectDomainName           string
		regionName                  string
		interfaceName               string
		floatingNetworkID           string
		publicNetworkName           string
	}

	cloudsYaml struct {
		Clouds map[string]cloud `yaml:"clouds"`
	}

	cloud struct {
		Auth       auth   `yaml:"auth"`
		RegionName string `yaml:"region_name"`
		Interface  string `yaml:"interface"`
	}

	auth struct {
		AuthURL                     string `yaml:"auth_url"`
		ApplicationCredentialID     string `yaml:"application_credential_id"`
		ApplicationCredentialName   string `yaml:"application_credential_name"`
		ApplicationCredentialSecret string `yaml:"application_credential_secret"`
		Username                    string `yaml:"username"`
		Password                    string `yaml:"password"`
		ProjectID                   string `yaml:"project_id"`
		ProjectName                 string `yaml:"project_name"`
		UserDomainName              string `yaml:"user_domain_name"`
		ProjectDomainName           string `yaml:"project_domain_name"`
	}
)
```

There is no need for exported types and respective for fields if applicable

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,273 @@
...
+		Namespace: cfg.ManagedCluster.Namespace,
+	}, openstackCluster)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return fmt.Errorf("OpenStackCluster %s/%s not found: %w",
+				cfg.ManagedCluster.Namespace, cfg.ManagedCluster.Name, err)
+		}
+		return fmt.Errorf("unable to get OpenStackCluster %s/%s: %w",
+			cfg.ManagedCluster.Namespace, cfg.ManagedCluster.Name, err)
+	}
```
zerospiel, line 89:
```suggestion
	if err := cfg.Client.Get(ctx, client.ObjectKey{
		Name:      cfg.ManagedCluster.Name,
		Namespace: cfg.ManagedCluster.Namespace,
	}, openstackCluster); err != nil {
		return fmt.Errorf("unable to get OpenStackCluster %s/%s: %w",
			cfg.ManagedCluster.Namespace, cfg.ManagedCluster.Name, err)
	}
```

I am not sure we will benefit from duplicating the not found error (`.. not found: ... not found`)

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Fetch the OpenStack secret
+	openstackSecret, err := fetchOpenStackSecret(ctx, cfg, credential)
+	if err != nil {
+		return fmt.Errorf("failed to fetch OpenStack secret: %w", err)
+	}
+
+	// Generate the CCM secret using the extracted cloudName
+	ccmSecret, err := generateOpenStackCCMSecret(ctx, cfg, openstackSecret, cloudName)
```
zerospiel, line 104:
```suggestion
	ccmSecret, err := generateOpenStackCCMSecret(openstackCluster, openstackSecret, cloudName)
```

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,273 @@
...
+		Namespace: credential.Spec.IdentityRef.Namespace,
+	}, openstackSecret); err != nil {
+		return nil, fmt.Errorf("failed to get OpenStack secret %s/%s: %w",
+			credential.Spec.IdentityRef.Namespace, credential.Spec.IdentityRef.Name, err)
+	}
+	return openstackSecret, nil
+}
+
+// Generate the CCM secret from the OpenStack secret
+func generateOpenStackCCMSecret(ctx context.Context, cfg *PropagationCfg, openstackSecret *corev1.Secret, cloudName string) (*corev1.Secret, error) {
```
zerospiel, line 135:
```suggestion
func generateOpenStackCCMSecret(osCluster *capo.OpenStackCluster, openstackSecret *corev1.Secret, cloudName string) (*corev1.Secret, error) {
```

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,273 @@
...
+	fields, err := extractCloudConfFields(parsedCloudsYaml, cloudName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract cloud.conf fields: %w", err)
+	}
+
+	// Fetch external network details from OpenStackCluster
+	externalNetwork, err := fetchExternalNetwork(ctx, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch external network details: %w", err)
+	}
```
zerospiel, line 187:
```suggestion
	externalNetwork := osCluster.Status.ExternalNetwork
```

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,273 @@
...
+	fields.FloatingNetworkID = externalNetwork.ID
+	fields.PublicNetworkName = externalNetwork.Name
+
+	// Render the cloud.conf secret
+	return renderCloudConf(cloudConfTemplate, fields)
+}
+
+// Parse the clouds.yaml content into structured types
+func parseCloudsYaml(data []byte) (*CloudsYaml, error) {
+	var cloudsYaml CloudsYaml
```
zerospiel, line 203:
```suggestion
	var cy cloudsYaml
```

Avoid shadowing (lint after the unexporting rename)

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("OpenStackCluster %s/%s not found",
+				cfg.ManagedCluster.Namespace, cfg.ManagedCluster.Name)
+		}
+		return nil, fmt.Errorf("unable to get OpenStackCluster %s/%s: %w",
+			cfg.ManagedCluster.Namespace, cfg.ManagedCluster.Name, err)
+	}
+	return openstackCluster.Status.ExternalNetwork, nil
+}
```
zerospiel, line 254: Could be removed, no need in fetching the same object within the same scope

internal/controller/managedcluster_controller.go
```diff
@@ -761,6 +761,25 @@ func (r *ManagedClusterReconciler) reconcileCredentialPropagation(ctx context.Co
 				Reason:  hmc.SucceededReason,
 				Message: "vSphere CCM credentials created",
 			})
+		case "openstack":
+			l.Info("OpenStack creds propagation start")
+			if err := credspropagation.PropagateOpenStackSecrets(ctx, propnCfg, credential); err != nil {
```
zerospiel, line 766: nit: the `".../credspropagation".PropagationCfg` symbol could be extended with the `credential`'s `Spec.IdentityRef` hence the signature of the `PropagateOpenStackSecrets` will be along with the rest of propagation functions. Up to the implementation and the code owning developer though

zerospiel (commented):
A couple of comments for network, reading, and memory simplifications

config/dev/openstack-credentials.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
zerospiel, line 1: Forgot to ask: do we need a doc on how to setup all of the envs here? I might have missed it in other PRs

internal/credspropagation/openstack.go
zerospiel: Note: the `cloudConfFields` type changes will also require the regarding changes to the template itself

internal/credspropagation/openstack.go
```diff
@@ -0,0 +1,273 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	capo "sigs.k8s.io/cluster-api-provider-openstack/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hmc "github.com/Mirantis/hmc/api/v1alpha1"
+)
+
+// Fields needed to populate the cloud.conf template
+type CloudConfFields struct {
```
eromanova, line 34: Can we use `AuthOptions` struct available in https://github.com/gophercloud/gophercloud/blob/v2.4.0/auth_options.go instead? Just a thought to consider.

internal/controller/managedcluster_controller.go
bnallapeta: Hmm, I am okay either ways. I guess we should do that (include identityRef into PropagationCfg struct) if it will be helpful for other providers in the future.
@eromanova thoughts? If you think its required, I can do that change in the same PR.

internal/credspropagation/openstack.go
bnallapeta: I did see this before. But it has only the auth fields. We would need to do a mix of importing this from gopher where maintaining our struct for the rest of the fields. That's why I chose to go with our custom struct so that its clean.

internal/credspropagation/openstack.go
bnallapeta: Fixed.

internal/credspropagation/openstack.go
bnallapeta: Same as above.

internal/credspropagation/openstack.go
bnallapeta: Up until the latest commit, I wasn't fetching OpenStackCluster resource. Since it was not there, I was using ctx and cfg. 

Have removed them now.

internal/credspropagation/openstack.go
bnallapeta: Removed.

internal/credspropagation/openstack.go
bnallapeta: Done.

bnallapeta:
Testing once again before I push the changes.

internal/credspropagation/openstack.go
zerospiel: i believe the API is stable, so supporting the own struct won't hurt in this case, considering introducing a new dependency for a mere struct, so I personally would agree with @bnallapeta on this and leave the struct

eromanova: I got it. Thank you!

internal/controller/managedcluster_controller.go
eromanova: I'd agree with Michael. Let's do it in this PR.

bnallapeta: Done.

bnallapeta:
Rebased the PR after the ManagedCluster renaming. Let's wait for the CI 🤞

internal/controller/clusterdeployment_controller.go
```diff
@@ -312,8 +312,9 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, mc *hmc
 		},
 		ChartRef: clusterTpl.Status.ChartRef,
 	}
-	if clusterTpl.Spec.Helm.ChartSpec != nil {
-		hrReconcileOpts.ReconcileInterval = &clusterTpl.Spec.Helm.ChartSpec.Interval.Duration
+	reconcileInterval := clusterTpl.Spec.Helm.ChartSpec.Interval.Duration
+	if reconcileInterval != 0 {
+		hrReconcileOpts.ReconcileInterval = &reconcileInterval
```
eromanova, line 317: Could you please remove this change? It seems to have been unintentionally introduced during the rebase process.

internal/controller/clusterdeployment_controller.go
```diff
@@ -354,11 +355,9 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, mc *hmc
...
 	}
 
-	if mc.Spec.PropagateCredentials {
-		if err := r.reconcileCredentialPropagation(ctx, mc); err != nil {
-			l.Error(err, "failed to reconcile credentials propagation")
-			return ctrl.Result{}, err
-		}
+	if err := r.reconcileCredentialPropagation(ctx, mc, cred); err != nil {
+		l.Error(err, "failed to reconcile credentials propagation")
+		return ctrl.Result{}, err
```
eromanova, line 360: Same

internal/credspropagation/common.go
```diff
@@ -86,7 +87,11 @@ func makeClientFromSecret(kubeconfSecret *corev1.Secret) (client.Client, error)
...
 		return nil, err
 	}
-	return client.New(restConfig, client.Options{
+	cl, err := client.New(restConfig, client.Options{
 		Scheme: scheme,
 	})
+	if err != nil {
+		return nil, err
+	}
+	return cl, nil
```
eromanova, line 96: Same

internal/credspropagation/openstack.go
eromanova: I'm not sure it's resolved. @bnallapeta could you please check?

bnallapeta: Fixed now.

eromanova (approved):

PR #699, opened 2024-11-28
Extend helm chart configuration for Templates

### LINKED ISSUE #615: Simplify BYO Templates
Currently there is a mismatch of templates shipped by 2A and templates that want to be added by a 2A operator.

2A templates only need to define a Template (let's say a `ClusterTemplate`) and in there the following config:
```
spec:
  helm:
    chartName: aws-standalone-cp
    chartVersion: 0.0.2
```

the required `HelmChart` is then automatically created through `reconcileHelmChart`
https://github.com/Mirantis/hmc/blob/fb83d23c3942bfa02727d0cc90f76295fe68dca6/internal/controller/template_controller.go#L299
with the assumption that the HelmRepo is the default repo.

If someone wants to bring their own templates they need to currently create 3 objects, as described in the documentation: https://mirantis.github.io/project-2a-docs/template/byo-templates/ :
- `ClusterTemplate` <-- obvious
- `HelmRepository` <-- also makes sense
- `HelmChart` <-- feels unnecessary

The ClusterTemplate then has the following spec:

```
spec:
  helm:
    chartRef:
      kind: HelmChart
      name: custom-template-chart
      namespace: default
```

And the biggest issue is the lack of the version of helmchart defined, which needs to happen inside the HelmChart object:

```
kind: HelmChart
spec:
  chart: custom-template-chart-name
  sourceRef:
    kind: HelmRepository
    name: custom-templates-repo
  version: 0.2.0
```

So this means if a 2A operator wants to upgrade their BYO Templates, they not only need to create a new `ClusterTemplate` Object, but also a new `HelmChart` object for every new version. 
This feels honestly too cumbersome and should be simplified.

A possible solution would be to allow ClusterTemplate with this config:

```
spec:
  helm:
    chartName: aws-standalone-cp
    chartVersion: 0.0.2
    chartRepo: custom-templates-repo # new field
```

And 2A will then automatically reconcile the `HelmChart` object for it (so exactly the same as we also do for the 2A templates)

eromanova:
This PR introduces the following changes in the `ClusterTemplate`, `ServiceTemplate`, and `ProviderTemplate` spec:

- New `spec.helm.chartSpec` section of the [HelmChartSpec](https://github.com/fluxcd/source-controller/blob/v1.4.1/api/v1beta2/helmchart_types.go#L34) type to support providing extended helm chart configuration. An example of the simplest allowed configuration:
```
spec:
  helm:
    chartSpec:
      chart: chart-name
```
An example of the extended configuration with the source 
```
spec:
  helm:
    chartSpec:
      chart: chart-name
      version: 1.0.0
      interval: 15m0s
      sourceRef:
        kind: HelmRepository
        name: helm-repo
```
The `version` parameter is optional (the latest chart version will be used if unset). The `sourceRef` and `interval` parameters from the `spec.helm.chart` will be automatically populated with defaults if unset:
```
interval: 10m0s
sourceRef:
  kind: HelmRepository
  name: hmc-templates
```

where `hmc-templates` is the name of the default `HelmRepository` created by the hmc controller.

- The `spec.helm.chartName` and `spec.helm.chartVersion` parameters were removed as unnecessary.

Closes #615


internal/controller/management_controller.go
```diff
@@ -142,13 +142,19 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
 			ChartRef:        template.Status.ChartRef,
 			DependsOn:       component.dependsOn,
 			TargetNamespace: component.targetNamespace,
 			CreateNamespace: component.createNamespace,
-		}); err != nil {
+		}
+		reconcileInterval := template.Spec.Helm.Chart.Interval.Duration
+		if reconcileInterval != 0 {
+			hrReconcileOpts.ReconcileInterval = &reconcileInterval
+		}
```
zerospiel, line 155: don't we want to set the default reconciliation interval here? as an analog to the `release` ctrl

internal/controller/managedcluster_controller.go
```diff
@@ -313,7 +313,13 @@ func (r *ManagedClusterReconciler) updateCluster(ctx context.Context, mc *hmc.Ma
 			UID:        mc.UID,
 		},
 		ChartRef: clusterTpl.Status.ChartRef,
-	})
+	}
+	reconcileInterval := clusterTpl.Spec.Helm.Chart.Interval.Duration
+	if reconcileInterval != 0 {
+		hrReconcileOpts.ReconcileInterval = &reconcileInterval
+	}
```
zerospiel, line 320: the same question about the default interval

api/v1alpha1/templates_common.go
```diff
@@ -30,27 +31,33 @@ const (
...
+	Kind: sourcev1.HelmRepositoryKind,
+	Name: DefaultRepoName,
+}
+
+// +kubebuilder:validation:XValidation:rule="(has(self.chart) && !has(self.chartRef)) || (!has(self.chart) && has(self.chartRef))", message="either chart or chartRef must be set"
 
 // HelmSpec references a Helm chart representing the HMC template
 type HelmSpec struct {
+	// Chart defines the desired state of the HelmChart to be created by the controller
+	Chart *sourcev1.HelmChartSpec `json:"chart,omitempty"`
```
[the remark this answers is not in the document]
eromanova: Yes. It'll be much better. I think we should do it. Thanks.

internal/controller/management_controller.go
eromanova: The default reconciliation interval is set by the admission mutation webhook (see changes in [internal/webhook/template_webhook.go](https://github.com/Mirantis/hmc/pull/699/files#diff-d026968b705a734fd88db7330857edc6977860ce0f9e044a203d4fc220d38c2e)). This field is required in the `HelmChartSpec` struct, it will always be populated, so I don't think we should set the defaults in the controller as well.

api/v1alpha1/templates_common.go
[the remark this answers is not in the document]
eromanova: It seems I spoke too soon. Embedding the `HelmChartSpec` struct causes issues with the current Kubebuilder validation for ensuring mutual exclusivity between the `ChartRef` and `HelmChartSpec` subfields. The `HelmChartSpec` struct includes required fields (`chart`, `sourceRef`, etc) that won’t be populated when `ChartRef` is set, leading to validation errors. To address this, I’ve renamed the `Chart` field to `ChartSpec` to avoid redundancy and prevent a `helm.chart.chart` naming issue.

zerospiel (approved):

internal/webhook/template_webhook.go
```diff
@@ -190,6 +201,28 @@ func (*ProviderTemplateValidator) ValidateDelete(_ context.Context, _ runtime.Ob
...
+	}
+	setHelmChartDefaults(template)
 	return nil
 }
+
+type templateCommon interface {
+	GetHelmSpec() *v1alpha1.HelmSpec
+}
+
+func setHelmChartDefaults(template templateCommon) {
```
zerospiel, line 217:
```suggestion
func setHelmChartDefaults(helmSpec *v1alpha1.HelmSpec) {
	if helmSpec == nil {
		return
	}
```

internal/webhook/template_webhook.go
```diff
@@ -190,6 +201,28 @@ func (*ProviderTemplateValidator) ValidateDelete(_ context.Context, _ runtime.Ob
...
+	helmSpec := template.GetHelmSpec()
```
zerospiel, line 218:
```suggestion
func setHelmChartDefaults(helmSpec *v1alpha1.HelmSpec) {
	if helmSpec == nil {
		return
	}
```

zerospiel (approved):

internal/controller/release_controller.go
```diff
@@ -249,7 +249,7 @@ func (r *ReleaseReconciler) reconcileHMCTemplates(ctx context.Context, releaseNa
 	if releaseName == "" {
 		releaseName = utils.ReleaseNameFromVersion(build.Version)
 		releaseVersion = build.Version
-		err := helm.ReconcileHelmRepository(ctx, r.Client, defaultRepoName, r.SystemNamespace, r.DefaultRegistryConfig.HelmRepositorySpec())
+		err := helm.ReconcileHelmRepository(ctx, r.Client, hmc.DefaultRepoName, r.SystemNamespace, r.DefaultRegistryConfig.HelmRepositorySpec())
```
Kshatrix, line 252: wdyt about keeping only helmrepository reference as a controller option and adding a default HelmRepository to the templates chart instead?

Kshatrix: will be discussed separately

Kshatrix (approved):

PR #701, opened 2024-12-02
Validate Release object removal

Kshatrix:
Resolves: #690

### DIFF
--- a/internal/webhook/release_webhook.go
+++ b/internal/webhook/release_webhook.go
@@ -1,0 +1,57 @@
+// Copyright 2024
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package webhook
+
+import (
+	"context"
+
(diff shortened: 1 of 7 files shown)


internal/webhook/release_webhook.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	mgmt := mgmtList.Items[0]
+	if mgmt.Spec.Release == release.Name {
+		return nil, fmt.Errorf("release %s is still in use", release.Name)
+	}
+
+	templates := release.Templates()
+	templatesInUse := []string{}
+	for _, p := range mgmt.Spec.Providers {
```
eromanova, line 79: Shouldn't we check core templates as well?

internal/webhook/release_webhook.go
```diff
@@ -0,0 +1,88 @@
...
+	if !ok {
+		return admission.Warnings{"Wrong object"}, apierrors.NewBadRequest(fmt.Sprintf("expected Release but got a %T", obj))
+	}
+	mgmtList := &hmcv1alpha1.ManagementList{}
+	if err := v.List(ctx, mgmtList); err != nil {
+		return nil, err
+	}
+	if len(mgmtList.Items) != 1 {
+		return nil, fmt.Errorf("expected 1 Management object, got %d", len(mgmtList.Items))
+	}
```
eromanova, line 70: Maybe it's worth allowing the Release deletion if there is no `Management` object yet? Ex. the HMC was installed but the Management object was not yet created, the new `Release` was published and the user wants to create a new release and remove an old one and then, create the `Management` object. What do you think?

Kshatrix: It wouldn't harm I guess, but realistically admission is not enabled until the `Management` object is reconciled, so should not really happen

eromanova (approved):

PR #712, opened 2024-12-04
Support credentials distribution

### LINKED ISSUE #692: Credentials Distribution System
Currently the `TemplateManagement` Object can distribute ClusterTemplates and ServiceTemplates to namespaces in order for them to be used in a `managedCluster` object in these namespaces.

See: https://github.com/Mirantis/hmc/blob/e38f5822cbefb05bfe00e200e90f68ca47f2592a/api/v1alpha1/templatemanagement_types.go#L42-L52

We would also need to be able to distribute credentials the same way. An example for clarification:

```
kind: TemplateManagement
metadata:
  name: hmc
spec:
  accessRules:
    - credential:
        - aws-cluster-identity-cred
      targetNamespaces:
        list:
          - default
```

would copy the credential `aws-cluster-identity-cred` to the namespace `default`.

eromanova:
This PR includes the following changes:
1. Rename `TemplateManagement` -> `AccessManagement`
2. `AccessManagement` spec is extended to receive the list of credential names to be distributed to target namespaces
3. Access management controller now supports credentials distribution

Closes #692


internal/controller/accessmanagement_controller.go
```diff
@@ -105,23 +110,38 @@ func (r *TemplateManagementReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+				keepCredentials[getNamespacedName(namespace, credentialName)] = true
+				if systemCredentials[credentialName] == nil {
+					errs = errors.Join(errs, fmt.Errorf("credential %s/%s is not found", r.SystemNamespace, credentialName))
+					continue
+				}
+				err = r.createCredential(ctx, namespace, credentialName, systemCredentials[credentialName])
+				if err != nil {
+					errs = errors.Join(errs, err)
+					continue
+				}
```
zerospiel, line 123: nit: could be simplified to oneliner (two with the `\n`) (applicable for the loops above)

```go

errs = errors.Join(errs, r.createCredential(ctx, namespace, credentialName, systemCredentials[credentialName]))
```

internal/controller/accessmanagement_controller.go
```diff
@@ -252,30 +295,54 @@ func (r *TemplateManagementReconciler) createTemplateChain(ctx context.Context,
...
+		},
+		Spec: *spec,
+	}
+	err := r.Create(ctx, target)
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return nil
+		}
+		return err
+	}
```
zerospiel, line 317: nit: could be simplified to (and applicable above/below depending on the `Ignore*` function), up to you whether to apply/skip

```go

	if err := r.Create(ctx, target); client.IgnoreAlreadyExists(err) != nil {
		return err
	}
```

zerospiel (approved):
just a couple optional nit comments, in general LGTM, approved

Kshatrix (approved):

PR #804, opened 2024-12-18
Fix azure.json generation: include node-subnet data

### LINKED ISSUE #730: Azure cloud provider only creates rules in the control plane NSG
I have an azure cluster created with the following manifest
```
kind: ManagedCluster
metadata:
  name: azure-reg1
  namespace: hmc-system
  labels:
    motel: "regional"
spec:
  template: azure-standalone-cp-0-0-3
  credential: azure-cluster-identity-cred
  config:
    controlPlaneNumber: 1
    workersNumber: 3
    location: "westus2"
    subscriptionID: "${AZURE_SUBSCRIPTION_ID}"
    controlPlane:
      vmSize: Standard_A4_v2
    worker:
      vmSize: Standard_A4_v2
    tenantID: "${AZURE_TENANT_ID}"
    clientID: "${AZURE_CLIENT_ID}"
    clientSecret: "${AZURE_CLIENT_SECRET}"
  servicesPriority: 100
  services:
    - template: ingress-nginx-4-11-3
      name: ingress-nginx
      namespace: ingress-nginx
    - name: cert-manager
      namespace: cert-manager
      template: cert-manager-1-16-1
      values: |
        crds:
          enabled: true
    - name: motel-regional
      namespace: motel
      template: motel-regional-0-1-1
      values: |
        victoriametrics:
          vmauth:
            ingress:
              host: reg-1.example.com
            credentials:
              username: motel
              password: motel
        grafana:
          ingress:
            enabled: true
            host: grafana.reg-1.example.com
        cert-manager:
          email: email@example.com

```

In result a nginx-ingress LB was created
```
ingress-nginx   ingress-nginx-controller                          LoadBalancer   10.103.165.119   4.155.90.207   80:30999/TCP,443:31409/TCP   75m
```

But the ingress-nginx-controller cannot receive any requests *probably* due to misconfiguration of azure-reg1-node-nsg NSG

a13x5:
Fixes #730

### DIFF
--- a/internal/credspropagation/azure.go
+++ b/internal/credspropagation/azure.go
@@ -63,6 +63,7 @@ func PropagateAzureSecrets(ctx context.Context, cfg *PropagationCfg) error {
 }
 
 func generateAzureCCMSecret(azureCluster *capz.AzureCluster, azureClIdty *capz.AzureClusterIdentity, azureSecret *corev1.Secret) (*corev1.Secret, error) {
+	subnetName, secGroup, routeTable := getAzureSubnetData(azureCluster)
 	azureJSONMap := map[string]any{
 		"cloud":                        azureCluster.Spec.AzureEnvironment,
 		"tenantId":                     azureClIdty.Spec.TenantID,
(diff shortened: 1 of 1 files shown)


a13x5:
Tested with LB service with `http-echo`

```
> kubectl get svc
NAME           TYPE           CLUSTER-IP      EXTERNAL-IP    PORT(S)        AGE
http-echo-lb   LoadBalancer   10.103.26.121   13.64.251.61   80:32290/TCP   116s
kubernetes     ClusterIP      10.96.0.1       <none>         443/TCP        28m
> curl http://13.64.251.61
"heyyyy"
```

internal/credspropagation/azure.go
```diff
@@ -94,3 +96,15 @@ func generateAzureCCMSecret(azureCluster *capz.AzureCluster, azureClIdty *capz.A
 
 	return makeSecret("azure-cloud-provider", metav1.NamespaceSystem, secretData), nil
 }
+
+func getAzureSubnetData(azureCluster *capz.AzureCluster) (subnetName, secGroup, routeTable string) {
+	subnetName = azureCluster.Name + "-node-subnet"
```
eromanova, line 101: This logic will apply to both hosted and standalone Azure clusters. But for the hosted case, the subnet name is different. Would it be better to retrieve the subnet name directly from the AzureCluster spec by identifying the subnet with the `node` role?
Or, the bug does not affect hosted clusters and we may completely ignore the hosted case?

a13x5: Yep you're rignt. We're passing different subnet explicitly to hosted...
That was a really dumb mistake from my side, thanks for noticing!

eromanova (approved):

PR #806, opened 2024-12-18
Add Azure AKS Cluster Template

### LINKED ISSUE #744: create AKS template

eromanova:
1. A new template for `azure-aks` clusters has been added.
2. The ClusterDeployment spec now supports the option to skip propagating credentials by setting `propagateCredentials: false`. This is particularly needed for AKS clusters, as they do not require credential propagation.
3. Added YAMLs with AKS credentials and managed cluster.

There are two API options for creating AKS clusters: `ManagedCluster` and `ASOManagedCluster`. The `ASOManagedCluster` approach was chosen because the `ManagedCluster` will eventually be deprecated. See [CAPZ ManagedCluster vs. ASOManagedCluster](https://capz.sigs.k8s.io/managed/managed#capzs-managed-cluster-versus-capzs-asomanaged-cluster).

### ASO Credentials
ASO (azure service operator) requires Azure credentials to be defined in a Kubernetes Secret and does not support reading creds from `AzureClusterIdentity` (source: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/docs/proposals/20230123-azure-service-operator.md#security-model).

Closes #744

### DIFF
--- a/templates/cluster/azure-aks/values.yaml
+++ b/templates/cluster/azure-aks/values.yaml
@@ -1,0 +1,106 @@
+# Cluster parameters
+clusterNetwork:
+  pods:
+    cidrBlocks:
+    - "10.244.0.0/16"
+  services:
+    cidrBlocks:
+    - "10.96.0.0/12"
+
+clusterIdentity:
+  name: ""
+
+# AKS cluster parameters
+apiServerAccessProfile:
+  authorizedIPRanges: []
+  disableRunCommand: false
+  enablePrivateCluster: false
+  enablePrivateClusterPublicFQDN: false
+  privateDNSZone: system
+
+autoUpgradeProfile:
+  nodeOSUpgradeChannel: "None"
+  upgradeChannel: "none"
+
+azureMonitorProfile:
+  metrics:
+    enabled: false
+    kubeStateMetrics:
+      metricAnnotationsAllowList: ""
+      metricLabelsAllowlist: ""
+
+dnsServiceIP: 10.96.0.10
+
+location: ""
+
+oidcIssuerProfile:
+  enabled: false
+
+securityProfile:
+  azureKeyVaultKms:
+    enabled: false
+    keyId: ""
+    keyVaultNetworkAccess: "Public"
+    keyVaultResourceReference: {}
+  defender:
+    logAnalyticsWorkspaceResourceReference: {}
+    securityMonitoring:
+      enabled: false
+  imageCleaner:
+    enabled: false
+    intervalHours: 1
+  workloadIdentity:
+    enabled: false
+
+serviceMeshProfile:
+  mode: Disabled
+  istio:
+    certificateAuthority:
+      certChainObjectName: ""
+      certObjectName: ""
+      keyObjectName: ""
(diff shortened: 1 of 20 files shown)


api/v1alpha1/managedcluster_types.go
```diff
@@ -84,6 +84,10 @@ type ManagedClusterSpec struct {
 	// By default the remaining services will be deployed even if conflict is detected.
 	// If set to true, the deployment will stop after encountering the first conflict.
 	StopOnConflict bool `json:"stopOnConflict,omitempty"`
+	// PropagateCredentials indicates whether credentials should be propagated
+	// for use by CCM (Cloud Controller Manager).
+	// Default: true.
```
zerospiel, line 89: i cannot see the kubebuilder default directive, nor where it is being defaulted via the mutation webhook, could you show me where it is being defaulted to such a value? and why should the field be nullable in the case when `true` is the default value?

eromanova: Initially I wanted to differentiate when the parameter is unset and consider `unset==true` in `PropagateCredentials() below`. This is because I wanted to keep this changes compatible with already existing clusters. But it's really not needed since the kubebuilder default directive does what I need (I thought it does not update already existing objects with the default parameter but seems it does). Thanks for pointing. Reworked.

api/v1alpha1/managedcluster_types.go
```diff
@@ -61,6 +61,10 @@ type ManagedClusterSpec struct {
 	Template string `json:"template"`
 	// Name reference to the related Credentials object.
 	Credential string `json:"credential,omitempty"`
+	// PropagateCredentials indicates whether credentials should be propagated
+	// for use by CCM (Cloud Controller Manager).
+	// +kubebuilder:default:=true
```
zerospiel, line 66:
```suggestion
	// +kubebuilder:default:=true

	// PropagateCredentials indicates whether credentials should be propagated
	// for use by CCM (Cloud Controller Manager).
```

Sorry for the very nit comment, it is up to you whether to change or not :)

zerospiel (dismissed):

zerospiel (dismissed):

templates/cluster/azure-aks/values.yaml
```diff
@@ -0,0 +1,106 @@
...
+    nodeLabels: {}
+    nodeTaints: []
+    osDiskSizeGB: 128
+    type: VirtualMachineScaleSets
+    vmSize: ""
+  user:
+    autoscaling:
+      enabled: false
+      minCount: 2
+      maxCount: 100
```
a13x5, line 92: NIT: The default value is a bit large IMO, the same `5` as in `system` probably a better default.

a13x5 (approved):
LGTM

templates/cluster/azure-aks/values.yaml
eromanova: I wanted to be close to the defaults that Azure suggests while creating the AKS cluster using their platform but I'm ok to decrease it.

Kshatrix (approved):

PR #809, opened 2024-12-18
Provide ability to add labels to `Cluster` object

wahabmk:
# Description

Provide ability to add labels to the `Cluster` object created when installing a clusterTemplate by:

1. Adding `.Values.ClusterLabels` to all clusterTemplate helm charts.
2. Using `ManagedCluster`'s own labels via the HMC controller if none have been defined in `.Values.ClusterLabels`.

# Testing
As can be seen from the output below, the `.metadata.labels` from `ManagedCluster` have been added to the `Cluster` object:
```sh
➜  ~ kubectl -n hmc-system get managedclusters.hmc.mirantis.com wali-dev-1 --show-labels
NAME         READY   STATUS                    LABELS
wali-dev-1   True    ManagedCluster is ready   labelA=A,labelB=B
➜  ~
➜  ~
➜  ~ kubectl -n hmc-system get cluster --show-labels                                
NAME         CLUSTERCLASS   PHASE         AGE   VERSION   LABELS
wali-dev-1                  Provisioned   13m             app.kubernetes.io/managed-by=Helm,helm.toolkit.fluxcd.io/name=wali-dev-1,helm.toolkit.fluxcd.io/namespace=hmc-system,labelA=A,labelB=B,sveltos-agent=present
```

### DIFF
--- a/templates/cluster/vsphere-standalone-cp/templates/cluster.yaml
+++ b/templates/cluster/vsphere-standalone-cp/templates/cluster.yaml
@@ -2,6 +2,9 @@ apiVersion: cluster.x-k8s.io/v1beta1
 kind: Cluster
 metadata:
   name: {{ include "cluster.name" . }}
+  {{- if .Values.clusterLabels }}
+  labels: {{- toYaml .Values.clusterLabels | nindent 4}}
+  {{- end }}
 spec:
   {{- with .Values.clusterNetwork }}
   clusterNetwork:
--- a/templates/cluster/vsphere-standalone-cp/values.yaml
+++ b/templates/cluster/vsphere-standalone-cp/values.yaml
@@ -10,6 +10,8 @@ clusterNetwork:
     cidrBlocks:
     - "10.96.0.0/12"
 
+clusterLabels: {}
+
 # vSphere cluster parameters
 clusterIdentity:
   name: ""
--- a/api/v1alpha1/clusterdeployment_types.go
+++ b/api/v1alpha1/clusterdeployment_types.go
@@ -15,6 +15,9 @@
 package v1alpha1
 
 import (
+	"encoding/json"
+	"fmt"
+
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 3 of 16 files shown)


templates/cluster/vsphere-standalone-cp/templates/cluster.yaml
```diff
@@ -2,6 +2,10 @@ apiVersion: cluster.x-k8s.io/v1beta1
 kind: Cluster
 metadata:
   name: {{ include "cluster.name" . }}
+  labels:
+    {{- with .Values.clusterLabels }}
+    {{- toYaml . | nindent 4 }}
```
zerospiel, line 7: nit: i think it'd look a little bit more beautiful with just `indent` instead of the `nindent` (but please verify it). applicable for the rest of the templates' clusters 

I also think this could be simplified to something like (but should be verified as well)
```yaml
  {{- if .Values.clusterLabels }}
  labels: {{- toYaml .Values.clusterLabels | nindent 4 }} # here nindent on purpose
  {{- end }}
```

templates/cluster/vsphere-standalone-cp/values.yaml
```diff
@@ -10,6 +10,8 @@ clusterNetwork:
     cidrBlocks:
     - "10.96.0.0/12"
 
+clusterLabels:
```
zerospiel, line 13: i am not sure but isn't that better (more explicit) to set the `{}` instead of nothing to set an explicit type of the object? the question is applicable for the rest of the charts

zerospiel (commented):
lgtm, 2 charts-related-mostly-nit comments

zerospiel (approved):

internal/controller/managedcluster_controller.go
```diff
@@ -297,13 +295,21 @@ func (r *ManagedClusterReconciler) updateCluster(ctx context.Context, mc *hmc.Ma
...
 
-	helmValues, err := setIdentityHelmValues(mc.Spec.Config, cred.Spec.IdentityRef)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("error setting identity values: %w", err)
+	if err := mc.AddHelmValues(func(values map[string]any) error {
+		values["clusterIdentity"] = cred.Spec.IdentityRef
+
+		if _, ok := values["clusterLabels"]; !ok {
+			// Use the ManagedCluster's own labels if not defined.
+			values["clusterLabels"] = mc.GetObjectMeta().GetLabels()
```
Kshatrix, line 303: It doesn't seem reliable since there's no guarantee that all cluster templates have this field (unless you want to make this field required).

internal/controller/managedcluster_controller.go
```diff
@@ -297,13 +295,21 @@ func (r *ManagedClusterReconciler) updateCluster(ctx context.Context, mc *hmc.Ma
...
 	}
 
-	helmValues, err := setIdentityHelmValues(mc.Spec.Config, cred.Spec.IdentityRef)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("error setting identity values: %w", err)
+	if err := mc.AddHelmValues(func(values map[string]any) error {
+		values["clusterIdentity"] = cred.Spec.IdentityRef
+
+		if _, ok := values["clusterLabels"]; !ok {
+			// Use the ManagedCluster's own labels if not defined.
```
Kshatrix, line 302: My understanding was that `clusterSelector` in `MultiClusterService` is supposed to refer hmc `ManagedCluster` instead of capi `Cluster` objects. In that case propagation should not be optional. Is my assumption valid?

wahabmk: Right now the `ClusterSelector metav1.LabelSelector` field is fed directly into the the [`.spec.clusterSelector`](https://github.com/projectsveltos/addon-controller/blob/ebff58a741cbc9ca1e78383fd7e8c7d9cfda82c6/api/v1beta1/spec.go#L593) field for the Sveltos ClusterProfile object. So the label selectors are matching `Cluster` objects.

internal/controller/managedcluster_controller.go
wahabmk: Yes, in this approach we are depending on the helm chart for cluster templates to use the `clusterLabels` value. We do need some way to assign labels to `Cluster` objects. Maybe the approach would be?
1. In the HMC controller, use the values defined in `clusterLabels` in the ManagedCluster spec and patch `Cluster` to apply those labels to it.
2. Or if `clusterLabels` is not defined, then use the ManagedCluster's own labels.

zerospiel (dismissed):
LGTM, the PR needs to be rebased and @Kshatrix should also check whether it's okay to assign labels to the CAPI `clusters`

gmlexx (dismissed):

mestadler:
Good to me!

wahabmk:
> LGTM, @Kshatrix could you check whether you've reconciled on the approach and if the initial is okay, then I'm ready to approve and merge the PR

In our last meeting @Kshatrix mentioned he didn't block the PR by requesting any changes on the PR so we can merge it

zerospiel (approved):

PR #810, opened 2024-12-18
Fix AWS hosted cluster template networking issue (fixes #788)

### LINKED ISSUE #788: AWS subnet tagging causes hosted clusters to not allow load balancer to be exposed
Currently the AWS provider will not allow a hosted cluster to create a load balancer that gets assigned an ip or host name. A possible solution is to change the provider template to define the KubernetesClusterTag or KubernetesClusterID variables in the AWS cloud config.

kylewuolle:
This PR adds the ability to set the management cluster name in the AWS hosted template. This is needed because of issue #788. Requires PR https://github.com/Mirantis/helm-charts/pull/2 to be merged and the resulting helm chart to be published. An upstream PR for the change to the chart has been created : https://github.com/kubernetes/cloud-provider-aws/pull/1085.

### DIFF
--- a/templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
+++ b/templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
@@ -58,7 +58,7 @@ spec:
               cloudConfig:
                 enabled: true
                 global:
-                  KubernetesClusterID: {{ .Values.managementClusterName }}
+                  KubernetesClusterID: {{ required ".Values.managementClusterName is required on AWS hosted deployment" .Values.managementClusterName }}
               # Removing the default `node-role.kubernetes.io/control-plane` node selector
               # TODO: it does not work
               nodeSelector:
--- a/templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
+++ b/templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
@@ -26,11 +26,11 @@ spec:
       extensions:
         helm:
           repositories:
-          - name: aws-cloud-controller-manager
+          - name: mirantis
             {{- if .Values.extensions.chartRepository }}
             url: {{ .Values.extensions.chartRepository }}
             {{- else }}
-            url: https://kubernetes.github.io/cloud-provider-aws
+            url: https://charts.mirantis.com
             {{- end }}
           - name: aws-ebs-csi-driver
             {{- if .Values.extensions.chartRepository }}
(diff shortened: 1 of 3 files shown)


templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
```diff
@@ -55,6 +55,10 @@ spec:
                 - --cluster-cidr={{ first .Values.clusterNetwork.pods.cidrBlocks }}
                 - --allocate-node-cidrs=true
                 - --cluster-name={{ include "cluster.name" . }}
+              cloudConfig:
+                enabled: {{ ne .Values.managementClusterName "" }}
+                global:
+                  KubernetesClusterID: {{ .Values.managementClusterName }}
```
a13x5, line 61: And then we can just have `required` if you want to have some guards on that value

a13x5 (commented):
Minor notes.
@kylewuolle did you research a possibility to pass `managementClusterName` automatically? 
For example we could do in the same way [we're passing credential values ](https://github.com/K0rdent/kcm/blob/main/internal/controller/clusterdeployment_controller.go#L796)

kylewuolle:
> Minor notes.
> @kylewuolle did you research a possibility to pass `managementClusterName` automatically? 
> For example we could do in the same way [we're passing credential values ](https://github.com/K0rdent/kcm/blob/main/internal/controller/clusterdeployment_controller.go#L796)
> 

The problem is I don't think we have that information during the hosted template deployment.

templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
```diff
@@ -55,6 +55,10 @@ spec:
                 - --cluster-cidr={{ first .Values.clusterNetwork.pods.cidrBlocks }}
                 - --allocate-node-cidrs=true
                 - --cluster-name={{ include "cluster.name" . }}
+              cloudConfig:
+                enabled: {{ ne .Values.managementClusterName "" }}
```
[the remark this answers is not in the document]
kylewuolle: good point

templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
kylewuolle: done!

a13x5: No I mean required in terms of [helm chart itself.](https://helm.sh/docs/chart_template_guide/function_list/#required) 

I'm not sure that schema validation alone will be enough.

a13x5 (approved):
LGTM

Issue #605, opened 2024-11-05
implement velero installation reconciliation


zerospiel:
depends on the #603

PR #815, opened 2024-12-19
Backup implementation: velero installation

zerospiel:
* new component labels on most components (required for backups' label selectors)
* wide range of new RBAC permission for the backup ctrl
* velero stack reconciliation
* adjusted backup user-facing roles
* added backup ctrl configuration in the hmc chart
* removed schema defaults in favor of mutation

*Note*: the controller is still turned off to avoid any potential regression.

#605


eromanova (commented):
A general question about the `hmc.mirantis.com/component: hmc` label and credentials backup. As I understood only objects with this label will be backed up. But I'm concerned about the credentials' identity references. Our controller does not modify the identity refs (Secrets, AzureClusterIdentity, etc) when it was provided by the user. Maybe the credentials controller should be adjusted to label those objects too?
---
Also, please add this label on newly added `config/dev/adoptedcluster-*.yaml` objects.

eromanova (commented):
Same question about the `HelmCharts` and `HelmRepositories`.
I haven't found any logic about labeling the generated HelmCharts (sorry if it's present and I've missed it).
When the user creates the BYO Template it can reference the helm chart from the custom `HelmRepository` (manually created). Should it also be labeled?
Please ignore my concerns if it's something that you're aware but it's not in the scope of this PR.

zerospiel:
> A general question about the `hmc.mirantis.com/component: hmc` label and credentials backup. As I understood only objects with this label will be backed up. But I'm concerned about the credentials' identity references. Our controller does not modify the identity refs (Secrets, AzureClusterIdentity, etc) when it was provided by the user. Maybe the credentials controller should be adjusted to label those objects too?
> 
> Also, please add this label on newly added `config/dev/adoptedcluster-*.yaml` objects.

Not quite, there will be many more label selectors, and those sentinel labels are just for objects that do not have any labels at the moment but should also be presented within a backup. One of the selectors also covers the `HelmRepositories` but not the BYO, the general recommendation (as I can see now) is `put a special label to all your BYO objects` but it'd be amended if required.

In particular, the question about the `credentials`' references: I had no intention to change permissions, and such objects referenced in a `credentials` object might have no labels, so I either will address it via the velero API or will add the sentinel label changing permissions along (the former is preferable). The problem is to be addressed in the next PR along with numerous enhancements to the installation and the controller itself.

Kshatrix (approved):

Issue #822, opened 2024-12-20
ManagedCluster controller test enhancement


BROngineer:
**Description**

Current test suite runs reconciliation and checks whether error occurred or not:

https://github.com/Mirantis/hmc/blob/2e92fb6880d955204e4e923d0ab92437e2719974/internal/controller/managedcluster_controller_test.go#L215-L226

However, reconciler exits after adding finalizer to the resource being reconciled:

```go
// managedcluster_controller.go

...
func (r *ManagedClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
    ...
    return r.reconcileUpdate(ctx, managedCluster)
}

func (r *ManagedClusterReconciler) reconcileUpdate(ctx context.Context, mc *hmc.ManagedCluster) (_ ctrl.Result, err error) {
...

    if controllerutil.AddFinalizer(mc, hmc.ManagedClusterFinalizer) {
        if err := r.Client.Update(ctx, mc); err != nil {
            return ctrl.Result{}, fmt.Errorf("failed to update managedCluster %s/%s: %w", mc.Namespace, mc.Name, err)
        }
        return ctrl.Result{}, nil
    }
...
```

**Expected behaviour**

Object should be reconciled until it eventually enters desired state defined by test-case parameters or test should fail by timeout.

PR #830, opened 2024-12-23
managedcluster-controller tests

BROngineer:
Fixes: #822

### DIFF
--- a/internal/controller/clusterdeployment_controller_test.go
+++ b/internal/controller/clusterdeployment_controller_test.go
@@ -15,12 +15,15 @@
 package controller
 
 import (
+	"context"
 	"time"
 
 	hcv2 "github.com/fluxcd/helm-controller/api/v2"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -32,11 +35,31 @@ import (
 	hmc "github.com/K0rdent/kcm/api/v1alpha1"
 )
 
+type fakeHelmActor struct{}
+
+func (f *fakeHelmActor) DownloadChartFromArtifact(_ context.Context, _ *sourcev1.Artifact) (*chart.Chart, error) {
+	return &chart.Chart{
+		Metadata: &chart.Metadata{
+			APIVersion: "v2",
+			Version:    "0.1.0",
+			Name:       "test-cluster-chart",
+		},
+	}, nil
+}
+
+func (f *fakeHelmActor) InitializeConfiguration(clusterDeployment *hmc.ClusterDeployment, _ action.DebugLog) (*action.Configuration, error) {
+	return &action.Configuration{}, nil
+}
+
+func (f *fakeHelmActor) EnsureReleaseWithValues(_ context.Context, _ *action.Configuration, _ *chart.Chart, _ *hmc.ClusterDeployment) error {
+	return nil
+}
+
 var _ = Describe("ClusterDeployment Controller", func() {
 	Context("When reconciling a resource", func() {
-		// const (
-		// 	helmChartURL = "http://source-controller.hmc-system.svc.cluster.local/helmchart/hmc-system/test-chart/0.1.0.tar.gz"
-		// )
+		const (
+			helmChartURL = "http://source-controller.hmc-system.svc.cluster.local/helmchart/hmc-system/test-chart/0.1.0.tar.gz"
+		)
 
 		// resources required for ClusterDeployment reconciliation
 		var (
@@ -235,7 +258,8 @@ var _ = Describe("ClusterDeployment Controller", func() {
 		AfterEach(func() {
 			By("Cleanup", func() {
 				controllerReconciler := &ClusterDeploymentReconciler{
-					Client: mgrClient,
+					Client:    mgrClient,
+					HelmActor: &fakeHelmActor{},
 				}
 
 				// Running reconcile to remove the finalizer and delete the ClusterDeployment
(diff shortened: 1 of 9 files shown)


internal/controller/template_controller_test.go
```diff
@@ -256,30 +256,33 @@ var _ = Describe("Template Controller", func() {
...
+			Expect(k8sClient.Get(ctx, key, mgmt)).To(Succeed())
 
 			By("Checking the management cluster appears")
 			Eventually(func() error {
 				if err := k8sClient.Get(ctx, client.ObjectKeyFromObject(mgmt), mgmt); err != nil {
 					return err
 				}
 
-				if l := len(mgmt.Status.AvailableProviders); l != 2 {
+				if l := len(mgmt.Status.AvailableProviders); l != 3 {
```
zerospiel, line 285: error has `2`, and at this line we have `3`. could you also elaborate why do we have to update the value? i mean, I see the three providers at the `suite_test.go:289` but at the moment of writing this comment, the reason is not clear. is it because the only instance of the `management` is to be used across all the tests (except the `mgmt-ctrl`) and at least one "expected"/"well-known provider" is expected? to me maybe it could also be beneficial to comment that explicitly

internal/controller/clusterdeployment_controller_test.go
```diff
@@ -164,65 +236,357 @@ var _ = Describe("ClusterDeployment Controller", func() {
...
+							// 	HaveField("Type", hmc.ReadyCondition),
+							// 	HaveField("Status", metav1.ConditionTrue),
+							// 	HaveField("Reason", hmc.SucceededReason),
+							// ),
+						))))
+				}).Should(Succeed())
+			})
+		})
+
+		// TODO (brongineer): Add test for ClusterDeployment reconciliation with Azure credentials
```
zerospiel, line 577: what do we expect here, file dedicated issues or refactor the code to maximize its reusability and then "copy-paste" the tests? i personally think that both of the proposals are time-wise costly and in general do not have much business value (having good coverage for the tier-1 provider)

zerospiel (commented):
minor comments, please consider to address

internal/controller/template_controller_test.go
```diff
@@ -256,30 +256,33 @@ var _ = Describe("Template Controller", func() {
...
-					Release: "test-release",
-				},
-			}
-			Expect(k8sClient.Create(ctx, mgmt)).To(Succeed())
-			mgmt.Status = hmcmirantiscomv1alpha1.ManagementStatus{
-				AvailableProviders: []string{someProviderName, otherProviderName},
-				CAPIContracts:      map[string]hmcmirantiscomv1alpha1.CompatibilityContracts{someProviderName: {capiVersion: someExposedContract}, otherProviderName: {capiVersion: otherExposedContract}},
-			}
-			Expect(k8sClient.Status().Update(ctx, mgmt)).To(Succeed())
+			// By("Creating a management cluster object with proper required versions in status")
```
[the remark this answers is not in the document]
BROngineer: Yep, thanks for pointing that. Forgot to cleanup.

internal/controller/template_controller_test.go
BROngineer: > is it because the only instance of the management is to be used across all the tests
not in tests, but controllers expect management object with hardcoded name to exist.

However, I just ran the tests without adding provider to management object's status, and they pass. So I'll revert this change. Also, we create the object, then manually update the status and check whether the status is in expected state. Without running controller which may mutate the object, these checks make no sense.

zerospiel: these checks are meant to ensure the objects have been actually updated and have no effect with further errors if any (e.g. `(1) create an object; (2) immediately check something; (3) object has been`, between `1` and `2` the `3` is not guaranteed hence the unexpected error might take place, hence it's better to have such checks)

internal/controller/clusterdeployment_controller_test.go
BROngineer: This and following todo's needed to not to forget that atm only one branch in creds propagation is tested. And actually it tests nothing:

```go
// clusterdeployment_controller.go L715
case "aws":
    l.Info("Skipping creds propagation for AWS")
```

> what do we expect here, file dedicated issues or refactor the code to maximize its reusability and then "copy-paste" the tests? i personally think that both of the proposals are time-wise costly and in general do not have much business value (having good coverage for the tier-1 provider)

hm, can't agree. without good coverage we can't predict the behaviour. Like did you know that clusterdeployment controller panics if for some reason helmchart status does not contain artifact field?

internal/controller/template_controller_test.go
BROngineer: > these checks are meant to ensure the objects have been actually updated and have no effect with further errors if any (e.g. (1) create an object; (2) object has been created; (3) check something, between 1 and 3 the step 2 is not guaranteed hence the unexpected error might take place, hence it's better to have such checks)

it would make sense in case there were running reconcilers in background which might mutate the object. Currently, this checks only tests that the k8s client in this concrete test was able to interact with api-server running for test. In other words, these checks do not test our code, but k8s client.

zerospiel: exactly. there are many checks of the same kind in the tests to avoid sporadic failures in the CI. better to have failures because of incorrect logic rather than having unrelated failures because of random whatever somewhere on the runners which caused sluggishness in a local running component

internal/helm/actor.go
```diff
@@ -0,0 +1,77 @@
...
+	clusterDeployment *v1alpha1.ClusterDeployment,
+	log action.DebugLog,
+) (*action.Configuration, error) {
+	getter := NewMemoryRESTClientGetter(a.Config, a.RESTMapper)
+	actionConfig := new(action.Configuration)
+	err := actionConfig.Init(getter, clusterDeployment.Namespace, "secret", log)
+	if err != nil {
+		return nil, err
+	}
+	return actionConfig, nil
```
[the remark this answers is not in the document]
BROngineer: @zerospiel is it ok for our code-style, that the function returning pointer and error, will return pointer to type's zero value along with an error? Isn't it more correct to return `nil, err` in case of error and `value, nil` in case no error occurred? 

Besides that, is it really worth packing all calls into a single-line return if it doesn't affect anything other than readability (which is very subjective measurement)?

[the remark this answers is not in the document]
zerospiel: just nitting, it was a suggestion obviously to my own style/preferences , up to the developer though

zerospiel (approved):

zerospiel:
@BROngineer pls update the PR regarding the project renaming ([here](https://github.com/K0rdent/kcm/actions/runs/12633148035/job/35198185480?pr=830#step:5:377) is the error)

zerospiel (approved):

PR #833, opened 2024-12-24
Modular infrastructure providers using inline Lua

s3rj1k:
This patch introduces a modular provider support system using inline Lua scripting. Key changes:

- Added provider registry system that enables modular provider support
- Added ability to load providers from YAML files with inline Lua scripts for credentials propagation
- Replaced hardcoded Azure provider logic with runtime loadable YAML config
- Removed Azure SDK dependencies


s3rj1k:
Marking this as draft, not closing, it has some value for Sveltos templates PR still

PR #841, opened 2024-12-27
Backup implementation: controller, plugins, collection, webhooks

zerospiel:
Backup implementation main part contains the following changes:

* install velero via flux rather than code
* adjust roles for the velero chart
* remove unnecessary controller values
* rename Backup to ManagementBackup
* remove Oneshot parameter from the Spec
* add StorageLocation to both Management and ManagementBackup
* remove unused types
* manage schedules in the ctrl instead of velero
* new source runner for schedules
* collect the required velero backup spec for the whole backup
* label Credential references (clusterIdentities) in order to include them in backup
* backup validation webhook
* amend backup controller logic with better objects handling
* fix bug in providertemplates ctrl when ownerreferences are being updated but requeue is not set
* add custom plugins set via mgmt spec
* rename k0smotron related provider labels to the correct ones from the k0sproject

To properly test the feature:

- have `k0rdent` instance (better with a `clusterdeployment` ready) with this PR
- install `backupstoragelocation` for `velero`, e.g.
```yaml
---
apiVersion: v1
data:
  # base64 decoded AK/SK, e.g. 3 lines:
  # [default]
  # aws_access_key_id = <key>
  # aws_secret_access_key = <secret_key>
  cloud: <base64>
kind: Secret
metadata:
  name: cloud-credentials
  namespace: kcm-system
type: Opaque
---
apiVersion: velero.io/v1
kind: BackupStorageLocation
metadata:
  name: default
  namespace: kcm-system
spec:
  config:
    region: <eu-central-1> # region with an s3 bucket
  default: true
  objectStorage:
    bucket: <bucket-name> # bucket name
  provider: aws
  credential:
    name: cloud-credentials
    key: cloud
```
- enable backup in the `management`, set `.spec.backup` to the something like `{enabled: true; schedule: "@every 5m"}`
- wait for the scheduled backup to be created
- fully drop the `k0rdent` instance (imitate a disaster)

The restoration process:
- install clean `k0rdent` installation (without `mgmt`/`releases`/etc)
- install `backup location` (see above) to be able to fetch `backups` from a cloud storage
- perform `velero restore` (e.g. `velero restore create <name> --existing-resource-policy update --from-backup <backup-name>`)
- wait for the restore in the `Completed` state (e.g. with `velero restore get`)
- done

#814


zerospiel:
Ok, I've figured out how to mitigate the issue with the `mutationwebhookconfigurations` and the template resources being declined to created: it needs to be done on the client side during the restoration process (we cannot control it) namely patch the deploy adding the `providertemplates` and `servicetemplates` resource to high priority list, so they are being applied before the `k0rdent` deployment (that problem does not affect the `clustertemplates` since they are alphabetically higher than `deployment.apps`)

The patch is as follows:
```
$ kubectl patch deploy -n velero velero \
--type='json' \
--patch='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--restore-resource-priorities=\"customresourcedefinitions,namespaces,storageclasses,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,datauploads.velero.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,endpoints,services,providertemplates.hmc.mirantis.com,servicetemplates.hmc.mirantis.com,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io,apps.kappctrl.k14s.io,packageinstalls.packaging.carvel.dev\""}]'
```

internal/controller/management_backup_controller.go
```diff
@@ -0,0 +1,324 @@
...
+
+	// NOTE: without installed CRDs it is impossible to initialize informers
+	// and the uncached client is required because it this point the manager
+	// still has not started the cache yet
+	uncachedCl, err := client.New(mgr.GetConfig(), client.Options{Cache: nil})
+	if err != nil {
+		return fmt.Errorf("failed to create uncached client: %w", err)
+	}
+
+	if err := r.config.InstallVeleroCRDs(uncachedCl); err != nil {
```
Kshatrix, line 147: we should install velero as part of the hmc chart as a dependency (similar to cert-manager)

internal/controller/backup/schedule.go
```diff
@@ -16,15 +16,202 @@ package backup
...
+
+	if scheduledBackup.Status.Reference == nil {
+		if scheduledBackup.CreationTimestamp.IsZero() || scheduledBackup.UID == "" {
+			l.Info("Creating scheduled ManagementBackup")
+			if err := c.cl.Create(ctx, scheduledBackup); err != nil {
+				return zeroResult, fmt.Errorf("failed to create scheduled ManagementBackup: %w", err)
+			}
+		}
+
+		veleroSchedule := &velerov1api.Schedule{
```
Kshatrix, line 56: I think we need to control execution ourselves (discussed seperately)

internal/controller/management_backup_controller.go
```diff
@@ -0,0 +1,324 @@
...
+			CreateFunc:  nil,
+			UpdateFunc: func(ctx context.Context, tue event.TypedUpdateEvent[client.Object], q workqueue.TypedRateLimitingInterface[ctrl.Request]) {
+				if tue.ObjectNew.GetNamespace() != r.config.GetVeleroSystemNamespace() || tue.ObjectNew.GetName() != backup.VeleroName {
+					return
+				}
+
+				q.Add(getManagementNameIfEnabled(ctx))
+			},
+		}).
+		Watches(&hmcv1alpha1.ClusterDeployment{}, handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, _ client.Object) []ctrl.Request {
```
Kshatrix, line 250: we won't need all of the watchers in that case

Kshatrix (changes requested):

internal/controller/backup/schedule.go
zerospiel: i've finally decided to go with the periodic updates since it is very complicated (and actually) unclear how to properly manipulate with a timer from another instance. i hope polling once a minute for a single API call is an okayish tradeoff compared to too complicated code. check it out

internal/controller/backup/collect.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	if err := cl.List(ctx, cldeploys, opts...); err != nil {
+		return nil, fmt.Errorf("failed to list ClusterDeployments: %w", err)
+	}
+
+	selectors := make([]*metav1.LabelSelector, len(cldeploys.Items)*2)
+	for i, cldeploy := range cldeploys.Items {
+		selectors[i] = selector(kcmv1alpha1.FluxHelmChartNameKey, cldeploy.Name)
+		selectors[i+1] = selector(clusterapiv1beta1.ClusterNameLabel, cldeploy.Name)
```
eromanova, line 127: The `i` variable corresponds to the index of `cldeploys.Items` but when you assign `selectors[i]` this value may already be assigned on the previous iteration by `selectors[i+1] = ...`. Is it correct?

internal/controller/backup/config.go
```diff
@@ -0,0 +1,63 @@
...
+	"errors"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kcmv1alpha1 "github.com/K0rdent/kcm/api/v1alpha1"
+)
+
+// Config holds required set of parameters of to successfully install Velero stack and manipulate with it.
```
eromanova, line 28: Looks like a typo
```suggestion
// Config holds required set of parameters to successfully install Velero stack and manipulate with it.
```

internal/controller/management_backup_controller.go
```diff
@@ -0,0 +1,147 @@
...
+				}
+			}
+		}
+
+		l.V(1).Info("No Management object exists or backup is disabled, nothing to do")
+		return ctrl.Result{}, nil
+	}
+
+	if mgmtBackup.CreationTimestamp.IsZero() || mgmtBackup.UID == "" { // mgmt exists
+		requestEqualsMgmt := mgmt != nil && req.Name == mgmt.Name && req.Namespace == mgmt.Namespace
```
eromanova, line 75: `req.Namespace == mgmt.Namespace` 
Management and ManagementBackup objects are cluster-scoped. Do we need this condition?

internal/controller/management_backup_controller.go
```diff
@@ -0,0 +1,147 @@
...
+		if !requestEqualsMgmt { // oneshot backup
+			l.V(1).Info("ManagementBackup object has not been found, nothing to do")
+			return ctrl.Result{}, nil
+		}
+
+		// required during creation
+		mgmtBackup.Name = req.Name
+		mgmtBackup.Namespace = req.Namespace
```
eromanova, line 83: Do we need to set namespace on cluster-scoped `ManagementBackup`?

api/v1alpha1/indexers.go
```diff
@@ -237,3 +239,19 @@ func extractOwnerReferences(rawObj client.Object) []string {
...
+
+// management backup indexers
+
+// ManagementBackupScheduledIndexKey indexer field name to extract only [ManagementBackup] objects
+// that meant to be scheduled.
+const ManagementBackupScheduledIndexKey = "k0rdent.scheduled-backup"
+
+func setupManagementBackupScheduledIndexer(ctx context.Context, mgr ctrl.Manager) error {
+	return mgr.GetFieldIndexer().IndexField(ctx, &ManagementBackup{}, ManagementBackupScheduledIndexKey, func(o client.Object) []string {
+		if _, err := strconv.ParseBool(o.GetAnnotations()[ScheduleBackupAnnotation]); err == nil {
```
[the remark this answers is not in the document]
zerospiel: i've just pushed the value to fill the index

internal/controller/backup/collect.go
zerospiel: right, thanks for noting, fixed!

internal/controller/management_backup_controller.go
zerospiel: for consistency, but yeah it is not required so I dropped it, resolving this and the same comment below

zerospiel:
@eromanova @Kshatrix I squashed both the commits into the single to ease the CR. Fine-tuned code blocks along with roles and manually tested from scratch, so now it is finally ready for a proper CR

templates/provider/kcm/values.yaml
```diff
@@ -95,3 +87,41 @@ cluster-api-operator:
...
+    enabled: false
+  upgradeCRDs: false
+  cleanUpCRDs: false
+  rbac:
+    create: false
+    clusterAdministrator: false
+  serviceAccount:
+    server:
+      create: false
+      name: velero-server-sa
```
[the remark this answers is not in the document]
zerospiel: `helm` does not allow to configure `values` with templates, unfortunately, but I think we could just go with default options which are:
```yaml
rbac:
  create: true
  clusterAdministrator: true
  clusterAdministratorName: cluster-admin
serviceAccount:
  server:
    create: true
```

considering the currently created cluster role has the same admin permissions. Let's stick to this and cover the caveats/howto in the docs

internal/controller/backup/reconcile.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *Reconciler) createScheduleBackup(ctx context.Context, mgmtBackup *kcmv1alpha1.ManagementBackup, nextAttemptTime time.Time) (ctrl.Result, error) {
+	now := time.Now()
+	backupName := mgmtBackup.TimestampedBackupName(now)
+
+	// TODO: NOTE: should we also set the ownership for the scheduled backups? i don't wanna do this to avoid accidental backups removal
```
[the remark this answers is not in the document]
zerospiel: i dropped the ownership in favor of a plain watcher, so the question is not valid now

templates/provider/kcm/values.yaml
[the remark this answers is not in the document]
zerospiel: I have dropped the `rbac` and the `serviceAccount` in favor of defaults (`clusterAdmin` `clusterrole`). This will be covered in the docs

zerospiel:
Should chart versions be updated because of the change to the provider label values?

api/v1alpha1/management_types.go
```diff
@@ -55,15 +55,19 @@ type Core struct {
...
+	// Default value is to backup at minute 0 past every 6th hour (0 */6 * * *).
 	Schedule string `json:"schedule,omitempty"`
 
-	// Flag to indicate whether the backup feature is enabled.
-	// If set to true, [Velero] platform will be installed.
-	// If set to false, creation or modification of Backups/Restores will be blocked.
+	// StorageLocation is the name of a [github.com/vmware-tanzu/velero/pkg/apis/velero/v1.StorageLocation]
+	// where the backup should be stored. It is propagated to the corresponding scheduled [ManagementBackup]
+	// unless set in the object.
+	StorageLocation string `json:"storageLocation,omitempty"`
```
Kshatrix, line 67: I'm a bit concerned that it overlaps with the same field in the backup object. let's discuss if we even need backup parameters in the management object at all

cmd/main.go
```diff
@@ -114,6 +115,7 @@ func main() {
 		"The name of the helm chart with KCM Templates.")
 	flag.BoolVar(&enableTelemetry, "enable-telemetry", true, "Collect and send telemetry data.")
 	flag.BoolVar(&enableWebhook, "enable-webhook", true, "Enable admission webhook.")
+	flag.BoolVar(&enableVelero, "enable-velero", true, "Enable Velero stack for management cluster backups.")
```
Kshatrix, line 118: I don't think we need it as part of the controller configuration. Similar approach to cer-manager and flux would be enough I think (enable/disable subchart)

Kshatrix (changes requested):

cmd/main.go
zerospiel: we need, if no chart is installed and without the flag, the controller starts and then fails because of watching a missing resource `velerov1.Backup`

api/v1alpha1/management_types.go
zerospiel: let's discuss, at the moment the ctrl sets up and manages the schedule hence no other way to configure "upstream" velero backups rather than via the mgmt object

Kshatrix (approved):

eromanova (approved):

PR #843, opened 2024-12-29
improve Makefile

### LINKED ISSUE #512: Makefile does not check for required binaries presence.
```
make cli-install
bash: line 1: jq: command not found
bash: line 1: jq: command not found
bash: line 1: jq: command not found
bash: line 1: jq: command not found
bash: line 1: jq: command not found
bash: line 1: jq: command not found
mkdir -p /media/share/Work/hmc/bin
curl -sL https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/download/v2.5.2/clusterawsadm_v2.5.2_linux_aarch64 -o /media/share/Work/hmc/bin/clusterawsadm
chmod +x /media/share/Work/hmc/bin/clusterawsadm
```

In this instance `jq` is missing, make should abort when required binary is not present on system

...
```
bash: line 3: unzip: command not found
```

ramessesii2:
- add jq to tools to be installed with `make cli-install`
    - exit with Error if a binary is not found.

Fixes #512


Makefile
```diff
@@ -399,7 +399,15 @@ dev-azure-nuke: envsubst azure-nuke ## Warning: Destructive! Nuke all Azure reso
 	@rm config/dev/azure-cloud-nuke.yaml
 
 .PHONY: cli-install
-cli-install: clusterawsadm clusterctl cloud-nuke envsubst yq awscli ## Install the necessary CLI tools for deployment, development and testing.
+cli-install: clusterawsadm clusterctl cloud-nuke envsubst yq awscli jq## Install the necessary CLI tools for deployment, development and testing.
+	@required_bins="clusterawsadm clusterctl cloud-nuke envsubst yq aws jq"; \
+	for bin in $$required_bins; do \
+		if ! command -v $$bin >/dev/null 2>&1; then \
+			$$bin ; \
```
zerospiel, line 406: I did not get it, why do we need to call a missing binary before printing an error and exiting?

nit: the comment does not have space between targets and its beginning (`jq##...`)

ramessesii2: Ah! that was only for debugging.
Removing it now. Thanks.

Makefile
```diff
@@ -441,6 +448,8 @@ export HELM
 KIND ?= $(LOCALBIN)/kind-$(KIND_VERSION)
 YQ ?= $(LOCALBIN)/yq-$(YQ_VERSION)
 export YQ
+JQ ?= $(LOCALBIN)/jq-$(JQ_VERSION)
+export JQ
```
a13x5, line 452: No need for export here

Makefile
```diff
@@ -399,7 +399,14 @@ dev-azure-nuke: envsubst azure-nuke ## Warning: Destructive! Nuke all Azure reso
...
 .PHONY: cli-install
-cli-install: clusterawsadm clusterctl cloud-nuke envsubst yq awscli ## Install the necessary CLI tools for deployment, development and testing.
+cli-install: clusterawsadm clusterctl cloud-nuke envsubst yq awscli jq ## Install the necessary CLI tools for deployment, development and testing.
+	@required_bins="clusterawsadm clusterctl cloud-nuke envsubst yq aws jq"; \
+	for bin in $$required_bins; do \
+		if ! command -v $$bin >/dev/null 2>&1; then \
+			echo "Error: '$$bin' is not installed or not in your PATH. Exiting."; \
+			exit 1; \
+		fi; \
+	done
```
a13x5, line 409: I don't understand the logic behind this. 
We just installed the tools needed (they're mentioned in the dependencies of the target) in the local bin directory and then we're testing that they're present in the PATH (they are not btw). Why?

a13x5 (changes requested):
Several notes:
- Some parts require additional explanation
- If we're installing `jq` we need to make sure that the same `jq` is used in all our scripts
- Issue with `unzip` reported in the #512 is not covered

Makefile
ramessesii2: Agreed. That's unnecessary. 
What I've been really upto is solving this isssue I've been seeing
```
make cli-install
bash: line 1: jq: command not found
bash: line 1: jq: command not found
bash: line 1: jq: command not found
bash: line 1: jq: command not found
bash: line 1: jq: command not found
bash: line 1: jq: command not found
```
The Makefile is structured in a way that certain variable assignments at the top level utilize external tools like jq and yq through $() functions.  These tools (jq and yq) are not installed when Make begins parsing the Makefile. Their installation is handled by the cli-install target, which occurs after the top-level variable assignments.

Secondly, #512 seems to point out to two issues:
1. The issue I've tried explaining above.
2. Missing unzip binary. To me it feels like adding unzip as a dependency would set a precedent for including other tools like curl. Does that make sense?

s3rj1k:
```
check-unzip:
    @which unzip > /dev/null 2>&1 || (echo "unzip is not installed" && exit 1)
```

ramessesii2: I can add that check. Thanks for the suggestion.

Makefile
```diff
@@ -489,22 +477,37 @@ $(HELM): | $(LOCALBIN)
 	curl -s --fail $(HELM_INSTALL_SCRIPT) | USE_SUDO=false HELM_INSTALL_DIR=$(LOCALBIN) DESIRED_VERSION=$(HELM_VERSION) BINARY_NAME=helm-$(HELM_VERSION) PATH="$(LOCALBIN):$(PATH)" bash
 
 $(FLUX_HELM_CRD): | $(EXTERNAL_CRD_DIR)
+	FLUX_HELM_VERSION ?= $(shell go mod edit -json | $(JQ) -r '.Require[] | select(.Path == "github.com/fluxcd/helm-controller/api") | .Version')
+	FLUX_HELM_NAME ?= helm
+	FLUX_HELM_CRD ?= $(EXTERNAL_CRD_DIR)/$(FLUX_HELM_NAME)-$(FLUX_HELM_VERSION).yaml
```
a13x5, line 482: I don't think it's justified here, since:
1. Syntax looks incorrect
2. You introduced duplication
3. You didn't add `jq` to the target requirements (I assume that was the idea)

I suggest you just change `jq` -> `$(JQ)` in the global definitions and leave it as it were.

a13x5 (changes requested):
Target specific vars look bad

Makefile
ramessesii2: Hey @a13x5, I've descoped the PR to address `jq dependecy` and `unzip verfication` to unblock the PR. 
I might raise another PR in future once I find a way to remove the duplication while addressing the original issue.

zerospiel:
closing because of the stale PR, the part of the initial issue regarding `jq` errors has already been resolved and the rest is here #1215

Issue #849, opened 2024-12-30
e2e Service deployment using ServiceTemplate as part of ClusterDeployment


BROngineer:
Deploy and validate deployment of the service, using ServiceTemplate included into `.spec.services` of ClusterDeployment

Issue #850, opened 2024-12-30
e2e Service deployment using ServiceTemplate after cluster deployment


BROngineer:
e2e test of MultiClusterService deployment

✅ tests deploying service after cluster(s) was created
✅ tests deploying service into multiple clusters simultaneously
✅ tests deploying service into clusters running by different cloud providers

Issue #871, opened 2025-01-07
e2e tests fail on admission check


a13x5:
# Issue

When running e2e tests there is an error, which prevents test to run

```
AWS Templates [It] should work with an AWS provider [provider:cloud, provider:aws]
/home/runner/work/kcm/kcm/test/e2e/provider_aws_test.go:75

  [FAILED] failed to create ClusterDeployment
  Unexpected error:
      <*errors.StatusError | 0xc0006dcc80>: 
      admission webhook "validation.clusterdeployment.hmc.mirantis.com" denied the request: the ClusterDeployment is invalid: the template is not valid: one or more required providers are not deployed yet: [bootstrap-k0smotron control-plane-k0smotron]
      {
          ErrStatus: {
              TypeMeta: {Kind: "Status", APIVersion: "v1"},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "admission webhook \"validation.clusterdeployment.hmc.mirantis.com\" denied the request: the ClusterDeployment is invalid: the template is not valid: one or more required providers are not deployed yet: [bootstrap-k0smotron control-plane-k0smotron]",
              Reason: "Forbidden",
              Details: nil,
              Code: 403,
          },
      }
  occurred
  In [It] at: /home/runner/work/kcm/kcm/test/e2e/provider_aws_test.go:85 @ 01/07/25 00:05:08.459
```

So it appears that we're trying to create `ClusterDeployment` before all other components are ready.

# Possible solution

Before applying `ClusterDeployment` we should check that all other objects (like Management, ClusterTemplate, ProviderTemplates and provider themselves) are ready. Currently we're relying on providers' Deployment status to figure that out, which isn't reliable and can't be used to detect k0rdent readiness.

**P.S.:** Since it was fixed before ( in #622 )it looks like the regression could be related to #847, since it significantly delays the deployment time of all components. Which actually emphasizes on how unreliable current readiness detection mechanism is.

PR #873, opened 2025-01-08
e2e test of managed service deployment

BROngineer:
- Add test of deploying service using ServiceTemplate included into ClusterDeployment

Fixes: #849 #850

### DIFF
--- a/test/e2e/clusterdeployment/servicevalidator.go
+++ b/test/e2e/clusterdeployment/servicevalidator.go
@@ -1,0 +1,65 @@
+// Copyright 2024
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clusterdeployment
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+
+	"github.com/K0rdent/kcm/test/e2e/kubeclient"
+)
+
+type ServiceValidator struct {
+	// managedClusterName is the name of managed cluster
+	managedClusterName string
+
+	// template being validated
+	template Template
+
+	// namespace is a namespace of deployed service
+	namespace string
+
+	// resourcesToValidate is a map of resource names and corresponding validation functions
+	resourcesToValidate map[string]resourceValidationFunc
+}
+
+func NewServiceValidator(clusterName, namespace string) *ServiceValidator {
+	return &ServiceValidator{
+		managedClusterName:  clusterName,
+		namespace:           namespace,
+		resourcesToValidate: make(map[string]resourceValidationFunc),
+	}
+}
+
+func (v *ServiceValidator) WithResourceValidation(resourceName string, validationFunc resourceValidationFunc) *ServiceValidator {
+	v.resourcesToValidate[resourceName] = validationFunc
+	return v
+}
+
+func (v *ServiceValidator) Validate(ctx context.Context, kc *kubeclient.KubeClient) error {
+	clusterKubeClient := kc.NewFromCluster(ctx, v.namespace, v.managedClusterName)
+
+	for res, f := range v.resourcesToValidate {
+		err := f(ctx, clusterKubeClient, res)
+		if err != nil {
+			_, _ = fmt.Fprintf(GinkgoWriter, "[%s/%s] validation error: %v\n", v.template, res, err)
+			return err
+		}
+		_, _ = fmt.Fprintf(GinkgoWriter, "[%s/%s] validation succeeded\n", v.template, res)
+	}
+	return nil
+}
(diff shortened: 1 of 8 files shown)


test/e2e/clusterdeployment/validate_deployed.go
```diff
@@ -273,3 +273,22 @@ func validateCCM(ctx context.Context, kc *kubeclient.KubeClient, clusterName str
 
 	return fmt.Errorf("%s Service does not yet have an external hostname", service.Name)
 }
+
+func ValidateService(ctx context.Context, kc *kubeclient.KubeClient, name string) error {
+	_, err := kc.Client.CoreV1().Services(kc.Namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
```
zerospiel, line 282: might it be useful to check if a request to the service's IP/in-cluster name actually returns something non-erroneous?

test/e2e/provider_aws_test.go
```diff
@@ -95,6 +97,20 @@ var _ = Describe("AWS Templates", Label("provider:cloud", "provider:aws"), Order
...
+				ResourceNameSuffix: "",
+				ValidationFunc:     clusterdeployment.ValidateService,
+			}).
+			WithResourceValidation("deployment", clusterdeployment.ManagedServiceResource{
+				ResourceNameSuffix: "",
+				ValidationFunc:     clusterdeployment.ValidateDeployment,
+			})
+		Eventually(func() error {
+			return serviceDeployedValidator.Validate(context.Background(), kc)
+		}).WithTimeout(30 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
```
zerospiel, line 112: 30 mins with 10 seconds interval looks a bit overwhelming, how much time does it take on practice to wait until deploys and services? might it be a good idea to increase the interval to maybe a minute or so?

zerospiel (commented):
In general LGTM, a couple of questions/comments, and the PR should be postponed until the e2e CI is ready (as of now and to my understanding it is blocked due to the #871). After addressing the issue, the PR should be then checked with the e2e tests job

test/e2e/clusterdeployment/servicevalidator.go
```diff
@@ -0,0 +1,85 @@
...
+	return &ServiceValidator{
+		managedClusterName:  clusterName,
+		managedServiceName:  serviceName,
+		namespace:           namespace,
+		resourcesToValidate: make(map[string]ManagedServiceResource),
+	}
+}
+
+func (v *ServiceValidator) WithResourceValidation(resourceName string, resource ManagedServiceResource) *ServiceValidator {
+	v.resourcesToValidate[resourceName] = resource
```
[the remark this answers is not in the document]
BROngineer: It's kinda part of "builder" for `ServiceValidator`, hence, no - it's not expected that anyone will try to apply options to the object during its creation concurrently.

test/e2e/clusterdeployment/validate_deployed.go
BROngineer: I assume that the goal is to ensure that the managed service was deployed, in other words that the resources were created. And checking that the service is functional is out of scope.

test/e2e/provider_aws_test.go
BROngineer: I'd prefer to keep interval reasonably short. Considering that in this case we poll external system, 10 seconds should be fine. Increasing the interval will cause unnecessary delay.

test/e2e/clusterdeployment/servicevalidator.go
```diff
@@ -0,0 +1,85 @@
...
+func (m ManagedServiceResource) resourceFullName(serviceName string) string {
+	if m.ResourceNameSuffix == "" {
+		return serviceName
+	}
+	return fmt.Sprintf("%s-%s", serviceName, m.ResourceNameSuffix)
+}
+
+type ServiceValidator struct {
+	// managedClusterName is the name of managed cluster
+	managedClusterName string
```
eromanova, line 43:
```suggestion
	clusterDeploymentName string
```

eromanova (commented):
Generally LGTM. One small question:
Should we also test service deployment for other providers or hosted environments? I’m asking because once https://github.com/k0rdent/kcm/pull/752 is implemented, some providers might be skipped, and service deployment may not be checked at all. But, if this code is expected to change soon, it’s fine as is for now.

test/e2e/clusterdeployment/validate_deployed.go
```diff
@@ -273,3 +273,22 @@ func validateCCM(ctx context.Context, kc *kubeclient.KubeClient, clusterName str
...
+}
+
+func ValidateDeployment(ctx context.Context, kc *kubeclient.KubeClient, name string) error {
+	dep, err := kc.Client.AppsV1().Deployments(kc.Namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if *dep.Spec.Replicas != dep.Status.ReadyReplicas {
```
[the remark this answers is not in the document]
BROngineer: Since there is a couple of validation functions with similar signature, it will be better to keep them consistent. So I added absent checks instead of reuse of function you referred.

zerospiel (approved):

BROngineer:
@zerospiel @eromanova @Kshatrix colleagues, e2e tests are flacky. What's especially annoying is that their stability depends on the cloud provider. GH actions history shows, that tests added in this PR passes in case kcm able to deploy the cluster. Pls approve and merge.

a13x5 (approved):
As discussed previously since tests are failing randomly (in other PRs as well) we can't wait indefinitely for the tests to pass.
Merging without green tests. Code itself looks good, but it must be revisited after service templates will be separated to the catalog repo

PR #876, opened 2025-01-09
Fix an issue where k0smotron enters a race condition on hosted templates

kylewuolle:
Simple fix to bump the k0smotron versions to 1.3.0 for all the components instead of just the one. The mismatch seems to cause a race condition when creating the etcd containers on hosted templates.


a13x5 (changes requested):
Please reduce the scope to just updating k0smotron and it's template version. 
No need to bumb hmc/hmc-template version here: this must be done during the release

a13x5 (approved):
LGTM

PR #877, opened 2025-01-09
Setup controllers dependent on Sveltos CRDs after its provider is installed

wahabmk:
# Description
Moved setup for `ClusterDeployment` and `MultiClusterService` controllers after Sveltos provider is installed.

# Verification
The logs no longer have the error:
```
2024-12-27T13:12:03Z ERROR controller-runtime.source.EventHandler if kind is a CRD, it should be installed before calling Start {"kind": "ClusterSummary.config.projectsveltos.io", "error": "no matches for kind \"ClusterSummary\" in version \"config.projectsveltos.io/v1beta1\""}
```

### DIFF
--- a/internal/controller/management_controller.go
+++ b/internal/controller/management_controller.go
@@ -39,6 +39,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	kcm "github.com/K0rdent/kcm/api/v1alpha1"
 	"github.com/K0rdent/kcm/internal/certmanager"
@@ -49,12 +50,14 @@ import (
 
 // ManagementReconciler reconciles a Management object
 type ManagementReconciler struct {
-	client.Client
-	Scheme                 *runtime.Scheme
-	Config                 *rest.Config
-	DynamicClient          *dynamic.DynamicClient
-	SystemNamespace        string
-	CreateAccessManagement bool
+	client                             client.Client
+	manager                            manager.Manager
+	scheme                             *runtime.Scheme
+	config                             *rest.Config
+	DynamicClient                      *dynamic.DynamicClient
+	SystemNamespace                    string
+	CreateAccessManagement             bool
+	sveltosDependentControllersStarted bool
 }
 
 func (r *ManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
@@ -62,7 +65,7 @@ func (r *ManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	l.Info("Reconciling Management")
 
 	management := &kcm.Management{}
-	if err := r.Get(ctx, req.NamespacedName, management); err != nil {
+	if err := r.client.Get(ctx, req.NamespacedName, management); err != nil {
 		if apierrors.IsNotFound(err) {
 			l.Info("Management not found, ignoring since object must be deleted")
 			return ctrl.Result{}, nil
@@ -84,14 +87,14 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *kcm.Manag
 	l := ctrl.LoggerFrom(ctx)
 
 	if controllerutil.AddFinalizer(management, kcm.ManagementFinalizer) {
-		if err := r.Client.Update(ctx, management); err != nil {
+		if err := r.client.Update(ctx, management); err != nil {
 			l.Error(err, "failed to update Management finalizers")
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, nil
 	}
 
-	if err := utils.AddKCMComponentLabel(ctx, r.Client, management); err != nil {
+	if err := utils.AddKCMComponentLabel(ctx, r.client, management); err != nil {
 		l.Error(err, "adding component label")
 		return ctrl.Result{}, err
 	}
@@ -111,7 +114,7 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *kcm.Manag
 		return ctrl.Result{}, err
 	}
 
-	components, err := getWrappedComponents(ctx, r.Client, management)
+	components, err := getWrappedComponents(ctx, r.client, management)
 	if err != nil {
 		l.Error(err, "failed to wrap KCM components")
 		return ctrl.Result{}, err
@@ -145,7 +148,7 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *kcm.Manag
 			continue
 		}
 		template := new(kcm.ProviderTemplate)
-		if err := r.Get(ctx, client.ObjectKey{Name: component.Template}, template); err != nil {
+		if err := r.client.Get(ctx, client.ObjectKey{Name: component.Template}, template); err != nil {
 			errMsg := fmt.Sprintf("Failed to get ProviderTemplate %s: %s", component.Template, err)
 			updateComponentsStatus(statusAccumulator, component, nil, errMsg)
 			errs = errors.Join(errs, errors.New(errMsg))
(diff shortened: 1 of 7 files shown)


internal/controller/management_controller.go
```diff
@@ -211,6 +218,47 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
...
+			currentNamespace := utils.CurrentNamespace()
+
+			l.Info(fmt.Sprintf("Provider %s has been successfully installed, so setting up controller for ClusterDeployment", hmc.ProviderSveltosName))
+			if err = (&ClusterDeploymentReconciler{
+				Client:          r.Manager.GetClient(),
+				Config:          r.Manager.GetConfig(),
+				DynamicClient:   dc,
+				SystemNamespace: currentNamespace,
+			}).SetupWithManager(r.Manager); err != nil {
+				return fmt.Errorf("failed to setup controller for ClusterDeployment: %w", err)
```
zerospiel, line 240: i think we should panic in that case, wdyt? the same comment for the second ctrl

internal/controller/management_controller.go
```diff
@@ -211,6 +218,47 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *hmc.Manag
 	return ctrl.Result{}, nil
 }
 
+func (r *ManagementReconciler) startDependentControllers(ctx context.Context, management *hmc.Management) error {
+	l := ctrl.LoggerFrom(ctx)
+
+	if !r.SveltosDependentControllersStarted {
+		if management.Status.Components[hmc.ProviderSveltosName].Success {
```
zerospiel, line 225: the nested blocks could be simplified
```
if started {
  return nil
}

if !mgmt.status.success {
   log
   return nil // btw, should we trigger requeue here?
}

// rest of the main logic
```


thoughts: the flag could be interchangeable with `sync.Once[Func]` or just an `atomic` but it probably might decrease readability with the former, atomic is just in case since I can't image the event in which the ctrl could run something in parallel (unless it is set explicitly by developers)

internal/controller/management_controller.go
```diff
@@ -50,11 +51,13 @@ import (
...
-	DynamicClient          *dynamic.DynamicClient
-	SystemNamespace        string
-	CreateAccessManagement bool
+	Manager                            manager.Manager
+	Scheme                             *runtime.Scheme
+	Config                             *rest.Config
+	DynamicClient                      *dynamic.DynamicClient
+	SystemNamespace                    string
+	CreateAccessManagement             bool
+	SveltosDependentControllersStarted bool
```
zerospiel, line 60: super nit up to you: most of these fields might be actually private and be set in the `setupwithmanager` method

```go
func (r *reconciler) setupwithmanager(mgr) error {
   r.mgr = mgr
   r.scheme = mgr.getscheme()
   r.sveltosStarted = false
   // etc
```

zerospiel (commented):
the approach looks good considering we cannot manually install the required crd beforehand somewhere in the `setup` section which is not an option here

internal/controller/management_controller.go
wahabmk: I used `sync.OnceFunc` with panics in the latest commit. With using it, I got rid of the `SveltosDependentControllersStarted` flag. Do you agree with its use?

zerospiel: i haven't noticed it yet, now i see. to me it looks okay! i'd only suggest to skip that huge `if` block as I suggested in the initial comment:

```
	if !management.Status.Components[kcm.ProviderSveltosName].Success {
            l.Info("Waiting for provider to be ready to setup contollers dependent on it", "provider", kcm.ProviderSveltosName))
            return
	}
	// syncOnce
```

s3rj1k (changes requested):

internal/controller/management_controller.go
```diff
@@ -211,6 +216,44 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *kcm.Manag
...
+
+	if !management.Status.Components[kcm.ProviderSveltosName].Success {
+		l.Info(fmt.Sprintf("Waiting for %s provider to be ready to setup contollers dependent on it", kcm.ProviderSveltosName))
+		return
+	}
+
+	sync.OnceFunc(func() {
+		dc, err := dynamic.NewForConfig(r.Manager.GetConfig())
+		if err != nil {
+			panic(fmt.Errorf("failed to create dynamic client: %w", err))
```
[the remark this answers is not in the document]
zerospiel: I missed it, that panic is definitely not required and should be requeued hence I think the whole approach is a bit superfluous

internal/controller/management_controller.go
```diff
@@ -211,6 +216,44 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *kcm.Manag
...
 
+func (r *ManagementReconciler) startDependentControllers(ctx context.Context, management *kcm.Management) {
+	l := ctrl.LoggerFrom(ctx)
+
+	if !management.Status.Components[kcm.ProviderSveltosName].Success {
+		l.Info(fmt.Sprintf("Waiting for %s provider to be ready to setup contollers dependent on it", kcm.ProviderSveltosName))
+		return
+	}
+
+	sync.OnceFunc(func() {
```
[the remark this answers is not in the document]
wahabmk: This function isn't intended to be used anywhere else though

internal/controller/management_controller.go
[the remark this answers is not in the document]
wahabmk: I could either use `sync.OnceValue` and return error or just stick with using a flag as done in the 1st commit.

internal/controller/management_controller.go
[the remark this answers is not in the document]
s3rj1k: yes, but defining it anonymously on every reconcile would be not very efficient, you also loose the ability to write unit-test for it and make this code read harder

internal/controller/management_controller.go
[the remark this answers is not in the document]
s3rj1k: sync.OnceValue for making sure flag is set only once would be better for this case IMO
(in case flag.Bool from CLI is for some reason is unacceptable)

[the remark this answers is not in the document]
s3rj1k: but I am not a codeowner :)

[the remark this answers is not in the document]
s3rj1k: so setting sync.OnceValue on defer and for intermediates, use localy-scoped variables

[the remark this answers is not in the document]
s3rj1k: @wahabmk not suggesting you do this, just in case

internal/controller/management_controller.go
```diff
@@ -49,20 +50,22 @@ import (
 
 // ManagementReconciler reconciles a Management object
 type ManagementReconciler struct {
```
s3rj1k, line 52: not sure why client is unexported and DynamicClient is exported?

internal/controller/management_controller.go
```diff
@@ -211,14 +222,53 @@ func (r *ManagementReconciler) Update(ctx context.Context, management *kcm.Manag
 	return ctrl.Result{}, nil
 }
 
+// startDependentControllers starts controllers that cannot be started
+// at process startup because of some dependency like CRDs being present.
+func (r *ManagementReconciler) startDependentControllers(ctx context.Context, management *kcm.Management) (requue bool, err error) {
```
s3rj1k, line 227: this is effectively a reconcile, so like starting reconcile in reconcile, not against that, just mentioning how far kubebuilder is from reality

no action needed here

zerospiel (dismissed):

internal/controller/management_controller.go
wahabmk: That is because dynamic client is [being set in main.go](https://github.com/k0rdent/kcm/blob/5c0ce3213dfc2c989e9ee4efeb266d0ee34fb21b/cmd/main.go#L244) so it needs to be exported. Although I guess I can create the dynamic client in the SetupWithManager function as well, that would allow it to be unexported. I'll do that

s3rj1k: I am pretty sure that even if you make all fields exported this won't have any impact as code is behind `internal` folder.

zerospiel: that does not matter, i've `super nit`'ed it bc of personal preference. no impact at all, I'm okay with either exported or unexported

let me know when the PR is ready to be merged

s3rj1k: I mean either way is ok, just let's be consistent with clients (both exported or both unexported)

s3rj1k (approved):

zerospiel (approved):

PR #883, opened 2025-01-10
Add authorize job which requires externally created PRs to be approve…

kylewuolle:
This PR enables the concept of approving externally created PRs for running e2e tests. It is based on a blog post : [https://datachain.ai/blog/testing-external-contributions-using-github-actions-secrets](https://datachain.ai/blog/testing-external-contributions-using-github-actions-secrets)

The setup involved here is first we create two environments: external and internal. The external environment will have a deployment protection rule enabled which will require a review from anyone on the team. This PR introduces a job called "authorize" which only runs after approval if the PR originates from a fork. Then, the e2e test jobs are marked as needing the authorize job to complete prior to running. In this way users on the team can still run e2e tests from either forked or unforked PRs since we can approve the job.

Here is an example of the protection in action:
vokoscreen-2025-01-10_09-52-32

Here's an example of a PR which is not forked:
vokoscreen-2025-01-10_10-06-37

For forked PRs from team members the flow will be the same as the first example but we can self approve the job run.

Resolves https://github.com/k0rdent/2a-internal/issues/17


zerospiel:
A quick clarification: so the new job `authorize` is valid only for the e2e-related jobs, and the flow is still the same, like to set the labels? However, only an authorized team member could set the label, right? Could you please elaborate on this? In other case, the approach looks ok to me

DinaBelova:
I think keeping labels is okay (if that is the case) as even for PRs produced by main dev team e2e tests run (and which exact one) may be different depending on the commit. So I think having labels as the instrument to specify what tests exactly to run seems ok.

@kylewuolle please confirm the labels approach was untouched ^^.

eromanova (commented):
@kylewuolle I like the approach, but could you please clarify something? So as I understand:
	•	For PRs created from forked repositories, additional approval from members of the k0rdent team is required
	•	For unforked PRs (pushed directly to the original kcm project), no additional protection is applied since only team members have permission to push directly to k0rdent.
Am I correct in understanding this? Specifically, I’m curious whether the protection will function correctly for PRs created directly in the repository.

kylewuolle:
> @kylewuolle I like the approach, but could you please clarify something? So as I understand: • For PRs created from forked repositories, additional approval from members of the k0rdent team is required • For unforked PRs (pushed directly to the original kcm project), no additional protection is applied since only team members have permission to push directly to k0rdent. Am I correct in understanding this? Specifically, I’m curious whether the protection will function correctly for PRs created directly in the repository.

Yes that's right. So unforked PRs will not require approval and forked ones will. Since every team member is able to approve PRs for e2e test runs (including their own) this will not cause issues where team members will need to get someone else to approve a test run. Hopefully that makes sense.

eromanova (approved):

DinaBelova:
please do not merge until we'll do the needed changes in the repo settings (envs setup)

a13x5 (approved):
I think that we can try that approach. LGTM

PR #888, opened 2025-01-13
Set default values for OpenStack resources 

bnallapeta:
Without these default values, cluster provisioning will not proceed. The defaults provided here are used by OpenStackCluster, OpenStackMachine. 

This is similar to how the other providers are currently set up.

### DIFF
--- a/config/dev/openstack-clusterdeployment.yaml
+++ b/config/dev/openstack-clusterdeployment.yaml
@@ -4,12 +4,12 @@ metadata:
   name: openstack-dev
   namespace: ${NAMESPACE}
 spec:
-  template: openstack-standalone-cp-0-0-1
+  template: openstack-standalone-cp-0-0-2
   credential: openstack-cluster-identity-cred
   config:
     controlPlaneNumber: 1
     workersNumber: 1
-    controlPlane:
+    controlplane:
       flavor: ${OPENSTACK_CONTROL_PLANE_MACHINE_FLAVOR}
       image:
         filter:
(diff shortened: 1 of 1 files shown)


zerospiel (approved):

ramessesii2 (approved):

eromanova (approved):

templates/cluster/openstack-standalone-cp/values.yaml
Kshatrix, line 1: Those defaults seem highly cloud-specific, how about we add them to the condig/dev example instead?

Kshatrix (changes requested):

templates/cluster/openstack-standalone-cp/values.yaml
bnallapeta: azure - https://github.com/k0rdent/kcm/blob/main/templates/cluster/azure-standalone-cp/values.yaml#L30-L35

aws - https://github.com/k0rdent/kcm/blob/main/templates/cluster/aws-standalone-cp/values.yaml#L34-L37

aws - https://github.com/k0rdent/kcm/blob/main/templates/cluster/aws-standalone-cp/values.yaml#L25-L27

Followed similar style to other providers. Even with this, users will be able to override them via the ClusterDeployment. But having a default (common values which are expected in most clouds) is good so that it won't fail to come up.

Kshatrix: it totally makes sense for a public cloud. os template is just different in that sense

Kshatrix (approved):

PR #893, opened 2025-01-14
Update cluster template versions

Kshatrix:


config/dev/vsphere-clusterdeployment.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     k0rdent.mirantis.com/component: kcm
 spec:
-  template: vsphere-standalone-cp-0-0-3
+  template: vsphere-standalone-cp-0-0-5
```
zerospiel, line 9: was the `3` -> `5` intentionally? the same goes for the change in the e2e tests

Kshatrix: I've updated to the current version. It wasn't updated last time it seems

zerospiel (approved):

Issue #919, opened 2025-01-17
Modular CAPI providers support


DinaBelova:
**Goals**
This epic focuses on enabling organizations to create and integrate custom cluster templates and Cluster API (CAPI) providers into k0rdent without modifying the core codebase. By providing an extension mechanism, integration framework, and comprehensive documentation, this solution will allow enterprises with custom infrastructure, compliance, or security requirements to meet their specific needs while maintaining the security and stability of the k0rdent platform.

- Enable customization: Allow organizations to create and integrate custom cluster templates and providers without modifying k0rdent’s core codebase.
- Support custom infrastructure: Provide a framework for custom CAPI provider integration that adheres to CAPI specifications.
- Simplify adoption: Deliver documentation, examples, and tools to make template and provider creation straightforward.
- Preserve security: Ensure custom templates and providers comply with the k0rdent security model.

**Major deliverables**
Extension Mechanism for Custom Templates
- Enable integration of organisation-specific cluster templates without modifying k0rdent.
- Provide validation and testing tools to ensure template compatibility and security.
- Document best practices for template creation and integration.

Integration Framework for Custom CAPI Providers
- Provide an interface to integrate custom providers that follow CAPI specifications.
- Offer support for direct integration of proprietary systems where required.
- Include testing tools to validate custom provider compatibility with k0rdent.

Documentation and Examples
- Comprehensive documentation for custom cluster template creation, provider integration, and verification processes.
- Runbooks covering:
    - Template creation and testing.
    - Provider integration workflows.
    - Verification and validation steps.

**Who it benefits**
Enterprises with Custom Infrastructure:
- Integrate proprietary cloud platforms and comply with organisation-specific requirements.

Organisations with Compliance and Security Needs:
- Build custom solutions that align with stringent security and regulatory standards.

Platform Teams:
- Extend cluster templates and configurations to meet specialised needs.

Mirantis:
- Reduce support overhead by enabling customers to independently manage templates and providers.

**Acceptance criteria**
Custom Template Integration
- Organisations can create and integrate custom cluster templates into k0rdent without modifying the core codebase.
- Validation and testing tools ensure templates are secure and compatible with k0rdent.
- Templates can extend existing configurations with organisation-specific requirements.

Custom CAPI Provider Integration
- Organisations can create and integrate custom providers following CAPI specifications.
- Integration framework supports proprietary systems without compromising k0rdent security.
- Custom providers can be validated and tested using provided tools.

Documentation and Tools
[...]

PR #925, opened 2025-01-20
Re:  Modular infrastructure providers using Sveltos templates

s3rj1k:
Modular infrastructure providers using Sveltos templates (https://github.com/k0rdent/kcm/issues/919)

- Consolidates cloud provider credential managers under a common interface.
- Adds reconciliation of infrastructure provider resources using Sveltos profile templates.
- Removes credential propagation logic.
- Migrates credential propagation logic into Sveltos profile template.
- Adopts `Backup` feature to new modular approach.
- Remove unused code.


s3rj1k:
@zerospiel PTAL (top) commit related  to `Backup`

internal/controller/credential_controller.go
eromanova, line 1: Currently, if we drop the credential controller the `Credential` objects will never be marked as ready. Thus, the `ClusterDeployment` creation will always be rejected by the admission webhook because the credentials are not ready.
Can we keep the credential controller and bring back the logic to verify the `identityRef` object exists? It does not contain any provider-specific code.

s3rj1k: >  Can we keep the credential controller and bring back the logic to verify the identityRef object exists? It does not contain any provider-specific code.

That will not allow us to make `manager` provider agnostic, there will be some hardcoded knowledge in code, either typed modules or untyped unstructured data that will still has to be based on specific provider as I understand this.

config/dev/openstack-credentials.yaml
```diff
@@ -22,10 +24,85 @@ kind: Credential
 metadata:
   name: openstack-cluster-identity-cred
   namespace: ${NAMESPACE}
+  labels:
+    k0rdent.mirantis.com/component: "kcm"
```
zerospiel, line 28: ditto about removing/leaving as is

config/dev/openstack-credentials.yaml
```diff
@@ -22,10 +24,85 @@ kind: Credential
...
+
+    {{- $$network_id := $$cluster.status.externalNetwork.id -}}
+    {{- $$network_name := $$cluster.status.externalNetwork.name -}}
+    ---
+    apiVersion: v1
+    kind: Secret
+    metadata:
+      name: openstack-cloud-config
+      namespace: kube-system
+    type: Opaque
```
zerospiel, line 76: would the already existing secret be updated with the strategic merge or would it be recreated? anyway i'd put the component label here explicitly

config/dev/vsphere-credentials.yaml
```diff
@@ -24,10 +28,76 @@ kind: Credential
 metadata:
   name: vsphere-cluster-identity-cred
   namespace: ${NAMESPACE}
+  labels:
+    k0rdent.mirantis.com/component: "kcm"
```
zerospiel, line 32: ditto about removing/leaving as is

config/dev/vsphere-credentials.yaml
```diff
@@ -24,10 +28,76 @@ kind: Credential
...
+    projectsveltos.io/template: "true"
+data:
+  configmap.yaml: |
+    {{- $$cluster := .InfrastructureProvider -}}
+    {{- $$identity := (getResource "InfrastructureProviderIdentity") -}}
+    {{- $$secret := (getResource "InfrastructureProviderIdentitySecret") -}}
+    ---
+    apiVersion: v1
+    kind: Secret
+    metadata:
```
zerospiel, line 58: the component label is missing here. i am not sure whether sveltos will properly recreate everything and in the correct moment that's why i'd put the label here to avoid potential inconsistencies

config/dev/vsphere-credentials.yaml
```diff
@@ -24,10 +28,76 @@ kind: Credential
...
+      name: vsphere-cloud-secret
+      namespace: kube-system
+    type: Opaque
+    data:
+      {{ printf "%s.username" $$cluster.spec.server }}: {{ index $$secret.data "username" }}
+      {{ printf "%s.password" $$cluster.spec.server }}: {{ index $$secret.data "password" }}
+    ---
+    apiVersion: v1
+    kind: Secret
+    metadata:
```
zerospiel, line 68: the component label is missing here. i am not sure whether sveltos will properly recreate everything and in the correct moment that's why i'd put the label here to avoid potential inconsistencies

config/dev/vsphere-credentials.yaml
```diff
@@ -24,10 +28,76 @@ kind: Credential
...
+        [VirtualCenter "{{ $$cluster.spec.server }}"]
+        insecure-flag = "true"
+        user = "{{ index $$secret.data "username" | b64dec }}"
+        password = "{{ index $$secret.data "password" | b64dec }}"
+        port = "443"
+        datacenters = ${VSPHERE_DATACENTER}
+    ---
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
```
zerospiel, line 86: the component label is missing here. i am not sure whether sveltos will properly recreate everything and in the correct moment that's why i'd put the label here to avoid potential inconsistencies

internal/controller/backup/collect_test.go
```diff
@@ -125,27 +125,18 @@ func Test_sortDedup(t *testing.T) {
 				{
 					MatchLabels: map[string]string{"cluster.x-k8s.io/provider": "control-plane-k0smotron"},
 				},
-				{
-					MatchLabels: map[string]string{"cluster.x-k8s.io/provider": "infrastructure-azure"},
-				},
```
zerospiel, line 130: what is the reason for removing the unrelated data from the test? besides, this will be in the final spec, so please, revert the changes

internal/controller/credential_controller.go
```diff
@@ -1,277 +0,0 @@
...
-
-		apimeta.SetStatusCondition(cred.GetConditions(), metav1.Condition{
-			Type:    kcm.CredentialReadyCondition,
-			Status:  metav1.ConditionFalse,
-			Reason:  kcm.FailedReason,
-			Message: errMsg,
-		})
-
-		return ctrl.Result{}, err
-	}
```
zerospiel, line 85: this code should exist

internal/controller/credential_controller.go
```diff
@@ -1,277 +0,0 @@
...
-
-	if err := utils.AddKCMComponentLabel(ctx, r.Client, clIdty); err != nil {
-		l.Error(err, "adding component label to the ClusterIdentity")
-		return ctrl.Result{}, err
-	}
-
-	if err := r.updateIdentityReferencedSecret(ctx, cred.Spec.IdentityRef); err != nil {
-		l.Error(err, "adding component label to the Secret from the Identity Reference")
-		return ctrl.Result{}, err
-	}
```
zerospiel, line 95: this code should be removed

internal/controller/credential_controller.go
```diff
@@ -1,277 +0,0 @@
...
-
-	cred := &kcm.Credential{}
-	if err := r.Client.Get(ctx, req.NamespacedName, cred); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
-	}
-
-	if err := utils.AddKCMComponentLabel(ctx, r.Client, cred); err != nil {
-		l.Error(err, "adding component label")
-		return ctrl.Result{}, err
-	}
```
zerospiel, line 56: this code MIGHT be removed if you want to mark the kcm `credentials` explicitly with labels but i'd leave it as it is right now

internal/controller/credential_controller.go
```diff
@@ -1,277 +0,0 @@
...
-
-	apimeta.SetStatusCondition(cred.GetConditions(), metav1.Condition{
-		Type:    kcm.CredentialReadyCondition,
-		Status:  metav1.ConditionTrue,
-		Reason:  kcm.SucceededReason,
-		Message: "Credential is ready",
-	})
-
-	return ctrl.Result{RequeueAfter: r.syncPeriod}, nil
-}
```
zerospiel, line 105: this code should exist

internal/controller/credential_controller.go
```diff
@@ -1,277 +0,0 @@
...
-	secret := new(corev1.Secret)
-	if err := cl.Get(ctx, secretKey, secret); err != nil {
-		return fmt.Errorf("failed to get Secret %s referenced in %s %s: %w", secretKey, identity.GetObjectKind().GroupVersionKind().Kind, client.ObjectKeyFromObject(identity), err)
-	}
-
-	if err := utils.AddKCMComponentLabel(ctx, cl, secret); err != nil {
-		return fmt.Errorf("failed to add component label: %w", err)
-	}
-	return nil
-}
```
zerospiel, line 268: this code should be removed

internal/controller/credential_controller.go
```diff
@@ -1,277 +0,0 @@
...
-			break
-		}
-	}
-
-	if err := r.Client.Status().Update(ctx, cred); err != nil {
-		return fmt.Errorf("failed to update Credential %s/%s status: %w", cred.Namespace, cred.Name, err)
-	}
-
-	return nil
-}
```
zerospiel, line 121: this code should exist

internal/controller/credential_controller.go
```diff
@@ -1,277 +0,0 @@
...
-
-// SetupWithManager sets up the controller with the Manager.
-func (r *CredentialReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	r.syncPeriod = 15 * time.Minute
-
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&kcm.Credential{}).
-		Complete(r)
-}
```
zerospiel, line 277: this code should exist

zerospiel (commented):
i forgot to explicitly mark 2 code blocks to continue existing

internal/controller/credential_controller.go
s3rj1k: we talked about `internal/controller/credential_controller` with @eromanova offline, I will keep first level of `ref` in controller, will resolve those in bulk

zerospiel: if you mean the `identityRef` i think this might not be a good idea since it might be required to have permissions to `update` some `fooidentities` resource which `manager` won't have hence one should manually change the corresponding cluster role which is not very convincing. to sum up, i'd only leave the `credentials` labeling and drop `identityRef` and all the nested ones. WDYT? if you agree, feel free to drop the following permissions from the  `templates/provider/kcm/templates/rbac/controller/roles.yaml`:

```yaml
  resources:
  - awsclusterstaticidentities
  - awsclustercontrolleridentities
  - awsclusterroleidentities
  - azureclusteridentities
  - vsphereclusteridentities
  verbs: {{ include "rbac.viewerVerbs" . | nindent 2 }}
  - update # required for the managementbackups # DROP ME
 ```

```yaml
  resources:
  - secrets
  verbs: {{ include "rbac.viewerVerbs" . | nindent 2 }}
  - update # required for the managementbackups # DROP ME
```

s3rj1k: I was thinking of keeping `Credentials` reconcile with labels and `ref` from `Credentials` to `something` and update that something label, but hmm, RBAC can be a problem, yea

s3rj1k: so I guess only 1st level labels on `Credentials` without touching `ref`? @eromanova ?

s3rj1k: I am actually not sure that it is worth dropping RBACs, we still would want to backup provider-specific object, what we don't want is provider-specific go-code, charts are fine, as they are customizable

zerospiel: these are not related to each other, `velero` has the dedicated cluster role, and we are discussing the `manager` role here, so these are 100% not required

config/dev/azure-credentials.yaml
```diff
@@ -29,10 +32,68 @@ kind: Credential
...
+      "vnetName" $$cluster.spec.networkSpec.vnet.name
+      "vnetResourceGroup" $$cluster.spec.networkSpec.vnet.resourceGroup
+    -}}
+    ---
+    apiVersion: v1
+    kind: Secret
+    metadata:
+      name: azure-cloud-provider
+      namespace: kube-system
+    type: Opaque
```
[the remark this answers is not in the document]
s3rj1k: this is managed/child cluster object, are we also doing backup for them?

config/dev/openstack-credentials.yaml
s3rj1k: same, this is managed/child cluster object

s3rj1k: it should be recreated by Sveltos from mgmt cluster

config/dev/azure-credentials.yaml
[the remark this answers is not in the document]
zerospiel: if the object does not exist on the management cluster, then the answer is no, and there is no need in labeling (if executing `kubectl get secret -n kube-system azure-cloud-provider` with the mgmt cluster kubeconfig will return `not found` then it's okay not to label)

config/dev/vsphere-credentials.yaml
s3rj1k: Sveltos reconciles that object, this is 100%, basically Sveltos is creating this object in the first place

config/dev/openstack-credentials.yaml
zerospiel: the same answer as here https://github.com/k0rdent/kcm/pull/925#discussion_r1923603071

if was under the impression that these objects would persist within the management cluster, otherwise it's okay not to label

internal/controller/backup/collect_test.go
s3rj1k: idea is to have no mention of other that `aws` providers in tests, it does not affect anything but just keeps codebase consistent and makes us remeber that we are provider-agnostic

config/dev/vsphere-credentials.yaml
zerospiel: ok, please resolve the same comments if they make no sense in terms of the management cluster. i was just trying to be over conscious just in case

internal/controller/backup/collect_test.go
zerospiel: 1. it is the test data (one could even name it `golden`)
2. the actual backup has these label selectors (a `clusterdeployment` object always has a provider)
3. please keep them since the data has no relation to providers or to plugins

config/dev/azure-credentials.yaml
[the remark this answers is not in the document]
s3rj1k: image
this is object template that is rendered by Sveltos, actual object that exists in management cluster is this one

image

config/dev/openstack-credentials.yaml
s3rj1k: resolving, keep one above for reference

internal/controller/backup/collect_test.go
zerospiel: if you want to rename `azure` to `aws` please go ahead, but the labels and the value-style should persist

config/dev/vsphere-credentials.yaml
s3rj1k: resolving, keep one above for reference

config/dev/vsphere-credentials.yaml
s3rj1k: resolving, keep one above for reference

internal/controller/backup/collect_test.go
s3rj1k: I think there are aws values in there, sure I can rename, no prob

internal/controller/credential_controller.go
```diff
@@ -55,53 +49,14 @@ func (r *CredentialReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-
-		apimeta.SetStatusCondition(cred.GetConditions(), metav1.Condition{
-			Type:    kcm.CredentialReadyCondition,
-			Status:  metav1.ConditionFalse,
-			Reason:  kcm.FailedReason,
-			Message: errMsg,
-		})
-
-		return ctrl.Result{}, err
-	}
```
zerospiel, line 85: i am not quite sure why is this block removed. if the reference is missing — obviously the credentials are incorrect, so why should we mark it as ready?

zerospiel (commented):
in general lgtm (except the credentials ctrl question), needs to be tested + at least `aws` backup (full) should be checked

internal/controller/credential_controller.go
zerospiel: i understand this might be related to the RBAC on the other hand (referring to the identity reference labeling) but otherwise the whole idea with the reference (and potentially with the `credentials` crd itself) does not make much sense at all

eromanova: Same question from my side. I think we should check the existence of `spec.identityRef` object and mark `Credential` as `ready` if the object exists or `not ready` if it's not. We can use `PartialObjectMetadata` to get the `identityRef`. For example, in a similar way it's currently done in https://github.com/k0rdent/kcm/blob/main/internal/webhook/template_webhook.go#L312 Or it does not work?

zerospiel: does the partial meta resolve the rbac "issue"? i mean unless there is explicit permission to `get` some `fooidentity` (mostly it is not the case but in theory, it might explode) I think the error with insufficient RBAC would arise

so in general, I dunno how to properly handle the `provider-agnostic` approach in this case since we had already ditched the labeling of `identityReference` to handle potentially insufficient RBAC permissions

eromanova: Oh, right. Partial meta will also return an error due to lack of access if RBAC is not configured properly. Ok, looks like the whole readiness check now does not make sense if we can't check the identity existence. I don't know, maybe keep things as is and consider credentials as always `Ready` without any internal checks (at least for now, to avoid lots of API/webhook changes). Not the best option but it can be discussed/reworked separately. @zerospiel what do you think?

zerospiel: let's ditch the logic then, then i'd explicitly ask also to drop the `requeueAfter` since it won't make sense (and the `requeuePeriod` field from the `reconciler` type). the open question to discuss: what should we do with the CRD itself and how to find a proper use-case for it? at the moment I'm just thinking of it as a status-less object that just holds and shows the current state.

a13x5: There was a specific requirement to validate that the referenced object in Credential exists and prohibit creation of the cluster if it's not.
If that requirement didn't changed this should be left as is.
Also these changes are out of scope of this PR

zerospiel: As discussed: we actually not sure what to do in such a case in terms of insufficient RBAC permissions. I cannot agree nor disagree that this is out of the scope of the PR, since `get`-ting out-of-tree provider resources might be an issue. I could only suggest that we'd stick to one of 2 options then:
1. Remove that logic (as @s3rj1k had done)
2. Revert it back (**without** labeling the `identityRef` but **with** getting/checking it) putting back the requeue AND creating a separated issue on how to process it properly (I've already suggested to make the creds status-less and do not enforce the webhook checks)

s3rj1k: FYI: There will be a followup on whole modular providers with a bit of less scope (size) of PR, so we can defer this until 2nd PR, we would still need to talk about it, so can also revisit this

s3rj1k: RBAC issue is about putting a list of cluster dependent objects into a chart, this is not a blocker in terms of current epic

s3rj1k: We still have them in cart, so I donno, we can bring back 1st level of ref-checking at any time

zerospiel: @a13x5 FYI: if some out-of-tree provider is trying to utilize the `kcm` there is a possibility to have some random (let's call it `FooIdentity`) kind as the `IdentityRef` in the `Credential`, hence the insufficient permissions problem will arise for this particular case and one should manually add the resource to the `kcm` cluster role which is not UX-friendly.

@s3rj1k will ya please revert the changes (only the basic part: label creds themselves, get identity reference, requeue, **no labeling code regarding identities** should be reverted) back so we can easily merge the PR to unlock follow-ups without breaking anything along with the current logic. I know it sounds frustrating but my decision was a bit premature

s3rj1k: questing is, do we cover with that non-recursive ref check, what we want, does it make sense to check only on object existence?

We can check for example if `azure-cluster-identity` exists but can't check for `azure-cluster-identity-secret` existence

zerospiel: That is a good question, there was no such logic beforehand and I've only implemented the nested refs labeling exclusively for the backup feature. I think that is the question for @a13x5 in this context, whether it might be/might not be a good idea to check all the references, not only the first level (e.g. AWS*Identities could represent several levels of references nesting)

s3rj1k: not for this PR but we can extend `Credential` object itself to have a list of referenced object, like all nested that would be flattened out and check them

a13x5: @zerospiel I understood the issue, but can we get object only partially without hitting RBAC issue? 
Because what we need here is to check if referenced object exists and prohibit cluster creation if it's not. Because later we will try to pass non-existing reference to the cluster and it will fail anyway, but in a background.

Also we can't check this reference further (like we wanted initially in the credential epic), since CAPI clusteridentity system is a total mess and some providers just use Secret and ClusterIdentity objects themselves aren't not standardized.

zerospiel: @a13x5 I believe we could check it further just like I implemented it for labeling nested references (check it out in [the `main` branch](https://github.com/k0rdent/kcm/blob/3ff082a2d84f96cec99634c60027285ba0ffadff/internal/controller/credential_controller.go#L125)). Regarding the partial checking, are your implying to check the `identityRef.Kind` and then just decide whether it's one of the suitable (namely well-known) kinds, then try to `get` it otherwise skip? I think it's just partial decision though it satisfies most of the cases IMO

s3rj1k: I've partially reverted change, so it should check for 1st level object existence without labeling it

zerospiel (approved):

internal/controller/credential_controller.go
a13x5: @zerospiel The issue with the ClusterIdentity is that the object can be pretty much anything (even just a Secret), because this is poorly standardized. 
That's why it was decided to offload these objects creation to the user. So user creates whatever ClusterIdentity it needs for provider it uses. 
In terms of the backup we can do exactly the same - just instruct the user to add the backup label (I assume the labeling thing is about that) in the credential docs.

From my point of view current state of this code is totally fine.

s3rj1k:
Tested manually on 
- AWS (AKS)
- Azure
- vSphere
- OS
seems to work, also tested Velero backup, seems to do backups, Restore failed for me but I don't think it's related to this PR

PS: We really need e2e tests operational, things get more and more complicated to test all that manually.

s3rj1k:
I've tested Backup restore against AWS, all seems to work fine

zerospiel (approved):

eromanova (approved):

a13x5 (approved):
LGTM
Tests are red because of issue in the tests themselves. Merging

PR #931, opened 2025-01-21
Support clusterLabels in OpenStack

wkonitzer:
Minor change to allow clusterLabels in the OpenStack templates, which will allow MultiClusterService templates to deploy, unblocking demo #4 in https://github.com/k0rdent/demos


a13x5 (commented):
@wkonitzer Please bump version in the `Chart.yaml` then run `make generate-all` and commit regenerated templates.

wkonitzer:
> @wkonitzer Please bump version in the `Chart.yaml` then run `make generate-all` and commit regenerated templates.

OK, hopefully all done!

bnallapeta (commented):
Looks good. But please update the values.schema.json file with the new field. 

Additionally, 

- you could provide sample values in config/dev/openstack-clusterdeployment.yaml if you wish.
- update the docs with this additional field - https://github.com/k0rdent/docs

a13x5:
@wkonitzer I see that @bnallapeta had a good point regarding `values.schema.json`
Could you please update that as well for consistency with other charts.
Docs are more like nice to have IMO

wkonitzer:
@bnallapeta and @a13x5 I've made the changes suggested, however please note https://github.com/k0rdent/kcm/issues/945 so I might have done it wrong. If so please let me know what to put and I'll correct.

bnallapeta (approved):

bnallapeta:
@wkonitzer Approved. But the unit tests failed. Please check.

wkonitzer:
@bnallapeta I fixed my part but the tests are failing with "template aws-hosted-cp-0-0-4 is still invalid"

zerospiel:
@wkonitzer there is a need to bump everything one more time https://github.com/k0rdent/kcm/pull/931#pullrequestreview-2565805427 because the template had already been bumped to 0-0-4

wkonitzer:
@zerospiel OK, I've bumped it all again - hopefully it's now good :)

config/dev/openstack-clusterdeployment.yaml
```diff
@@ -4,9 +4,11 @@ metadata:
   name: openstack-dev
   namespace: ${NAMESPACE}
 spec:
-  template: openstack-standalone-cp-0-0-3
+  template: openstack-standalone-cp-0-0-5
   credential: openstack-cluster-identity-cred
   config:
+    clusterLabels:
+      k0rdent: demo
```
zerospiel, line 11: do we need it? i mean, there is already the scheme and the chart itself, so I believe we already have enough (apart from the docs)

wkonitzer: @bnallapeta suggested to add it above as an example.

zerospiel: we could leave it here but empty, so both cases are covered: we have an example, and no undesired labels applied by default

to sum up

```yaml
spec:
  # ...
  config:
    clusterLabels: {}
  # ...
```

wkonitzer: @zerospiel seems fair enough-  let's do that. Changes made.

zerospiel (approved):

Kshatrix (approved):

PR #944, opened 2025-01-22
Load provider definition dynamically

s3rj1k:
Adds ability to provide `provider` configuration from loadable YAML file.


a13x5:
I would prefer to put provider yamls into ConfigMap instead of baking it into the image. 
CM could be added to the kcm chart along with changes to deployment to mount it.

s3rj1k:
> I would prefer to put provider yamls into ConfigMap instead of baking it into the image.

True, but two issues with that:

- Unit-tests require this to be loaded during testing
- List has to be loaded during early manager start [here](https://github.com/k0rdent/kcm/blob/main/internal/controller/release_controller.go#L204), and `ensureManagement` is only run once if I get this correctly from [here](https://github.com/k0rdent/kcm/blob/main/internal/controller/release_controller.go#L105-L111)

s3rj1k:
@zerospiel Please also take a look at this

s3rj1k:
> CM could be added to the kcm chart along with changes to deployment to mount it.

Wait, maybe I misunderstood, just mounting CMs into image? If so, yea that should work, this is was the end-goal actually

a13x5:
> Wait, maybe I misunderstood, just mounting CMs (that have inline YAMLs in them) into image?

Yes, I meant just move these yamls to CM and mount this CM as a volume in Deployment. 
Controller shouldn't get this CM via API it could just read the files in the same way it does now.

Makefile
```diff
@@ -258,6 +262,8 @@ kcm-deploy: helm
 
 .PHONY: dev-deploy
 dev-deploy: ## Deploy KCM helm chart to the K8s cluster specified in ~/.kube/config.
+	@mkdir -p $(PROVIDER_TEMPLATES_DIR)/kcm/files/providers
+	@cp -a providers/*.yml $(PROVIDER_TEMPLATES_DIR)/kcm/files/providers/
```
eromanova, line 266: Don't we need to do that before the kcm template generation instead? Right now, those changes will be effective on `dev-apply` only but what about the released version of kcm?

Dockerfile
```diff
@@ -30,6 +30,7 @@ RUN go mod download
 COPY cmd/main.go cmd/main.go
 COPY api/ api/
 COPY internal/ internal/
+COPY providers/ providers/
```
eromanova, line 33: Is it still needed? As I see, the image does not contain the providers anymore, it's mounted as a configmap now.

s3rj1k: Yea, needed to run build and units, this is for 1st stage of docker file

Makefile
s3rj1k: Can you suggest a correct place to put those commands in? (Make Target)
I had a hard time finding this one.

s3rj1k:
related: https://github.com/k0rdent/docs/pull/98

eromanova (commented):
LGTM. @a13x5 could you please also take a final look and approve the PR if no objections?

internal/providers/yaml.go
```diff
@@ -0,0 +1,91 @@
...
+	"os"
+	"path/filepath"
+	"slices"
+
+	"gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GVK represents the GroupVersionKind structure in YAML.
+type GVK struct {
```
zerospiel, line 28: nit: these types and functions (not methods ofc) might be unexported, I know this is the `internal` but still do not understand why these should be exposed if there are no callers outside the package

config/dev/aks-credentials.yaml
```diff
@@ -25,3 +25,13 @@ spec:
     kind: Secret
     name: azure-aks-credential
     namespace: ${NAMESPACE}
+---
+apiVersion: v1
+kind: ConfigMap
```
zerospiel, line 30: i do not quite understand the purpose of this object along with the similar in the `config/dev/aws-credentials.yaml`, could you elaborate on this?

internal/projectroot/path.go
```diff
@@ -12,28 +12,16 @@
...
 
 import (
-	"k8s.io/apimachinery/pkg/runtime/schema"
+	"path/filepath"
+	"runtime"
 )
 
-type ProviderAzure struct{}
+var (
+	_, b, _, _ = runtime.Caller(0)
```
zerospiel, line 23: this is not a stable solution (it works and I am not aware of another approach) in terms of the file location, so i'd explicitly comment on the relative file path in which this particular file should be placed, namely `internal/projectroot/path.go` and it is not recommended to move the file to another path

zerospiel (commented):
i have a couple questions, though looks good, I also ask @eromanova and @a13x5 to look on this once again just to ensure everything is also aligned to them

config/dev/aks-credentials.yaml
zerospiel: discussed personally

internal/providers/yaml.go
zerospiel: personal preference

zerospiel (approved):

zerospiel (approved):

s3rj1k:
related: https://github.com/k0rdent/docs/pull/101

Makefile
```diff
@@ -335,6 +335,12 @@ dev-release:
 
 .PHONY: dev-adopted-creds
 dev-adopted-creds: envsubst
+	@if [ -z "$$KUBECONFIG_DATA" ]; then \
+		echo "Error: KUBECONFIG_DATA environment variable is not set"; \
+		echo "Please set its value to Base64 encoded kubeconfig file:"; \
+		echo "  export KUBECONFIG_DATA=\$$(cat kubeconfig | base64 -w 0)"; \
+		exit 1; \
+	fi
 	@NAMESPACE=$(NAMESPACE) $(ENVSUBST) -i config/dev/adopted-credentials.yaml | $(KUBECTL) apply -f -
```
eromanova, line 344: Let's pass the `-no-unset` parameter to the `envsubst` command and remove the `KUBECONFIG_DATA` check above.

s3rj1k:
```
### Adopted Cluster Setup

To "adopt" an existing cluster first obtain the kubeconfig file for the cluster.
Then set the `DEV_PROVIDER` to "adopted". Export the kubeconfig file as a variable by running the following:

`export KUBECONFIG_DATA=$(cat kubeconfig | base64 -w 0)`

The rest of the deployment procedure is same for all providers.
```

Seems that this is on devdoc, so check can indeed be removed

eromanova (approved):

Issue #916, opened 2025-01-17
[bug] Sveltos CRDs are not updated between releases


Kshatrix:
Sveltos chart keeps all sveltos CRDs in a /crd directory. Once installed, they are no longer managed, so any updates to these CRDs won't be delivered with a new kcm release.

PR #949, opened 2025-01-23
move projectsveltos crds into kcm chart

BROngineer:
Fixes: #916


zerospiel (approved):
lgtm someone else should review

templates/provider/kcm-templates/files/release.yaml
```diff
@@ -21,5 +21,7 @@ spec:
       template: cluster-api-provider-aws-0-0-5
     - name: cluster-api-provider-openstack
       template: cluster-api-provider-openstack-0-0-2
+    - name: projectsveltos-crds
```
Kshatrix, line 24: would it make sense to add it to the kcm chart instead of a separate "provider"?

BROngineer: Thought about it and I'm not sure. Maybe it would be better to place both of them - projectsveltos and projectsveltos-crds - into kcm chart (tbh, I didn't check whether there are anything depending on projectsveltos provider). But I do not see any advantages in placing them separately.

BROngineer: Checked - actually it do.

Kshatrix: we can keep sveltos crds in the kcm chart and disable crds creation in the sveltos chart then as an option

BROngineer: @Kshatrix using separate chart would also allow more accurate versioning: in this pr there is "main" projectsveltos chart with version v0.45.0 and crds chart with the same version. Thus it will be easy to track whether crds version matches main chart. Moving crds as regular files under kcm chart would eliminate that possibility.
In summary: I'd prefer to keep crds as a chart and as close to main projectsveltos chart as possible.

Kshatrix: sveltos is not a capi provider. the only reason it's in the list of providers is the requirement of a hardcoded namespace. when/if the problem resolves, it makes sense to move it to the main kcm chart as well. having sveltos-crds as a "provider" makes even less sense imo, but overall it's up to you

BROngineer: > the only reason it's in the list of providers is the requirement of a hardcoded namespace. when/if the problem resolves, it makes sense to move it to the main kcm chart as well

@Kshatrix Fair enough.

templates/provider/kcm/templates/projectsveltos-crds.yaml
```diff
@@ -0,0 +1,9 @@
+{{- range $path, $_ := .Files.Glob "projectsveltos-crds/*.yaml" }}
```
Kshatrix, line 1: please add an option to disable sveltos crds creation

Kshatrix:
We'd also need to keep those CRDs in sync with our currently used sveltos version (have a make target for that). could be added later though.

Kshatrix (approved):

PR #957, opened 2025-01-24
Use remote "k0rdent/catalog" service templates

Josca:


Makefile
```diff
@@ -266,6 +270,7 @@ dev-deploy: ## Deploy KCM helm chart to the K8s cluster specified in ~/.kube/con
 		$(YQ) eval -i '.controller.defaultRegistryURL = "$(REGISTRY_REPO)"' config/dev/kcm_values.yaml; \
 	fi; \
 	$(MAKE) kcm-deploy KCM_VALUES=config/dev/kcm_values.yaml
+	$(MAKE) k0rdent-core-deploy
```
eromanova, line 273: If we deploy `k0rdent-core` this way, it will only be installed when we deploy the dev cluster using `make dev-deploy` or `make dev-apply` (which requires repository cloning, etc.). This is not the desired behavior.

We need to ensure the `k0rdent-core` Helm chart is installed as part of the released version of the KCM.

This component should be installed in a similar manner to other ProviderTemplates (handled by the management controller). The `k0rdent-core` component must be defined in the `Release` and `Management` objects, with their APIs extended.

eromanova (changes requested):

Makefile
eromanova: Nevermind. There is a better option. Discussed in Slask.

zerospiel:
@Josca is the PR still relevant having the #964? otherwise, please close this PR

Issue #906, opened 2025-01-16
[bug] template-controller sometimes prevents proper management cleanup


Kshatrix:
It should be aware of the ongoing cleanup and stop any reconciles.

PR #961, opened 2025-01-27
Stop reconciles when Management is deleting

Kshatrix:
Resolves: #906


internal/controller/template_controller.go
```diff
@@ -359,19 +386,22 @@ func (r *TemplateReconciler) getHelmChartFromChartRef(ctx context.Context, chart
...
 }
 
-func (r *ClusterTemplateReconciler) validateCompatibilityAttrs(ctx context.Context, template *kcm.ClusterTemplate) error {
-	management := new(kcm.Management)
-	if err := r.Client.Get(ctx, client.ObjectKey{Name: kcm.ManagementName}, management); err != nil {
+func (r *TemplateReconciler) getManagement(ctx context.Context, template templateCommon) (*kcm.Management, error) {
+	management := &kcm.Management{}
+	if err := r.Get(ctx, client.ObjectKey{Name: kcm.ManagementName}, management); err != nil {
 		if apierrors.IsNotFound(err) {
 			_ = r.updateStatus(ctx, template, "Waiting for Management creation to complete validation")
```
eromanova, line 393: I think now it would be better to move the template status update outside of the `getManagement` function, but it’s up to you.

internal/controller/accessmanagement_controller.go
```diff
@@ -42,6 +42,15 @@ func (r *AccessManagementReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	l := ctrl.LoggerFrom(ctx)
 	l.Info("Reconciling AccessManagement")
 
+	management := &kcm.Management{}
+	if err := r.Get(ctx, client.ObjectKey{Name: kcm.ManagementName}, management); err != nil {
```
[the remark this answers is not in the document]
Kshatrix: It's either mgmt is not yet created or already removed and in the former case we should not skip reconciles. It makes sense to tune backoff though, but as a separate task

internal/controller/backup/periodic.go
```diff
@@ -114,6 +114,14 @@ var errEmptyList = errors.New("no items available to enqueue")
...
 // either are schedules or are not yet completed.
 func (r *Runner) enqueueSchedulesOrIncompleteBackups(ctx context.Context) error {
+	management := &kcmv1alpha1.Management{}
+	if err := r.cl.Get(ctx, client.ObjectKey{Name: kcmv1alpha1.ManagementName}, management); err != nil {
+		return fmt.Errorf("failed to get Management: %w", err)
+	}
+	if !management.DeletionTimestamp.IsZero() {
+		return nil
+	}
+
```
zerospiel, line 124: there is inconsistency: this is a periodic runner, so "reconciles" continues execution each minute, so either add the same check in the Reconciler or remove this one

internal/controller/template_controller.go
eromanova: Discussed on slack. It’s not needed since the function is internal to the template reconciler, and updating the status within it is acceptable.

internal/controller/backup/periodic.go
zerospiel: missed the `internal/controller/management_backup_controller.go`

zerospiel (approved):

eromanova (approved):

PR #967, opened 2025-01-28
Expose configuration options for Openstack volumes in the template

### LINKED ISSUE #965: [enhancement] Expose configuration options for Openstack volumes in the template
**Description**
Currently CAPO capabilities on using Boot from Volume feature are not exposed in the ClusterTemplate while some user may want/require this feature.

**Expected behaviour**
Boot from volume is not enabled by default but end user can configure it via ClusterDeployment object for either controlPlane or worker or both. Same should be possible for bastion.

Link to the CAPO documentation regarding the requested feature:
https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration.html#boot-from-volume

MirgDenis:
Expose rootVolume option in Openstack standalone control plane cluster template.
Resolves #965

### DIFF
--- a/config/dev/openstack-clusterdeployment.yaml
+++ b/config/dev/openstack-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: openstack-dev
   namespace: ${NAMESPACE}
 spec:
-  template: openstack-standalone-cp-0-0-5
+  template: openstack-standalone-cp-0-0-6
   credential: openstack-cluster-identity-cred
   config:
     clusterLabels: {}
--- a/templates/cluster/openstack-standalone-cp/Chart.yaml
+++ b/templates/cluster/openstack-standalone-cp/Chart.yaml
@@ -6,7 +6,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.0.5
+version: 0.0.6
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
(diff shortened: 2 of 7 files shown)


zerospiel (commented):
lgtm, please bump the version in the `config/dev/openstack-clusterdeployment.yaml` to the corresponding one

MirgDenis:
JFYI this patch solves https://github.com/k0rdent/kcm/issues/965
Unfortunately I could not find the way to link this PR to issue

wkonitzer:
@MirgDenis do you want to add example config to 

‎config/dev/openstack-clusterdeployment.yaml ?

Also, you might wish to update the k0rdent docs with the new configuration and add it to the OpenStack templates in the demo repo, perhaps as an additional demo or a template on Demo1?

(In the "description" for your PR, you can add "Fixes: #965") and that should link it.

MirgDenis:
@wkonitzer I think it's better to examples as simple as possible. For sure we will have more options added, maybe even all keys from OpenStackMachine will be added in template, so I think no point in one more example.

Commit in doc is good idea, I will prepare a patch.

wkonitzer:
@MirgDenis I disagree - I want to see an example of it somewhere so I'm not having to try and figure it out from all the docs, templates etc. I already have people interested in this update, so I'd like to make it as simple as possible for them (as non-coders), to get going.

MirgDenis:
@wkonitzer undertsood. But in my opinion there are no place for for such an example. config/dev has only one simplest scenario, and docs have only API ref.
There should be some examples directory with list of optional scenario like:

- openstack-cluster-filtered-image.yaml
- openstack-cluster-port-options.yaml
- openstack-cluster-sec-groups.yaml
- openstack-cluster-ssh.yaml
- openstack-cluster-root-volume.yaml

And also examples can be grouped by control/worker attribute. I think such folder is out of scope of this PR.

bnallapeta (approved):
Looks good. Please squash the commits.

a13x5 (commented):
You should put new values into `values.yaml` with an empty/default value.  This is convention we're using for other charts, so please add it for consistency.

eromanova (approved):

PR #968, opened 2025-01-28
Fix Iss945 - clusterLabels definition in values.schema.json missing from templates

wkonitzer:
Resolves Issue 945 - clusterLabels definition in values.schema.json missing from templates.
Also, make usage of clusterLabels consistent across providers and dev templates.


wkonitzer:
Going to recreate PR with clean history,

PR #969, opened 2025-01-28
Resolves Issue 945 - clusterLabels definition in values.schema.json missing

### LINKED ISSUE #945: [bug] clusterLabels definition in values.schema.json missing from templates
In the values.schema.json for cluster templates, e.g. AWS (https://github.com/k0rdent/kcm/blob/main/templates/cluster/aws-standalone-cp/values.schema.json) there is no definition for clusterLabels, but it is defined/used in the values.yaml.

We should update the charts to make them consistent internally.

wkonitzer:
Fixes #945 - clusterLabels definition in values.schema.json missing from templates.

This PR also ensures consistent usage of clusterLabels across providers and development templates.


templates/cluster/aws-eks/values.schema.json
Kshatrix, line 1: jsonschema update would also require a chart version bump

Kshatrix (changes requested):

templates/cluster/aws-eks/values.schema.json
wkonitzer: @Kshatrix is that just for aws-eks (chart version bump), or should I do it for all providers? Essentially I was just correcting something that was missing but already present in the values.yaml (so was working).

wkonitzer: @Kshatrix, OK, I bumped the chart version for EKS as you requested above. I also did the same for all other providers to keep them aligned.

Kshatrix (approved):

PR #971, opened 2025-01-28
Extend providers interface to support multiple cluster GVKs

s3rj1k:
PR extend providers interface to support multiple cluster GVKs and improves Dev-experience via Makefile changes.

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -649,12 +649,12 @@ func (r *ClusterDeploymentReconciler) releaseCluster(ctx context.Context, namesp
 
 	// Associate the provider with it's GVK
 	for _, provider := range providers {
-		gvk := providersloader.GetClusterGVK(provider)
-		if !gvk.Empty() {
+		gvks := providersloader.GetClusterGVKs(provider)
+		if len(gvks) == 0 {
 			continue
 		}
 
-		cluster, err := r.getCluster(ctx, namespace, name, gvk)
+		cluster, err := r.getCluster(ctx, namespace, name, gvks...)
 		if err != nil {
 			if provider == "aws" && apierrors.IsNotFound(err) {
 				return nil
@@ -665,7 +665,7 @@ func (r *ClusterDeploymentReconciler) releaseCluster(ctx context.Context, namesp
 
 		found, err := r.objectsAvailable(ctx, namespace, cluster.Name, gvkMachine)
 		if err != nil {
-			return err
+			continue
 		}
 
 		if !found {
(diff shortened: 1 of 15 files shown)


internal/controller/clusterdeployment_controller.go
```diff
@@ -697,21 +697,38 @@ func (r *ClusterDeploymentReconciler) getInfraProvidersNames(ctx context.Context
...
 	}
 
-	return &itemsList.Items[0], nil
+	switch len(foundClusters) {
+	case 0:
+		return nil, fmt.Errorf("no cluster found with name %s in namespace %s for any of the provided GroupVersionKinds", name, namespace)
+	case 1:
+		return foundClusters[0], nil
+	default:
+		return nil, fmt.Errorf("found multiple clusters with name %s in namespace %s across different GroupVersionKinds", name, namespace)
```
eromanova, line 730: I'm not sure we should exit with an error in >1 clusters found. (In theory, there can be >1 cluster declared in one cluster template). Currently, this function is only used to block the removal of cluster deployment if any children object exists. This function should be refactored to either loop through all the GVKs, get all the clusters of each kind, and block the cluster deployment removal if any of the child objects still exist. But it's not in the scope of this task.

For now, let's get the first one found cluster across all GVKs and exit fast with it. And exit with an error if no clusters were found.

s3rj1k: 2 clusters with same name but different ClusterGVK?

eromanova: Yes, it sounds weird, but a template can technically define two clusters with different GVK (ex, AWS+EKS). While our templates don’t, some BYO templates might. I’d prefer to ensure the controller handles this scenario gracefully rather than failing with an error.

internal/providers/yaml.go
```diff
@@ -24,18 +24,11 @@ import (
...
-	Kind    string `yaml:"kind"`
-}
-
 // YAMLProviderDefinition represents a YAML-based provider configuration.
 type YAMLProviderDefinition struct {
-	Name                 string   `yaml:"name"`
-	ClusterGVK           GVK      `yaml:"clusterGVK"`
-	ClusterIdentityKinds []string `yaml:"clusterIdentityKinds"`
+	Name                 string                    `yaml:"name"`
+	ClusterGVK           []schema.GroupVersionKind `yaml:"clusterGVKs"`
```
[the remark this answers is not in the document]
s3rj1k: missed that, thanks

internal/controller/clusterdeployment_controller.go
s3rj1k: label selector, hmph, ok

s3rj1k: something like that?

eromanova (approved):

internal/controller/clusterdeployment_controller.go
eromanova: Yes, thank you.

Issue #956, opened 2025-01-24
adjust kcm repo content after service templates migration to catalog


DinaBelova:
relates to https://github.com/k0rdent/catalog/issues/1

Josca:
https://github.com/k0rdent/kcm/pull/957

zerospiel:
Completed by the https://github.com/k0rdent/kcm/pull/982

PR #982, opened 2025-01-29
Add "catalog-core" provider to reference core servicetemplates

Josca:
k0rdent/kcm#956, k0rdent/catalog#1

Service templates were moved to a separated [k0rdent/catalog](https://github.com/k0rdent/catalog) repo.
This PR removes original service templates and adding them as kcm-templates dependency.


templates/provider/kcm-templates/files/templates/catalog-core.yaml
```diff
@@ -1,14 +1,14 @@
 apiVersion: k0rdent.mirantis.com/v1alpha1
-kind: ServiceTemplate
+kind: ProviderTemplate
 metadata:
-  name: dex-0-19-1
+  name: catalog-core-1-0-0
```
Kshatrix, line 4: I don't think ProviderTempalte is an appropriate place for that. If we want to keep kcm and catalog releases aligned, the easiest way for now is to make it a subchart of the kcm-templates chart

Kshatrix (changes requested):

templates/provider/kcm-templates/files/templates/catalog-core.yaml
Josca: Ok, I will update that. I did that on @a13x5 's suggestion to make that part of release object.

Josca: Updated

zerospiel (approved):

a13x5 (approved):
LGTM

Kshatrix (approved):

Issue #751, opened 2024-12-10
backups e2e tests framework development


zerospiel:
implement e2e testing covering at least the following cases:
1. backup a cluster, create a clean one, restore a cluster, ensure both are identical, remove the cloud data
2. upgrade case: initiate the upgrade, ensure a backup has been created automatically, wait for the upgrade process to end, restore the cluster, and ensure the version of the cluster equals the pre-upgrade version

zerospiel:
Should be implemented as part of 3d party system integration task

PR #1007, opened 2025-01-31
Integration (envtest) tests for backups

zerospiel:
* add basic envtest for the backup ctrl
* add tests for the periodic runner


eromanova (commented):
LGTM. 1 small comment.

eromanova (approved):

moshiurH:
It looks like the [release.yaml](https://github.com/k0rdent/kcm/blob/main/.github/workflows/release.yml) should run a basic upgrade and rollback test. I am not sure if this is already implemented.

zerospiel:
> how are these integration tests meant to be run?

these are tests implemented to be run in the `envtest`, basically they are unit tests but because there is another system (`envtest` having its own etcd and basic k8s) many call such tests "integration". hence the clarification in the title. just run `make test` target

> Are they to be run by some sort of CI/CD like Github Actions?

correct, they run each time a commit is pushed

> I am not sure if this is already implemented.

It has not, there is the issue https://github.com/k0rdent/kcm/issues/751 describing the minimal set of tests to run, the implementation should be some kind of a dedicated framework to perform such a complicated scenario, including the upgrade via delivering a "new" one (a fake one, or decrease a version in the `release` object preliminary and then upgrade to the current one, it does not matter)

Issue #900, opened 2025-01-15
[bug] Kubectl apply installation failed on leader election


MirgDenis:
**Describe the bug**
Installation using rendered helm template, i.e. kubectl apply <URL>/install.yaml failed on leader election for cert-manager
```
E0115 15:01:00.255546       1 leaderelection.go:436] error retrieving resource lock kube-system/cert-manager-controller: leases.coordination.k8s.io "cert-manager-controller" is forbidden: User "system:serviceaccount:kcm-system:kcm-cert-manager" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "kube-system"
```
and cainjector
```
E0115 15:01:13.590420       1 leaderelection.go:436] error retrieving resource lock kube-system/cert-manager-cainjector-leader-election: leases.coordination.k8s.io "cert-manager-cainjector-leader-election" is forbidden: User "system:serviceaccount:kcm-system:kcm-cert-manager-cainjector" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "kube-system"
```

Further investigation showed that rendered install.yaml has "leftovers" from chart defaults
```
$ grep -r kube-system install.yaml -C3
          imagePullPolicy: IfNotPresent
          args:
            - --v=2
            - --leader-election-namespace=kube-system
          ports:
            - containerPort: 9402
              name: http-metrics
--
          args:
            - --v=2
            - --cluster-resource-namespace=$(POD_NAMESPACE)
            - --leader-election-namespace=kube-system
            - --acme-http01-solver-image=quay.io/jetstack/cert-manager-acmesolver:v1.16.2
            - --max-concurrent-challenges=60
          ports:
``` 

**To Reproduce**
1. Create kind cluster
2. Run `kubectl apply -f https://github.com/K0rdent/kcm/releases/download/v0.0.7/install.yaml`
3. Check `kcm-cert-manager-startupapicheck` job (it will be in Failed state), check kcm-controller-manager (it will be in CrashLoopBackOff state), checkcainjector and cert-manager logs (they will have RBAC related errors)

**Expected behavior**
I would expect that KCM to be install without CrashLoopBack for kcm-contoller-manager and cert-manager and cainjector did not have RBAC related errors in its logs.

**Additional context**
One of possible solutions is to add global.leaderElection.namespace in cert-manager and cainjector charts values, however I'm not sure how to handle "templatization". If we add it to the values, then namespace will be hardcoded, however in Makefile it's expected that it use Release.namespace
```
[...]

PR #1014, opened 2025-01-31
Remove install.yaml usage from README

Kshatrix:


MirgDenis:
@Kshatrix If you do not plan to add it back, then you can consider closing this ticket https://github.com/k0rdent/kcm/issues/900

PR #1034, opened 2025-02-04
Support for providing a testing configuration to the workflow

### LINKED ISSUE #642: Rework providers e2e testing trigger
Currently, our provider E2E testing is triggered by the `test e2e` PR label. However, there is no straightforward way to provide overrides for the testing configuration. There is a need to test more complex configurations, such as deploying two clusters with different templates on a single management cluster, testing the deployment of the N-1 template, and upgrading to the N template, among others.

**UPD:**

**NEW PROPOSAL:**
The initial proposal is slightly outdated, as we now use an `authorize` job that requires approval from authoritative reviewers. When approving a deployment, reviewers can leave a comment to provide a custom testing configuration. If no configuration is specified, the default testing configuration will be used.

**INITIAL PROPOSAL:**
I suggest the following changes:

Trigger provider E2E testing when the PR was commented starting with the `/test-e2e` trigger word followed by an optional testing configuration in a YAML format. I’m also considering combining both conditions: to trigger the provider testing on a PR, it should contain both the `test e2e` label and a comment containing `/test-e2e`. Since labels on a PR can only be added by our team (afaik, but I may be mistaken), this will help avoid unnecessary test runs triggered by someone else, not part of our team. We can keep triggering provider e2e testing on the existence of the `test e2e` label if there is no need to provide the testing config.

Other jobs: `build` and `controller-e2etest` may be triggered on each PR regardless of the `test e2e` label or `/test-e2e` comment present. This task is only about the adapting the github build workflow. Testing configuration support will be implemented in #641.

eromanova:
This PR updates the `authorize` job to retrieve the testing configuration from the approval comment. The comment should be provided during the approval process for deployment in the `external` environment.

* The comment must contain valid YAML content representing the testing configuration.
* If the comment is empty, the default testing configuration will be used, which deploys a single AWS standalone cluster.
* If the comment contains invalid YAML, an error will occur and the job will fail.

Below is an example that demonstrates how this process should work:

https://github.com/user-attachments/assets/84971961-4dcc-4912-aa22-12ad030d5254

Closes #642

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -28,8 +28,26 @@ jobs:
       github.event.pull_request.head.repo.full_name != github.repository) &&
       'external' || 'internal' }}
     runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ github.token }}
+    outputs:
+      config: ${{ steps.get_config.outputs.config }}
     steps:
-      - run: echo ✓
+      - name: Get Testing Configuration
+        id: get_config
+        run: |
+          config=$(gh api "/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/approvals" --jq '.[0] | .comment')
+          if [ -z "$config" ]; then
+            exit 0
+          fi
+
+          echo "Provided testing configuration:"
+          echo "$config"
+          if ! echo "$config" | yq e > /dev/null; then
+            echo "Invalid testing configuration provided"
+            exit 1
+          fi
+          echo "config=$(echo -n "$config" | base64)" >> $GITHUB_OUTPUT
   build:
     concurrency:
       group: build-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


zerospiel (approved):
LGTM just one subtle thing to clarify for myself:
so after filing a PR, one should leave a comment (extra comment, not the PR description itself, right?) with a valid YAML test configuration? can the comment contain formatting like (backslashes are to avoid escape graves)
```
\```yaml
code
\```
```

or should the commit omit any possible formatting?

eromanova:
Thanks for the review, @zerospiel. Let me clarify a bit.
The comment should be provided during the approval process for deployment in the external environment (not a comment on a pull request). The content of the first approval comment will be used as the testing configuration. I’ll create a document describing this process in more detail.

Here is a screen recording that demonstrates how this process should work (same `authorize` workflow from my personal project).

https://github.com/user-attachments/assets/d9074033-d023-42c5-a0c3-d97e888c298e

Issue #875, opened 2025-01-09
[bug] multiple services included into `ClusterDeployment` are deployed until first failed deployment only


BROngineer:
**Describe the bug**
When multiple service templates are defined in `ClusterDeployment` resource, they are deployed sequentially until the first failure occurs. Then charts following failed one will never be deployed.

REF issue in projectsveltos: https://github.com/projectsveltos/addon-controller/issues/936

**To Reproduce**
1. Apply `ClusterDeployment` with multiple services included. Add service that will fail in the middle of the list. For instance, cert-manager which will not run properly without CRDs enabled.
2. Observe services deployed into managed cluster: only those services that come before the cert manager in the template list (including the cert manager) will be deployed

<details>
<summary>ClusterDeployment example</summary>

   ```yaml
   apiVersion: hmc.mirantis.com/v1alpha1
   kind: ClusterDeployment
   metadata:
     name: foo
   spec:
     ...
     serviceSpec:
       services:
       - name: managed-external-secrets  # will run normally
         namespace: external-secrets
         template: external-secrets-0-11-0
       - name: managed-cert-manager  # will fail to run normally
         namespace: cert-manager
         template: cert-manager-1-16-2
       - name: managed-ingress-nginx  # will not be deployed
         namespace: ingress-nginx
         template: ingress-nginx-4-11-3
     servicesPriority: 100
     stopOnConflict: false
   ```

</details>

**Expected behavior**
A deployment attempt must be made against all services defined in the `ClusterDeployment` resource.

BROngineer:
Flag `ContinueOnError` added in projectsveltos v0.46.0 release, which in general solves the issue. The default value for flag equals to "false".
Need to expose the flag to k0rdent resources.

PR #1038, opened 2025-02-04
expose ContinueOnError flag for services spec

BROngineer:
Fixes: #875

### DIFF
--- a/api/v1alpha1/multiclusterservice_types.go
+++ b/api/v1alpha1/multiclusterservice_types.go
@@ -105,6 +105,7 @@ type ServiceSpec struct {
 	DriftExclusions []sveltosv1beta1.DriftExclusion `json:"driftExclusions,omitempty"`
 
 	// +kubebuilder:default:=false
+
 	// ContinueOnError specifies if the services deployment should continue if an error occurs.
 	ContinueOnError bool `json:"continueOnError,omitempty"`
 }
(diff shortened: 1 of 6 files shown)


eromanova (approved):

api/v1alpha1/multiclusterservice_types.go
```diff
@@ -103,6 +103,10 @@ type ServiceSpec struct {
 	DriftIgnore []libsveltosv1beta1.PatchSelector `json:"driftIgnore,omitempty"`
 	// DriftExclusions specifies specific configurations of resources to ignore for drift detection.
 	DriftExclusions []sveltosv1beta1.DriftExclusion `json:"driftExclusions,omitempty"`
+
+	// +kubebuilder:default:=false
+	// ContinueOnError specifies if the services deployment should continue if an error occurs.
```
zerospiel, line 108:
```suggestion
	// +kubebuilder:default:=false

	// ContinueOnError specifies if the services deployment should continue if an error occurs.
```

formatting to avoid appearing as part of the comment. Why do we need the explicit default? I recall asking a similar question, and IIRC, the answer was similar to "to ensure more clarity" (or smth), is there the same intention?

BROngineer: Correct. Will reformat.

zerospiel (approved):

PR #1041, opened 2025-02-04
pulling projectsveltos CRDs of version defined in Chart.yaml

BROngineer:
Fixes: #1040


Makefile
```diff
@@ -91,7 +91,16 @@ templates-generate:
 	@hack/templates.sh
 
 .PHONY: generate-all
-generate-all: generate manifests templates-generate add-license
+generate-all: generate manifests templates-generate add-license projectsveltos-crds
+
+.PHONY: projectsveltos-crds
+projectsveltos-crds: yq
```
eromanova, line 97: Makefile already has `$(SVELTOS_CRD)` target to fetch sveltos CRDs for testing that does almost the same code but it places the CRDs in a different directory. Can we reuse the existing target?

eromanova, line 97:
```suggestion
projectsveltos-crds: $(YQ)
```

Makefile
```diff
@@ -91,7 +91,16 @@ templates-generate:
...
+	@$(eval PROJECTSVELTOS_RELEASE=$(shell $(YQ) '.version' < $(PROVIDER_TEMPLATES_DIR)/projectsveltos/Chart.yaml))
+	@echo "Pulling CRDs from projectsveltos release $(PROJECTSVELTOS_RELEASE)"
+	@curl -sL https://raw.githubusercontent.com/projectsveltos/sveltos/v$(PROJECTSVELTOS_RELEASE)/manifest/crds/sveltos_crds.yaml -o $(PROVIDER_TEMPLATES_DIR)/kcm/projectsveltos-crds/crds.yaml
+	@sed '$$d' $(PROVIDER_TEMPLATES_DIR)/kcm/projectsveltos-crds/crds.yaml | \
+	yq -s '"$(PROVIDER_TEMPLATES_DIR)/kcm/projectsveltos-crds/" + .metadata.name + ".yaml"'
```
eromanova, line 102:
```suggestion
	$(YQ) -s '"$(PROVIDER_TEMPLATES_DIR)/kcm/projectsveltos-crds/" + .metadata.name + ".yaml"'
```

Makefile
BROngineer: why replace `yq` target with `$(YQ)` var?

BROngineer: sure. I'll update new target to reuse it.

Makefile
```diff
@@ -91,7 +91,11 @@ templates-generate:
 	@hack/templates.sh
 
 .PHONY: generate-all
-generate-all: generate manifests templates-generate add-license
+generate-all: generate manifests templates-generate add-license projectsveltos-crds
+
+.PHONY: projectsveltos-crds
+projectsveltos-crds: sveltos-crds
```
eromanova, line 97: I think it can be simplified
```suggestion
projectsveltos-crds: $(SVELTOS_CRD)
```
and the line 524 can be removed.

BROngineer: The thing is it does not work as you suggested. Maybe local mac issue, but I didn't manage to run target in format `rojectsveltos-crds: $(SVELTOS_CRD)`. Thus I decided to keep as in PR, bc in that format it will work.

eromanova (approved):

PR #1057, opened 2025-02-08
Add Docker provider and cluster template

### LINKED ISSUE #1080: [enhancement] CAPD infrastructure provider
I have problems with testing basic KCM functionality without provider which has low hardware footprint (unlike Openstack) and is completely free (unlike AWS, Azure, vSphere).

I would like to have CAPD provider [1] in KCM, it deploys k8s node as one container in Docker i.e. has lowest hw footprint and is free. 

Alternative solution would be to have providers and cluster templates independent from KCM, so that anyone could bring own provider at any time on their own environment. But I think it's more like epic kind of ticket, and require a lot more work.

**Additional context**
Also I think it would be beneficial for KCM functional tests since it would be possible to run then locally without interacting with non-free cloud providers.

[1] https://github.com/kubernetes-sigs/cluster-api/tree/main/test/infrastructure/docker

MirgDenis:
Closes #1080

### DIFF
--- a/config/dev/docker-credentials.yaml
+++ b/config/dev/docker-credentials.yaml
@@ -1,0 +1,31 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: docker-cluster-secret
+  namespace: ${NAMESPACE}
+  labels:
+    k0rdent.mirantis.com/component: "kcm"
+type: Opaque
+---
+apiVersion: k0rdent.mirantis.com/v1alpha1
+kind: Credential
+metadata:
+  name: docker-stub-credential
+  namespace: ${NAMESPACE}
+spec:
+  description: Docker Credentials
+  identityRef:
+    apiVersion: v1
+    kind: Secret
+    name: docker-cluster-secret
+    namespace: ${NAMESPACE}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: docker-cluster-credential-resource-template
+  namespace: ${NAMESPACE}
+  labels:
+    k0rdent.mirantis.com/component: "kcm"
+  annotations:
+    projectsveltos.io/template: "true"
--- a/providers/docker.yml
+++ b/providers/docker.yml
@@ -1,0 +1,21 @@
+# Copyright 2024
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: docker
+clusterGVKs:
+  - group: infrastructure.cluster.x-k8s.io
+    version: v1beta1
+    kind: DockerCluster
+clusterIdentityKinds:
+  - Secret
--- a/templates/cluster/docker-hosted-cp/Chart.yaml
+++ b/templates/cluster/docker-hosted-cp/Chart.yaml
@@ -1,0 +1,19 @@
+apiVersion: v2
+name: docker-hosted-cp
+description: |
+  A KCM template to deploy a k0s cluster on Docker with hosted control plane.
+type: application
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "v1.32.1+k0s.0"
+annotations:
+  cluster.x-k8s.io/provider: infrastructure-docker, control-plane-k0sproject-k0smotron, bootstrap-k0sproject-k0smotron
+  cluster.x-k8s.io/bootstrap-k0sproject-k0smotron: v1beta1
+  cluster.x-k8s.io/control-plane-k0sproject-k0smotron: v1beta1
+  cluster.x-k8s.io/infrastructure-docker: v1beta1
--- a/templates/cluster/docker-hosted-cp/templates/dockercluster.yaml
+++ b/templates/cluster/docker-hosted-cp/templates/dockercluster.yaml
@@ -1,0 +1,9 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: DockerCluster
+metadata:
+  name: {{ include "cluster.name" . }}
+  annotations:
+    cluster.x-k8s.io/managed-by: k0smotron
+  finalizers:
+  - k0rdent.mirantis.com/cleanup
+spec:
--- a/templates/cluster/docker-hosted-cp/templates/dockermachinetemplate.yaml
+++ b/templates/cluster/docker-hosted-cp/templates/dockermachinetemplate.yaml
@@ -1,0 +1,7 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: DockerMachineTemplate
+metadata:
+  name: {{ include "dockermachinetemplate.name" . }}
+spec:
+  template:
+    spec: {}
--- a/templates/cluster/docker-hosted-cp/templates/machinedeployment.yaml
+++ b/templates/cluster/docker-hosted-cp/templates/machinedeployment.yaml
@@ -1,0 +1,26 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: MachineDeployment
+metadata:
+  name: {{ include "machinedeployment.name" . }}
+spec:
+  clusterName: {{ include "cluster.name" . }}
+  replicas: {{ .Values.workersNumber }}
+  selector:
+    matchLabels:
+      cluster.x-k8s.io/cluster-name: {{ include "cluster.name" . }}
+  template:
+    metadata:
+      labels:
+        cluster.x-k8s.io/cluster-name: {{ include "cluster.name" . }}
+    spec:
+      bootstrap:
+        configRef:
+          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+          kind: K0sWorkerConfigTemplate
+          name: {{ include "k0sworkerconfigtemplate.name" . }}
+      clusterName: {{ include "cluster.name" . }}
+      version: {{ (split "+" .Values.k0s.version)._0 }}
+      infrastructureRef:
+        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+        kind: DockerMachineTemplate
+        name: {{ include "dockermachinetemplate.name" . }}
--- a/templates/cluster/docker-hosted-cp/values.yaml
+++ b/templates/cluster/docker-hosted-cp/values.yaml
@@ -1,0 +1,20 @@
+workersNumber: 1
+
+clusterLabels: {}
+
+clusterNetwork:
+  pods:
+    cidrBlocks:
+    - "192.168.0.0/16"
+  services:
+    cidrBlocks:
+    - "10.128.0.0/12"
+  serviceDomain: "cluster.local"
+
+# K0smotron parameters
+k0smotron:
+  service:
+    type: NodePort
+
+k0s:
+  version: v1.32.1+k0s.0
--- a/templates/provider/cluster-api-provider-docker/Chart.yaml
+++ b/templates/provider/cluster-api-provider-docker/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: cluster-api-provider-docker
+description: A Helm chart for Cluster API provider Docker
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "v1.9.4"
+annotations:
+  cluster.x-k8s.io/provider: infrastructure-docker
+  cluster.x-k8s.io/v1beta1: v1beta1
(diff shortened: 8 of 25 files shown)


templates/cluster/docker-hosted-cp/templates/machinedeployment.yaml
```diff
@@ -0,0 +1,26 @@
...
+      labels:
+        cluster.x-k8s.io/cluster-name: {{ include "cluster.name" . }}
+    spec:
+      bootstrap:
+        configRef:
+          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+          kind: K0sWorkerConfigTemplate
+          name: {{ include "k0sworkerconfigtemplate.name" . }}
+      clusterName: {{ include "cluster.name" . }}
+      version: {{ (split "+" .Values.k0s.version)._0 }}
```
zerospiel, line 22: i am not aware of such a function in `helm` templates and do not completely understand the last part adding `._0` to the string, could you elaborate on this? could the `{{ regexReplaceAll "\\+k0s.+$" .Values.k0s.version "" }}` be used instead as it is in other cluster templates?

templates/provider/kcm-templates/files/release.yaml
```diff
@@ -21,5 +21,7 @@ spec:
       template: cluster-api-provider-aws-0-1-0
     - name: cluster-api-provider-openstack
       template: cluster-api-provider-openstack-0-1-1
+    - name: cluster-api-provider-docker
+      template: cluster-api-provider-docker-0-1-0
```
zerospiel, line 25: do we actually need the provider in the release? isn't it kind of a testing CAPI provider? anyway, i'd wait other reviewers' opinion on that

zerospiel (changes requested):
I think we should also add development credentials/policy references/cluster deployment/make targets/etc. in addition to the charts

zerospiel:
Please provide an issue link describing the initial problem and why it is required from the beginning

a13x5:
It would be nice to have some justification/explanation (like why it's needed, who will benefit etc.) for that in the description of the PR.
If there is an issue for that please link it.

templates/provider/cluster-api-provider-docker/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "v1.9.4"
+annotations:
+  cluster.x-k8s.io/provider: infrastructure-docker
+  cluster.x-k8s.io/v1beta2: v1beta2
```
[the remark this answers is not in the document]
MirgDenis: Fixed.

providers/docker.yml
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: docker
+clusterGVKs:
+  - group: infrastructure.cluster.x-k8s.io
+    version: v1beta2
```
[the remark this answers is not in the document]
MirgDenis: Fixed

templates/cluster/docker-hosted-cp/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "v1.30.0+k0s.0"
+annotations:
+  cluster.x-k8s.io/provider: infrastructure-docker, control-plane-k0sproject-k0smotron, bootstrap-k0sproject-k0smotron
+  cluster.x-k8s.io/bootstrap-k0sproject-k0smotron: v1beta1
+  cluster.x-k8s.io/control-plane-k0sproject-k0smotron: v1beta1
+  cluster.x-k8s.io/infrastructure-docker: v1beta2
```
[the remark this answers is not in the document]
MirgDenis: Fixed

templates/cluster/docker-hosted-cp/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+type: application
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "v1.30.0+k0s.0"
```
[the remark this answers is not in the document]
MirgDenis: Switched to common version of k0s

templates/cluster/docker-hosted-cp/values.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - "10.128.0.0/12"
+  serviceDomain: "cluster.local"
+
+# K0smotron parameters
+k0smotron:
+  service:
+    type: NodePort
+
+k0s:
+  version: v1.30.0+k0s.0
```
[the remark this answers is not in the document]
MirgDenis: Switched to common version of k0s

templates/cluster/docker-hosted-cp/templates/machinedeployment.yaml
MirgDenis: Split creates map with index keys and we can get elements by using "_N" index. Here is [1] the doc.

[1] https://masterminds.github.io/sprig/string_slice.html

templates/provider/kcm-templates/files/release.yaml
MirgDenis: >do we actually need the provider in the release?
No we do not, but I could not make it work without adding it to release.

>isn't it kind of a testing CAPI provider?
Yes, it is. And my intention is to use it as testing provider in KCM.

MirgDenis:
> I think we should also add development credentials/policy references/cluster deployment/make targets/etc. in addition to the charts

Regarding development credentials/policyRefs, as far as I understand CAPD does not require/use any credentials, in case of this patch I used Secret just as stub for credentials, Secret can be empty. Ideally I would like to have one more exception in CLD webhook [1] to allow CAPD without any credentials.

I agree about cluster deployment and make target for credentials. Will ad them.

[1] https://github.com/k0rdent/kcm/blob/main/internal/webhook/clusterdeployment_webhook.go#L283-L289

MirgDenis:
> It would be nice to have some justification/explanation (like why it's needed, who will benefit etc.) for that in the description of the PR. If there is an issue for that please link it.

#1080 Here is ticket with feature request

config/dev/docker-credentials.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
```
zerospiel, line 1: i think we should also provide an empty config map for sveltos to properly find resources, one could find examples in `aws` or `aks` credential files

templates/provider/kcm-templates/files/release.yaml
zerospiel: > without adding it to release.
Exactly, and since it is a testing provider, one could add it explicitly and manually to the release object. I do not think making it available for each user is a good idea

zerospiel:
> > I think we should also add development credentials/policy references/cluster deployment/make targets/etc. in addition to the charts
> 
> Regarding development credentials/policyRefs, as far as I understand CAPD does not require/use any credentials, in case of this patch I used Secret just as stub for credentials, Secret can be empty. Ideally I would like to have one more exception in CLD webhook [1] to allow CAPD without any credentials.
> 
> I agree about cluster deployment and make target for credentials. Will add them.
> 
> [1] https://github.com/k0rdent/kcm/blob/main/internal/webhook/clusterdeployment_webhook.go#L283-L289

I do not think that another exception is required, since you have already provided CAPD as a pluggable provider and described its identity kinds, it should work just as it is right now

zerospiel:
@s3rj1k could you also take a quick look regarding a new pluggable provider, is everything in place as it should be?

config/dev/docker-credentials.yaml
MirgDenis: You mean something like "openstack-cloud-config-resource-template" or rather "<identity_ref>-resource-template"?

zerospiel:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: docker-cluster-credential-resource-template
  namespace: ${NAMESPACE}
  labels:
    k0rdent.mirantis.com/component: "kcm"
  annotations:
    projectsveltos.io/template: "true"
```

This should be enough

templates/cluster/docker-hosted-cp/templates/dockermachinetemplate.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: DockerMachineTemplate
+metadata:
+  name: {{ include "dockermachinetemplate.name" . }}
+spec:
+  template:
+    spec: {}
```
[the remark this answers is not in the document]
MirgDenis: Empty spec worked out in case of simple deployment.
Maybe it makes sense to add templates over spec, but I think it's not critical for testing provider + if provider is merged anyone can build own cluster template with non-empty spec, and we can fix it with next releases of cluster template in case if we actually need something from spec.

config/dev/docker-credentials.yaml
s3rj1k: Yes, this is needed, either in docs or in config/dev/...

s3rj1k: https://github.com/k0rdent/docs/pull/89/files

templates/provider/kcm-templates/files/release.yaml
s3rj1k: I see no problem in releasing it in Chart, if there is concern about someone using it in production we can handle that in docs by mentioning that Docker provider is mostly used for Demo/Testing/Development purposes, please don't use it in production.

templates/cluster/docker-hosted-cp/values.yaml
```diff
@@ -0,0 +1,20 @@
+workersNumber: 1
+
+clusterLabels: {}
+
+clusterNetwork:
+  pods:
+    cidrBlocks:
+    - "192.168.0.0/16"
```
s3rj1k, line 8: Would be nice to have the ability to make network configurable via Environ, can be some override, could be covered purely in docs

templates/provider/kcm-templates/files/release.yaml
MirgDenis: @s3rj1k Actually I would prefer not to have it in release. Because some may not read docs to the point where WARNING is placed + I really think it better not to have testing provider in release.

Will I be able to add Docker provider to already deployed KCM, and expect to be able to deploy Docker clusters?

s3rj1k: In that case I see no reason to merge this PR, this can be out-off-tree provider

zerospiel: Yeah, I agree with @s3rj1k regarding merging the provider to the repo, we could add it as a testing one within the `test` directory or figure out how to enable it because clearly, this will come in handy to have a minimal provider. As of now (this PR) the provider is listed in both `management` and `release` objects (the former because the CAPD is now being listed) hence we support it, which I discourage from the first place: it is the testing provider.

To sum up, the provider should be excluded from both: the release and providers list (`providers/docker.yml`). We should figure out how to properly install it for test purposes though

templates/cluster/docker-hosted-cp/values.yaml
MirgDenis: Sorry, did not get what is needed. AWS hosted cluster template also has cidrBlocks set in it's values. Could you please elaborate?

templates/provider/kcm-templates/files/release.yaml
s3rj1k: I see big benefit in having this in release, specifically this provider, this would help a lot with demos and showcasing where I don't have access to infra.

If we don't add that to released chart, I see very little benefit, internally we already have a bunch of providers to test things on, all of should have access to at least one of infra/providers. External contributors would be able to use that provider only if they build from source and they would still need to test on something that is production-targeted.

MirgDenis: Yes, I agree as well.
BYO provider is also an option. @zerospiel @s3rj1k Do I need to create separate ticket for that, or it can be tracked in linked ticket?

zerospiel: @MirgDenis, there is no need for another issue, the current one might be adjusted accordingly to the changes when we decide how to 

Ok, I think this should be raised to the mgmt so that I will bring this up today

templates/cluster/docker-hosted-cp/templates/dockercluster.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
```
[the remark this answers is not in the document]
MirgDenis: Yes, you were right. Without finalizer it could not properly deleted all resources. Added finalizer.

MirgDenis:
Had to change k0s version to 1.32.1 because kindest/node image and k0s image has different release versions.

Also added dockerclusters to RBAC because without it kmc-controller-manager had errors related to operations with dockerclsuters.

templates/provider/kcm-templates/files/release.yaml
zerospiel: The current approach to the new provider being added to the release is approved.

zerospiel:
> Had to change k0s version to 1.32.1 because kindest/node image and k0s image has different release versions.

Could you elaborate on this? If we are talking about the dev experience, the current default k8s version of the server version (kind) is 1.32.0, so I do not understand 2 things at the moment:
- why does it even matter? the mgmt and child clusters' versions may differ and it should always be ok
- where is the `1.32.1` version coming from?

MirgDenis:
> > Had to change k0s version to 1.32.1 because kindest/node image and k0s image has different release versions.
> 
> Could you elaborate on this? If we are talking about the dev experience, the current default k8s version of the server version (kind) is 1.32.0, so I do not understand 2 things at the moment:
> 
>     * why does it even matter? the mgmt and child clusters' versions may differ and it should always be ok
> 
>     * where is the `1.32.1` version coming from?

In case of CAPD Machine object is container running `kindest/node` [1] image, and ControlPlane is containers running `k0sproject/k0s` [2] images on mgmt. As far as I understand we do not bootstrap k0s on worker nodes, instead we rely on what is inside kindest/node image.
Ideally `kindest/node` and `k0sproject/k0s` must use same version of k8s. However they do not have same tags available.

For example:
`k0sproject/k0s` has 1.31.5 image tag, `kindest/node` does not have it.
or vice versa
`kindest/node` has 1.31.6 image tag, `k0sproject/k0s` does not have it.

So I found common tag for both images, and it turned out to be 1.32.1.

[1] https://hub.docker.com/r/kindest/node/tags
[2] https://hub.docker.com/r/k0sproject/k0s/tags

zerospiel (dismissed):

zerospiel:
@MirgDenis could you please proceed with the conficts/outdated files to merge the PR?

zerospiel (approved):

Issue #953, opened 2025-01-24
Adopted local kind cluster doesn't work


Josca:
**Describe the bug**
I tried to adopt local "kind" cluster as adopted cluster for k0rdent testing. It should be possible according to [Adopted cluster guide](https://github.com/k0rdent/docs/blob/419a3a1b254f885fbb9ce8a0db756dc69bd98118/docs/usage/adopt-existing-cluster.md#adopting-existing-kubernetes-cluster-guide).
Local setup:
 - kind k0rdent management cluster created using [demo setup](https://github.com/k0rdent/demos?tab=readme-ov-file#demo-cluster-setup).
 - a second kind cluster to be used as a managed cluster.

**To Reproduce**
1. Create secret and credential resource:
~~~yaml
apiVersion: v1
data:
  value: YXNkc2RmYXNkZgo= # invalid kubeconfig - but it's still accepted
kind: Secret
metadata:
  name: adopted-cluster-kubeconf
  namespace: k0rdent
  labels:
    k0rdent.mirantis.com/component: "kcm"
type: Opaque
---
apiVersion: k0rdent.mirantis.com/v1alpha1
kind: Credential
metadata:
  name: adopted-cluster-cred
  namespace: k0rdent
spec:
  description: Adopted Credentials
  identityRef:
    apiVersion: v1
    kind: Secret
    name: adopted-cluster-kubeconf
    namespace: k0rdent
~~~
2. Create adopted cluster `ClusterDeployment` based on `adopted-cluster-0-0-2` template:
~~~yaml
apiVersion: k0rdent.mirantis.com/v1alpha1
kind: ClusterDeployment
metadata:
  name: kind-adopted-cluster
  namespace: k0rdent
spec:
  template: adopted-cluster-0-0-2
  credential: adopted-cluster-cred
  # dryRun: <"true" or "false": defaults to "false">
  config:
    clusterLabels:
      k0rdent: demo
  serviceSpec:
    services:
      - template: ingress-nginx-4-11-0
        name: ingress-nginx
        namespace: ingress-nginx
~~~
4. Check clusterdeployment cluster status, it shows cluster ready:
~~~bash
k get clusterdeployments.k0rdent.mirantis.com 
# NAME                   READY   STATUS
# kind-adopted-cluster   True    ClusterDeployment is ready
k get clustersummaries.config.projectsveltos.io -A # ... but no cluster summaries
# No resources found
~~~
- Services are not installed
- sveltos agent is not being installed to manged cluster
[...]

PR #1066, opened 2025-02-11
Add sveltos cluster status to the clusterdeployment conditions

kylewuolle:
Simple change to add the sveltos cluster status to the clusterdeployment conditions for adopted clusters. This fixes issue #953.

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -27,12 +27,14 @@ import (
 	fluxconditions "github.com/fluxcd/pkg/runtime/conditions"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	sveltosv1beta1 "github.com/projectsveltos/addon-controller/api/v1beta1"
+	libsveltosv1beta1 "github.com/projectsveltos/libsveltos/api/v1beta1"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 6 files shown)


internal/controller/clusterdeployment_controller.go
```diff
@@ -388,13 +390,71 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, mc *kcm
...
+		Resource: "sveltosclusters",
+	}
+
+	sveltosClusters, err := r.DynamicClient.Resource(gvr).Namespace(clusterDeployment.Namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{kcm.FluxHelmChartNameKey: clusterDeployment.Name}).String(),
+	})
+	if err != nil {
+		if !errors.As(err, &status.ResourceNotFoundError{}) {
+			return false, fmt.Errorf("failed to get sveltos cluster status: %w", err)
+		}
```
zerospiel, line 406: that would never be the case, the k8s std method does not return such an error

internal/controller/clusterdeployment_controller.go
```diff
@@ -388,13 +390,71 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, mc *kcm
...
 }
 
+func (r *ClusterDeploymentReconciler) updateSveltosClusterCondition(ctx context.Context, clusterDeployment *kcm.ClusterDeployment) (bool, error) {
+	gvr := schema.GroupVersionResource{
+		Group:    libsveltosv1beta1.GroupVersion.Group,
+		Version:  libsveltosv1beta1.GroupVersion.Version,
+		Resource: "sveltosclusters",
+	}
+
+	sveltosClusters, err := r.DynamicClient.Resource(gvr).Namespace(clusterDeployment.Namespace).List(ctx, metav1.ListOptions{
```
zerospiel, line 400: i do not fully understand, why cannot we utilize the typed client. it'd be much easier, IMO. are there any limitations?

internal/controller/clusterdeployment_controller.go
```diff
@@ -388,13 +389,54 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, mc *kcm
...
+	sveltosClusters := &libsveltosv1beta1.SveltosClusterList{}
+
+	err := r.Client.List(ctx, sveltosClusters, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{kcm.FluxHelmChartNameKey: clusterDeployment.Name}),
+	})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return true, fmt.Errorf("failed to get sveltos cluster status: %w", err)
+		}
+	}
```
zerospiel, line 402:
```suggestion
	if err := r.Client.List(ctx, sveltosClusters, &client.ListOptions{
		LabelSelector: labels.SelectorFromSet(map[string]string{kcm.FluxHelmChartNameKey: clusterDeployment.Name}),
	}); err != nil {
		return true, fmt.Errorf("failed to get sveltos cluster status: %w", err)
	}
```

redundant check, the instance itself is not null, the list operation does not return such an error basically (until the resource itself is missing which won't arise)

zerospiel (approved):

PR #1078, opened 2025-02-14
[e2e] Test cluster upgrades

### LINKED ISSUE #313: Add automated tests for managed clusters upgrades

eromanova:
1. Support to skip upgrade sequence validation (in case the user does not want to use our access management system). By default, the validation is enabled. In e2e tests, this validation is disabled to allow upgrade verification between any pair of templates.
2. In e2e testing automatically select the template to deploy. If no template is specified in the config, the system will automatically select one. For standard tests, the latest template is used. If an upgrade test is requested, the latest template serves as the upgrade target, while the previous version is the source.
3. Support to test ClusterDeployment upgrade. The procedure is as follows:
* Deploy kcm
* Installing stable templates via `make stable-templates`, which sets up a stable KCM HelmRepository, downloads, and installs templates using the stable tag.
* Creating a ClusterDeployment and waiting for it to become ready.
* Updating the `spec.template` in ClusterDeployment.
* Verifying the upgrade process.
4. Several minor fixes

Closes #313


Makefile
```diff
@@ -10,6 +10,8 @@ IMG_TAG = $(shell echo $(IMG) | cut -d: -f2)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.32.0
 
+KCM_STABLE_VERSION ?= v0.1.0
```
zerospiel, line 13: can't we automate it by grabbing the latest tag e.g. `git ls-remote --tags --sort v:refname --exit-code --refs https://github.com/k0rdent/kcm | tail -n1 | cut -d '/' -f3` but that wouldn't work with release candidate releases so might need some tweaking

Makefile
```diff
@@ -137,7 +139,8 @@ test-e2e: cli-install ## Run the e2e tests using a Kind k8s instance as the mana
 	@if [ "$$GINKGO_LABEL_FILTER" ]; then \
 		ginkgo_label_flag="-ginkgo.label-filter=$$GINKGO_LABEL_FILTER"; \
 	fi; \
-	KIND_CLUSTER_NAME="kcm-test" KIND_VERSION=$(KIND_VERSION) go test ./test/e2e/ -v -ginkgo.v -ginkgo.timeout=3h -timeout=3h $$ginkgo_label_flag
+	KIND_CLUSTER_NAME="kcm-test" KIND_VERSION=$(KIND_VERSION) VALIDATE_CLUSTER_UPGRADE_PATH=false \
```
zerospiel, line 142: i do not completely understand why is `VALIDATE_CLUSTER_UPGRADE_PATH=false` required in this particular case, could you elaborate?

eromanova: Sure.
Right now, cluster upgrades between templates are blocked by admission if the upgrade sequence isn’t properly defined in the `ClusterTemplateChain`. I was looking for a better way to test cluster upgrades in e2e:
1. Provide default `ClusterTemplateChain`. We could include a predefined chain in the kcm template and add it to e2e `AccessManagement`. This would ensure all upgrades are defined, but developers would need to update it whenever template versions change.
2. Build a dynamic Chain in e2e. We could generate a `ClusterTemplateChain` during tests and validate access management as well. Initially, this felt excessive, but it does make sense actually.
3. `Disable Validation`. Skip the upgrade sequence check entirely.

I prefer the third approach:
* It’s simpler and improves developer experience by allowing upgrades between any templates without manually updating any chains.
* It provides flexibility since users might not always want to enforce upgrade sequence validation. This enhancement could be useful beyond just testing.

zerospiel: sorry, i mixed up the target, i thought it was just the `test` recipe and actually it is `test-e2e`. no questions then

Makefile
zerospiel: may we stick from `curl` and `jq` to something more generic like `git`/`cut`/`tail`? i believe both do the same (both don't consider releases candidates, right?) but the former uses `jq` which "might" not exist yet in the moment of calculation

eromanova: I see your point. `curl -s https://api.github.com/repos/k0rdent/kcm/tags` fetches the same data as the command you mentioned. By release candidates, you mean pre-releases, right? It retrieves a complete list of available tags, including both releases and pre-releases, which I believe fits this use case. But sure, we can avoid using curl and jq.

zerospiel: correct, it'd be good to avoid pre-releases but i hope we can figure it out later

zerospiel (approved):

Issue #1077, opened 2025-02-14
[bug] openstack provider 0.1.0 stuck on waiting correct providerID


prazumovsky:
**Describe the bug**
ClusterDeployment could not become Ready because node spec.providerID and machine.spec.providerID are not equal.

clusterdeployment stuck on

```$ kubectl --kubeconfig kcm.kube -n kcm-system get clusterdeployment  
NAME              READY   STATUS
pr-ceph-k0rdent   False   pr-ceph-k0rdent-md: Minimum availability requires 3 replicas, current 1 available. ```

why? because machines stuck on

```$ kubectl --kubeconfig kcm.kube -n kcm-system get machine pr-ceph-k0rdent-cp-1 -o yaml
...
spec:
...
  providerID: openstack://RegionOne/05224519-f863-4857-a971-5384bd879ab8
...
status:
...
    - lastTransitionTime: "2025-02-14T13:45:30Z"
      message: Waiting for a Node with spec.providerID openstack://RegionOne/05224519-f863-4857-a971-5384bd879ab8
        to exist
      observedGeneration: 3
      reason: InspectionFailed
      status: Unknown
      type: NodeHealthy
    - lastTransitionTime: "2025-02-14T13:45:30Z"
      message: Waiting for a Node with spec.providerID openstack://RegionOne/05224519-f863-4857-a971-5384bd879ab8
        to exist
      observedGeneration: 3
      reason: InspectionFailed
      status: Unknown
      type: NodeReady
```

because a corresponding node has incorrect unmatched providerID

```
$ kubectl --kubeconfig kcm.child.kube get node pr-ceph-k0rdent-cp-1 -o yaml | grep providerID
  providerID: openstack:///05224519-f863-4857-a971-5384bd879ab8
```

**To Reproduce**
1. Create required openstack cloud.yaml secret, credentials and resource-template cm.
2. Create clusterdeployment from the provided file. [k0rdent-cluster-deployment.txt](https://github.com/user-attachments/files/18800515/k0rdent-cluster-deployment.txt)
4. Frequently (3 out of 5 runs) clusterdeployment stuck.

**Expected behavior**
ClusterDeployment becomes Ready.

**Additional context**
Creds for the first step:

[k0rdent-resource-template.txt](https://github.com/user-attachments/files/18800585/k0rdent-resource-template.txt)

[k0rdent-child-secret.txt](https://github.com/user-attachments/files/18800622/k0rdent-child-secret.txt)

[...]

PR #1091, opened 2025-02-17
capo: Fix missing region in providerID on cluster nodes

prazumovsky:
Sometimes node deployment could be slow therefore openstack ccm sets providerID without region in format `openstack:///<instanceID>` instead of `openstack://<region>/<instanceID>`. Enable OS_CCM_REGIONAL to openstack ccm always set region to providerID.

-----

Fixes https://github.com/k0rdent/kcm/issues/1077

### DIFF
--- a/config/dev/openstack-clusterdeployment.yaml
+++ b/config/dev/openstack-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: openstack-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
-  template: openstack-standalone-cp-0-1-3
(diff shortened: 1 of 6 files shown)


eromanova (commented):
LGTM. @bnallapeta could you please also take a look?

wkonitzer:
@prazumovsky should we add

"default": true

to the json?

bnallapeta (commented):
Looks good. But this requires a version bump too.

prazumovsky:
> @prazumovsky should we add
> "default": true
> to the json?

I don't see that "default" parameter is used anywhere in values.schema.json all over helm charts in templates/cluster. Thus I think we shouldn't add it for consistency.

Issue #1090, opened 2025-02-17
[bug] CAPO 0.12.0 misses ORC crds


prazumovsky:
**Describe the bug**
ORC was removed from CAPO and moved to another repo:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/releases/tag/v0.12.0

while bumping cluster-api-provider-openstack to 0.12.0 version: https://github.com/k0rdent/kcm/commit/6af684110c69ace62d1942ba1a6a6ad10dbde661

Therefore we have repeatable CAPO restarts with the following log lines:
```
E0217 18:51:22.669307       1 kind.go:71] "if kind is a CRD, it should be installed before calling Start" err="no matches for kind \"Image\" in version \"openstack.k-orc.cloud/v1alpha1\"" logger="controller-runtime.source.EventHandler" kind="Image.openstack.k-orc.cloud"
E0217 18:51:32.667736       1 kind.go:71] "if kind is a CRD, it should be installed before calling Start" err="no matches for kind \"Image\" in version \"openstack.k-orc.cloud/v1alpha1\"" logger="controller-runtime.source.EventHandler" kind="Image.openstack.k-orc.cloud"
```

**To Reproduce**

exec on main branch:
```
DEV_PROVIDER=openstack make dev-apply
```

**Expected behavior**
cluster-api-provider-openstack is running and active.

**Additional context**
After I applied these steps issue gone:

https://github.com/k-orc/openstack-resource-controller?tab=readme-ov-file#installation

PR #1093, opened 2025-02-18
capo: Add orc v1.0.0 file with all required objects

prazumovsky:
Fixes https://github.com/k0rdent/kcm/issues/1090

-----

CAPO v0.12.0 misses openstack-resource-controller with all required objects and CRDs. Add orc v1.0.0 `install.yaml` as a template to capo chart.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,10 @@ CONTAINER_TOOL ?= docker
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+TEMPLATES_DIR := templates
+PROVIDER_TEMPLATES_DIR := $(TEMPLATES_DIR)/provider
+CLUSTER_TEMPLATES_DIR := $(TEMPLATES_DIR)/cluster
+
 .PHONY: all
 all: build
 
@@ -90,8 +94,21 @@ kcm-dist-release: helm yq
 templates-generate:
 	@hack/templates.sh
 
+CAPO_DIR := $(PROVIDER_TEMPLATES_DIR)/cluster-api-provider-openstack
+CAPO_ORC_VERSION ?= $(shell grep 'orcVersion:' $(CAPO_DIR)/values.yaml | cut -d '"' -f 2)
+CAPO_ORC_TEMPLATE := "$(CAPO_DIR)/templates/orc-$(shell echo $(CAPO_ORC_VERSION) | sed 's/\./-/g').yaml"
+.PHONY: capo-orc-fetch
+capo-orc-fetch: yq
+	@if ! test -s "$(CAPO_ORC_TEMPLATE)"; then \
+	  	curl -L --fail -s https://github.com/k-orc/openstack-resource-controller/releases/download/v$(CAPO_ORC_VERSION)/install.yaml -o $(CAPO_ORC_TEMPLATE); \
+	  	$(YQ) -i '(select(.kind == "Deployment") | .spec.template.spec.containers.[0].image) |= sub("$(CAPO_ORC_VERSION)"/"{{ .Values.orcVersion }}")' $(CAPO_ORC_TEMPLATE); \
+	  	printf '%s\n' "{{ if (eq .Values.orcVersion \"$(CAPO_ORC_VERSION)\") }}" | cat - $(CAPO_ORC_TEMPLATE) > $(CAPO_ORC_TEMPLATE).tmp; \
+	  	mv $(CAPO_ORC_TEMPLATE).tmp $(CAPO_ORC_TEMPLATE); \
+	  	echo "{{- end }}" >> $(CAPO_ORC_TEMPLATE); \
+	fi
+
 .PHONY: generate-all
-generate-all: generate manifests templates-generate add-license projectsveltos-crds
+generate-all: generate manifests templates-generate add-license projectsveltos-crds capo-orc-fetch
 
 .PHONY: projectsveltos-crds
 projectsveltos-crds: sveltos-crds yq
@@ -135,10 +152,6 @@ add-license: addlicense
 	$(ADDLICENSE) -c "" -ignore ".github/**" -ignore "config/**" -ignore "templates/**" -ignore "bin/**" -ignore ".*" .
 
 ##@ Package
-
-TEMPLATES_DIR := templates
-PROVIDER_TEMPLATES_DIR := $(TEMPLATES_DIR)/provider
-CLUSTER_TEMPLATES_DIR := $(TEMPLATES_DIR)/cluster
 
 CHARTS_PACKAGE_DIR ?= $(LOCALBIN)/charts
 $(CHARTS_PACKAGE_DIR): | $(LOCALBIN)
--- a/templates/provider/cluster-api-provider-openstack/Chart.yaml
+++ b/templates/provider/cluster-api-provider-openstack/Chart.yaml
@@ -13,7 +13,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.1
+version: 0.1.2
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
--- a/templates/provider/cluster-api-provider-openstack/templates/orc-1-0-0.yaml
+++ b/templates/provider/cluster-api-provider-openstack/templates/orc-1-0-0.yaml
@@ -1,0 +1,856 @@
+{{ if (eq .Values.orcVersion "1.0.0") }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: orc
+    control-plane: controller-manager
+  name: orc-system
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.16.4
+  name: images.openstack.k-orc.cloud
+spec:
+  group: openstack.k-orc.cloud
+  names:
+    kind: Image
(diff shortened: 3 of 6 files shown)


zerospiel (commented):
LGTM. @bnallapeta PTAL

bnallapeta (commented):
Looks good. But this isn't leaving room for choosing the version of ORC for installation. 

I think we should either do:

a) sub-charts and add ORC as a dependency for OpenStack.
b) keep it simple and add versioning capability in [values.yaml](https://github.com/k0rdent/kcm/blob/main/templates/provider/cluster-api-provider-openstack/values.yaml) for orc.

zerospiel:
Correct. I'd personally go with the second option since there is no ORC chart (at least i couldn't find it)

prazumovsky:
@bnallapeta @zerospiel how we could versioning only one file? just make something like {{ if orcVersion == "v1.0.0" }} then deploy all monolith file?

zerospiel:
I was thinking about a make target, idk something similar to 
```yaml
# values.yaml
orcVersion: 1.0.0
```

```makefile
target: ORC_VERSION ?= v$(shell grep 'orcVersion:' $(path_to_values) | cut -d '"' -f 2)
target:
  curl --fail -s https://github.com/orc/release/path/$(ORC_VERSION) > $(path_to_orc_install_file)
```

prazumovsky:
@zerospiel but how it will help folks who are not deploying with Makefile but using stable version of kcm?

zerospiel:
Correct, I suggest this approach to be exclusively for dev purposes (to ease the bump-version scenario). I do not see another option to manage the `install.yaml` since it is being published as is in raw format after being generated by `kustomize` (as of my understanding), hence it is not feasible to version it in any other way. I hope one day the maintainers would suggest something more manageable, in particular, a helm chart

s3rj1k:
> @zerospiel but how it will help folks who are not deploying with Makefile but using stable version of kcm?

They will have single-tested version that works and was tested before release was published

prazumovsky:
@zerospiel please review my suggestion then, I updated commit. If it is not applicable - I'll rework it to "only dev purposes" as you suggested.

zerospiel:
Unfortunately, it is not what I'd expect as a developer, I still would have to perform manual steps such as fetching a newly released version. The bumping process should be straightforward and consist of a minimal amount of manual steps to ensure an error-proof approach. All of this is only applicable until a proper ORC chart is available.

bnallapeta:
I agree with @prazumovsky. I think we need to focus on supporting this to a regular user who would do a `helm install`. Supporting dev workflow with Makefile can be done as a secondary thing imo.

What I have in my mind is this: 

For every release, the only thing that is changing is the [image version](https://github.com/k0rdent/kcm/pull/1093/files#diff-f8c4286f6060707818ab7b75352b46e78d94aff1d23a8f0025175d0e36d9eddeR853) in orc.

So, let's do this in values.yaml:
```
orcVersion:  "1.0.0"
```

And then substitute this in orc.yaml [here](https://github.com/k0rdent/kcm/pull/1093/files#diff-f8c4286f6060707818ab7b75352b46e78d94aff1d23a8f0025175d0e36d9eddeR853). 

```
        command:
        - /manager
        image: quay.io/orc/openstack-resource-controller: {{ .Values.orcVersion }}
```

This way, even if we want to include more fields in the future, we can easily modify this. (image tag, image repo etc).

And I don't think we need this file anymore: templates/provider/cluster-api-provider-openstack/templates/orc.yaml

zerospiel:
@bnallapeta we could do this though it is not sustainable: there is no guarantee (e.g. like the v1 promise in Go) that a new controller version does not have any breaking changes (it doesn't matter whether CRDs or controller flags or labels or the deployment setup, etc.), meaning that controller version itself does not set the rest of the objects up which are connected in one or another way, hence not user nor developer would benefit from such a knob. We could ensure users that we use some fixed version of ORC here and there, and that it is not movable (e.g. nobody cares that we use some fixed versions of CSI/CCM in different providers)

prazumovsky:
@zerospiel @bnallapeta that's why I added this way:

From user perspective we have official install.yaml in capo chart which is versioned by `orcRelease` and a separate file to load from capo chart files. Install.yaml currently is the only way to deliver orc from official repo and I am not a fan of changing it - better download full new version instead of changing existing one because it could be any changes in rbac or crds which are not covered by only image version change.

From dev perspective if a new version comes, we use change orcRelease and make generate to download a new version to capo chart files and nothing else.

Win-win :)

Makefile
```diff
@@ -90,8 +94,19 @@ kcm-dist-release: helm yq
...
+CAPO_DIR := $(PROVIDER_TEMPLATES_DIR)/cluster-api-provider-openstack
+CAPO_ORC_VERSION ?= v$(shell grep 'orcRelease:' $(CAPO_DIR)/values.yaml | cut -d '"' -f 2)
+CAPO_ORC_FILE := "$(CAPO_DIR)/files/orc-$(shell echo $(CAPO_ORC_VERSION) | sed 's/\./\-/g' | cut -c2-).yaml"
+.PHONY: capo-orc-fetch
+capo-orc-fetch:
+	@if [ -f "$(CAPO_ORC_FILE)" ]; then \
+  		echo "ORC installation file with version $(CAPO_ORC_VERSION) already exists in $(CAPO_DIR) chart."; \
+  	else \
+		curl -L --fail -s https://github.com/k-orc/openstack-resource-controller/releases/download/$(CAPO_ORC_VERSION)/install.yaml -o $(CAPO_ORC_FILE); \
+	fi
```
zerospiel, line 106:
```suggestion
CAPO_ORC_FILE := "$(CAPO_DIR)/files/orc-$(shell echo $(CAPO_ORC_VERSION) | sed 's/\./-/g' | cut -c2-).yaml"
.PHONY: capo-orc-fetch
capo-orc-fetch:
	@test -s "$(CAPO_ORC_FILE)" || curl -L --fail -s https://github.com/k-orc/openstack-resource-controller/releases/download/$(CAPO_ORC_VERSION)/install.yaml -o $(CAPO_ORC_FILE)
```

might be simplified (no need in echo; no need to escape `-` sym)

zerospiel:
I agree to allow the setting of the version of the controller image. One should perform such a version change on its responsibility. The problem with the dev approach is it requires subtle substitutes that I am keen to avoid.

@prazumovsky i believe @bnallapeta meant to set the controller image version itself which is now hardcoded. This will require performing some `grep` magic finding the `Deployment` entry, then the template's image, and substituting the hardcoded version (e.g. `v1.0.0`) with the helm value such as `v{{ .Values.orcRelease}}`. 

I am content with the current approach having the hardcoded version though. So it is up to you to agree whether to implement or not implement the aforementioned substitute.

prazumovsky:
> @prazumovsky i believe @bnallapeta meant to set the controller image version itself which is now hardcoded. This will require performing some grep magic finding the Deployment entry, then the template's image, and substituting the hardcoded version (e.g. v1.0.0) with the helm value such as v{{ .Values.orcRelease}}.

OK, will do along with review comment resolving.

> Or did I get it wrong and you suggest that a user should deliver its own install.yaml along with the chart? I cannot see how it will work since at that point the capo provider chart already includes the file (we prepared some version with the v1.0.0 orc version and then released it), and I do not quite understand how to swap the old one with a new one

No-no, you got it right - user gets only versions which we are delivering (e.g. for now we are supporting and delivering only install.yaml v1.0.0 by downloading orc-1-0-0.yaml to capo chart) and dev could easily perform orc upgrade by 2 steps: change `orcRelease` and `make generate`.

bnallapeta:
yes, we could go with just 1.0.0 hard-coded and document that in docs and move forward with the PR. 

I don't see why we would need to load a file separately with the versioning in that case.

Makefile
```diff
@@ -90,8 +94,22 @@ kcm-dist-release: helm yq
 templates-generate:
 	@hack/templates.sh
 
+CAPO_DIR := $(PROVIDER_TEMPLATES_DIR)/cluster-api-provider-openstack
+CAPO_ORC_VERSION ?= $(shell grep 'orcVersion:' $(CAPO_DIR)/values.yaml | cut -d '"' -f 2)
+CAPO_ORC_TEMPLATE := "$(CAPO_DIR)/templates/orc-$(shell echo $(CAPO_ORC_VERSION) | sed 's/\./-/g').yaml"
+.PHONY: capo-orc-fetch
+capo-orc-fetch: yq
+	@if test -s "$(CAPO_ORC_TEMPLATE)"; then \
+		echo "ORC installation file with version $(CAPO_ORC_VERSION) already exists in $(CAPO_DIR) chart."; \
```
zerospiel, line 103: i do not think we need this echo

Makefile
```diff
@@ -90,8 +94,22 @@ kcm-dist-release: helm yq
...
+	else \
+	  	curl -L --fail -s https://github.com/k-orc/openstack-resource-controller/releases/download/v$(CAPO_ORC_VERSION)/install.yaml -o $(CAPO_ORC_TEMPLATE); \
+	  	$(YQ) -i '(select(.kind == "Deployment") | .spec.template.spec.containers.[0].image) |= sub("$(CAPO_ORC_VERSION)"/"{{ .Values.orcVersion }}")' $(CAPO_ORC_TEMPLATE); \
+	  	sed  -i '1i {{ if (eq .Values.orcVersion "$(CAPO_ORC_VERSION)") }}' $(CAPO_ORC_TEMPLATE); \
```
zerospiel, line 107: this will not work on bsd-based `sed` binary, in particular, darwin. since not all of us has `gsed` (especially substituted in PATH). i will try to find another approach and then suggest it

zerospiel: that should work
```makefile
...
		printf '{{ if (eq .Values.orcVersion "$(CAPO_ORC_VERSION)") }}\n' > $(CAPO_ORC_TEMPLATE).bak && \
		cat $(CAPO_ORC_TEMPLATE) >> $(CAPO_ORC_TEMPLATE).bak && \
		mv $(CAPO_ORC_TEMPLATE).bak $(CAPO_ORC_TEMPLATE) && \
```

whether to put `&&` everywhere or a plain `; ` is up to you

prazumovsky:
@zerospiel changed.

zerospiel (approved):

PR #1108, opened 2025-02-20
Enable Openstack MachineTemplate rotation

### LINKED ISSUE #1107: [enhancement] Machine template rotation
The problem is that I cannot change MachineTemplate. In particular I cannot change base image (which may have some CVE), I cannot change flavor of instances. In case if I try to do it I will get failed chart release and following logs
```
      OpenStackMachineTemplate spec.template.spec field is immutable. Please create
      a new resource instead. Ref doc: https://cluster-api.sigs.k8s.io/tasks/change-machine-template.html. '
```
Seems like rotation of machine templates is valid operation from ClusterAPI point of view

I would like to be able to change worker/controlplane machine template spec, and expect that it will:
1) Create new machine template
2) Delete old machine template
3) Change ClusterDeployment status according to what is happening on data plane. Specifically I would like to have Updating status in ClusterDeployment object during rotation of MachineTemplate and I would expect Ready state only when all machines in MachineTemplate are ready.

Last point may be optional, but my concern is that operators may break something if they try to change something in the middle of rotation process, changes in status would allow us to implement blocking mechanism in that case. 

Alternative solution is to have different images, flavors or other machine parameters embedded in clusterTemplate releases, it would allow I to change machine spec, but that would limit operators to release period.

[1] https://release-0-3.cluster-api.sigs.k8s.io/tasks/change-machine-template

MirgDenis:
Closes #1107

### DIFF
--- a/templates/cluster/openstack-standalone-cp/templates/_helpers.tpl
+++ b/templates/cluster/openstack-standalone-cp/templates/_helpers.tpl
@@ -3,11 +3,11 @@
 {{- end }}
 
 {{- define "openstackmachinetemplate.controlplane.name" -}}
-    {{- include "cluster.name" . }}-cp-mt
+    {{- include "cluster.name" . }}-cp-mt-{{ .Values.controlPlane.image | toString | sha256sum | trunc 8 }}
 {{- end }}
 
 {{- define "openstackmachinetemplate.worker.name" -}}
-    {{- include "cluster.name" . }}-worker-mt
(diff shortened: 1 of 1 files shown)


zerospiel (approved):
Every branch must be up-to-date with the `main` to be merged

zerospiel:
@MirgDenis I've preliminarily merged it, could you prepare a new PR updating the cluster template version (`Chart.yaml`) and call the `generate-all` target along with changing the version in the dev config `config/dev/openstack-clusterdeployment.yaml`??

MirgDenis:
@zerospiel Sure, will do it. I was also surprised by how fast it was merged.

PR #1114, opened 2025-02-24
Adjust admission validation for the k0smotron infrastructure provider

### LINKED ISSUE #1113: Adjust admission validation for the new remote provider
The KCM code should be updated to support the new remote cluster template, including adapting credentials validation and the pluggable providers list.

eromanova:
1. Adjust admission validation for the k0smotron infrastructure provider
2. Fix incorrect control plane ref in the `remote-server` template

Closes #1113

### DIFF
--- a/internal/webhook/clusterdeployment_webhook.go
+++ b/internal/webhook/clusterdeployment_webhook.go
@@ -280,25 +280,23 @@ func isCredMatchTemplate(cred *kcmv1.Credential, template *kcmv1.ClusterTemplate
 	const secretKind = "Secret"
 
 	for _, provider := range template.Status.Providers {
-		if provider == providersloader.InfraPrefix+"internal" {
-			if idtyKind != secretKind {
-				return errMsg(provider)
-			}
-
+		if !strings.HasPrefix(provider, providersloader.InfraPrefix) {
 			continue
 		}
-
-		idtys, found := providersloader.GetClusterIdentityKinds(provider)
-		if !found {
-			if strings.HasPrefix(provider, providersloader.InfraPrefix) {
-				return fmt.Errorf("unsupported infrastructure provider %s", provider)
+		infraProviderName := strings.TrimPrefix(provider, providersloader.InfraPrefix)
+		if infraProviderName == "internal" || infraProviderName == "k0sproject-k0smotron" {
+			if idtyKind != secretKind {
+				return errMsg(infraProviderName)
 			}
+		}
 
-			continue
+		idtys, found := providersloader.GetClusterIdentityKinds(infraProviderName)
+		if !found {
+			return fmt.Errorf("unsupported infrastructure provider %s", infraProviderName)
 		}
 
 		if !slices.Contains(idtys, idtyKind) {
-			return errMsg(provider)
+			return errMsg(infraProviderName)
 		}
 	}
 
--- a/internal/webhook/clusterdeployment_webhook_test.go
+++ b/internal/webhook/clusterdeployment_webhook_test.go
@@ -390,7 +390,7 @@ func TestClusterDeploymentValidateCreate(t *testing.T) {
 					template.WithValidationStatus(v1alpha1.TemplateValidationStatus{Valid: true}),
 				),
 			},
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -149,7 +149,7 @@ lint-fix: golangci-lint fmt vet ## Run golangci-lint linter and perform fixes
 
 .PHONY: add-license
 add-license: addlicense
-	$(ADDLICENSE) -c "" -ignore ".github/**" -ignore "config/**" -ignore "templates/**" -ignore "bin/**" -ignore ".*" .
+	$(ADDLICENSE) -c "" -ignore ".github/**" -ignore "config/**" -ignore "templates/**" -ignore "providers/**" -ignore "bin/**" -ignore ".*" .
```
s3rj1k, line 152: are we stopping adding copyrights for `providers/` ?

internal/providers/yaml.go
```diff
@@ -27,6 +27,7 @@ import (
 // YAMLProviderDefinition represents a YAML-based provider configuration.
 type YAMLProviderDefinition struct {
 	Name                 string                    `yaml:"name"`
+	ComponentName        string                    `yaml:"componentName,omitempty"`
```
s3rj1k, line 30: feels like extra step for people that want to use out-of-tree feature, can we reuse `Name` field, without adding extra configurable field?

s3rj1k: also not very clear what is the difference between `Name` and `ComponentName` without looking into actual code, same concern about `people that want to use out-of-tree feature`

Makefile
eromanova: Since we don’t set copyrights for the template YAMLs, I thought we could follow the same approach for providers, as those lines seem inaccurate in the resulting ConfigMap. However, this is just my perspective, and I won’t insist on reverting it. Do you think we should keep it?

s3rj1k: I think that for `licensing` we should get approval, I am OK with removing also

internal/providers/yaml.go
eromanova: I see your point, it may not be the best approach. But, we currently have inconsistencies in provider names. In cluster templates, we agreed that charts should have annotations following the format `<providerType>-<providerName>` in CAPI terms (e.g., `infrastructure-k0sproject-k0smotron`). And the Credential validation is based on the provider name in CAPI terms.

With k0smotron, the component name in the release is `k0smotron`, but the infrastructure provider name in CAPI terms is `k0sproject-k0smotron`. This means that for credential validation, we need to check if `infrastructure-k0sproject-k0smotron` is exposed while simultaneously registering the provider as `k0smotron`.

I think we should either standardize naming so that the `component name` matches the `CAPI provider name` or introduce a way to customize these mappings. However, since the component exposing a provider may not always be prefixed with `cluster-api-provider-`, I kept this solution. @s3rj1k @zerospiel what do you think would be the better option here?

s3rj1k: IMO, we definitely should make some naming standardization, keep in mind that this specific feature is exposed to users that don't want to deal with code at all, so we should keep that struct definition as simple as possible even if this will require some jumping around internally with `k0sproject-k0smotron`

zerospiel: i'd agree on that with @s3rj1k that we have to unify the approach and/or rename the provider from `k0smotron` to its CAPI name `k0sproject-k0smotron` to eliminate the confusion

eromanova: Thanks, guys. I'll update the PR accordingly.

eromanova: Renaming k0smotron should be discussed with the management team and thoroughly tested to ensure backward compatibility, as it involves changes to the Release. Also, the k0smotron team has moved the project to a separate repository. Before renaming it in k0rdent, I want to confirm whether they will also change the CAPI provider name. I’ll create a follow-up task for this specific change. For now, to unblock remote server deployment, let's keep the k0smotron provider as is with some validation adjustments.

Makefile
eromanova: Let's keep it as is for now and discuss it with the management team later.

internal/providers/yaml.go
zerospiel: I agree on that hence resolving

s3rj1k:
LGTM, have no ACLs to approve

s3rj1k (approved):

zerospiel (approved):

PR #1129, opened 2025-02-25
Implement IPAM provider and include it in vsphere cluster templates.

kylewuolle:
This change resolves issue https://github.com/k0rdent/2a-internal/issues/7 by creating a new provider template for the CNCF cluster API IPAM provider (https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster) and incorporating it into the vsphere standalone and hosted cluster templates.


templates/cluster/vsphere-hosted-cp/values.yaml
Kshatrix, line 1: Please update versions of the charts (Chart.yaml)

Kshatrix (changes requested):

PR #1144, opened 2025-03-03
Update sveltos to v0.49.0

Kshatrix:


zerospiel:
We may want to drop our internal kludges for the CRDs because there are several sveltos' charts, including a dedicated one for them. I have also mentioned that [here](https://github.com/k0rdent/kcm/issues/1116#issuecomment-2685654630) and that the new version automatically addresses the issue

wahabmk:
Let's also remove https://github.com/k0rdent/kcm/blob/main/templates/provider/projectsveltos/values.yaml, it has default values only so don't need it.

Kshatrix:
@zerospiel makes sense, although let's have a separate task for that

PR #1167, opened 2025-03-05
Add GCP provider template

### LINKED ISSUE #1146: ProviderTemplate for GCP

eromanova:
Closes #1146


Kshatrix (approved):

providers/gcp.yml
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: gcp
+clusterGVKs:
+  - group: infrastructure.cluster.x-k8s.io
+    version: v1beta2
```
zerospiel, line 18: is that correct? i can only see the `v1beta1` version of the `GCPCluster`. the same goes with the `GCPManagedCluster` kind

zerospiel, line 18: is that correct? i can only see the v1beta1 version of the GCPCluster. the same goes with the GCPManagedCluster kind

zerospiel (changes requested):

providers/gcp.yml
eromanova: Right. My mistake. I'll fix that. Thanks, Michael.

zerospiel (approved):

PR #1168, opened 2025-03-05
Migrate to a dedicated sveltos CRDs chart

### LINKED ISSUE #1116: Non-backing off call to https://webhook-service.projectsveltos.svc:443/convert?timeout=30s
**What happened**:
Installed KCM 0.1.0 on K0s Server Version: v1.32.1+k0s. After system has settled, something is still calling https://webhook-service.projectsveltos.svc:443/convert?timeout=30s every second. This causes 100% of 1 vCPU to be used when the system is idle.
```
journalctl --since "1m ago" | grep -i webhook-service.projectsveltos.svc | grep -i E0217 | wc -l
59

Feb 17 18:20:54 ip-172-31-14-170 k0s[11008]: time="2025-02-17 18:20:54" level=info msg="W0217 18:20:54.762850   26115 reflector.go:569] storage/cacher.go:/config.projectsveltos.io/profiles: failed to list config.projectsveltos.io/v1alpha1, Kind=Profile: conversion webhook for config.projectsveltos.io/v1beta1, Kind=Profile failed: Post \"https://webhook-service.projectsveltos.svc:443/convert?timeout=30s\": service \"webhook-service\" not found" component=kube-apiserver stream=stderr
Feb 17 18:20:54 ip-172-31-14-170 k0s[11008]: time="2025-02-17 18:20:54" level=info msg="E0217 18:20:54.762872   26115 cacher.go:478] cacher (profiles.config.projectsveltos.io): unexpected ListAndWatch error: failed to list config.projectsveltos.io/v1alpha1, Kind=Profile: conversion webhook for config.projectsveltos.io/v1beta1, Kind=Profile failed: Post \"https://webhook-service.projectsveltos.svc:443/convert?timeout=30s\": service \"webhook-service\" not found; reinitializing..." component=kube-apiserver stream=stderr
```

**What you expected to happen**:
Whatever is calling the service should exponentially back-off or stop. OR the sveltos webhook-service should be deployed. It doesn't look like the webhook-service is deployed
```
ubuntu@ip-172-31-14-170:~$ kubectl get service -A  | grep -i webhook
kcm-system       azureserviceoperator-webhook-service                          ClusterIP   10.106.162.39    <none>        443/TCP                  4d1h
kcm-system       capa-webhook-service                                          ClusterIP   10.98.132.52     <none>        443/TCP                  4d1h
kcm-system       capi-operator-webhook-service                                 ClusterIP   10.106.0.11      <none>        443/TCP                  4d1h
kcm-system       capi-webhook-service                                          ClusterIP   10.108.254.154   <none>        443/TCP                  4d1h
kcm-system       capo-webhook-service                                          ClusterIP   10.108.199.178   <none>        443/TCP                  4d1h
kcm-system       capv-webhook-service                                          ClusterIP   10.108.130.81    <none>        443/TCP                  4d1h
kcm-system       capz-webhook-service                                          ClusterIP   10.96.211.117    <none>        443/TCP                  4d1h
kcm-system       kcm-cert-manager-webhook                                      ClusterIP   10.109.104.165   <none>        443/TCP,9402/TCP         4d1h
kcm-system       kcm-webhook-service                                           ClusterIP   10.107.89.11     <none>        443/TCP                  4d1h
```

**Where can this issue be corrected? (optional)**
Two fixes need to be made. Fix the call so it does not fail. Fix the call so it exponentially backs off if the service is down to not impact CPU so much.

**How to reproduce it (as minimally and precisely as possible)**:
Happens on basic install. https://docs.k0rdent.io/latest/quickstart-1-mgmt-node-and-cluster/

**Anything else we need to know?**:

### LINKED ISSUE #1163: Migrate to a dedicated sveltos crds chart

zerospiel:
* add a new sveltos chart with crds
* remove extra make targets to download and separate CRDs
* remove flux helm-ctrl skip crds related code
* update sveltos to 0.50.0

Closes #1163 
Closes #1116 transitively due to `sveltos` bump to `>=0.48.0`


templates/provider/projectsveltos/Chart.yaml
```diff
@@ -2,9 +2,12 @@ apiVersion: v2
 name: projectsveltos
 description: A Helm chart to refer the official projectsveltos helm chart
 type: application
-version: 0.49.0
-appVersion: "0.49.0"
+version: 0.49.1
+appVersion: "0.49.1"
 dependencies:
+  - name: sveltos-crds
```
Kshatrix, line 8: afair we couldn't add crds to the same sveltos chart because some of the resources created in the chart already required those crds to be installed. is it no longer the case?

templates/provider/projectsveltos/Chart.yaml
```diff
@@ -2,9 +2,12 @@ apiVersion: v2
...
+    version: ~0.49.1
```
Kshatrix, line 9: let's keep the version strict

Kshatrix (changes requested):

templates/provider/projectsveltos/Chart.yaml
zerospiel: If we are to separate them, then it'd lead to either a separated quasi provider (that is not an option back according to the prior discussions) or extra logic during the reconciliation of provider templates (i.e,. install the `sveltos-crds` chart that has to be already presented in the registry somehow).

The case you are asking about is still valid, though:
- `crds` directory is still presented within the `projectsveltos` chart, so during the initial installation there is no such problem
- the `sveltos-crds` charts take the ownership over the CRDs after the initial installation, so during the upgrade path, there should be no such an issue with the CRDs upgrading/missing
- during the upgrade, the "default" objects had already been created so that also won't impact the process

The only issue I could assume is removing the `crds` directory from the `projectsveltos` chart, which might lead to the error above but at the same time, the CRDs from the `sveltos-crds` should already be installed, that is why I do not think there would be a persistent error

zerospiel: I have checked the upgrade in both ways: via the `helm upgrade --take-ownership` (because it is the default setting for the flux's `helmreleases` upgrade action) and via the upgrading from `0.1.0` to a fake `0.2.0` with the updated versions, both variants had the updated CRDs with the following metadata:

```yaml
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.17.2
    meta.helm.sh/release-name: projectsveltos
    meta.helm.sh/release-namespace: projectsveltos
  creationTimestamp: "2025-03-11T16:19:35Z"
  generation: 1
  labels:
    app.kubernetes.io/instance: projectsveltos
    app.kubernetes.io/name: sveltos-crds
    app.kubernetes.io/version: 0.49.0
    helm.sh/chart: sveltos-crds-0.49.1
    helm.toolkit.fluxcd.io/name: projectsveltos
    helm.toolkit.fluxcd.io/namespace: kcm-system
```

templates/provider/projectsveltos/Chart.yaml
```diff
@@ -2,9 +2,12 @@ apiVersion: v2
 name: projectsveltos
 description: A Helm chart to refer the official projectsveltos helm chart
 type: application
-version: 0.49.0
-appVersion: "0.49.0"
+version: 0.49.1
+appVersion: 0.49.0
```
eromanova, line 6: Shouldn't we bump the `appVersion` as well?

Makefile
```diff
@@ -513,7 +509,7 @@ VELERO_VERSION ?= $(shell go mod edit -json | jq -r '.Require[] | select(.Path =
 VELERO_BACKUP_NAME ?= velero.io_backups
 VELERO_BACKUP_CRD ?= $(EXTERNAL_CRD_DIR)/$(VELERO_BACKUP_NAME)-$(VELERO_VERSION).yaml
 
-SVELTOS_VERSION ?= v$(shell grep 'appVersion:' $(PROVIDER_TEMPLATES_DIR)/projectsveltos/Chart.yaml | cut -d '"' -f 2)
+SVELTOS_VERSION ?= v$(shell grep 'appVersion:' $(PROVIDER_TEMPLATES_DIR)/projectsveltos/Chart.yaml | cut -d ' ' -f 2)
```
eromanova, line 512: The `SVELTOS_VERSION` variable is now used solely to fetch the sveltos external CRDs to bin/crd. I believe it would be more appropriate to use the version of the `projectsveltos-crds` dependency rather than the `appVersion` of the `projectsveltos` chart. What do you think?

eromanova (commented):
LGTM in general

templates/provider/projectsveltos/Chart.yaml
zerospiel: Bumped both chart version and the application version to the `0.50.0`

eromanova (approved):

PR #1180, opened 2025-03-12
Use `sveltosv1beta1.HelmChart` directly

s3rj1k:
closes: https://github.com/k0rdent/kcm/issues/1181

### DIFF
--- a/internal/sveltos/profile.go
+++ b/internal/sveltos/profile.go
@@ -40,7 +40,7 @@ type ReconcileProfileOpts struct {
 	OwnerReference       *metav1.OwnerReference
 	SyncMode             string
 	LabelSelector        metav1.LabelSelector
-	HelmChartOpts        []HelmChartOpts
+	HelmCharts           []sveltosv1beta1.HelmChart
 	TemplateResourceRefs []sveltosv1beta1.TemplateResourceRef
 	PolicyRefs           []sveltosv1beta1.PolicyRef
 	DriftIgnore          []libsveltosv1beta1.PatchSelector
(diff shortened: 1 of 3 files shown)


internal/sveltos/profile.go
```diff
@@ -261,34 +249,7 @@ func GetSpec(opts *ReconcileProfileOpts) (*sveltosv1beta1.Spec, error) {
...
-			// and InsecureSkipTLSVerify were both set, so verify before removing.
-			helmChart.RegistryCredentialsConfig.InsecureSkipTLSVerify = false
-		}
-
-		helmChart.Values = hc.Values
-		helmChart.ValuesFrom = hc.ValuesFrom
-
-		spec.HelmCharts = append(spec.HelmCharts, helmChart)
-	}
+	spec.HelmCharts = append(spec.HelmCharts, opts.HelmChartOpts...)
```
zerospiel, line 252: I would suggest changing the names accordingly (there are no `options` now, these are the full `helmcharts`) across all the package/callers

Also, this particular line is redundant, the helm charts might be effectively set into the `spec` right away (line nr 236):

```go
HelmCharts: make([]sveltosv1beta1.HelmChart, 0, len(opts.HelmChartOpts)),
# HelmCharts: opts.HelmChartOpts,
```

zerospiel (commented):
LGTM, couple small comments. I also ask to add the issue describing the problem/benefits/etc.

internal/sveltos/profile.go
```diff
@@ -261,35 +249,6 @@ func GetSpec(opts *ReconcileProfileOpts) (*sveltosv1beta1.Spec, error) {
...
-				CredentialsSecretRef:  hc.CredentialsSecretRef,
-			},
-		}
-
-		if hc.PlainHTTP {
-			// InsecureSkipTLSVerify is redundant in this case.
-			// At the time of implementation, Sveltos would return an error when PlainHTTP
-			// and InsecureSkipTLSVerify were both set, so verify before removing.
-			helmChart.RegistryCredentialsConfig.InsecureSkipTLSVerify = false
-		}
```
zerospiel, line 285: i haven't noticed it from the first glance, could you add such a statement to the `GetHelmCharts` function as well, somewhere after the 

```go
opt := sveltosv1beta1.HelmChart{ // opt -> helmChart?
// ... 
```

internal/sveltos/profile.go
zerospiel: Yep, thanks. Please add the description or attach the issue so it'd be easy to backtrack the "why" afterwards

internal/sveltos/profile.go
s3rj1k: done

zerospiel (approved):
Waiting for the PR's description
UPD. Done

PR #1193, opened 2025-03-13
Add the workflow for code review by AI

sotpuschennikov:
This workflow is designed to run python script to obtain review code from AI


.github/workflows/ai-codereview.yml
```diff
@@ -0,0 +1,61 @@
...
+    branches: [ "main" ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          repository: Mirantis/ai-ci-scripts
```
zerospiel, line 13: this does not seem to be a [working step](https://github.com/k0rdent/kcm/actions/runs/13845349129/job/38770623599?pr=1193)

.github/workflows/ai-codereview.yml
```diff
@@ -0,0 +1,61 @@
+name: AI code review
+on:
+  pull_request:
+    branches: [ "main" ]
```
zerospiel, line 4: should we execute the jobs on each and every PR? i'd suggest to skip some paths/extensions (e.g. `**.md`, `config/**`, `hack/**`, etc.)

.github/workflows/ai-codereview.yml
```diff
@@ -0,0 +1,61 @@
...
+          MAX_TOKENS: '10000'
+          TEMPERATURE: '0.6'
+          VENV_DIR_LOC: '.venv'
+          LANG: 'go'
+          PATH_TO_AI_CODEREVIEW_SCRIPT: 'ai-ci-scripts/ai-codereview/ai-codereview.py'
+          PATH_TO_JINJA_TEMPLATE_FILE: 'ai-ci-scripts/ai-codereview/prompt_template.j2'
+        run: |
+          #!/bin/bash -e
+          pushd 'ai-ci-scripts'
+          changedFiles=$(git show --pretty='format:' --name-only | grep .py)
```
zerospiel, line 30: Could you elaborate on the command? I do not understand how it actually fetches the changed files, how it behaves in case of multiple commits, and why only python files are being considered if `go` files are to be reviewed?

I would suggest using [tj-actions/changed-files](https://github.com/marketplace/actions/changed-files) for that purpose

.github/workflows/ai-codereview.yml
```diff
@@ -0,0 +1,61 @@
...
+          popd
+          sudo apt install python3-virtualenv
+          python3 -m venv ${VENV_DIR_LOC}
+          set +x
+          source ${VENV_DIR_LOC}/bin/activate
+          pip3 install -q openai jinja2
```
zerospiel, line 36: I am not aware of those particular packages. Is there any chance to use either some actions or cache the packets so that each time, there is no need to download them? Otherwise it is okay to me

zerospiel (changes requested):

.github/workflows/ai-codereview.yml
r0mik: https://adamj.eu/tech/2023/11/02/github-actions-faster-python-virtual-environments/

.github/workflows/ai-codereview.yml
r0mik: @zerospiel JFYI https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised

.github/workflows/ai-codereview.yml
```diff
@@ -0,0 +1,87 @@
...
+        id: codereview
+        if: steps.changed-python-files.outputs.any_changed == 'true'
+        env:
+          AI_API_TOKEN: ${{ secrets.AI_API_TOKEN }}
+          AI_API_MODEL: ${{ secrets.AI_API_MODEL }}
+          AI_API_BASE_URL: ${{ secrets.AI_API_BASE_URL }}
+          MAX_TOKENS: '10000'
+          TEMPERATURE: '0.6'
+          VENV_DIR_LOC: '.venv'
+          LANG: 'python'
```
sotpuschennikov, line 45: need to be changed to Go

.github/workflows/ai-codereview.yml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          python-version: '3.12'
+      - name: Restore cached virtualenv
+        uses: actions/cache/restore@v4
+        with:
+          key: venv-${{ runner.os }}-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('ai-ci-scripts/ai-codereview/requirements.txt') }}
+          path: .venv     
+      - name: Run ai-codereview script
+        id: codereview
+        if: steps.changed-python-files.outputs.any_changed == 'true'
```
sotpuschennikov, line 37: need to be changed to changed-golang-files

.github/workflows/ai-codereview.yml
```diff
@@ -12,32 +17,48 @@ jobs:
         with:
           repository: Mirantis/ai-ci-scripts
           path: ai-ci-scripts
+      - name: Get all changed golang files
+        id: changed-golang-files
+        uses: tj-actions/changed-files@v45
```
zerospiel, line 22: As discussed previously and because of the [security breach](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised), could we depend on a revision rather than the version? Something similar to the `tj-actions/changed-files@<revision> # v45`

.github/workflows/ai-codereview.yml
```diff
@@ -48,14 +69,19 @@ jobs:
                 --lang "${LANG}" \
                 --template-file "${PATH_TO_JINJA_TEMPLATE_FILE}"
           done
-          echo "review_result=$(cat ./message.txt | tr '\n' ' ')" >> $GITHUB_OUTPUT
+          echo "See the code review here ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" >> ./summary.txt
```
zerospiel, line 72: nit: I'd suggest adding a link in the markdown format to beautify the comment: `See the code review [here](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})`

.github/workflows/ai-codereview.yml
s3rj1k: Why not just write inline bash script for that? Why depending on some 3rd party actions that already have bad security history?

zerospiel: That's also okay. The previous version didn't come with a proper collection of changed files through multiple commits. It does not matter.

.github/workflows/ai-codereview.yml
```diff
@@ -0,0 +1,87 @@
+name: AI Codereview
+on:
+  pull_request:
```
zerospiel, line 3: @eromanova could you elaborate on this, we should set `pull_request_target` here as it is in the other jobs, right?

zerospiel (approved):

.github/workflows/ai-codereview.yml
```diff
@@ -0,0 +1,102 @@
+name: AI Codereview
+on:
+  pull_request_target:
+    branches: 
+      - main
```
eromanova, line 5: Maybe include `*-release` branches as well?

zerospiel: Good catch

sotpuschennikov: Done

zerospiel (approved):

zerospiel:
@sotpuschennikov I'm going to drop the WF for now because it does not work as expected, please check out an example of a failed job [here](https://github.com/k0rdent/kcm/actions/runs/14127780220/job/39580670202?pr=1309)

PR #1194, opened 2025-03-14
Add GCP hosted cluster template

### LINKED ISSUE #1149: ClusterTempate for hosted-controlplane GCP

eromanova:
This PR contains the cluster template for the hosted GCP cluster.
It does not contain a CSI component yet. It'll be enabled in the follow-up PR once https://github.com/Mirantis/helm-charts/pull/4 is landed and the CSI helm chart is published.

Closes #1149

### DIFF
--- a/templates/cluster/gcp-hosted-cp/templates/gcpcluster.yaml
+++ b/templates/cluster/gcp-hosted-cp/templates/gcpcluster.yaml
@@ -1,0 +1,18 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: GCPCluster
+metadata:
+  name: {{ include "cluster.name" . }}
+  finalizers:
+    - k0rdent.mirantis.com/cleanup
+spec:
+  project: {{ .Values.project }}
+  region: {{ .Values.region }}
+  network:
+    name: {{ .Values.network.name }}
+    mtu: {{ .Values.network.mtu }}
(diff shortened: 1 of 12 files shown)


templates/cluster/gcp-hosted-cp/templates/gcpcluster.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: GCPCluster
+metadata:
+  name: {{ include "cluster.name" . }}
+  finalizers:
+    - k0rdent.mirantis.com/cleanup
```
zerospiel, line 6: out of curiosity: is the finalizer actually required? are there any orphans after a `gcpcluster` object removal?

zerospiel (approved):

templates/cluster/gcp-hosted-cp/templates/gcpcluster.yaml
eromanova: I didn’t encounter any leftover resources after removing the GCPCluster, but I decided to keep this finalizer as we do for other providers. Just as a precaution in case of any unnoticed issues. Since the controller already includes the necessary cleanup logic, no additional code changes are required.

zerospiel: ok got it thanks

PR #1200, opened 2025-03-17
Update k0s version to v1.32.2+k0s.0

Kshatrix:


eromanova (approved):

zerospiel (commented):
Shouldn't the dev configs also be bumped with the corresponding versions of the cluster templates to avoid `not found` errors during the dev setup?

eromanova (approved):

zerospiel (approved):

Issue #1043, opened 2025-02-05
[enhancement] add capability to deploy services using sources other than HelmChart


BROngineer:
thread: https://miracloud.slack.com/archives/C075Z4V0TA5/p1738743515927559

tl;dr kcm does not support deploying services from git repositories

s3rj1k:
Below **does not** work:
```
cat <<'EOF' | kubectl apply -f -
apiVersion: source.toolkit.fluxcd.io/v1
kind: GitRepository
metadata:
  name: prometheus-community
  namespace: kcm-system
  labels:
    k0rdent.mirantis.com/managed: "true"
spec:
  interval: 10m0s
  url: https://github.com/prometheus-community/helm-charts.git
  ref:
    tag: prometheus-25.8.2
---
apiVersion: k0rdent.mirantis.com/v1alpha1
kind: ServiceTemplate
metadata:
  name: prometheus-25-8-2
  namespace: kcm-system
spec:
  helm:
    chartSpec:
      chart: ./charts/prometheus
      version: 25.8.2
      interval: 10m0s
      reconcileStrategy: ChartVersion
      sourceRef:
        kind: GitRepository
        name: prometheus-community
EOF

kubectl patch clusterdeployment -n kcm-system openstack --type=merge --patch "$(cat <<'EOF'
spec:
  serviceSpec:
    services:
      - template: prometheus-25-8-2
        name: prometheus
        namespace: prometheus
EOF
)"

kubectl get secret -n kcm-system openstack-kubeconfig -o=jsonpath={.data.value} | base64 -d > openstack.kubeconfig

kubectl --kubeconfig openstack.kubeconfig get pods -n prometheus
```

BROngineer:
@s3rj1k sveltos currently does not support helm chart installation from git repo, unforunately

s3rj1k:
How is Sveltos related to this? Charts are handled by Flux components last time I've cheked

BROngineer:
In general - yes. However we create sveltos CRs - Profile/ClusterProfile out of our ServiceTemplates. Sveltos CRs has three fields related to artifacts to be deployed:
- helmCharts
- policyRefs
- kustomizationRefs

helmRefs in turn contains the list of structs where the following fields specified:
```yaml
helmCharts:
  - repositoryURL:    https://kyverno.github.io/kyverno/
    repositoryName:   kyverno
    chartName:        kyverno/kyverno
    chartVersion:     v3.3.3
[...]

PR #1203, opened 2025-03-17
Allow to use sources other than HelmChart in ServiceTemplates

BROngineer:
### Overview

This PR allows to use various sources for services deployment aside from Helm chart.
Apart from that, authorization definition was added to `ServiceTemplate` to allow using private repositories.
Changes were made to only extend the existing resources, to avoid necessity to introduce next API version and implementing of conversion between versions.

### API changes:

- added `serviceTemplate.spec.kustomize` field;
- added `serviceTemplate.spec.resources` field;
- added `serviceTemplate.spec.authorization` field;

`.spec.helm`, `.spec.kustomize` and `.spec.resources` fields are mutual exclusive.

`.spec.kustomize` and `.spec.resources` fields has the same underlying structure, which defines:
- `path` field reflecting the directory containing resources manifests or kustomization;
- `deploymentType` field defining whether resources should be installed locally or to remote cluster;
- `localSourceRef` field defining reference to local resource such as ConfigMap or Secret;
- `remoteSourceSpec` field defining configuration of flux source object such as GitRepository, Bucket or OCIRepository;

`.spec.authorization` field contains references to local Secrets which should contain credentials and/or certificates to be used to access remote source.

`.spec.kustomization` will be added to `.spec.kustomizationRefs` and `.spec.resources` will be added to `.spec.policyRefs` of the projectsveltos `[cluster]profile` object which is created during `ClusterDeployment` or `MultiClusterService` reconciliation.

### ServiceTemplate with kustomization example

Here is an example of service template definition, containing kustomization to fetch from git repository:

```yaml
kind: ServiceTemplate
metadata:
  name: hello-world
  namespace: kcm-system
spec:
  kustomize:
    deploymentType: Remote
    path: ./helloWorld
    remoteSourceSpec:
      git:
        url: https://github.com/gianlucam76/kustomize.git
        ref:
          branch: main
```


s3rj1k:
> > JFYI: Sveltos is bringing [this](https://github.com/k0rdent/kcm/issues/1043) natively, I would hold-back
> 
> @s3rj1k what exactly sveltos brings natively?

https://github.com/projectsveltos/addon-controller/commit/dad4b69cdef867c66395fe54b40ec786c858c38d

BROngineer:
> > > JFYI: Sveltos is bringing [this](https://github.com/k0rdent/kcm/issues/1043) natively, I would hold-back
> > 
> > 
> > @s3rj1k what exactly sveltos brings natively?
> 
> [projectsveltos/addon-controller@dad4b69](https://github.com/projectsveltos/addon-controller/commit/dad4b69cdef867c66395fe54b40ec786c858c38d)

so still unclear what exactly wrong with this PR. Ok, sveltos will support installing helm charts from different sources. Perfect, but are you sure we want to be capable to install **only** helm charts?

internal/sveltos/profile.go
```diff
@@ -226,6 +231,91 @@ func GetHelmCharts(ctx context.Context, c client.Client, namespace string, servi
...
+		tmplRef := client.ObjectKey{Name: svc.Template, Namespace: namespace}
+		if err := c.Get(ctx, tmplRef, tmpl); err != nil {
+			return nil, fmt.Errorf("failed to get ServiceTemplate %s: %w", tmplRef.String(), err)
+		}
+
+		if tmpl.Spec.Kustomize == nil {
+			continue
+		}
+
+		if !tmpl.Status.Valid {
```
wahabmk, line 258: I don't think we should use `.status.valid` here as it is currently only relevant for helmchart source. Maybe we should update template reconciliation so that the valid field is updated for kustomize and resources sources as well if that's feasible.

internal/webhook/template_webhook.go
```diff
@@ -187,7 +187,9 @@ func (*ServiceTemplateValidator) Default(_ context.Context, obj runtime.Object)
 	if !ok {
 		return apierrors.NewBadRequest(fmt.Sprintf("expected ServiceTemplate but got a %T", obj))
 	}
-	setHelmChartDefaults(template.GetHelmSpec())
+	if template.Spec.Helm != nil {
+		setHelmChartDefaults(template.GetHelmSpec())
+	}
```
wahabmk, line 192: The `setHelmChartDefaults` function already [returns if helmSpec == nil](https://github.com/k0rdent/kcm/blob/ae7a42ff853e3b12e67ca8d40b4ebda03a6442b9/internal/webhook/template_webhook.go#L322-L324).

internal/controller/templatechain_controller.go
```diff
@@ -154,7 +154,10 @@ func (r *TemplateChainReconciler) ReconcileTemplateChain(ctx context.Context, te
 				return ctrl.Result{}, fmt.Errorf("type assertion failed: expected ServiceTemplate but got %T", source)
 			}
 			spec := serviceTemplate.Spec
-			spec.Helm = kcm.HelmSpec{ChartRef: serviceTemplate.Status.ChartRef}
+			if spec.Helm == nil {
+				continue
+			}
```
wahabmk, line 159: If we continue if `spec.Helm == nil` then the template will not be created below, even if .spec.Kustomize or .spec.Resources have been specified.

wahabmk: Also do we need to do similar for ClusterTemplate case?

internal/sveltos/profile.go
wahabmk: Yeah the reconciler is setting `.status.valid` within the `updateStatus` function. What I wanted to point out though it that currently `updateStatus` is only being called in relation to helm chart in the [`ReconcileTemplate`](https://github.com/k0rdent/kcm/blob/5f73672be8a5a96f4f3696a50c289120921aa550/internal/controller/template_controller.go#L234) function. So the `.status.valid` field is being set only for helm chart but used here for kustomize (and resources below). Hope that clarifies my concern.

api/v1alpha1/source_types.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	sourcev1 "github.com/fluxcd/source-controller/api/v1"
+	sourcev1beta2 "github.com/fluxcd/source-controller/api/v1beta2"
+)
+
+// EmbeddedGitRepositorySpec is the partial spec for a GitRepository reused from
+// the GitRepositorySpec provided by fluxcd/source-controller API. It reflects all
+// fields defined in the GitRepositorySpec except those containing the references
+// to a Secret with authorization credentials.
```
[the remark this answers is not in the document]
wahabmk: If in the future if we have to use source-specific fields, we can always then create a new custom struct with fields from the original flux struct + our own new fields that we need without breaking the API. Maybe other reviewers can also chime in.

internal/sveltos/profile.go
wahabmk: I missed reviewing the new servicetemplate_reconciler.go file because github did not show it by default 😄. The `.status.valid` is handled in there so this is good.

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+	hashBytes := hasher.Sum(nil)
+	hash := hex.EncodeToString(hashBytes)
+
+	// Use first 8 characters of the hash
+	shortHash := hash[:8]
+	sourceName := fmt.Sprintf("%s-%s", templateNamespacedName.Name, shortHash)
+
+	// Ensure the name doesn't exceed 63 characters (Kubernetes name limit)
+	if len(sourceName) > 253 {
```
wahabmk, line 507: Was this meant to be 63 instead?

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	ctx context.Context,
+	template *kcm.ServiceTemplate,
+	ref *kcm.RemoteSourceSpec,
+) error {
+	// we'll create a new Bucket object for every ServiceTemplate. This is needed because we cannot reuse the
+	// same Bucket objects for multiple ServiceTemplates due to possible difference between bucket parameters.
+	// Since there would be no labels to identify created Bucket objects, they won't be reconciled by the flux
+	// controller. The OwnerReference will be set to the ServiceTemplate object, so that the Bucket object will be
+	// deleted when the ServiceTemplate is deleted.
+	bucketName, err := GenerateSourceName(client.ObjectKeyFromObject(template))
```
wahabmk, line 350: Generating random name works, but just wondering, if there is a 1-to-1 relationship between a servicetemplate and bucket, will it be more user-friendly to generate the name as `<servicetemplate-name>-bucket` instead?

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+
+const conditionReady = "Ready"
+
+// ServiceTemplateReconciler reconciles a *Template object
+type ServiceTemplateReconciler struct {
+	TemplateReconciler
+}
+
+func (r *ServiceTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	l := ctrl.LoggerFrom(ctx)
```
zerospiel, line 54: from the log's perspective, i see a lot of messages, like `Kustomization reconciliation finished` followed by `Resources reconciliation finished` and that is almost all of the messages (apart from the main part of the controller and mutations). In the case of many events (and there will be many of them because of the ownees) these messages will spam the controller suggesting no useful information.

Can we do the following:
- separate debug messages from non-debug with the `V(1)` (that verbosity is enough I think)
- add more log messages (also depending on the importance with either `V(0)` or `V(1)`)

I believe that would provide more information and value on what is going on in general and also some debug information with increased verbosity.

Let me know what you think

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	r.defaultRequeueTime = 1 * time.Minute
+
+	return ctrl.NewControllerManagedBy(mgr).
+		WithOptions(controller.TypedOptions[ctrl.Request]{
+			RateLimiter: ratelimit.DefaultFastSlow(),
+		}).
+		For(&kcm.ServiceTemplate{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Watches(&sourcev1beta2.OCIRepository{}, handler.EnqueueRequestsFromMapFunc(getServiceTemplateForEventSource)).
+		Watches(&sourcev1.GitRepository{}, handler.EnqueueRequestsFromMapFunc(getServiceTemplateForEventSource)).
+		Watches(&sourcev1.Bucket{}, handler.EnqueueRequestsFromMapFunc(getServiceTemplateForEventSource)).
```
zerospiel, line 532: should we limit the events being watched to just `create` and `update`?

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+	template.Status = status
+	return err
+}
+
+func (r *ServiceTemplateReconciler) reconcileRemoteSource(ctx context.Context, template *kcm.ServiceTemplate) error {
+	ref := template.RemoteSourceSpec()
+	if ref == nil {
+		return errors.New("remote source ref is undefined")
+	}
```
zerospiel, line 254: the variable can't be `nil` at this point, or is it only a compliment check for the compiler?

applicable for the local source method

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+		err = fmt.Errorf("unsupported source kind %s", ref.Kind)
+	}
+
+	if err != nil {
+		status.TemplateValidationStatus.Valid = false
+		status.TemplateValidationStatus.ValidationError = err.Error()
+	} else {
+		status.TemplateValidationStatus.Valid = true
+	}
+	template.Status = status
```
zerospiel, line 246: I would say this is the right place for a `defer`, something like

```go
func (r *receiver) methodName(arguments ...struct{}) (err error) {
  defer func() {
    if err != nil { failed; } else { valid; }
    template.Status = status
  }
  
  switch ref.Kind {
  case "Kind1": 
    if err := r.Client.Get(); err != nil { return fmt.Errorf("wrapme: %w", err);} ;
  default:
    return fmt.Errorf("unexpected")
}
```

that way it would be easier to read and no need to preserve the global (to the method) scope of the `err`, no need in the `break` statements, etc. Just plain `return` from each case

Let me know what you think

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	kcm "github.com/K0rdent/kcm/api/v1alpha1"
+	"github.com/K0rdent/kcm/internal/utils"
+	"github.com/K0rdent/kcm/internal/utils/ratelimit"
+)
+
+const conditionReady = "Ready"
+
+// ServiceTemplateReconciler reconciles a *Template object
```
eromanova, line 48:
```suggestion
// ServiceTemplateReconciler reconciles a ServiceTemplate object
```

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+
+func (r *ServiceTemplateReconciler) reconcileLocalSource(ctx context.Context, template *kcm.ServiceTemplate) error {
+	ref := template.LocalSourceRef()
+	if ref == nil {
+		return errors.New("local source ref is undefined")
+	}
+
+	name, namespace := ref.Name, ref.Namespace
+	if namespace == "" {
+		namespace = r.SystemNamespace
```
zerospiel, line 158: it should be the `template`'s namespace

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+			Name:      repositoryName,
+			Namespace: template.Namespace,
+		},
+	}
+	key := client.ObjectKeyFromObject(gitRepository)
+	if err := r.Get(ctx, key, gitRepository); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get GitRepository %s: %w", key, err)
+		}
+	}
```
zerospiel, line 292: redundant call

eromanova, line 292: Is it necessary before using `CreateOrUpdate`? I think it's better to run it before `template.Status.Valid` update (line 333) to get the actual `gitRepository` status.

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+			Name:      bucketName,
+			Namespace: template.Namespace,
+		},
+	}
+	key := client.ObjectKeyFromObject(bucket)
+	if err := r.Get(ctx, key, bucket); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get Bucket %s: %w", key, err)
+		}
+	}
```
zerospiel, line 365: redundant call

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+			Name:      ociRepositoryName,
+			Namespace: template.Namespace,
+		},
+	}
+	key := client.ObjectKeyFromObject(ociRepository)
+	if err := r.Get(ctx, key, ociRepository); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get OCIRepository %s: %w", key, err)
+		}
+	}
```
zerospiel, line 440: redundant call

eromanova (commented):
Minor comments

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
...
+		maxParentNameLen := 253 - 9 // 8 for hash + 1 for hyphen
+		truncatedParentName := templateNamespacedName.Name
+		if len(truncatedParentName) > maxParentNameLen {
+			truncatedParentName = truncatedParentName[:maxParentNameLen]
+		}
+		sourceName = fmt.Sprintf("%s-%s", truncatedParentName, shortHash)
+	}
+
+	return sourceName, nil
+}
```
zerospiel, line 519: The function could be simplified, but I do not understand its goal in the first place: why cannot we just use `template`'s name? There won't be any collisions, because there is 1:1 mapping between service templates and remove objects

Simplification:
```go
func GenerateSourceName(templateNamespacedName client.ObjectKey) string {
	sha := sha256.Sum256([]byte(templateNamespacedName.String()))
	shortSha := hex.EncodeToString(sha[:])[:8]

	sourceName := fmt.Sprintf("%s-%s", templateNamespacedName.Name, shortSha)
	if len(sourceName) <= validation.DNS1123SubdomainMaxLength {
		return sourceName
	}

	charsFromName := validation.DNS1123SubdomainMaxLength - 9 // 8 for hash + 1 for hyphen
	if len(templateNamespacedName.Name) > charsFromName {
		sourceName = fmt.Sprintf("%s-%s", templateNamespacedName.Name[:charsFromName], shortSha)
	}

	return sourceName
}
```

where `validation` is `"k8s.io/apimachinery/pkg/util/validation"`

zerospiel (changes requested):

internal/sveltos/profile.go
```diff
@@ -149,6 +150,10 @@ func GetHelmCharts(ctx context.Context, c client.Client, namespace string, servi
 			return nil, fmt.Errorf("failed to get ServiceTemplate %s: %w", tmplRef.String(), err)
 		}
 
+		if tmpl.Spec.Helm == nil {
```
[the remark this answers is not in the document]
zerospiel: Makes sense. Okay to me but it should be done within the same release version, so as soon as possible

internal/controller/servicetemplate_controller.go
```diff
@@ -0,0 +1,580 @@
+// Copyright 2024
```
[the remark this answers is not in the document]
zerospiel: >  I do not see a big difference in how other stuff is implemented and this new controller

Correct, the controllers have been implemented and we want to start decreasing the scope, dividing the new functionality (new as of "a new controller" for instance) into separate packages. There are too many variables/structures/functions each related to a specific controller but available for other controllers. That makes no sense, only adding extra cognitive complexity. For example, there was a constant for the requeue period used in one of the controllers, and suddenly it started to be used in other controllers and it was erroneous because the defaults of one controller are not defaults for other controllers. 

I do not insist on refactoring this in the PR but insist on doing so at least in a follow-up PR (not necessarily within the release since the functionality would be the same)

api/v1alpha1/source_types.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2024
```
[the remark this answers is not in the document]
zerospiel: Then there is the `templates_common.go` file containing common structures shared by multiple API types

api/v1alpha1/servicetemplate_types.go
```diff
@@ -28,21 +30,202 @@ const (
...
+
+	// Git is the definition of git repository source.
+	Git *EmbeddedGitRepositorySpec `json:"git,omitempty"`
+
+	// Bucket is the definition of bucket source.
+	Bucket *EmbeddedBucketSpec `json:"bucket,omitempty"`
+
+	// OCI is the definition of OCI repository source.
+	OCI *EmbeddedOCIRepositorySpec `json:"oci,omitempty"`
+}
```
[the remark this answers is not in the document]
zerospiel: There is a linter called `fieldalignment` and it actually runs but is suppressed for the `api` directory because the generated top-level structs are "not aligned" according to it. I have just noticed several non-aligned types and align them manually

api/v1alpha1/servicetemplate_types.go
```diff
@@ -28,21 +30,202 @@ const (
...
+	// Name is the name of the remote source.
+	Name string `json:"name"`
+
+	// Namespace is the namespace of the remote source.
+	Namespace string `json:"namespace"`
+
+	// Artifact is the artifact that was generated from the template source.
+	Artifact *sourcev1.Artifact `json:"artifact,omitempty"`
+
+	// ObservedGeneration is the latest generation observed by the controller.
```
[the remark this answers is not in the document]
zerospiel: I meant to clarify the description a bit, sorry for the ambiguity

internal/controller/servicetemplate_controller.go
zerospiel: Up to the implementer

api/v1alpha1/servicetemplate_types.go
```diff
@@ -28,21 +30,202 @@ const (
...
+// +kubebuilder:validation:XValidation:rule="has(self.remoteSourceSpec) ? !has(self.localSourceRef): true",message="LocalSource and RemoteSource are mutually exclusive."
+// +kubebuilder:validation:XValidation:rule="has(self.localSourceRef) || has(self.remoteSourceSpec)",message="One of LocalSource or RemoteSource must be specified."
+
+// +kubebuilder:validation:XValidation:rule="has(self.localSourceRef) ? !has(self.remoteSourceSpec): true",message="LocalSource and RemoteSource are mutually exclusive."
+// +kubebuilder:validation:XValidation:rule="has(self.remoteSourceSpec) ? !has(self.localSourceRef): true",message="LocalSource and RemoteSource are mutually exclusive."
+// +kubebuilder:validation:XValidation:rule="has(self.localSourceRef) || has(self.remoteSourceSpec)",message="One of LocalSource or RemoteSource must be specified."
+
+// ResourceSpec defines the desired state of Resource
+type ResourceSpec struct {
+	// +required
```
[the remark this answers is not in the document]
zerospiel: Discussed in private

zerospiel (approved):
LGTM, some changes to come from a follow-up PR
@eromanova pls consider to review the PR

PR #1219, opened 2025-03-18
chore(ci): add triage labeling and stale workflows

DinaBelova:


DinaBelova:
@zerospiel comparing with reverted https://github.com/k0rdent/kcm/pull/1199 
- added extra check on "needs-triage" label existence before removing it (you hit https://github.com/k0rdent/kcm/pull/1199#issuecomment-2730108019 because you removed it manually together with adding triage-ok and me not handling it)
- added correct permissions on stale workflow (caught it in my test repo)

zerospiel (approved):

PR #1242, opened 2025-03-20
Add template invalidity metric

wahabmk:
# Description

Add a metric to track currently invalid templates. Example:
```
# HELP kcm_template_invalidity Number of invalid templates
# TYPE kcm_template_invalidity gauge
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="aws-hosted-cp-0-1-5",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="aws-standalone-cp-0-1-5",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="azure-aks-0-1-1",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="azure-hosted-cp-0-1-5",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="azure-standalone-cp-0-1-5",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="docker-hosted-cp-0-1-3",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="gcp-hosted-cp-0-1-3",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="gcp-standalone-cp-0-1-3",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="openstack-standalone-cp-0-1-8",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="remote-cluster-0-1-3",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="vsphere-hosted-cp-0-1-5",template_namespace="kcm-system"} 1
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="vsphere-standalone-cp-0-1-5",template_namespace="kcm-system"} 1
```

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -59,7 +59,13 @@ func init() {
 	)
 }
 
-func TrackMetricTemplateUsageSet(ctx context.Context, templateKind, templateName, parentKind string, parent metav1.ObjectMeta, value float64) {
+//nolint:revive // false-positive
+func TrackMetricTemplateUsage(ctx context.Context, templateKind, templateName, parentKind string, parent metav1.ObjectMeta, inUse bool) {
+	var value float64
+	if inUse {
+		value = 1
+	}
+
 	metricTemplateUsage.With(prometheus.Labels{
 		metricLabelTemplateKind:    templateKind,
 		metricLabelTemplateName:    templateName,
@@ -78,7 +84,13 @@ func TrackMetricTemplateUsageSet(ctx context.Context, templateKind, templateName
 	)
 }
 
-func TrackMetricTemplateInvaliditySet(ctx context.Context, templateKind, templateNamespace, templateName string, value float64) {
+//nolint:revive // false-positive
+func TrackMetricTemplateInvalidity(ctx context.Context, templateKind, templateNamespace, templateName string, valid bool) {
+	var value float64
+	if !valid {
+		value = 1
+	}
+
 	metricTemplateInvalidity.With(prometheus.Labels{
 		metricLabelTemplateKind:      templateKind,
 		metricLabelTemplateNamespace: templateNamespace,
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -59,66 +59,36 @@ func init() {
 	)
 }
 
-func TrackMetricTemplateUsageSet(ctx context.Context, templateKind, templateName, parentKind string, parent metav1.ObjectMeta) {
+func TrackMetricTemplateUsageSet(ctx context.Context, templateKind, templateName, parentKind string, parent metav1.ObjectMeta, value float64) {
 	metricTemplateUsage.With(prometheus.Labels{
 		metricLabelTemplateKind:    templateKind,
 		metricLabelTemplateName:    templateName,
 		metricLabelParentKind:      parentKind,
 		metricLabelParentNamespace: parent.Namespace,
 		metricLabelParentName:      parent.Name,
-	}).Set(1)
+	}).Set(value)
 
-	ctrl.LoggerFrom(ctx).V(1).Info("Tracking template usage metric (set to 1)",
+	ctrl.LoggerFrom(ctx).V(1).Info("Tracking template usage metric",
 		metricLabelTemplateKind, templateKind,
 		metricLabelTemplateName, templateName,
 		metricLabelParentKind, parentKind,
 		metricLabelParentNamespace, parent.Namespace,
 		metricLabelParentName, parent.Name,
+		"value", value,
 	)
 }
 
-func TrackMetricTemplateUsageDelete(ctx context.Context, templateKind, templateName, parentKind string, parent metav1.ObjectMeta) {
-	metricTemplateUsage.Delete(prometheus.Labels{
-		metricLabelTemplateKind:    templateKind,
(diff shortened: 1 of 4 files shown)


internal/metrics/metrics.go
```diff
@@ -42,9 +43,19 @@ var metricTemplateUsage = prometheus.NewGaugeVec(
 	[]string{metricLabelTemplateKind, metricLabelTemplateName, metricLabelParentKind, metricLabelParentNamespace, metricLabelParentName},
 )
 
+var metricTemplateInvalidity = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Namespace: kcm.CoreKCMName,
```
eromanova, line 48: Just out of curiosity, is this correct? The value of `kcm.CoreKCMName` is set to `kcm`. Shouldn’t it match the namespace of the k0rdent installation or it's another parameter?

zerospiel: It should not. It is the prom ns, read it as a "prefix" for each metric: `kcm_<metric_name>`

internal/metrics/metrics.go
```diff
@@ -83,3 +94,31 @@ func TrackMetricTemplateUsageDelete(ctx context.Context, templateKind, templateN
 		metricLabelParentName, parent.Name,
 	)
 }
+
+func TrackMetricTemplateInvaliditySet(ctx context.Context, templateKind, templateNamespace, templateName string) {
```
zerospiel, line 98: can these 2 funcs be conjugated into a single? and I just realized that this metric and the previous (usage) are both removed, is that right? from my understanding, this operation deletes the time series, which means there won't be any history, hence no possibility of constructing any data based upon the metrics. i might be wrong though, could you elaborate on that? in any way, this is a tradeoff between history and storage, but in my opinion, the overall cardinality is not enormous (<7) and there won't be dozens of thousands of templates so that should be okay in terms of storage and metrics flattening

wahabmk: My reasoning is that these metrics are only ever set to 1 (binary): one measures the number of invalid templates and the other (in https://github.com/k0rdent/kcm/pull/1190) measures the number of templates currently in use. So if a template is no longer invalid then it is no longer of interest to measure for invalidity. If the values were not binary and within a range (e.g. cpu) then I think setting it to 0 would be better.

So taking the example output in the description of this PR, if the "aws-standalone-cp-0-1-5" becomes valid again, it's entry should be removed instead of it becoming 0 like so:
```sh
kcm_template_invalidity{template_kind="ClusterTemplate",template_name="aws-standalone-cp-0-1-5",template_namespace="kcm-system"} 0
```

Having said this, I am quite new to prometheus/grafana so I am happy to be corrected, like if there will be problems with visualizing the data if I am deleting some labels instead of setting them to 0.

zerospiel: I'd stick to the `Set(0)` for both gauges nevertheless. It is very tricky to [handle missing timeseries](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics) (via the `absent` and a user must know that they are absent which is not a general approach). Besides prometheus could struggle to mark the remove metrics rather than processing them as stale.

I would also suggest thoroughly thinking about the labels we want to put in the metrics, for the invalid gauge having 3 labels is okay, for the usage gauge having 5 labels might potentially be a problem but only for gigantic clusters (dozens of thousands of templates).

To sum up, I vote for the `Set(1)` and `Set(0)` correspondingly.

wahabmk: Yeah I thought about if I can reduce the labels for template usage metric:
```
template_kind, template_name, parent_kind, parent_namespace, parent_name
```
But I am struggling to. Maybe parent_kind can be removed as it will either be ClusterDeployment or MultiClusterService....

zerospiel: I think it is okay, it multiplies the theoretical metrics only by two, so it is not a big deal. The current labels with the given conditions are okay at the moment.

internal/metrics/metrics.go
```diff
@@ -42,44 +43,52 @@ var metricTemplateUsage = prometheus.NewGaugeVec(
...
+
 func init() {
 	metrics.Registry.MustRegister(
 		metricTemplateUsage,
+		metricTemplateInvalidity,
 	)
 }
 
-func TrackMetricTemplateUsageSet(ctx context.Context, templateKind, templateName, parentKind string, parent metav1.ObjectMeta) {
+func TrackMetricTemplateUsageSet(ctx context.Context, templateKind, templateName, parentKind string, parent metav1.ObjectMeta, value float64) {
```
zerospiel, line 62: a small suggestion to consider here and for the invalidity function: will it be ok to replace the `value` parameter with something `inUse bool` (and `isInvalid bool`), it'd mitigate potential misusage (highly likely the linter will fail and it is okay to disable it in this case with the `nolint:<name> // false-positive`)

let me know what you think

eromanova (approved):

PR #1250, opened 2025-03-21
Add GCP GKE cluster template

### LINKED ISSUE #1147: ClusterTemplate for GKE

eromanova:
This PR adds the basic GCP GKE cluster template. 

Closes #1147


templates/cluster/gcp-gke/templates/gcpmanagedcluster.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: GCPManagedCluster
+metadata:
+  name: {{ include "cluster.name" . }}
+  annotations:
+    # This annotation is required to prevent the premature deletion
+    # of GCPManagedCluster. Without it, GCPManagedCluster
+    # may be deleted before the MachinePool, causing the MachinePool
+    # deletion to get stuck due to the missing cluster.
```
zerospiel, line 9: is this valid having the finalizer? the same question goes for the rest of such annotations

eromanova: Yes, it is still valid. The KCM controller removes the blocking finalizer only after all Machines are deleted.
In GKE, MachinePool objects can become orphaned when all cluster objects are removed at once (during HelmRelease removal). Because `GCPManagedCluster` and `GCPManagedControlPlane` are deleted before MachinePools.
One possible solution is to adapt the KCM controller to wait for MachinePools before releasing the cluster object. However, this will not help much since `GCPManagedControlPlane` is also deleted too early. Additionally, the KCM controller currently manages only the Cluster deletion process with this finalizer (not the ControlPlane), and modifying it to handle control plane deletion would require significant changes.

zerospiel: That's clear. The annotation should stay. The question is then out of curiosity: what happens on the release removal, I mean the MP objects would become orphan, correct? And who's responsibility to remove then? Should this behavior be accordingly reflected in the docs?

eromanova: While testing the ClusterDeployment deletion I discovered that yes, the MP objects remain orphaned and CAPI + CAPG are trying to remove them. CAPG controller is responsible for the removal of `GCPManagedMachinePool` and it is giving errors like: GCPManagedControlPlane <namespace/name> does not exist. CAPI controller does the same for `MachinePool` object and it's arguing on missing `Cluster` object. Or something like this (didn't save logs before dropping the management cluster, need another reproduce). I think, the issue should be forwarded to the CAPG project. But I'll take a deeper look at the CAPG project, maybe it was fixed already.

zerospiel (approved):

zerospiel:
@eromanova the annotation could be adjusted in a follow up PR if any

PR #1288, opened 2025-03-26
AccessManagement ctrl rearrangement

### LINKED ISSUE #1234: accessmanagement validations to reconciliations

zerospiel:
* rearrange am reconciler code
* simplify am reconciler code
* unrelated: field alignment and descriptions in other API types

Closes #1234

### DIFF
--- a/internal/controller/accessmanagement_controller.go
+++ b/internal/controller/accessmanagement_controller.go
@@ -38,7 +38,7 @@ type AccessManagementReconciler struct {
 	SystemNamespace string
 }
 
-func (r *AccessManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+func (r *AccessManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	l := ctrl.LoggerFrom(ctx)
 	l.Info("Reconciling AccessManagement")
 
@@ -53,13 +53,8 @@ func (r *AccessManagementReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	accessMgmt := &kcm.AccessManagement{}
 	if err := r.Get(ctx, req.NamespacedName, accessMgmt); err != nil {
-		if apierrors.IsNotFound(err) {
-			l.Info("AccessManagement not found, ignoring since object must be deleted")
-			return ctrl.Result{}, nil
-		}
-
-		l.Error(err, "Failed to get AccessManagement")
-		return ctrl.Result{}, err
+		l.Error(err, "unable to fetch AccessManagement")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
 	if updated, err := utils.AddKCMComponentLabel(ctx, r.Client, accessMgmt); updated || err != nil {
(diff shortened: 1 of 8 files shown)


internal/controller/accessmanagement_controller.go
```diff
@@ -38,58 +40,94 @@ type AccessManagementReconciler struct {
...
+	accessMgmt := &kcm.AccessManagement{}
+	if err := r.Get(ctx, req.NamespacedName, accessMgmt); err != nil {
+		l.Error(err, "unable to fetch AccessManagement")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// NOTE (zerospiel): do we want to allow removal anyway (via an anno)? does still makes sense to block the removal if admission is disabled?
+	if controllerutil.AddFinalizer(accessMgmt, kcm.AccessManagementFinalizer) {
+		if err := r.Client.Update(ctx, accessMgmt); err != nil {
+			l.Error(err, "failed to update Management finalizers")
```
eromanova, line 56:
```suggestion
			l.Error(err, "failed to update Access Management finalizers")
```

internal/controller/accessmanagement_controller.go
```diff
@@ -38,58 +40,94 @@ type AccessManagementReconciler struct {
...
+		accessMgmt.Status.Error = deletionForbidden
 		accessMgmt.Status.ObservedGeneration = accessMgmt.Generation
-		err = errors.Join(err, r.updateStatus(ctx, accessMgmt))
-	}()
+
+		if !equality.Semantic.DeepEqual(accessMgmt.Status, cs) {
+			return ctrl.Result{}, errors.Join(errors.New(deletionForbidden), r.updateStatus(ctx, accessMgmt))
+		}
+
+		return ctrl.Result{}, errors.New(deletionForbidden) // always return error to trigger events to remove the finilizer on mgmt deletion/absence
```
eromanova, line 101:
```suggestion
		return ctrl.Result{}, errors.New(deletionForbidden) // always return error to trigger events to remove the finalizer on mgmt deletion/absence
```

internal/controller/accessmanagement_controller.go
```diff
@@ -38,58 +40,94 @@ type AccessManagementReconciler struct {
...
 	l := ctrl.LoggerFrom(ctx)
 	l.Info("Reconciling AccessManagement")
 
+	accessMgmt := &kcm.AccessManagement{}
+	if err := r.Get(ctx, req.NamespacedName, accessMgmt); err != nil {
+		l.Error(err, "unable to fetch AccessManagement")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// NOTE (zerospiel): do we want to allow removal anyway (via an anno)? does still makes sense to block the removal if admission is disabled?
```
eromanova, line 53: I think, we should still block the AccessManagement deletion until the Management object is not removed. If we allow deletion of the AccessManagement object when Management still exists, there is a case when someone accidentally removed `am` that managed some templatechains or credential distribution. And if this object is lost, the initial state of the `am` spec may be very hard to restore. However, all the template chains should remain (since the reconciler will not be triggered anymore).

Another concern is about accidental removal of `am`. Do we want the reconciler to continue distributing the template chains/credentials in case the `am` object was accidentally removed? If so, we should adapt the reconciler accordingly (currently, it blocks all the reconciliation until the `Management` is not removed if I'm not mistaken). 

@zerospiel What do you think?

zerospiel: >  And if this object is lost, the initial state of the am spec may be very hard to restore. However, all the template chains should remain (since the reconciler will not be triggered anymore).

I would say that is not the problem at all. Since we just won't restore everything, one should create a new `am` object with the same rules. Regarding the remaining chains, why should they even remain if each of them must be owned by a given `am`, in my opinion? After the creation of a new `am` with the same rules, the same template chain objects would appear again.

> Do we want the reconciler to continue distributing the template chains/credentials in case the am object was accidentally removed?

At the moment (before this PR), the `delete` event is not allowed until the `mgmt` exists, any `update` events of an `am` object will reconcile the chains though. With this PR the `delete` (if the webhook is disabled) will mark the object as removed and no further reconciliation would be possible because the `deletionTimestamp` is set only once and cannot be removed. Here we have different behavior with enabled webhook: with the latter, the deletion is not possible hence updates to an `am` object are available, otherwise, the updates make no sense in that matter hence no sense in reconciliation. 

That is why I raised the question: do we even want to block the deletion? If the `am` owned its chains there won't even be such problems — GC will collect all of the objects and it's user's (admin's to be precise) responsibility to control the objects.

To sum up: either leave it as it is right now in the PR or remove the webhook completely in that matter and mark chains to be created as owned.

zerospiel: > But if we agree that accidental object removal is not our problem, then it's ok (otherwise, the user should enable admission webhook, right?)

Correct, I would say that should not concern us if the webhook is disabled. But in this case the behaviour between disabled and enabled webhook will differ which is the opposite of the epic's goal. So the question: do we want to leave deletion block in the controller (as it is in the PR with introducing the "hacky" finalizer)?

eromanova: > do we want to leave the deletion block in the controller (as it is in the PR with introducing the "hacky" finalizer) or just change nothing and allow deletion with disabled webhhok?

Ok, I think we should find the option that suits us better as regards the UX. If we keep this "hacky" finalizer, and the user accidentally removes the `am` this will provide extra complexity since `am` will become non-operational (since chains and creds will no longer be created/updated), the deletion can't be canceled, and `am` can't be removed (without dropping the finalizer manually) to be recreated afterwards.

I'm now more inclined to drop this block with the finalizer from the controller and allow the `am` removal, but I don't like the fact that dependant chains and creds can be undesirably removed. But again, the user should enable the admission to prevent this.

PR #1289, opened 2025-03-26
Several minor fixes

eromanova:
Contains:
* fix objects description for GCP templates
* several minor fixes in config/dev examples
* fixes management components readiness check (collect false condition even if the message is unset)
* do not release ClusterDeployment until Cluster is removed

### DIFF
--- a/internal/controller/management_controller.go
+++ b/internal/controller/management_controller.go
@@ -486,7 +486,7 @@ func getFalseConditions(gp capioperatorv1.GenericProvider) []string {
 		if cond.Status == corev1.ConditionTrue {
 			continue
 		}
-		msg := fmt.Sprintf("condition %s has status:%s", cond.Type, cond.Status)
+		msg := fmt.Sprintf("condition %s is in status %s", cond.Type, cond.Status)
 		if cond.Message != "" {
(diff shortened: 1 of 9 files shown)


zerospiel (approved):

internal/controller/management_controller.go
```diff
@@ -480,11 +480,17 @@ func isProviderReady(gp capioperatorv1.GenericProvider) bool {
...
-			messages = append(messages, cond.Message)
+	conditions := gp.GetStatus().Conditions
+	messages := make([]string, 0, len(conditions))
+	for _, cond := range conditions {
+		if cond.Status == corev1.ConditionTrue {
+			continue
+		}
+		msg := fmt.Sprintf("condition %s has status:%s", cond.Type, cond.Status)
+		if cond.Message != "" {
+			msg += ": " + cond.Message
```
zerospiel, line 491:
```suggestion
		msg := fmt.Sprintf("condition %s is in status '%s'", cond.Type, cond.Status) // or even without single quotes
		if cond.Message != "" {
			msg += ": " + cond.Message
```

nit ui

zerospiel (approved):

PR #1291, opened 2025-03-27
TemplateChains reconciles admission checks

### LINKED ISSUE #1239: templatechains validations to reconciliations

zerospiel:
* add common check to validate template chains
* add status to templatechains
* add debug logs

Closes #1239


internal/controller/templatechain_controller.go
```diff
@@ -113,6 +113,13 @@ func (r *TemplateChainReconciler) ReconcileTemplateChain(ctx context.Context, te
 		return ctrl.Result{}, nil
 	}
 
+	// NOTE (zeropspiel): should we either introduce status with {valid, warnings} or update the chains with some error label?
```
eromanova, line 116: I think, yes, when the webhook is disabled, it makes sense to have a `status` with `valid` and `warnings` fields. Initially, we didn't add it since an invalid chain simply couldn't be created due to admission validation.

api/v1alpha1/templatechain_common.go
```diff
@@ -14,12 +14,25 @@
...
+	"slices"
+)
+
+// TemplateChainSpec defines the desired state of *TemplateChain
 type TemplateChainSpec struct {
 	// SupportedTemplates is the list of supported Templates definitions and all available upgrade sequences for it.
 	SupportedTemplates []SupportedTemplate `json:"supportedTemplates,omitempty"`
 }
 
+// TemplateChainSpec defines the desired state of *TemplateChain
```
eromanova, line 28:
```suggestion
// TemplateChainSpec defines the observed state of *TemplateChain
```
or something like this

eromanova (commented):
One small comment

api/v1alpha1/templatechain_common.go
eromanova, line 28:
```suggestion
// TemplateChainStatus defines the observed state of *TemplateChain
```
or something like this

zerospiel: amended, thanks for noticing

eromanova (approved):

PR #1314, opened 2025-03-28
Handle `Release` object delete request

### LINKED ISSUE #1238: release validations to reconciliations

zerospiel:
* trigger mgmt on release create/delete
* set not ready to mgmt if release is not found
* do not reconcile mgmt if release is not found
* fix admission wh disabled by default
* release ctrl sets release owner ref to kcm-templates
  helm release in order to create a new one when
  a new release is created; this action does not remove
  other components since the mgmt will not be
  reconciled when the release object is absent
* chore: kubebuilder project and redundant defaulters

There is nothing to do with the webhook validations reconciliation. Here goes the same idea as in the #1288: there are currently no finalizers for the `releases` CR and hence it does not make much sense to fully block the reconciliation of a `Release` object and force a user to delete it.

To properly address (re)creation of a `Release` the `Management` object should not be reconciled to avoid removal of the other components. Deletion of a `Release` object will remove the `kcm-templates` `HelmRelease` if the corresponding value `controller.createTemplates` had been set to `true`. (Re)creation of a `Release` object will create a new `kcm-templates` `HelmRelease` properly (re)installing the `providertemplates` without extra frictions to actually re-trigger already existing `kcm-templates` `HelmRelease`. And because the `Management` will wait until there is a `Release` object, no destructive cleanups would be performed.

**Warning**: the new watcher for the `Release` kind is set only if the webhook controller is disabled, otherwise it isn't required.

These changes in behavior would be described in the docs: turning off the webhook controller might change the behavior of the `kcm` but at the same time there is no forcing in deletion of other objects sticking to some predefined (by the controller) order of objects, hence the requirement of the epic still happening 

Closes #1238

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -17,6 +17,10 @@ resources:
   kind: ClusterDeployment
   path: github.com/K0rdent/kcm/api/v1alpha1
   version: v1alpha1
+  webhooks:
+    defaulting: true
+    validation: true
+    webhookVersion: v1
 - api:
     crdVersion: v1
   controller: true
@@ -25,6 +29,10 @@ resources:
   kind: Management
   path: github.com/K0rdent/kcm/api/v1alpha1
   version: v1alpha1
+  webhooks:
+    defaulting: true
+    validation: true
+    webhookVersion: v1
 - api:
     crdVersion: v1
   domain: k0rdent.mirantis.com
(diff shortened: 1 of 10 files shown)


eromanova (approved):

eromanova:
This PR looks good. The handling of Release deletion (as part of [k0rdent/kcm#1238](https://github.com/k0rdent/kcm/issues/1238)) was discussed separately.

eromanova (approved):

PR #1317, opened 2025-03-28
Enable fallback KubeVirt software emulation

s3rj1k:
https://github.com/kubevirt/kubevirt/blob/main/docs/software-emulation.md#software-emulation


eromanova (approved):

DinaBelova:
@s3rj1k can you please elaborate  - do we even want to include CAPK to the list of in-tree providers? My assumption was to add it only as out of tree option to be frank and I missed if there was a decision to land it in-tree.

Issue #1240, opened 2025-03-20
templates (cluster, service, provider) validations to reconciliations


zerospiel:

zerospiel:
JFYI https://github.com/k0rdent/kcm/pull/1337#discussion_r2027190371

PR #1337, opened 2025-04-02
Reconcile MCS and CD validations

zerospiel:
* MCS validates valid refs to servicetemplates
* CD validates valid refs to servicetemplates
* CD validates valid cross-namespace references
* introduce new services references validation condition
* flatten code corresponding to the services conds
* rearrange MCS code for the sake of clarity


internal/controller/clusterdeployment_controller.go
```diff
@@ -505,48 +505,65 @@ func getProjectPolicyRefs(mc *kcm.ClusterDeployment, cred *kcm.Credential) []sve
...
+}
+
 // updateServices reconciles services provided in ClusterDeployment.Spec.ServiceSpec.
 func (r *ClusterDeploymentReconciler) updateServices(ctx context.Context, cd *kcm.ClusterDeployment) (_ ctrl.Result, err error) {
 	l := ctrl.LoggerFrom(ctx)
 	l.Info("Reconciling Services")
 
+	r.initServicesConditions(cd)
+
+	{
```
wahabmk, line 544: Why these extra curly braces?

zerospiel: To declare the vars in this particular block of the code without affecting the rest of the method(s)

internal/controller/clusterdeployment_controller.go
```diff
@@ -505,48 +505,65 @@ func getProjectPolicyRefs(mc *kcm.ClusterDeployment, cred *kcm.Credential) []sve
...
+		nsErr := validation.ClusterDeployCrossNamespaceServicesRefs(ctx, cd)
+		tplErr := validation.ServicesHaveValidTemplates(ctx, r.Client, cd.Spec.ServiceSpec.Services, cd.Namespace)
+		merr := errors.Join(nsErr, tplErr)
+		r.setCondition(cd, kcm.ServicesReferencesValidationCondition, merr)
+		if merr != nil {
+			l.Error(merr, "failed to validate services, will not retrigger this error")
```
wahabmk, line 550: I am thinking what if:

1. User creates CD referencing `myapp` ServiceTemplate, which has `.status.valid=false`.
2. Then afterwards, the `myapp` ServiceTemplate becomes valid.

In this scenario, the CD would not get re-triggered even when the ServiceTemplate is now valid.

zerospiel: that is to be done within the #1240 but I could do this as part of this PR, what would you prefer having the context?

zerospiel: To be precise, the idea is in the [mgmt controller](https://github.com/k0rdent/kcm/blob/main/internal/controller/management_controller.go#L951-L960) here

internal/controller/multiclusterservice_controller.go
```diff
@@ -83,43 +84,37 @@ func (r *MultiClusterServiceReconciler) Reconcile(ctx context.Context, req ctrl.
...
+func (*MultiClusterServiceReconciler) initServicesConditions(mcs *kcm.MultiClusterService) (changed bool) {
+	for _, typ := range [3]string{kcm.SveltosClusterProfileReadyCondition, kcm.FetchServicesStatusSuccessCondition, kcm.ServicesReferencesValidationCondition} {
+		if apimeta.SetStatusCondition(&mcs.Status.Conditions, metav1.Condition{
+			Type:               typ,
+			Status:             metav1.ConditionUnknown,
+			Reason:             kcm.ProgressingReason,
+			ObservedGeneration: mcs.Generation,
+		}) {
+			changed = true
+		}
```
wahabmk, line 96: Why not just return?
```suggestion
		return apimeta.SetStatusCondition(&mcs.Status.Conditions, metav1.Condition{
			Type:               typ,
			Status:             metav1.ConditionUnknown,
			Reason:             kcm.ProgressingReason,
			ObservedGeneration: mcs.Generation,
		})
```

zerospiel: We have to change all of the conditions, not just some of them

wahabmk: Oh yeah, missed that.

wahabmk (approved):

PR #1349, opened 2025-04-03
Update capa version to v2.8.2

Kshatrix:


zerospiel (approved):
According to the [v2.8.1](https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/tag/v2.8.1) we should be good to go with the update since we already have the `machineset.cluster.x-k8s.io/skip-preflight-checks` anno set and have updated the CAPI version to 1.9.x long before the CAPA bump

PR #1354, opened 2025-04-04
Change the e2e-tests to a reusable structure

### LINKED ISSUE #426: DRY GitHub actions workflow with a matrix
Right now our workflow has 3 e2e jobs that basically run the same workflow with a couple minor differences, notably:

* `env.GINKGO_LABEL_FILTER`
* `runs-on`
* env blocks for different providers

We should ideally refactor the workflow into a [matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) to simplify the workflow and make maintaining each workflow easier.

Once we have a matrix we can further break out the jobs into different providers, so that each provider test runs concurrently with their own statuses.

nwneisen:
Closes #426 
A continuation of #1326


nwneisen:
@eromanova I had to recreate this PR. I was able to test a default controller run on my fork but I didn't manage to trigger the cloud runs. I'm still trying a couple things but please take a look otherwise.

PR #1360, opened 2025-04-08
ci: build binaries on each merge event

va4st:
Introduce binaries build on each push to provide artifacts for testing


.github/workflows/post-merge-push.yaml
```diff
@@ -0,0 +1,61 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.vars.outputs.version }}
+    permissions:
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ''
```
nwneisen, line 24: The checkout workflow will default to the code that triggered the action. This isn't needed and leaves some ambiguity if it is still checking out the default code or doing something else with an empty string.

nwneisen (commented):
Instead of copying the build job in `.github/workflows/build_test.yml`, why not make it a reusable workflow? That way you know the code is being built the same in a PR and once it is merged into main.

va4st:
@nwneisen is there a need to reuse post-merge workflow? This wf will only be triggered when the code merged. So, there no chance to push something that not merged into repo. 
Code for the PR will not be built each time. It requires specific label to build and test.

nwneisen:
@va4st It depends how many bugs you like to look for. To be clear, I'm referring to the building of the binary, not the pushing of the image and chart.

You've now introduced a second path with potentially different args for the build. I imagine the release will eventually get its own build process also. Someone will come along and only change one of them resulting in a different binary that passes in a PR but not in main, and even worse, not for a release.

va4st:
@nwneisen I can't actually understand the concern.
1. This wf will be used only for the merges inside main branch
2. This wf will push binaries only in CI repo
3. Building is the same as in build_test
4. And not, wf it will not same - inside build_test few differences, related to variables that needed for tests. This wf is not customisable with variables, so it will not be reused in build_test.

s3rj1k:
> reusable workflow

One thing that is not nice with that, logs are ugly

s3rj1k:
> throwing all of the logs for each step into one job

yea, was referring to this

va4st:
@nwneisen ok, more differences
```yaml
jobs:
  build:
      - name: Checkout repository
        uses: actions/checkout@v4
        with:
          ref: ${{fromJSON(steps.pr.outputs.result).merge_commit_sha}} <<< in post-merge-push we does not have pr at all.
          fetch-depth: 0
      - name: Setup Go
        uses: actions/setup-go@v5
        with:
          go-version-file: 'go.mod'
          cache: true # default
      - name: Set up Buildx
        uses: docker/setup-buildx-action@v3
      - name: Login to GHCR
        uses: docker/login-action@v3.4.0
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
      - name: Get outputs
        id: vars
        run: |
          GIT_VERSION=$(git describe --tags --always)
          echo "version=${GIT_VERSION:1}" >> $GITHUB_OUTPUT
          echo "clusterprefix=ci-$(date +%s | cut -b6-10)" >> $GITHUB_OUTPUT <<< That was noted by you
      - name: Build and push KCM controller image to the public repository
        uses: docker/build-push-action@v6
        if: ${{ env.PUBLIC_REPO == 'true' }} <<< This step is mandatory
        with:
          build-args: |
            LD_FLAGS=-s -w -X github.com/K0rdent/kcm/internal/build.Version=${{ steps.vars.outputs.version }}
          context: .
          platforms: linux/amd64
          tags: |
            ghcr.io/k0rdent/kcm/controller-ci:${{ steps.vars.outputs.version }}
          push: true
          cache-from: type=gha
          cache-to: type=gha,mode=max
      - name: Prepare and push KCM template charts to the public repository
        if: ${{ env.PUBLIC_REPO == 'true' }} <<< And this step mandatory also
        run: |
          make kcm-chart-release
          make helm-push
```
I added <<< to mention the changes. 

By some reason you does not mentioning that steps like Lint and getting PR ref also does not exist in post-merge-push.
So, generally I would agree that we doing actually the same. But besides, we the way we doing - differs. Some steps removed at all. build_test requires authorize, post-merge-push not requires. Ofc it can be parametrised and be reusable but it's not the case when it should be done imo.

nwneisen:
@va4st I think you are confusing my suggestion to make a workflow step reusable with making the entire workflow trigger for both situations. 

Linting, unit tests, and authorize would stay in the parent workflow and pass values to the called step workflows. Though I think linting and unit tests should be made reusable also.

This lets you build up parent workflows that just handle the triggers in different situations but run consistent builds and testing with the reusable workflow steps:

``` Merge.yml
name: Merge

on:
  push:
    branches:
      - main

jobs:
  setup-vars:
    uses: MirantisContainers/shared-workflows/.github/workflows/vars.yml@main
    with:
      env-config-path: .github/development.env # No values are actually used ATM
  vet:
    uses: ./.github/workflows/vet.yml
  static-analysis:
    uses: ./.github/workflows/static-analysis.yml
  unit-test:
      uses: ./.github/workflows/unit.yml
  integration:
    uses: ./.github/workflows/integration.yml
    secrets: inherit
    with:
      checkout_ref: ${{ github.ref }}
      cni_provider: "calico"
  build-and-publish-to-github:
    if: ${{ always() && contains(join(needs.*.result, ','), 'success') }} # if all `needs` jobs are successful
    needs: [setup-vars, vet, static-analysis, unit-test, integration]
    uses: ./.github/workflows/build-and-publish-to-github.yml
    secrets: inherit
    with:
      version: ${{ needs.setup-vars.outputs.version }}
      commit: ${{ needs.setup-vars.outputs.commit }}
      date: ${{ needs.setup-vars.outputs.date }}
```

``` Release.yml
name: Release

on:
  release:
    types: [prereleased]

jobs:
  setup-vars:
    uses: MirantisContainers/shared-workflows/.github/workflows/vars.yml@main
    with:
      env-config-path: .github/development.env # No values are actually used ATM
  vet:
    uses: ./.github/workflows/vet.yml
  static-analysis:
    uses: ./.github/workflows/static-analysis.yml
  unit-test:
    uses: ./.github/workflows/unit.yml
  integration:
    uses: ./.github/workflows/integration.yml
    secrets: inherit
    with:
      checkout_ref: ${{ github.ref }}
      cni_provider: "calico"
  test-migration:
    needs: [vet, static-analysis, unit-test]
    uses: ./.github/workflows/test-migration.yml
    secrets: inherit
    with:
      checkout_ref: ${{ github.ref }}
      platforms: '["ubuntu_22.04", "rhel_8.10", "rocky_9.4"]'
  build-and-publish-to-github:
    if: ${{ always() && contains(join(needs.*.result, ','), 'success') }} # if all `needs` jobs are successful
    needs: [setup-vars, vet, static-analysis, unit-test, integration, test-migration]
    uses: ./.github/workflows/build-and-publish-to-github.yml
    secrets: inherit
    with:
      version: ${{ needs.setup-vars.outputs.version }}
      commit: ${{ needs.setup-vars.outputs.commit }}
      date: ${{ needs.setup-vars.outputs.date }}
```

It also lets you setup a test setup so that changes to the callable workflows can be tested without having to merge into main to actually run them.

.github/workflows/post-merge-push.yaml
```diff
@@ -0,0 +1,58 @@
...
+  build-and-push:
+    concurrency:
+      group: build-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
+    name: Build and Unit Test
+    env:
+      REGISTRY_REPO: oci://ghcr.io/k0rdent/kcm/charts-ci
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.vars.outputs.version }}
```
a13x5, line 17: This is extra. Since you have a single job here.

.github/workflows/post-merge-push.yaml
```diff
@@ -0,0 +1,58 @@
...
+    permissions:
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+          cache: true # default
```
a13x5, line 27: You don't need that. Build runs inside docker

.github/workflows/post-merge-push.yaml
```diff
@@ -0,0 +1,58 @@
...
+        uses: docker/login-action@v3.4.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Get outputs
+        id: vars
+        run: |
+          GIT_VERSION=$(git describe --tags --always)
+          echo "version=${GIT_VERSION:1}" >> $GITHUB_OUTPUT
```
a13x5, line 42: Arguably `${GIT_VERSION#v}` better shows the intention here

.github/workflows/post-merge-push.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - main
+
+jobs:
+  build-and-push:
+    concurrency:
+      group: build-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
+    name: Build and Unit Test
+    env:
+      REGISTRY_REPO: oci://ghcr.io/k0rdent/kcm/charts-ci
```
a13x5, line 14: NIT: it's better to explicitly put env to the step which uses it, right now it's the `Prepare and push KCM template charts to the public repository` step only

.github/workflows/post-merge-push.yaml
```diff
@@ -0,0 +1,58 @@
...
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-and-push:
+    concurrency:
+      group: build-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
```
a13x5, line 11: This is extra, since you only have only a `build-and-push` job

.github/workflows/post-merge-push.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Set up Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Login to GHCR
+        uses: docker/login-action@v3.4.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Get outputs
+        id: vars
```
a13x5, line 39: NIT: Name like `Get version` with id `version` looks much nicer

.github/workflows/post-merge-push.yaml
```diff
@@ -0,0 +1,58 @@
...
+          platforms: linux/amd64
+          tags: |
+            ghcr.io/k0rdent/kcm/controller-ci:${{ steps.vars.outputs.version }}
+          push: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+      - name: Prepare and push KCM template charts to the public repository
+        run: |
+          make kcm-chart-release
+          make helm-push
```
a13x5, line 58: Here we're hoping that the `VERSION` variable will be set in the same way in Makefile as it was set in the `vars` step. I would like to give up this hope and set VERSION env here to the `${{ steps.vars.outputs.version }}`

Also we're building `ghcr.io/k0rdent/kcm/controller-ci` image, but it looks like chart will be build with the default `ghcr.io/k0rdent/kcm/controller`, since we're only patching the tag version. Extra image name patch is needed here

a13x5 (changes requested):
Mostly pointed a stuff to remove. Some nit comments (I know the parts of the workflow were mostly copied from the other one, but I don't think we need to copy weird parts)

Makefile
```diff
@@ -9,6 +9,8 @@ IMG_REPO = $(shell echo $(IMG) | cut -d: -f1)
 IMG_TAG = $(shell echo $(IMG) | cut -d: -f2)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.32.0
+# Default image repo for kcm values during build: image.repository
+KCM_CHART_IMAGE_REPO ?= ghcr.io/k0rdent/kcm/controller
```
a13x5, line 13: Since this Makefile is quite busy already let's not introduce new variable, but reuse the `IMG_REPO`
I propose to not modify `set-kcm-version` target, but add new target `set-kcm-repo` with:
```makefile
$(YQ) eval '.image.repository = "$(IMG_REPO)"' -i $(PROVIDER_TEMPLATES_DIR)/kcm/values.yaml
```
and call it separately from the workflow along with `IMG` (yes the one with a full image name) env variable populated.

The rationale here is that we use `set-kcm-version` as a part of release process and it will make it little bit more confusing, since all `.image` values literally defined in the kcm chart will be overwritten.

So ultimately: I consider usage of `IMG` variable optional (since it's not ideal as well), but separate target is a must.

a13x5 (commented):
Added note regarding Makefile changes.
The workflow looks good.

a13x5 (approved):
LGTM

PR #1362, opened 2025-04-08
feat: ClusterDeployment reconciles validations

zerospiel:
* add servicetemplates watcher to cld controller
  and mcs controller if validations are turned off
* add clustertemplate watcher to cld controller
  if validations are turned off
* mcs controller reconciles validations if they are turned off
* cld controller reconciles validations if they are turned off
  - k8s compatibility failure results in template ready cond
  - credentials validation failure results in creds ready cond
* cld controller: if validation WH is turned off and validation
  has failed it won't be retriggered due to watchers
* cld controller: if validation WH is turned off and cluster template
  is not valid it won't be retriggered due to clustertemplate watcher
* fixed always-ready creds condition in cld

NOTES:

* the mutations are not reconciled
* the upgrade sequence validation is not reconciled

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -40,12 +40,14 @@ import (
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	crcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	kcm "github.com/K0rdent/kcm/api/v1alpha1"
 	"github.com/K0rdent/kcm/internal/helm"
@@ -76,6 +78,8 @@ type ClusterDeploymentReconciler struct {
 	SystemNamespace string
 
 	defaultRequeueTime time.Duration
+
+	IsDisabledValidationWH bool // is webhook disabled set via the controller flags
 }
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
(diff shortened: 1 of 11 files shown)


internal/controller/clusterdeployment_controller.go
```diff
@@ -215,109 +210,104 @@ func (r *ClusterDeploymentReconciler) reconcileUpdate(ctx context.Context, cd *k
...
 	// template is ok, propagate data from it
 	cd.Status.KubernetesVersion = clusterTpl.Status.KubernetesVersion
 
-	apimeta.SetStatusCondition(cd.GetConditions(), metav1.Condition{
-		Type:    kcm.TemplateReadyCondition,
-		Status:  metav1.ConditionTrue,
-		Reason:  kcm.SucceededReason,
-		Message: "Template is valid",
-	})
+	var cred *kcm.Credential
```
eromanova, line 234: nit: Do we really need to initialize `cred` var here? It's used at line 255, and then at line 297 it's reassigned

internal/controller/clusterdeployment_controller.go
```diff
@@ -215,109 +210,104 @@ func (r *ClusterDeploymentReconciler) reconcileUpdate(ctx context.Context, cd *k
...
+		l.Info("Validating ClusterDeployment upgrade sequence")
+		upgrades, err := r.getAvailableUpgrades(ctx, clusterTpl)
+		if err != nil {
+			l.Error(err, "failed to get available upgrades")
+			return ctrl.Result{}, err
+		}
+
+		var templateErr error
+		if !slices.Contains(upgrades, cd.Spec.Template) {
+			templateErr = fmt.Errorf("the upgrade sequence is not allowed: ClusterDeployment %s cannot be upgraded to %s", client.ObjectKeyFromObject(cd), cd.Spec.Template)
```
eromanova, line 245: I'm afraid the upgrade sequence validation will not work properly in the current implementation if the webhook is disabled. Right now, we don’t verify whether `spec.template` has changed. Instead, the controller always checks if the current template is available for an upgrade.
What happens if I haven’t configured any `ClusterTemplateChain`, and no upgrades are available? In that case, the controller will continuously fail, since the initial template isn’t present in the list of valid upgrade targets.
Or am I mistaken?

eromanova: Also, I guess we should add a watcher on ClusterTemplateChains as well (to catch the situation when the upgrade sequence for a particular template is updated). What do you think?

eromanova: Another concern is that we probably shouldn't validate k8s compatibility and other checks if the cluster is already deployed. What do you think?

eromanova (commented):
Could you please rename your PR to be standardized according to https://www.conventionalcommits.org/en/v1.0.0-beta.2/#summary?
```
feat: ClusterDeployment reconciles validations
```

internal/controller/clusterdeployment_controller.go
zerospiel: this is done to avoid double-fetch, if the webhook is disabled a `Credential` object is being fetch and assigned to the variable during the validation and then reused to extract identity reference

internal/controller/clusterdeployment_controller.go
zerospiel: You are right about checking the new values rather than the old ones, it should be reworked. The problem is that I am not sure in which way: either we just drop this check or we validate a new (the current one) template reference against the `status` if it is not empty. The second option sounds preferable. If the latter is okay then the chains must be watched, that is correct 

> if the cluster is already deployed

The current implementation of the WH does not take this into account, should we also change the approach? What is the definition of the "already deployed Cluster"? Is it cluster with any `status` and without `dryRun` set, or is it one with the ready status?

zerospiel: Removed upgrade sequence validations since they might be ambiguous and error-prone. Left the compatibility check as it is unclear how to determine whether a cluster has already been deployed and how to distinguish updates when it has already been deployed

eromanova: Discussed offline. Decided to drop the upgrade sequence validation from the reconciler.

eromanova (approved):

PR #1363, opened 2025-04-08
fix: Ensure nil registry credentials config when no values are set

byDimasik:
Resolves https://github.com/k0rdent/2a-internal/issues/75

### DIFF
--- a/internal/sveltos/profile_test.go
+++ b/internal/sveltos/profile_test.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"testing"
 
+	fluxcdmeta "github.com/fluxcd/pkg/apis/meta"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


internal/sveltos/profile_test.go
```diff
@@ -44,3 +46,58 @@ func Test_priorityToTier(t *testing.T) {
 		})
 	}
 }
+
+func Test_emptyRegistryCredentialsConfig(t *testing.T) {
```
zerospiel, line 50: both test functions could be flattened into a single

zerospiel (commented):
lgtm but someone else from the ksm should also take a look

a13x5 (commented):
LGTM, but please rebase and resolve all comments before merging

internal/sveltos/profile_test.go
byDimasik: I would like to keep them separate for easier maintenance. There are scenarios when we expect this config to be nil, and all such scenarios should be covered in a separate test for better visibility. The whole bug was about the config not being nil when it's supposed to be, so, having separate tests for nil and non-nil scenarios ensures we can easily cover all nil cases in one place.

byDimasik:
@zerospiel @BROngineer @a13x5 

Thank you for the reviews! I've addressed all the comments and rebased main to resolve conflicts. Could you please take another look and put a ✅ if you have no more questions so I can proceed with merging? Thanks again!

zerospiel (approved):

eromanova (approved):

a13x5 (approved):

PR #1365, opened 2025-04-09
style: add PR template

zerospiel:


zerospiel:
The instructions still must persist I assume. The checkboxes and other stuff might clog the comment itself. More or less projects (the huge ones) have either a set of instructions to be deleted or the same combined with simple **bold** titles like **Why do we need this PR?**. And that is all. With checkboxes like "I followed the commit guidelines; I followed the 76 chars guideline" and so forth it won't look clear.

Though I am open to adding those **bold** titles if they make sense in that matter

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,17 @@
...
+* The first PR comment (this one) clearly describes what it does and why we need it
+* A new Template or updates to an existing one should be followed
+  with calling `make generate-all` to ensure proper updates;
+  corresponding changes should be reflected in the `config/dev` directory
+* A reference to a corresponding issue to be fixed or updated
+  added via either `Fixes #1234` or `Updates #1234` (the latter if this is not a complete fix)
+* If referring to a repo other than `k0rdent/kcm` add the `owner/repo#issue_number` syntax:
+  `Fixes k0rdent/kof#1234`
+* Fixes or updates to issues might be chained: `Fixes #1234, fixes #1235`
+* Delete these instructions once you have read and applied them
```
eromanova, line 17: I think using checkboxes is a better approach compared to deleting the instructions after reading them (feels more developer-friendly for me 🙂 ) . We should definitely keep these items, but present them as guidance (for example, as a comment here in the template) and then follow it with something like this:

## Summary

`<!-- Describe your changes here in a clear, concise way. -->`

##  Related Issues

`<!-- Link any related issues here -->`
Fixes #

## Checklist

- [ ] I've read the CONTRIBUTING guide, and my PR follows it
- [ ] I've run `make templates-generate lint test`
- [ ] I’ve added or updated tests
etc.

## Notes to Reviewer

`<!-- Anything you'd like the reviewer to know, focus on, or questions you have -->`

zerospiel: I got it. Still, these MD comment lines need to be removed.

Unfortunately, we do not have a CONTRIBUTING guide at the moment, so it feels like it should be provided beforehand, right?

zerospiel: Anyway, I'd suggest starting with something at the moment to avoid empty PRs or without issues or without generation and etc. 

We could introduce contributing guide as a next step and then change the template accordingly introducing the checkboxes and etc. in general I am not against such an approach

zerospiel: oh sorry, your comment was at the same time as mine 😄 

So what is the final suggestion then?

eromanova: > I'd suggest starting with something at the moment to avoid empty PRs

I totally agree.

> So what is the final suggestion then?

Ok, I’m good with moving forward with the current approach until we have a properly defined contributing guide.

PR #1376, opened 2025-04-11
improvement: process absent *Templates

### LINKED ISSUE #1240: templates (cluster, service, provider) validations to reconciliations

zerospiel:
* fix cld controller deletion logic to avoid dropping the finalizer before the rest of the cleanup
* small refactor in the cld ctrl

If admission validation webhook is turned off:

* mgmt ctrl processes absent ProviderTemplates and does not retrigger
* mcs ctrl processes absent ServiceTemplates and does not retrigger
* cld ctrl processes absent ServiceTemplates and ClusterTemplates and does not retrigger
* fix incorrect requests in the servicetemplates watcher in the mcs ctrl
* cld ctrl now watches creation of Cluster and ServiceTemplates
* cld ctrl does not proceed deletion if ClusterTemplate is missing stopping the finalizer deletion

Fixes #1240

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -61,7 +61,10 @@ import (
 	"github.com/K0rdent/kcm/internal/utils/validation"
 )
 
-var ErrClusterNotFound = errors.New("cluster is not found")
+var (
+	errClusterNotFound         = errors.New("cluster is not found")
+	errClusterTemplateNotFound = errors.New("cluster template is not found")
+)
 
 type helmActor interface {
 	DownloadChartFromArtifact(ctx context.Context, artifact *sourcev1.Artifact) (*chart.Chart, error)
@@ -101,7 +104,7 @@ func (r *ClusterDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	if !clusterDeployment.DeletionTimestamp.IsZero() {
 		l.Info("Deleting ClusterDeployment")
-		return r.Delete(ctx, clusterDeployment)
+		return r.reconcileDelete(ctx, clusterDeployment)
 	}
 
 	management := &kcm.Management{}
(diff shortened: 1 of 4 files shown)


internal/controller/clusterdeployment_controller.go
```diff
@@ -665,85 +677,96 @@ func (r *ClusterDeploymentReconciler) Delete(ctx context.Context, cd *kcm.Cluste
...
-		if err = r.Client.Get(ctx, client.ObjectKeyFromObject(cd), cluster); apierrors.IsNotFound(err) {
-			l.Info("Removing Finalizer", "finalizer", kcm.ClusterDeploymentFinalizer)
-			if controllerutil.RemoveFinalizer(cd, kcm.ClusterDeploymentFinalizer) {
-				if err := r.Client.Update(ctx, cd); err != nil {
-					return ctrl.Result{}, fmt.Errorf("failed to update clusterDeployment %s/%s: %w", cd.Namespace, cd.Name, err)
-				}
-			}
-			l.Info("ClusterDeployment deleted")
+
+		l.Info("HelmRelease still exists, retrying")
```
eromanova, line 695: With the current implementation, the kcm-controller will not continue the deletion reconciliation when the HelmRelease still exists. HelmRelease will not be deleted until either the timeout is reached or all the objects are deleted. The `provider cluster` will not be deleted until the `releaseProviderCluster` function is called. So, here is a circular dependency:
```
HelmRelease is not deleted until CAPI Cluster exists <-> CAPI Cluster will not be released until HelmRelease is gone
```
and the deletion will be stuck until the HelmRelease is gone due to timeout.

eromanova (commented):
LGTM, except for 1 moment about ClusterDeployment deletion logic (resolved).
Also, @wahabmk @kylewuolle @BROngineer could you please also review the services deletion logic improvements?

internal/controller/clusterdeployment_controller.go
zerospiel: So that is the previous implementation and the current one (as of this PR), right? Because the previous one was performing the same loop depending on a `HelmRelease` object. Hence the question: what should be done first?

eromanova: Yes, the previous implementation wasn’t entirely correct, but the controller initially didn’t stop further reconciliation if the HelmRelease still existed. This part is mostly correct. But, when the HelmRelease still exists, we shouldn’t return early. Instead, we should continue with the next steps.

zerospiel: I got it! Fixed. Thanks for pointing that out, I haven't been aware of such a dependency.

zerospiel: There is one thing though that concerns me: if `Machine` objects still exist (CAPI) then we continue the execution but I think that we should not continue it and throw an error (probably a sentinel one) hence trying to release the provider's cluster at first, shouldn't we?

eromanova: No, I don't think we should throw an error: the controller will be stuck. The circular dependency still exists: now, the `releaseProviderCluster` is called before the HelmRelease deletion. But the Machines will never be triggered for the deletion until the HelmRelease is triggered for the deletion. And we will never get to see machines' deletion happen since it'll never be triggered for the deletion. So, we should continue executing the next steps.

zerospiel:
[Results](https://github.com/k0rdent/kcm/actions/runs/14444509815/job/40502111151?pr=1376)

eromanova (approved):

PR #1394, opened 2025-04-17
feat: implement PluggableProvider for dynamic infrastructure provider management

s3rj1k:
This PR introduces a new `PluggableProvider` CRD that allows dynamic registration and management of CAPI infrastructure providers.

The change moves away from hardcoded provider definitions in the `/providers` directory to a CRD-based approach with a dedicated controller and updates the `Management` object to reflect a need for ad-hoc provider registration and configuration.

This effectively gives out-of-tree provider UX as follows:

1. Install k0rdent via Helm
2. Wait for Management object readiness
3. Install out-of-tree provider chart (foobar-pp)
4. Wait for Management and provider specific object readiness
5. Ready to use out-of-tree provider

Provider charts example list:

- cluster-api-provider-foobar (similar to what is in-tree)
- foobar-standalone-cp (similar to what is in-tree)
- foobar (provider specific chart for runtime, is installed as dependency from cluster-api-provider-foobar)
- foobar-pp (entrypoint chart, contains `HelmRepository`, `ClusterTemplate`, `ProviderTemplate`, `PluggableProvider`, can be replace by plain manifest and installed via `kubectl`)


api/v1alpha1/management_types.go
```diff
@@ -137,8 +137,12 @@ type ManagementStatus struct {
 	BackupName string `json:"backupName,omitempty"`
 	// Release indicates the current Release object.
 	Release string `json:"release,omitempty"`
+
+	// RequestedProviders holds all requested CAPI providers.
+	RequestedProviders []Provider `json:"requestedProviders,omitempty"`
```
zerospiel, line 142: Why do we need another field? What is the difference between available in this context? Can we still request providers via the spec? 

minor: cannot it be just a `[]string` rather than the current slice of structs?

api/v1alpha1/pluggableprovider_types.go
```diff
@@ -0,0 +1,96 @@
...
+	// ClusterAPIProviderLabel is the label used to identify cluster API providers
+	ClusterAPIProviderLabel = "k0rdent.mirantis.com/capi-provider"
+	// ClusterAPIProviderOverrideAnnotation is the annotation used to override the default cluster API provider name
+	ClusterAPIProviderOverrideAnnotation = "k0rdent.mirantis.com/capi-provider-override"
+)
+
+// GroupVersionKind unambiguously identifies a kind. It doesn't anonymously include GroupVersion
+// to avoid automatic coercion. It doesn't use a GroupVersion to avoid custom marshalling
+// Note: mirror of https://github.com/kubernetes/apimachinery/blob/v0.32.3/pkg/runtime/schema/group_version.go#L140-L146
+type GroupVersionKind struct {
```
zerospiel, line 43: why should we copy-paste the basic API structure to our public API?

api/v1alpha1/pluggableprovider_types.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Description provides a human-readable explanation of what this provider does
+	Description string `json:"description,omitempty"`
+
+	Component `json:",inline"`
+}
+
+// PluggableProviderStatus defines the observed state of PluggableProvider
+type PluggableProviderStatus struct {
+	Infrastructure string `json:"infrastructure"`
```
zerospiel, line 65: do we need the status at all? if so, would it be ok to indicate readiness as we do in the current implementation?

api/v1alpha1/pluggableprovider_types.go
```diff
@@ -0,0 +1,96 @@
...
+	CAPI           string `json:"capi"`
```
eromanova, line 66: Could you please elaborate on what exactly this `CAPI` field means?
From the current implementation, I think it may be really confusing. For example, `projectsveltos` is not a CAPI provider, but it has this field equal to `projectsveltos`.
I believe we should have an API to determine the related ProviderTemplate, and the CAPI provider name (if it's a CAPI provider).

---
I can suggest the following:
1. Rename this field to `template` or something like this to reflect the ProviderTemplate linked to this pluggable provider. Each provider should be linked to a provider template.
2.  Either change the annotations specified above to support overriding the provider template name only (it will be `cluster-api-provider-<pprovName>` by default, or can be overwritten with an annotation, for the sveltos, for example), or provide another API (ex., introduce a new boolean field `isCAPI` that will mark that this is a CAPI provider and generate the template name based on it: `cluster-api-provider-<pprovName>` or simply <pprovName>).
What do you think?

api/v1alpha1/pluggableprovider_types.go
eromanova, line 65: Isn't the `status.Infrastructure` field excessive? I didn't find any occurrences where it's in use (except for specifying it in the status). Sorry if it's present and I'm blind.
Anyway, it may also be confusing since the pluggable provider may expose not only the infrastructure, but also the control plane and bootstrap capi providers.
Sveltos does not expose any of it. What's the purpose of this field?

api/v1alpha1/management_types.go
eromanova: So, am I right that currently, the list of providers to install is being fetched from two places (mgmt spec + all created pluggable providers)? But if I want to override one of the provider configuration, I still should do it in the management spec.

Also, the current implementation does not imply an easy way to disable one of the providers. Yes, I can do it by the removal of the PluggableProvider object. But what if I want to disable it temporarily and enable it after some time?

--- 
I think, better UX would be:
1. Keep the list of enabled providers in the management spec (if the provider is not present in the management spec, it won't be installed) with a single point of configuration.
2. This extra status field is redundant in this case (instead, the `spec.providers` can be used)

So, my main question: do we really need to use both the management spec and the list of all pluggable providers as a source?

api/v1alpha1/pluggableprovider_types.go
eromanova: I got it. In the `PluggableProviders` API, this field is needed for CLI visibility only. If so, I still think it's worth either showing all of the providers exposed (control plane, bootstrap as well) or nothing. It may be confusing.

api/v1alpha1/pluggableprovider_types.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2025
```
zerospiel, line 1: I will do the proper CR later. At the moment, I just wanted to point, that the CR has been added manually and not via the `kubebuilder`, hence the `PROJECT` now lacks the new CR. I suggest at least adding it manually with according settings if there are any validations/mutations or whether it is cluster-wide.

api/v1alpha1/pluggableprovider_types.go
eromanova: > I don't think that this makes sense, if you need to override provider name in this specific format, just create new object with expected name.
Yeah, I mean we still need any way to specify the template for the pluggable provider object. Currently, it's done via annotations. We can agree that the pluggable provider name should always be equal to the template name (which is better, I think, and will correspond with the provider name from the release object). But it requires changes in the current PluggableProviders (for example, currently the name is `aws` but the related template is `cluster-api-provider-aws-XXX`).

api/v1alpha1/pluggableprovider_types.go
eromanova: Ok, let's keep this status if you find it helpful. But since this field is being used for visibility only, then, I'd suggest removing the `infrastructure-` label: https://github.com/k0rdent/kcm/pull/1394/files#diff-7d0f26d17cb9555fff321ee38e7a02717d02525dc5d256f167d4ca653ac622b0R8 from the sveltos pluggable provider then. Sveltos does not expose any infrastructure provider.

api/v1alpha1/pluggableprovider_types.go
zerospiel: Now there is divergance: the CR is cluster-scoped though the PROJECT file states opposite

internal/controller/management_controller.go
```diff
@@ -100,6 +101,59 @@ func (r *ManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	return r.update(ctx, management)
 }
 
+func (r *ManagementReconciler) getRequestedProvidersList(ctx context.Context, management *kcm.Management) ([]kcm.Provider, error) {
```
wahabmk, line 104: Nit. Perhaps this function should be named `getExistingProvidersList`?

api/v1alpha1/management_types.go
zerospiel: I think the main question here is: who is supposed to use pprov? should it be a lead or an engineer? what are the expectations? i'd reckon that a lead is responsible for adding new providers to a mgmt cluster. if that is the case, then the only place for adding/showing should be. here is a pure example (but still): consider a pprov object is a PV, adding it to the mgmt spec is like creating a PVC selecting the PV. otherwise it is not ambiguous but rather confusing: what is the entrypoint for a provider?

i would vote that the mgmt should be the only place here to manage providers. what if one decided to use another provider template for the same provider? mgmt covers that case either

zerospiel: if we would stick to the latter, then I think the API should also be changed accordingly (removing the possibility to refer a provider template)

internal/controller/management_controller.go
```diff
@@ -100,6 +101,59 @@ func (r *ManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		return err
+	}
+
+	if equality.Semantic.DeepEqual(rprov, management.Status.RequestedProviders) {
+		return nil
+	}
+
+	management.Status.RequestedProviders = rprov
+
+	return r.updateStatus(ctx, management)
```
zerospiel, line 154: potential conflict: there will be 2 updates to the same resource version within a single event

I suggest removing it

api/v1alpha1/pluggableprovider_types.go
eromanova: Discussed in DM.

internal/controller/management_controller.go
```diff
@@ -100,6 +101,59 @@ func (r *ManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	return r.update(ctx, management)
 }
 
+func (r *ManagementReconciler) getRequestedProvidersList(ctx context.Context, management *kcm.Management) ([]kcm.Provider, error) {
+	list := slices.Clone(management.Spec.Providers)
+
+	existingProviders := make(map[string]struct{}, len(list))
+	for _, provider := range list {
+		existingProviders[provider.Name] = struct{}{}
+	}
```
zerospiel, line 110: perf: utilizing slices and traversing through them is more performant compared to maps (before the `1.24` at least) on relatively small slices

internal/controller/management_controller.go
```diff
@@ -100,6 +101,59 @@ func (r *ManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+	var objList kcm.PluggableProviderList
+
+	if err := r.Client.List(ctx, &objList, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{
+			kcm.GenericComponentNameLabel: kcm.GenericComponentLabelValueKCM,
+		}),
+	}); err != nil {
```
zerospiel, line 118:
```suggestion
        if err := r.Client.List(ctx, &objList, client.MatchingLabels{kcm.GenericComponentNameLabel: kcm.GenericComponentLabelValueKCM}); err != nil {
```

internal/controller/management_controller.go
```diff
@@ -100,6 +101,59 @@ func (r *ManagementReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	for _, el := range objList.Items {
+		if _, exists := existingProviders[el.Name]; !exists {
+			list = append(list,
+				kcm.Provider{
+					Name:      el.Name,
+					Component: el.Spec.Component,
+				},
+			)
+
+			existingProviders[el.Name] = struct{}{}
```
zerospiel, line 131: no need to add: the objects list has unique names (due to cluster scope), hence if a value is not presented, it won't be duplicated. unless there are collisions between in-tree and out-of-tree components having the same name. in that case, we should report an error to the condition and stop reconciling

internal/controller/management_controller.go
```diff
@@ -1133,6 +1193,23 @@ func (r *ManagementReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		setupLog.Info("Validations are disabled, watcher for ProviderTemplate objects is set")
 	}
 
+	managedController.Watches(&kcm.PluggableProvider{}, handler.EnqueueRequestsFromMapFunc(func(context.Context, client.Object) []ctrl.Request {
```
zerospiel, line 1196: nit: might be added directly during construction of the `managedController` variable to improve readability

internal/utils/validation/cd.go
```diff
@@ -94,10 +93,22 @@ func ClusterDeployCredential(ctx context.Context, cl client.Client, cd *kcmv1.Cl
...
-	return cred, isCredIdentitySupportsClusterTemplate(cred, clusterTemplate)
+	return cred, isCredIdentitySupportsClusterTemplate(ctx, cl, cred, clusterTemplate)
 }
 
-func isCredIdentitySupportsClusterTemplate(cred *kcmv1.Credential, clusterTemplate *kcmv1.ClusterTemplate) error {
+func getProviderClusterIdentityKinds(ctx context.Context, cl client.Client, infrastructureProviderName string) []string {
+	obj := &kcmv1.PluggableProvider{}
+
+	if err := cl.Get(ctx, client.ObjectKey{
+		Name: strings.TrimPrefix(infrastructureProviderName, kcmv1.InfrastructureProviderPrefix),
```
zerospiel, line 103: perf: nothing to trim in this context: the parameter already implies there is no prefix

internal/utils/validation/mgmt.go
```diff
@@ -49,7 +49,7 @@ func GetIncompatibleContracts(ctx context.Context, cl client.Client, release *kc
 	}
 
 	incompatibleContracts := strings.Builder{}
-	for _, p := range mgmt.Spec.Providers {
+	for _, p := range mgmt.Status.RequestedProviders {
```
zerospiel, line 52: that won't work in case of the disabled validation: that status might (and won't) contain the providers, resulting in successful validation of potentially erroneous configuration

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -186,6 +186,25 @@ rules:
...
+  verbs:
+  - update
+- apiGroups:
+  - k0rdent.mirantis.com
+  resources:
+  - pluggableproviders/status
+  verbs:
+  - get
+  - patch
+  - update
```
zerospiel, line 207: Shouldn't we add user-facing RBAC? the question relates to my original question about who is supposed to use the new resource

internal/controller/pluggableprovider_controller.go
```diff
@@ -0,0 +1,126 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+
+	kcm "github.com/K0rdent/kcm/api/v1alpha1"
+	"github.com/K0rdent/kcm/internal/utils"
+	"github.com/K0rdent/kcm/internal/utils/ratelimit"
+)
+
+const defaultSyncPeriod = 5 * time.Minute
```
zerospiel, line 34: Please hide this in the `setup` method, otherwise, it is visible across thousands of lines of code in this package. The package is gigantic and is to be refactored; until that moment, we must refrain from adding package-level abominable variables/constants

internal/controller/pluggableprovider_controller.go
```diff
@@ -0,0 +1,126 @@
...
+func (r *PluggableProviderReconciler) addLabels(ctx context.Context, pprov *kcm.PluggableProvider) error {
+	if !utils.AddLabel(pprov, kcm.GenericComponentNameLabel, kcm.GenericComponentLabelValueKCM) {
+		return nil
+	}
+	if err := r.Update(ctx, pprov); err != nil {
+		return fmt.Errorf("failed to update %s %s labels: %w",
+			pprov.GetObjectKind().GroupVersionKind().Kind, client.ObjectKeyFromObject(pprov), err)
+	}
+
+	return nil
```
zerospiel, line 76:
```suggestion
	_, err := utils.AddKCMComponentLabel(ctx, r.Client, pprov)
	return err
```

internal/controller/pluggableprovider_controller.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (r *PluggableProviderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, err error) {
+	l := ctrl.LoggerFrom(ctx)
+	l.Info("PluggableProvider reconcile start")
+
+	pprov := &kcm.PluggableProvider{}
```
zerospiel, line 100: i see that if there is no management object, there is nothing to reconcile. but if management object happens to appear after a pprov object has been created and reconciled, then this object won't be reconciled again further. i believe we should return an error if no mgmt is not presented, like we do for the rest of the controllers 

I suggest such behavior in order to avoid clogging the main queue with permanent requeues

internal/controller/pluggableprovider_controller.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if err := r.addLabels(ctx, pprov); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	defer func() {
+		err = errors.Join(err, r.updateStatus(ctx, pprov))
+	}()
+
+	return ctrl.Result{RequeueAfter: r.syncPeriod}, nil
```
zerospiel, line 113:
```suggestion
	return ctrl.Result{RequeueAfter: r.syncPeriod}, errors.Join(err, r.updateStatus(ctx, pprov))
```

zerospiel (dismissed):

templates/provider/kcm/templates/rbac/controller/roles.yaml
zerospiel: There is no difference in that matter, having the upcoming release date. It is either all or nothing. So there is no difference for me. Besides, the question relates to the API change in general

internal/controller/management_controller.go
zerospiel: Correct, it updates the status and ends the whole reconciliation until new changes. The proposed solution has the different behavior

api/v1alpha1/management_types.go
zerospiel: The process that I have in mind:

- a lead (or whoever TBD) adds a new pprov object
- the same person changes (patches) the mgmt spec appending a newly added provider to already presented

That is the question of design, I suppose

eromanova (approved):

eromanova:
Looks good overall. Additional improvements will be addressed in follow-up patches as discussed.

zerospiel (approved):
approving due to the requirement to approve. i personally against merging the pr in the current state. i still have unresolved questions in both ways: code and the design. a written promise to fix everything in a follow up pr has been given

internal/controller/management_controller.go
zerospiel: ok that is valid also for the 1.24

```
goos: darwin
goarch: arm64
pkg: github.com/zerospiel/playground
cpu: Apple M1 Pro
BenchmarkDirectSearch-10    	10353183	       115.4 ns/op	     384 B/op	       2 allocs/op
BenchmarkMap-10             	 6311917	       189.8 ns/op	     384 B/op	       2 allocs/op
PASS
```

```go
var smallList = []string{
	"projectsveltos",
	"cluster-api-provider-aws",
	"cluster-api-provider-azure",
	"cluster-api-provider-docker",
	"cluster-api-provider-gcp",
	"cluster-api-provider-k0sproject-k0smotron",
	"cluster-api-provider-openstack",
	"cluster-api-provider-vsphere",
}

var anotherSmallList = []string{
	"cluster-api-provider-custom0",
	"cluster-api-provider-custom1",
	"cluster-api-provider-aws",
	"cluster-api-provider-azure",
}

func BenchmarkDirectSearch(b *testing.B) {
	for b.Loop() {
		newslice := slices.Clone(smallList)
		for _, v := range anotherSmallList {
			if !slices.Contains(smallList, v) {
				newslice = append(newslice, v)
			}
		}
	}
}

func BenchmarkMap(b *testing.B) {
	for b.Loop() {
		newslice := slices.Clone(smallList)
		existing := make(map[string]struct{}, len(newslice))
		for _, v := range newslice {
			existing[v] = struct{}{}
		}
		for _, v := range anotherSmallList {
			if _, ok := existing[v]; !ok {
				newslice = append(newslice, v)
			}
		}
	}
}
```

internal/utils/validation/cd.go
zerospiel: initially, there was no need to trim given that the previous code block had the line `infraProviderName := providerName[len(providers.InfraPrefix):]`, which has been removed for some reason. due to the forced merge that removal led to #1438

zerospiel: anyway, it was addressed

internal/controller/management_controller.go
zerospiel: Resolved [here](https://github.com/k0rdent/kcm/pull/1434/files#diff-7fbd16e56d0b84de5285626a796c9df0598bebbc52d3ef34bcbc9cf84e8e8c91R150)

internal/controller/management_controller.go
zerospiel: This ability is done via the mgmt, I mean, if there is by any change for some reason one `pprov` object named `cluster-api-provider-aws` (so, out-of-tree provider) which we have in-tree, and the values from the OOT are required, then one can just create a new `ProviderTemplate` and then reference it as a custom in the `mgmt.spec`. So to my understanding, there is no need in such ability

templates/provider/kcm/templates/rbac/controller/roles.yaml
zerospiel: I do not know. Who are the anticipated creators of the `pprov` objects? I was under the impression that this must be discussed and settled during the design by feature designers. Since the PR does not provide any information on that, I do not have the answer.

In any case, I suppose it must be `global-admin` since the resource is cluster-scoped

internal/controller/pluggableprovider_controller.go
zerospiel: Since the main issue with the clogged queue is indeed fixed, it is up to the implementor. If the absence of the `Management` object aligns with the feature's business logic, then resolve the question.

templates/provider/kcm/templates/rbac/controller/roles.yaml
zerospiel: Here `templates/provider/kcm/templates/rbac/user-facing`, there are also enough examples for that role

internal/controller/pluggableprovider_controller.go
zerospiel: No issues according to the current controller implementation in terms of errors. The current approach with errors is that there is absolutely no reason to reconcile for the majority (if not all) of the controllers. If your particular controller is just okay without the management object, then everything is okay

Feel free to resolve if that answers your question

templates/provider/kcm/templates/rbac/controller/roles.yaml
zerospiel: `status` subresource is missing, some of the verbs are redundant. basically the editor would look like

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    k0rdent.mirantis.com/aggregate-to-global-admin: "true"
  name: {{ include "kcm.fullname" . }}-<resource>-editor-role
rules:
- apiGroups:
  - k0rdent.mirantis.com
  resources:
  - pluggableproviders
  - pluggableproviders/status
  verbs: {{ include "rbac.editorVerbs" . | nindent 6 }}
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    k0rdent.mirantis.com/aggregate-to-global-viewer: "true"
  name: {{ include "kcm.fullname" . }}-<resource>-viewer-role
rules:
- apiGroups:
  - k0rdent.mirantis.com
  resources:
  - pluggableproviders
  - pluggableproviders/status
```

zerospiel: Just in case: put these in separate files, like it is already done for other resources.

PR #1400, opened 2025-04-18
fix: do not init ClusterDeployment condition if it's already set

eromanova:
This PR fixes ClusterDeployment conditions and initializes them only if the condition is unset.


eromanova:
@zerospiel When we reinitialize the condition list on every reconcile, all conditions are overwritten, making it difficult to track the actual moment a condition’s status changes.

For example, I originally noticed this with the `HelmReleaseReady` condition. On each reconcile loop, this condition is reset to `Unknown`. After that, the controller checks the `HelmRelease` readiness and sets the condition accordingly, triggering an event here: https://github.com/k0rdent/kcm/blob/main/internal/controller/clusterdeployment_controller.go#L320.

The issue is that, without this fix, the event is emitted on every reconcile because `SetStatusCondition` returns true (indicating the condition was updated). But in reality, I want the event to be generated only once, at the moment the condition actually transitions, not every time it’s re-evaluated.

zerospiel:
I see, thanks for the explanation. I would suggest then to implement the same for the other methods as well ([here](https://github.com/k0rdent/kcm/blob/main/internal/controller/clusterdeployment_controller.go#L516) and [here](https://github.com/k0rdent/kcm/blob/main/internal/controller/multiclusterservice_controller.go#L88)) to avoid different implementations of the same approach. And I also suggestcommentingt on that so that it would be clear why such checks exist

internal/controller/clusterdeployment_controller.go
```diff
@@ -378,7 +378,15 @@ func (r *ClusterDeploymentReconciler) validateConfig(ctx context.Context, cd *kc
 }
 
 func (*ClusterDeploymentReconciler) initClusterConditions(cd *kcm.ClusterDeployment) (changed bool) {
+	conditions := make(map[string]struct{}, len(cd.Status.Conditions))
+	for _, cond := range cd.Status.Conditions {
+		conditions[cond.Type] = struct{}{}
+	}
+
 	for _, typ := range [4]string{kcm.CredentialReadyCondition, kcm.HelmReleaseReadyCondition, kcm.HelmChartReadyCondition, kcm.TemplateReadyCondition} {
+		if _, found := conditions[typ]; found {
```
zerospiel, line 387: wouldn't the following be simplier:

```go
if apimeta.FindStatusCondition(cd.Status.Conditions, typ) != nil {
    continue
}
```

zerospiel (approved):

PR #1422, opened 2025-04-23
fix(openstack-standalone-cp): make ssh config more uniform

chramb:
This patch replaces `sshPublicKey` in template values with `sshKeyName`
also used by upstream capi provider as well as bastion in this template.

values.schema.json now marks `sshPublickKey` as deprecated

By utilizing if-else block in template this change is backwards compatible
still allowing for use of `sshPublicKey` which might be needed for 
example when upgrading from 0-1-0 to 0-2-0

### DIFF
--- a/templates/cluster/openstack-standalone-cp/values.yaml
+++ b/templates/cluster/openstack-standalone-cp/values.yaml
@@ -22,8 +22,8 @@ identityRef:
 
 bastion:
   enabled: false
-  sshKeyName: ""
   spec:
+    sshKeyName: ""
     providerID: ""
     flavor: ""
     image:
--- a/templates/cluster/openstack-standalone-cp/values.yaml
+++ b/templates/cluster/openstack-standalone-cp/values.yaml
@@ -22,6 +22,7 @@ identityRef:
 
 bastion:
   enabled: false
+  sshKeyName: ""
   spec:
     providerID: ""
     flavor: ""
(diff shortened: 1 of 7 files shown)


zerospiel (changes requested):
The minimal requirements for the PR are not met: templates aren't regenerated, there are no changes to the appropriate configs, etc.

If any changes are introduced to a template, its version is to be bumped in the corresponding `Chart.yaml`.

Before making a PR and removing the given instructions, please read them and ensure they are followed.

chramb:
Sorry for some reason I thought templates in PR template meant controller templates :/ - fixed now.

Also do I see it correctly that e2e tests don't include OpenStack in testing?

zerospiel (approved):

templates/cluster/openstack-standalone-cp/values.yaml
```diff
@@ -22,6 +22,7 @@ identityRef:
 
 bastion:
   enabled: false
+  sshKeyName: ""
```
zerospiel, line 25: the value is unused, is it expected?

eromanova (approved):

chramb:
Yea looks like my editor didn't expand tabs properly, should be correct now, I didn't want to pass it through jq because it expanded arrays into multiple lines.

I *think* it's ok now

zerospiel (approved):

PR #1429, opened 2025-04-24
feat: Add global registry values passthrough

### LINKED ISSUE #1428: Global variables passtrough
Provide a way to pass global parameters, such as `global.registry` to all templates configuration in the system.
This is needed to provide uniform way to pass registry information across templates in case of using single registry.
Currently the following values should be passed:

```yaml
global:
  registry: "some registry host"
  k0sURL: "some k0s url"
```

The parameters must set using a flags in the controller and there should be a possibility to disable these parameters creation.

a13x5:
Resolves #1428 

> [!NOTE]
> I intentionally haven't done any refactoring to keep this trivial. So the function `updateCluster` of clusterdeploymetn controller is triggering `gocycle` and probably should be splitted into several parts, but this is way out of scope and for now I just bumped threshold of `gocycle` to 31 (which is shitty, but will do)

> [!NOTE]
> The `defaultRegistryURL` flag probably should be renamed, since now it may look confusing. Now I left this out of the PR.


internal/controller/clusterdeployment_controller.go
```diff
@@ -271,6 +274,13 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, cd *kcm
 			"namespace":  cred.Spec.IdentityRef.Namespace,
 		}
 
+		if r.EnableGlobalVals {
```
eromanova, line 277: Isn't this boolean parameter excessive? Can it be replaced with:
```
if r.GlobalRegistry != "" || r.GlobalK0sURL != ""
```
or something like this?

internal/controller/management_controller.go
```diff
@@ -759,7 +766,30 @@ func applyKCMDefaults(config *apiextensionsv1.JSON) (*apiextensionsv1.JSON, erro
...
+			"registry": registry,
+		},
+	}
+
+	raw, err := json.Marshal(chartutil.CoalesceTables(values, globalValues))
+	if err != nil {
+		return nil, err
+	}
+
+	return &apiextensionsv1.JSON{Raw: raw}, nil
```
eromanova, line 789: We already have similar logic in the `enableAdditionalComponents` function. I believe this code block can be moved to this function to avoid extra JSON marshalling and unmarshalling (but the function should probably be renamed).

eromanova: This is optional. I know you wanted to make minimal changes. This can be refactored later.

.golangci.yml
```diff
@@ -316,6 +316,8 @@ linters:
         - suite-subtest-run
         - suite-thelper
         - useless-assert
+    gocyclo:
+      min-complexity: 31
```
zerospiel, line 320: please leave as it was, add `nolint:gocyclo` directive to the method

internal/controller/management_controller.go
zerospiel: I would suggest adding `TODO:` and `WARN:` or even `kludge` to highlight that it is to be refactored

zerospiel:
> The defaultRegistryURL flag probably should be renamed, since now it may look confusing. Now I left this out of the PR.

I think we should decide on that before the release

internal/controller/management_controller.go
a13x5: Yes functions `applySveltosDefaults`, `applyKCMDefaults` and `applyGlobalRegistry` could be generalized an put in one place (like in `enableAdditionalComponents`). 
Currently I did it as it was done before in a similar situation. I will put visible comments to all these parts of code and will create a new task for backlog, so it will have a chance to not become permanent.

a13x5:
> I think we should decide on that before the release
Renamed in a separate commit. It touches quite a few things though

internal/controller/management_controller.go
a13x5: Created #1431

internal/controller/management_controller.go
```diff
@@ -822,6 +852,16 @@ func getWrappedComponents(mgmt *kcm.Management, release *kcm.Release) ([]compone
 			}
 		}
 
+		// TODO: Refactor this kludge. Generalize default values passing.
+		if r.GlobalRegistry != "" || r.GlobalK0sURL != "" {
```
eromanova, line 856: I think we should keep only `r.GlobalRegistry != ""` condition here

internal/controller/clusterdeployment_controller.go
```diff
@@ -271,6 +274,13 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, cd *kcm
 			"namespace":  cred.Spec.IdentityRef.Namespace,
 		}
 
+		if r.GlobalRegistry != "" || r.GlobalK0sURL != "" {
+			values["global"] = map[string]any{
+				"registry": r.GlobalRegistry,
```
eromanova, line 279: I think still, extra conditions are required here: set `registry` only if `r.GlobalRegistry` is not empty, and set `k0sURL` only if `r.GlobalK0sURL` is not empty. But I guess these parameters work in a bundle only, right?

eromanova (commented):
Two more small comments. The other code looks good to me.

internal/controller/clusterdeployment_controller.go
a13x5: I assume that they will be passed together into `ClusterDeployment` config. But I guess it won't hurt to add an extra check

eromanova (approved):
LGTM. But I doubt that with these changes, the k0rdent upgrade from 0.2.0 -> 0.3.0 may not automatically work without manual actions from the user due to the controller parameter renaming. @a13x5 could you please check it? If my assumption is right, we will need to update the k0rdent upgrade docs accordingly.

a13x5:
@eromanova If user set this flag to some custom value before it will affect the update, so user must set new flag before updating. 
I marked the commit with `!` and explicitly mentioned parameter renaming + I will add that to the release notes

Issue #1301, opened 2025-03-27
[enhancement] ensure ServiceTemplateChain functionality


BROngineer:
### Description

Ensure that services' upgrades properly handled in case corresponding `ServiceTemplateChain` exists

### DoD

- Available upgrades reflected in resources' statuses
- Upgrades performed according to defined chain

PR #1433, opened 2025-04-24
feat: enable services upgrades based on ServiceTemplateChains

BROngineer:
This PR introduces changes in API and controllers' behaviour to address the necessity to ensure services upgrade paths.

### API changes:
- added `.spec.version` field to `ServiceTemplate`. The field is optional at the moment, supposed to become mandatory. This field is needed to encapsulate service template version, after support of sources other than Helm charts was added.
- added `.spec.serviceSpec.services[].templateChain` field to `ClusterDeployment` / `MultiClusterService` (hereafter `CD`/`MCS` respectively). This field is optional. If set, the possible upgrade paths can be computed. If not set, changing the `.spec.serviceSpec.services[].template` field value is about replacing the application rather than upgrade.
- added `.status.servicesUpgradePaths` field to `CD`/`MCS`. This field reflects possible upgrade for services with defined template chain
- added `.sourceSpec.localSourceRef.namespace` field for `ServiceTemplate`. This allows cross-namespace reference to flux sources, when `ServiceTemplate` is populated across namespaces using `ServiceTemplateChain` and access management. Cross-namespace references required to keep secrets, certificates and other credentials, needed to access flux source, in system namespace.

### Controllers & Webhooks:
- `ServiceTemplateChain` controller populates `ServiceTemplates` across namespaces, replacing flux source definitions or references with cross-namespace references to flux source objects existing in system namespace.
- `CD`/`MCS` validating hooks ensures template chains defined for services are valid.
- `CD`/`MCS` controllers validate whether template chains defined for services are valid or not.
- `CD`/`MCS` controllers validate whether desired services' templates match observed (installed) templates in terms of computed upgrade paths.

Fixes: #1301 

### Out-Of-Scope:
- ensuring sequential upgrade: with given implementation it's almost impossible or at least will significantly increase the complexity of the controllers' code. Therefore, we might need to introduce intermediate resource, something like `ServiceSet` (`ClusterServiceSet`) which in turn will spawn Sveltos `Profiles` or `ClusterProfiles`. This will allow, first, validate the upgrade path for templates defined in `CD`/`MCS` spec. Second, it will allow to implement flow for sequential upgrade.

### DIFF
--- a/api/v1alpha1/templatechain_common_test.go
+++ b/api/v1alpha1/templatechain_common_test.go
@@ -77,8 +77,7 @@ func TestTemplateChainSpec_TemplateUpgradePath(t *testing.T) {
 					Name: "foo-0-2-0",
 				},
 			}},
-			templateName: "foo-0-1-0",
-			// expectedUpgradePath: [][]string{{"foo-0-2-0"}},
+			templateName:        "foo-0-1-0",
 			expectedUpgradePath: []UpgradePath{{Versions: []string{"foo-0-2-0"}}},
 		},
 		"upgrade-path-2": {
@@ -155,12 +154,6 @@ func TestTemplateChainSpec_TemplateUpgradePath(t *testing.T) {
 				},
 			}},
 			templateName: "foo-0-1-0",
-			// expectedUpgradePath: [][]string{
-			// 	{"foo-0-2-0"},
-			// 	{"foo-0-3-0"},
-			// 	{"foo-0-4-0"},
-			// 	{"foo-0-4-0", "foo-1-0-0"},
-			// },
 			expectedUpgradePath: []UpgradePath{
 				{
 					Versions: []string{"foo-0-2-0"},
@@ -210,10 +203,6 @@ func TestTemplateChainSpec_TemplateUpgradePath(t *testing.T) {
 				},
 			}},
 			templateName: "foo-0-3-0",
-			// expectedUpgradePath: [][]string{
-			// 	{"foo-0-4-0"},
-			// 	{"foo-0-4-0", "foo-1-0-0"},
-			// },
 			expectedUpgradePath: []UpgradePath{
 				{
 					Versions: []string{"foo-0-4-0"},


internal/utils/validation/services.go
```diff
@@ -29,17 +32,112 @@ import (
...
 	return errs
 }
+
+func ValidateUpgradePaths(services []kcmv1.Service, upgradePaths []kcmv1.ServiceUpgradePaths) error {
+	observedTemplatesMap := make(map[string]struct {
+		Template     string
+		UpgradePaths [][]string
+	}, len(upgradePaths))
+	for _, observedService := range upgradePaths {
+		observedTemplatesMap[observedService.ServiceName] = struct {
```
wahabmk, line 106: Should the key be namespaced-name?

internal/utils/validation/services.go
```diff
@@ -29,17 +32,112 @@ import (
...
 	}
 
 	return errs
 }
+
+func ValidateUpgradePaths(services []kcmv1.Service, upgradePaths []kcmv1.ServiceUpgradePaths) error {
+	observedTemplatesMap := make(map[string]struct {
+		Template     string
+		UpgradePaths [][]string
+	}, len(upgradePaths))
```
wahabmk, line 104: We could also do the following instead of creating a new struct type:
```go
	observedTemplatesMap := make(map[string]kcmv1.ServiceUpgradePaths, len(upgradePaths))
```

api/v1alpha1/multiclusterservice_types.go
```diff
@@ -166,12 +170,24 @@ type ServiceStatus struct {
 type MultiClusterServiceStatus struct {
 	// Services contains details for the state of services.
 	Services []ServiceStatus `json:"services,omitempty"`
+	// ServicesUpgradePaths contains details for the state of services upgrade paths.
+	ServicesUpgradePaths []ServiceUpgradePaths `json:"servicesUpgradePaths,omitempty"`
```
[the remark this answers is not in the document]
wahabmk: Maybe my example wasn't clear enough.

<table>
<tr>
<td> Currently In this PR </td> <td> Proposed </td>
</tr>
<tr>
<td>

```yaml
status:
  services:
    - clusterName: clusterA
      clusterNamespace: kcm-system
      conditions: [...]
    - clusterName: clusterB
      clusterNamespace: kcm-system
      conditions: [...]
  serviceUpgradePaths:
    - serviceName: foo
      template: foo-0-1-0
      upgradePaths:
        - ["foo-0-2-0"]
  newfeatureA:
    newfeatureASubfeature1: ...
    newfeatureASubfeature2: ... 
  newfeatureB: ...
  observedGeneration: 1
  conditions: [...]
```

</td>
<td>

```yaml
status:
  services:
    serviceUpgradePaths:
      - serviceName: foo
        template: foo-0-1-0
        upgradePaths:
          - ["foo-0-2-0"]
    newfeatureA:
        newfeatureASubfeature1: ...
        newfeatureASubfeature2: ... 
    newfeatureB: ...
    serviceStatus:   # <---- originally named services, renamed to serviceStatus
      - clusterName: clusterA
        clusterNamespace: kcm-system
        conditions: [...]
      - clusterName: clusterA
        clusterNamespace: kcm-system
        conditions: [...]
  observedGeneration: 1
  conditions: [...]
```

</td>
</tr>
</table>

My main motivation for this proposal is to group all objects related to beach-head services together, rather than adding them directly under `.status`. For ClusterDeployment especially, this would help a lot as users will expect to find everything beach-head services related under `.status.services`.

internal/utils/validation/services.go
wahabmk: I meant we could do away with the anonymous struct and just re-use `kcmv1.ServiceUpgradePaths`, it's up to you.

api/v1alpha1/templatechain_common.go
```diff
@@ -71,3 +76,84 @@ func (s *TemplateChainSpec) IsValid() (warnings []string, ok bool) {
...
 	return warnings, len(warnings) == 0
 }
+
+// findAllUpgradePaths returns all possible upgrade paths from the given template
+func (s *TemplateChainSpec) findAllUpgradePaths(templateName string) ([][]string, error) {
+	// Build a map for lookup of supported templates by name
+	templateMap := make(map[string]SupportedTemplate)
+	for _, template := range s.SupportedTemplates {
+		templateMap[template.Name] = template
+	}
```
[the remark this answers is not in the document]
wahabmk: Yes. However, this makes the function more correct in its intended purpose even if the constraints get removed. 🙂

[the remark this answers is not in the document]
wahabmk: A test case for multiple entries should be added in any case. If you also enhance the function, the test case won't have to fail even if constraints are changed. This is good defensive programming. It is not overly redundant either because with this you're solving for a case within the function, instead of the function having to rely on outside factors. I don't see any harm in adding it, but I won't press the issue any further.

api/v1alpha1/multiclusterservice_types.go
```diff
@@ -186,8 +186,14 @@ type ServiceUpgradePaths struct {
...
-	// UpgradePaths contains details for the state of service upgrade paths.
-	UpgradePaths [][]string `json:"upgradePaths,omitempty"`
+	// AvailableUpgrades contains details for the state of available upgrades.
+	AvailableUpgrades []UpgradePath `json:"availableUpgrades,omitempty"`
+}
+
+// UpgradePath contains details for the state of service upgrade paths.
+type UpgradePath struct {
+	// Versions contains the list of versions that service can be upgraded to.
+	Versions []string `json:"upgradePaths,omitempty"`
```
wahabmk, line 196: Should this be `Name` or `TemplateName` instead because it is currently holding values such as `foo-0-2-0` etc, which are template names (not versions).

api/v1alpha1/templatechain_common_test.go
```diff
@@ -144,11 +155,25 @@ func TestTemplateChainSpec_TemplateUpgradePath(t *testing.T) {
...
-				{"foo-0-2-0"},
-				{"foo-0-3-0"},
-				{"foo-0-4-0"},
-				{"foo-0-4-0", "foo-1-0-0"},
+			// expectedUpgradePath: [][]string{
+			// 	{"foo-0-2-0"},
+			// 	{"foo-0-3-0"},
+			// 	{"foo-0-4-0"},
+			// 	{"foo-0-4-0", "foo-1-0-0"},
+			// },
```
wahabmk, line 163: Missed removing this commented code

api/v1alpha1/templatechain_common_test.go
```diff
@@ -185,9 +210,17 @@ func TestTemplateChainSpec_TemplateUpgradePath(t *testing.T) {
 				},
 			}},
 			templateName: "foo-0-3-0",
-			expectedUpgradePath: [][]string{
-				{"foo-0-4-0"},
-				{"foo-0-4-0", "foo-1-0-0"},
+			// expectedUpgradePath: [][]string{
+			// 	{"foo-0-4-0"},
+			// 	{"foo-0-4-0", "foo-1-0-0"},
+			// },
```
wahabmk, line 216: And here

wahabmk (approved):

wahabmk (approved):

PR #1434, opened 2025-04-25
fix: Rework `PluggableProvider` controller reconcile

s3rj1k:
PR Reworks `PluggableProvider` controller reconcile by adding additional watchers for `ProviderTemplate` and `Management` object update events.

### DIFF
--- a/internal/controller/pluggableprovider_controller.go
+++ b/internal/controller/pluggableprovider_controller.go
@@ -16,15 +16,20 @@ package controller
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	kcm "github.com/K0rdent/kcm/api/v1alpha1"
 	"github.com/K0rdent/kcm/internal/utils"
@@ -41,10 +46,12 @@ func (r *PluggableProviderReconciler) getProviderTemplate(ctx context.Context, p
 	if pprov.Spec.Template != "" {
 		return pprov.Spec.Template
 	}
+
 	management := &kcm.Management{}
 	if err := r.Get(ctx, client.ObjectKey{Name: kcm.ManagementName}, management); err != nil {
 		return ""
 	}
+
 	return management.Status.Components[pprov.Name].Template
 }
 
@@ -53,7 +60,9 @@ func (r *PluggableProviderReconciler) getExposedProviders(ctx context.Context, p
 	if template == "" {
 		return "", nil
 	}
+
 	templateObj := &kcm.ProviderTemplate{}
+
 	err := r.Get(ctx, types.NamespacedName{Name: template}, templateObj)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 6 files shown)


internal/controller/pluggableprovider_controller.go
```diff
@@ -72,44 +81,141 @@ func (r *PluggableProviderReconciler) updateStatus(ctx context.Context, pprov *k
...
+
+	var pprov kcm.PluggableProvider
+
+	if err := r.Get(ctx, req.NamespacedName, &pprov); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		l.Error(err, "PluggableProvider reconcile error")
+		return ctrl.Result{RequeueAfter: r.syncPeriod}, err
```
zerospiel, line 126:
```suggestion
	if err := r.Get(ctx, req.NamespacedName, &pprov); err != nil {
		l.Error(err, "PluggableProvider reconcile error")
		return ctrl.Result{}, client.IgnoreNotFound(err)
```

Having non-zero result with an error does not make sense and lead to warning errors, please return an empty result

```
Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler
```


So, in general, `syncPeriod` might be removed completely from the struct

internal/controller/pluggableprovider_controller.go
```diff
@@ -72,44 +81,141 @@ func (r *PluggableProviderReconciler) updateStatus(ctx context.Context, pprov *k
...
+		l.Error(err, "PluggableProvider update status error")
+		return ctrl.Result{RequeueAfter: r.syncPeriod}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *PluggableProviderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	l := ctrl.LoggerFrom(ctx)
+	l.V(6).Info("PluggableProvider reconcile start")
```
zerospiel, line 116:
```suggestion
	l.V(1).Info("PluggableProvider reconcile start")
```

We do not use `6` level, either `0` or `1`.

internal/controller/pluggableprovider_controller.go
```diff
@@ -72,44 +81,141 @@ func (r *PluggableProviderReconciler) updateStatus(ctx context.Context, pprov *k
...
+	if err := r.addLabels(ctx, pprov); err != nil {
+		l.Error(err, "PluggableProvider adding labels error")
+		return ctrl.Result{RequeueAfter: r.syncPeriod}, err
 	}
 
-	if err := r.addLabels(ctx, pprov); err != nil {
-		return ctrl.Result{}, err
+	if err := r.updateStatus(ctx, pprov); err != nil {
+		l.Error(err, "PluggableProvider update status error")
+		return ctrl.Result{RequeueAfter: r.syncPeriod}, err
```
zerospiel, line 108: Having non-zero result with an error does not make sense and lead to warning errors, please return an empty result

```
Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler
```

internal/controller/pluggableprovider_controller.go
```diff
@@ -72,44 +81,141 @@ func (r *PluggableProviderReconciler) updateStatus(ctx context.Context, pprov *k
...
 	l := ctrl.LoggerFrom(ctx)
-	l.Info("PluggableProvider reconcile start")
+	l.Info("PluggableProvider object event")
 
-	pprov := &kcm.PluggableProvider{}
-	if err := r.Get(ctx, req.NamespacedName, pprov); err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+	if err := r.addLabels(ctx, pprov); err != nil {
+		l.Error(err, "PluggableProvider adding labels error")
+		return ctrl.Result{RequeueAfter: r.syncPeriod}, err
```
zerospiel, line 103: Having non-zero result with an error does not make sense and lead to warning errors, please return an empty result

```
Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler
```

zerospiel (changes requested):
This is the first part of the review so far.

At the moment, the following comments are not yet addressed:

[1](https://github.com/k0rdent/kcm/pull/1394#discussion_r2058737174): to be discussed
[2](https://github.com/k0rdent/kcm/pull/1394#discussion_r2058712039): RBAC-related question, the corresponding user-facing RBAC permissions are to be created
[3](https://github.com/k0rdent/kcm/pull/1394#discussion_r2058683937): redundant `map` usage, also aligns with the [performance optimization](https://github.com/k0rdent/kcm/pull/1394#discussion_r2058674041) if the map should actually be removed there

internal/controller/pluggableprovider_controller.go
```diff
@@ -72,44 +81,141 @@ func (r *PluggableProviderReconciler) updateStatus(ctx context.Context, pprov *k
...
+
+		return resp
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(controller.TypedOptions[ctrl.Request]{
 			RateLimiter: ratelimit.DefaultFastSlow(),
 		}).
 		For(&kcm.PluggableProvider{}).
+		Watches(&kcm.ProviderTemplate{}, handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []ctrl.Request {
```
zerospiel, line 160: Isn't it enough to watch exclusively `Management` object as the only source of providers? In the current implementation, `providertemplates` are dependent on the `Management` object, which is already watched

internal/controller/pluggableprovider_controller.go
```diff
@@ -72,44 +81,141 @@ func (r *PluggableProviderReconciler) updateStatus(ctx context.Context, pprov *k
...
+				if !ok {
+					return false
+				}
+
+				newObj, ok := e.ObjectNew.(*kcm.Management)
+				if !ok {
+					return false
+				}
+
+				return !equality.Semantic.DeepEqual(oldObj.Status, newObj.Status)
```
zerospiel, line 214: I suppose this is redundant, and we just want to trigger a new event only if `management.status.components` has changed

zerospiel (changes requested):

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -186,7 +186,7 @@ rules:
   - get
   - patch
   - update
-# (s3rj1k) RBAC rules for `pluggableproviders` should be expanded
+# (s3rj1k): consider expanding RBAC rules for `pluggableproviders`
```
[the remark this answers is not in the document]
zerospiel: These permissions have nothing to do with the user-facing RBAC. User-facing RBAC permissions should be decided by feature implementers, not the code owners, we do not own the feature you're implementing.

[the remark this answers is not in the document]
zerospiel: So yes, the comment should be removed

internal/controller/pluggableprovider_controller.go
zerospiel: No, I suggest the opposite: to decrease the number of events: `return !equality.Semantic.DeepEqual(oldObj.Status.Components, newObj.Status.Components)`

internal/controller/pluggableprovider_controller.go
zerospiel: The only thing I am a little concerned now is that the `management` controller now watches `pprov` (basically unconditionally), the latter watches the former (conditionally). I just hope there won't be any circular events.

@eromanova please take a note that the `mgmt`-ctrl and `pprov`-ctrl watch each other, just in case. I would suggest test it in a way

internal/controller/pluggableprovider_controller.go
zerospiel: I didn't get it with:

> install deadlocks,

Anyway, we could introduce a precheck for the lengths at least, if they differ, then we might return `true` earlier without complex checks:


```go
if len(oldObj.Status.Components) != len(newObj.Status.Components) {
    return true
}
// other check(s)
```



The original question is because the `management.status` gets too many updates

zerospiel: Ok, so let me clarify: during the installation there are `pprov` objects (w/o status) and (potentially, it depends on the installation settings) `mgmt` object, the latter does not get updates until the former objects get updates, is that right? From the code that I see right now, there is no dependency on the `pprov` objects' statuses whatsoever. So if someone creates a `mgmt` object, it should get updates, including `status.components`, which then should lead to the informational field updates of `pprov` objects, isn't that right? Or do I miss something? Could you please provide me the dependency that introduces the aforementioned event-deadlock?

p.patch
```diff
@@ -0,0 +1,40 @@
+diff --git a/internal/controller/management_controller.go b/internal/controller/management_controller.go
```
zerospiel, line 1: Please remove the patch

internal/controller/pluggableprovider_controller.go
zerospiel: I've installed the dev version with the proposed change, no problems in regards to `mgmt` readiness or absence of providers. Everything is in place

templates/provider/kcm/templates/rbac/user-facing/pluggableproviders-vieweryaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
zerospiel, line 1: file extension typo `templates/provider/kcm/templates/rbac/user-facing/pluggableproviders-vieweryaml` -> `templates/provider/kcm/templates/rbac/user-facing/pluggableproviders-viewer.yaml`

templates/provider/kcm/templates/rbac/user-facing/pluggableproviders-viewer.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    k0rdent.mirantis.com/aggregate-to-global-viewer: "true"
+  name: {{ include "kcm.fullname" . }}-pluggableproviders-viewer-role
+rules:
+- apiGroups:
+  - k0rdent.mirantis.com
+  resources:
+  - pluggableproviders
+  - pluggableproviders/status
```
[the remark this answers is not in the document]
zerospiel: updated the top comment. the rules' verbs

internal/controller/pluggableprovider_controller.go
zerospiel: sorrz i missed the link, here is the right with checking [the map](https://github.com/k0rdent/kcm/pull/1434#discussion_r2060364733) . I've checked with the length precheck and then checking only the map diff, worked pretty well. and as I ve mentioned earlier, I do not see any deadlock from the code point, and i'd appreciate if you provide it so we are on the same page

I will continue the review in 2 weeks if it would be still actual

internal/controller/pluggableprovider_controller.go
zerospiel: im okay to remove it

I will continue the review in 2 weeks if it is still be the case

eromanova: It’s worth watching the `ProvideTemplate` to get the update of the exposed provider as soon as it is set in the ProviderTemplate status, and update the status of the `PluggableProvider` faster. I think it’s worth keeping this watcher, but simply check the length of providers array (it’s not that part that frequently changes, I think DeepEqual is excessive):
```
                             UpdateFunc: func(e event.UpdateEvent) bool {
				if !utils.HasLabel(e.ObjectNew, kcm.GenericComponentNameLabel) {
					return false
				}

				oldObj, ok := e.ObjectOld.(*kcm.ProviderTemplate)
				if !ok {
					return false
				}

				newObj, ok := e.ObjectNew.(*kcm.ProviderTemplate)
				if !ok {
					return false
				}

				return len(oldObj.Status.Providers) != len(newObj.Status.Providers)
```

internal/controller/pluggableprovider_controller.go
eromanova: The management watcher was reworked to trigger reconciliation only if `Status.Components` has been changed (first, it checks the length, then DeepEqual)

eromanova (approved):

PR #1435, opened 2025-04-25
fix: optimize calls to management status updates

s3rj1k:
PR removes redundant call to management object status update, fixes `DisabledValidation` branching for PluggableProviders


zerospiel:
Please cherry-pick the commit onto #1434 and close the PR

This is not Gerrit, and commits (plural) are allowed in a single PR, so it would be much easier to make a code review

PR #1440, opened 2025-04-25
fix: unify infrastructure providers naming

### LINKED ISSUE #1438: [bug] The controller does not remove the blocking finalizer
**Describe the bug**
On the latest main branch, the ClusterDeployment deletion is broken; the controller doesn't remove the blocking finalizer anymore.

https://github.com/k0rdent/kcm/commit/6f18145925d8fcafa93dd588e76fe66baa8c6723

eromanova:
There was an inconsistency in the infrastructure providers' naming
(for example, searched for `infrastructure-aws` in one place and `aws` in another),
which breaks the removal of the blocking finalizer from the ClusterDeployment.

Now, the full provider name with the infrastructure prefix is being used.

Also, added an indexer for the PluggableProvider object to fetch all the exposed infrastructure providers.

Fixes #1438

### DIFF
--- a/api/v1alpha1/indexers.go
+++ b/api/v1alpha1/indexers.go
@@ -17,6 +17,7 @@ package v1alpha1
 import (
 	"context"
 	"errors"
+	"strings"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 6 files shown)


zerospiel (approved):

eromanova:
Verified: https://github.com/k0rdent/kcm/actions/runs/14663034437/job/41151646378

zerospiel (approved):

PR #1450, opened 2025-04-28
fix: several fixes for PluggableProviders implementation

### LINKED ISSUE #1445: Move PluggableProvider objects to corresponding ProviderTemplates

### LINKED ISSUE #1446: Bring back full control of provider configuration to the Management object
Currently it's a mix of Management and PluggableProvider overrides.
It should be only possible to enable/disable provider through the Management object.
ProviderTemplate values overrides should be exclusively handled in the Management object.

eromanova:
This PR contains the following changes:
1. Moves all the in-tree PluggableProviders to the corresponding provider templates
2. The Management object remains a single source of enabled providers. The `RequestedProviders` field was removed from the status. Each occurrence was replaced with `spec.Providers`.
3. Dropped several unused watchers to avoid circular dependencies (pprov controller watches mgmt, mgmt controller watches pprov)
4. Removed unused Makefile targets
5. Additionally, reworked the process of fetching the exposed providers for the `PluggableProvider` object (instead of getting the ProviderTemplate in the pluggable provider controller, extended the Management Status with the `ExposedProviders` for each component, and get the exposed providers from here).

Manually verified the installation and upgrade. No issues found. The installation works for both in-tree and out-of-tree providers (verified on hetzner). The K0rdent upgrade from 0.2.0 to the latest changes now does not require any additional manual actions.

OOT provider installation:
```
helm install hetzner-pp oci://ghcr.io/k0rdent-oot/oot/charts/hetzner-pp -n kcm-system
kubectl patch mgmt kcm \
  --type='json' \
  -p='[
    {
      "op": "add",
      "path": "/spec/providers/-",
      "value": {
        "name": "cluster-api-provider-hetzner",
        "template": "cluster-api-provider-hetzner-0-2-1",
      }
    }
  ]'
```

Fixes #1445
Fixes #1446


s3rj1k:
I suggest renaming PluggableProviders to something else, with this change they are not pluggable, effectively this replaces ConfigMap definitions that where before with CRDs but keeps around manual flow of enabling each installed provider.

a13x5 (approved):
Generally LGTM.
Changes look reasonable as a fix, but since there is no requirements I'm aware of pluggable providers I can't say for sure if this is within the requirements.
Approve, since it fixes broken upgrade.

Kshatrix (approved):

api/v1alpha1/provider_types.go
```diff
@@ -19,8 +19,8 @@ import (
 )
 
 const (
-	// PluggableProviderKind represents the kind for pluggable providers
-	PluggableProviderKind = "PluggableProvider"
+	// ProviderKind represents the kind for providers
+	ProviderKind = "Provider"
```
Kshatrix, line 23: I'd name it as `ProviderRegistration` or `ProviderInterface`

Kshatrix (approved):

Kshatrix (approved):

PR #1464, opened 2025-04-30
feat: Improve the output for AI code review workflow

sotpuschennikov:
* Add separate step to show results
* Add step to archive results
* Add summary to the comment


.github/workflows/ai-codereview.yml
```diff
@@ -23,6 +29,13 @@ jobs:
               pull_number: context.payload.pull_request.number,
             });
             return pullRequest
+      - name: Get job name
+        id: vars
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # In that particular case it will work because we have only one job. It will also work if we'll have that job first in a row. But there no unified solution for that
+          echo "jobname=$(gh api "/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/jobs" --jq '.jobs[0] | .name')" >> $GITHUB_OUTPUT
```
eromanova, line 38: I'm wondering if we can find this job by its ID (`{{ github.job }}`) in the latter step instead of saving the `jobname` variable as an output and using it later. I assume the job ID is more reliable than this implementation, but I'm not sure it'll work. Could you please double-check?
If it works, this step can be omitted.

.github/workflows/ai-codereview.yml
```diff
@@ -86,31 +102,56 @@ jobs:
...
+          mv output.txt output.md
+          mv summary.txt summary.md
+          rich output.md
+          python3 -m markdown output.md -f output.html
+          echo -e "\nThere is some code review from AI. See the code review [here]($JOB_URL)" >> ./summary.md
+      - name: Upload Artifact
+        id: upload_artifact
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 #v4
+        with:
+          name: my-artifact
```
eromanova, line 126: Maybe change the artifact name for uniqueness and clarity? For example:
```suggestion
          name: ai-codereview-report
```

.github/workflows/ai-codereview.yml
```diff
@@ -64,6 +77,7 @@ jobs:
           PATH_TO_REQUIREMENTS_FILE: 'ai-ci-scripts/ai-codereview/requirements.txt'
           ALL_CHANGED_FILES: ${{ steps.changed-golang-files.outputs.all_changed_files }}
           GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number || github.event.issue.number }}
```
eromanova, line 80: This variable looks unused? Is it required?

.github/workflows/ai-codereview.yml
sotpuschennikov: Unfortunately for that workflow
```jobs:
    ai-codereview:
        name: "AI Code Review"```
we will have different values for job and jobname:
github.job = "ai-codereview" 
jobname = "AI Code Review"

PR #1477, opened 2025-05-06
feat: migrate to v1beta1 api version of k0rdent

Kshatrix:


templates/provider/kcm-templates/files/templates/cluster-api-provider-aws.yaml
```diff
@@ -1,14 +1,14 @@
 apiVersion: k0rdent.mirantis.com/v1alpha1
```
a13x5, line 1: Looks like we need to adjust `hack/templates.sh` script. Looks like it still has alpha version there

a13x5 (commented):
`hack/templates.sh` must be adjusted to generate beta version templates.
Rest looks good

templates/provider/kcm-templates/files/templates/cluster-api-provider-aws.yaml
Kshatrix: this one has to stay on alpha to not break upgrades

a13x5 (approved):
LGTM
Missed that templates must have aplha version on the release to not break the upgrades.

Issue #1488, opened 2025-05-08
[enh] extend KSM API to allow provider-agnostic approach


BROngineer:
### Goals

- allow to declaratively define the set of resources required for deployment provider (sveltos, flux, etc) to operate
- add API resource which will define intermediate state for services' deployment to allow dependency management and sequential updates

PR #1489, opened 2025-05-08
enh: StateManagementProvider API and controller

BROngineer:
This PR extends API in part of state management:

- new API type `StateManagementProvider`. The goal of this type is to define the set of resources required for the deployment provider (e.g. projectsveltos) to operate;
- controller for `StateManagementProvider` type;

This PR does not affect any KCM functionality at the moment.

Fixes: #1488

### DIFF
--- a/internal/controller/statemanagementprovider/reconciler.go
+++ b/internal/controller/statemanagementprovider/reconciler.go
@@ -59,9 +59,17 @@ const (
 	emptyConditionMessage = ""
 )
 
+type (
+	DiscoveryClientFunc func(*rest.Config) (discovery.DiscoveryInterface, error)
+	DynamicClientFunc   func(*rest.Config) (dynamic.Interface, error)
+)
+
 // Reconciler reconciles a StateManagementProvider object
 type Reconciler struct {
 	client.Client
+
+	discoveryClientFunc DiscoveryClientFunc
+	dynamicClientFunc   DynamicClientFunc
 
 	config          *rest.Config
 	timeFunc        func() time.Time
@@ -140,6 +148,13 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 	r.config = mgr.GetConfig()
 
+	r.discoveryClientFunc = func(config *rest.Config) (discovery.DiscoveryInterface, error) {
+		return discovery.NewDiscoveryClientForConfig(config)
+	}
+	r.dynamicClientFunc = func(config *rest.Config) (dynamic.Interface, error) {
+		return dynamic.NewForConfig(config)
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(controller.TypedOptions[ctrl.Request]{
 			MaxConcurrentReconciles: 10,
@@ -172,7 +187,7 @@ func (r *Reconciler) ensureRBAC(ctx context.Context, smp *kcmv1beta1.StateManage
 		l.V(1).Info("Finished ensuring RBAC", "duration", time.Since(start))
 	}()
 
-	adapterGVR, err := gvrFromResourceReference(ctx, r.config, smp.Spec.Adapter)
+	adapterGVR, err := r.gvrFromResourceReference(ctx, r.config, smp.Spec.Adapter)
 	if err != nil {
 		reason = kcmv1beta1.StateManagementProviderRBACFailedToGetGVKForAdapterReason
 		message = fmt.Sprintf("Failed to ensure RBAC: %v", err)
@@ -181,7 +196,7 @@ func (r *Reconciler) ensureRBAC(ctx context.Context, smp *kcmv1beta1.StateManage
 	gvrList := []schema.GroupVersionResource{adapterGVR}
 
 	for _, provisioner := range smp.Spec.Provisioner {
-		provisionerGVR, err := gvrFromResourceReference(ctx, r.config, provisioner)
+		provisionerGVR, err := r.gvrFromResourceReference(ctx, r.config, provisioner)
 		if err != nil {
 			reason = kcmv1beta1.StateManagementProviderRBACFailedToGetGVKForProvisionerReason
 			message = fmt.Sprintf("Failed to ensure RBAC: %v", err)
@@ -279,7 +294,8 @@ func (r *Reconciler) ensureServiceAccount(ctx context.Context, smp *kcmv1beta1.S
 
 	desiredSA := &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: smp.Name + serviceAccountSuffix,
+			Name:      smp.Name + serviceAccountSuffix,
+			Namespace: r.SystemNamespace,
 		},
 	}
 
@@ -384,7 +400,7 @@ func (r *Reconciler) ensureAdapter(ctx context.Context, config *rest.Config, smp
 		l.V(1).Info("Finished ensuring adapter", "duration", time.Since(start))
 	}()
 
-	adapter, err := getReferencedObject(ctx, config, smp.Spec.Adapter)
+	adapter, err := r.getReferencedObject(ctx, config, smp.Spec.Adapter)
 	if err != nil {
 		reason = kcmv1beta1.StateManagementProviderFailedToGetResourceReason
 		message = fmt.Sprintf("Failed to get adapter object: %v", err)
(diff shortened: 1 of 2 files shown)


PROJECT
```diff
@@ -288,4 +288,12 @@ resources:
   kind: ProviderInterface
   path: github.com/k0rdent/kcm/api/v1beta1
   version: v1beta1
+- api:
+    crdVersion: v1
+    namespaced: false
```
zerospiel, line 293: nit: i was under the impression if a resource is cluster wide, then the field is actually omit by the `kubebuilder`

api/v1alpha1/templatechain_common.go
```diff
@@ -152,6 +152,7 @@ func (s *TemplateChainSpec) UpgradePaths(templateName string) ([]UpgradePath, er
 	// Convert map back to slice
 	result := make([]UpgradePath, 0, len(uniquePaths))
 	for _, path := range uniquePaths {
+		slices.Sort(path)
```
zerospiel, line 155: shouldn't this change be rather in the `v1beta1`?

api/v1beta1/statemanagementprovider_types.go
```diff
@@ -0,0 +1,204 @@
...
+	Version string `json:"version"`
+
+	// Resources is the list of resources under given APIVersion
+	Resources []string `json:"resources"`
+}
+
+// StateManagementProviderStatus defines the observed state of StateManagementProvider
+type StateManagementProviderStatus struct {
+	// Conditions is a list of conditions for the state management provider
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
zerospiel, line 167: shouldn't we add the following as recommended by the type definition?

```yaml
// +patchMergeKey=type
// +patchStrategy=merge
// +listType=map
// +listMapKey=type
```

there are other resources that also miss these keys (that is just a notice that the current code base also misses it)

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -5,6 +5,36 @@ metadata:
   labels:
   {{- include "kcm.labels" . | nindent 4 }}
 rules:
+- apiGroups:
```
zerospiel, line 8: The comment is not related to this particular file but rather to the user permissions: who is supposed to use the new resource? Should we add user-facing roles for it?

zerospiel (commented):
Just a brief CR with very basic comments. Will go through after reading the design and the business logic in the whole

kylewuolle (approved):

zerospiel:
@BROngineer I reckon this PR can wait a little before merging to the `main` branch? Since a release branch will be based on the current `main`, and we do not want to introduce only parts of a feature. Is it okay to wait until the release branch?

internal/controller/statemanagementprovider/reconciler.go
```diff
@@ -0,0 +1,647 @@
...
+		l.V(1).Info("Finished ensuring provisioner", "duration", time.Since(start))
+	}()
+
+	var (
+		provisioner *unstructured.Unstructured
+		ready       bool
+		err         error
+	)
+	provisionersReady := true
+	for _, item := range smp.Spec.Provisioner {
```
wahabmk, line 368: A general question about this approach came to mind while reviewing this PR. 

Taking the provisioner references as an example, suppose if a resource referenced in `smp.Spec.Provisioner` is no longer available. In such a case, we should run the reconcile loop for the StateManagementProvider to reflect that this resource is no longer available. But looking at this PR, the reconcile loop will not be triggered in such a case.

This is true for `smp.Spec.ProvisionerCRDs` and `smp.Spec.Adapter` as well.

wahabmk (approved):

wahabmk (approved):

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -5,6 +5,36 @@ metadata:
...
+  - list
+  - update
+  - bind
+  - escalate
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts/token
+  verbs:
+  - create
```
zerospiel, line 37: duplicates the lines below

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -5,6 +5,36 @@ metadata:
...
+  - clusterroles
+  - clusterrolebindings
+  verbs:
+  - create
+  - watch
+  - get
+  - list
+  - update
+  - bind
+  - escalate
```
zerospiel, line 31: aren't these verbs too wide in terms of the `clusterrole` itself? I wonder if that is even possible to decouple these rules into fine-grained `roles` instead of a single `clusterrole`. If that is not feasible then just resolve the question

internal/controller/statemanagementprovider/reconciler.go
```diff
@@ -0,0 +1,647 @@
...
+		l.Info("StateManagementProvider reconciled", "duration", time.Since(start))
+	}()
+
+	fillConditions(smp, r.timeFunc())
+
+	// We'll ensure RBAC resources first and return in case of an error. Without RBAC
+	// resources, we'll not be able to reconcile other resources.
+	if reconcileErr := r.ensureRBAC(ctx, smp); reconcileErr != nil {
+		record.Warnf(smp, smp.Generation, kcmv1beta1.StateManagementProviderFailedRBACEvent,
+			"Failed to ensure RBAC for %s %s: %v", kcmv1beta1.StateManagementProviderKind, client.ObjectKeyFromObject(smp), reconcileErr)
```
zerospiel, line 118: a generic question about the `events`: won't these be "fired" always on the error, meaning there might be duplicates potentially? might it be a good idea to base the `events` on unique `conditions`, lest we get duplicates?

> !

This question is applicable to all of the `events` records

internal/controller/statemanagementprovider/reconciler.go
```diff
@@ -0,0 +1,647 @@
...
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      smp.Name + serviceAccountSuffix,
+				Namespace: r.SystemNamespace,
+			},
+		},
+	}
+	op, err := controllerutil.CreateOrUpdate(ctx, r.Client, binding, func() error {
```
zerospiel: a generic question regarding `CreateOrUpdate`: isn't it redundant to call the function in such a manner? I mean, e.g. for a `ClusterRoleBinding`, the `RoleRef` field is immutable, and it does not make much sense to call the function potentially trying to change the field (this might be avoided by checking if `RoleRef` is already set in the closure function of the `CreateOrUpdate`). For the `ClusterRole` I reckon it is okay, always to attach the same rules, though it might be checked beforehand in another way, avoiding usage of the `CreateOrUpdate`.

The question arises because the `CreateOrUpdate` function in general makes more harm than good, and in some cases might lead to redundant updates or even errors. This is mostly an open question, though I believe that it is up to the implementer whether to use or avoid the function

zerospiel: Attaching #1568 here for the sake of an example

templates/provider/kcm/templates/rbac/controller/roles.yaml
BROngineer: I thought about the same - wide permissions, clusterrole, etc. However, resources defined in the `StateManagementProvider` object could be cluster-scoped. Thus, we need to create a clusterrole to grant access to them. Hence, I'll keep it as-is.

internal/controller/statemanagementprovider/reconciler.go
BROngineer: Tbh, I can't get what you want to say. If you mean that event emitting should be based on some `.status.condition[].*`, then it means to break the convention to not to reconcile an object basing on it's status.

Aside from that I do not see any problems in events duplication - it's a common practice in kubernetes, when you describe an object and see (xN) events. Also, from observability point of view continuously firing event might be more indicative than single event.

internal/controller/statemanagementprovider/reconciler.go
BROngineer: Yeah, as we already faced to an issue with createOrUpdate it might be better to avoid it until it is really necessary. Thanks for catch!

internal/controller/statemanagementprovider/reconciler.go
zerospiel: I meant that if the same error is occurring continuously, e.g., hours-long, producing hundreds (or thousands) of the same events, it is not very useful. In most of the currently existing controllers, the events are emitted only if a condition has changed. On the other hand, if a custom rate limiter is configured in a way that eventually back-off duration becomes high enough not to generate the very same event once in several seconds, then it is fine

Anyway, I was just asking if such a behavior (e.g., having `(x1000) EventName` for each of an object) is expected, if so, then please resolve the question

internal/controller/statemanagementprovider/reconciler.go
BROngineer: @zerospiel could you take a look on the updated code. Thx

zerospiel: Looks good

zerospiel (commented):
LGTM

PR #1496, opened 2025-05-12
chore: bump openstack ORC to 2.1.0

zerospiel:
* bump OpenStack ORC resources to v2.1.0
* preserve the v1.0.0 manifest for bw-comp
* pin ORC version in the Makefile
* bump CAPI OS provider version accordingly
* small field alignments at ProviderInterface
  and Management API types (alpha,beta)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,15 +101,11 @@ templates-generate:
 
 .PHONY: capo-orc-fetch
 capo-orc-fetch: CAPO_DIR := $(PROVIDER_TEMPLATES_DIR)/cluster-api-provider-openstack
-capo-orc-fetch: CAPO_ORC_VERSION ?= $(shell grep 'orcVersion:' $(CAPO_DIR)/values.yaml | cut -d '"' -f 2)
+capo-orc-fetch: CAPO_ORC_VERSION := 2.1.0
 capo-orc-fetch: CAPO_ORC_TEMPLATE := "$(CAPO_DIR)/templates/orc-$(shell echo $(CAPO_ORC_VERSION) | sed 's/\./-/g').yaml"
-capo-orc-fetch: yq
+capo-orc-fetch:
 	@if ! test -s "$(CAPO_ORC_TEMPLATE)"; then \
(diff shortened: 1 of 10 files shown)


s3rj1k:
Would be nice to have point release for charts, people use Apple Silicon and ask questions about why k0rdent Management does not work, one of reasons old ORC that has no ARM images

templates/provider/cluster-api-provider-openstack/templates/orc-1-0-0.yaml
```diff
@@ -1,856 +0,0 @@
-{{ if (eq .Values.orcVersion "1.0.0") }}
```
eromanova, line 1: If `orcVersion` was explicitly set to `1.0.0` in the management spec (I know this was the default value and overriding it might not make sense, but this scenario can still occur), and CAPO was upgraded from 0.3.1 to 0.3.2, all orc components will be removed because the condition no longer matches.
Wouldn’t it make sense to keep the support for `orc-1-0-0` as well?

zerospiel: It makes sense in that way. I would rather rework the whole approach, completely removing this knob, otherwise we'll end with dozens of manifests for the BW compatibility.

My idea is:
1 remove the knob (`orcVersion`)
2 drop the `if` statement (since there won't be such a value)
3 pin the orc version in the `makefile`

File names might still be versioned to be explicit about the current versioning 

Let me know what you think

zerospiel: Of course, the proposed solution implies that a potential user of the `orcVersion` value must remove it. But I would embrace it that was,y having (almost) no users of the value

eromanova: I think it makes sense. Let's proceed with the new approach.

zerospiel: So I've preserved the 1.0.0 manifest, added a new one without any changes to it, namely no `if` statements and placeholders

eromanova (approved):

PR #1546, opened 2025-05-22
ci: aireview timeout + concurrency cancel

zerospiel:
Sets timeout for the AI-Review WF and forces to be canceled if new commits / rebases arrive

### DIFF
--- a/.github/workflows/ai-codereview.yml
+++ b/.github/workflows/ai-codereview.yml
@@ -13,6 +13,10 @@ jobs:
       issues: write
       pull-requests: write
     runs-on: ubuntu-latest
+    concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/ai-codereview.yml
```diff
@@ -13,6 +13,10 @@ jobs:
       issues: write
       pull-requests: write
     runs-on: ubuntu-latest
+    concurrency:
+      group: cloud-${{ github.head_ref || github.run_id }}
```
eromanova, line 17: Let's name this group something to be unique, for example: `ai-codereview-${{ github.head_ref || github.run_id }}`. The current name intersects with the cloud E2E tests group, and I'm not sure it won't block each other.

eromanova (approved):

Issue #1555, opened 2025-05-26
[bug] orc templates ignore global.registry


Kshatrix:
orc-controller-manager in the capi-openstack provider chart does not consider .Values.global.registry

PR #1566, opened 2025-05-27
fix: use global.registry in orc templates

Kshatrix:
Fixes: #1555


templates/provider/cluster-api-provider-openstack/templates/orc-2-1-0.yaml
```diff
@@ -5117,7 +5117,7 @@ spec:
         - --health-probe-bind-address=:8081
         command:
         - /manager
-        image: quay.io/orc/openstack-resource-controller:v2.1.0
+        image: {{ default "quay.io" (and .Values.global .Values.global.registry) }}/orc/openstack-resource-controller:v2.1.0
```
zerospiel, line 5120: This will be overwritten by any next `make capo-orc-fetch` call with an updated `CAPO_VERSION`. I think we should fix it within the recipe itself. Also, does it make sense to partially hardcode the rest of the path (along with the image's version) in case of `global.registry`?

Extra: I think now we could safely drop the `orc-1-0-0.yaml` and stop versioning the file name since only we control this manifest.

PR #1567, opened 2025-05-27
feat: add non-root volumes to AWS Templates

### LINKED ISSUE #1482: [enhancement] add support for extra volume in aws-standalone-cp clustertemplate

When deploying a clustertemplate using `aws-standalone-cp-0-2-1`, there is only one volume (the root volume) deployed with each node.

When deploying Ceph, it depends on the availability of a node-attached volume which has not been formatted/partitioned. Specifically the worker nodes, in the case of Ceph.

While it is possible to create and attach a volume to the worker nodes outside of k0rdent, this seems like something that k0rdent should be managing.

This issue appears to have considered this previously (`.spec.nonRootVolumes`), but ultimately no action was taken in this regard: #168

This PR appears to address the equivalent feature in `openstack-standalone-cp`: #1059 

1. Expose `.spec.nonRootVolumes` (or something equivalent to `additionalBlockDevices` used in `openstack-standalone-cp`) for the clusterdeployment nodes (both worker and control-plane nodes).
2. Document this feature and provide a simple example that uses it.

1. Create and attach EBS volume via AWS UI -- manual and error-prone.
2. Write Python code using `kubernetes` and `boto3` libs to identify the instance IDs of the worker nodes, create a volume of the desired size, then attach each new volume to each worker node. (No partitioning or formatting required at this time.)

**Additional context**
n/a

zerospiel:
Added `nonRootVolumes` configuration to the AWS ClusterTemplates (standalone, EKS, hosted).

Updated JSON Schema validations to be generated along with instructions on how to do it.

Bumped the corresponding templates' versions.

Generalized the approach of the `aws-standalone-cp` to set up OIDC, preserving backward-compatibility

Fixes #1482

### DIFF
--- a/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
+++ b/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
@@ -15,15 +15,19 @@ spec:
       - --enable-cloud-provider
       - --kubelet-extra-args="--cloud-provider=external"
       - --disable-components=konnectivity-server
+    {{- $auth := .Values.k0s.auth }}
+    {{- $files := .Values.k0s.files }}
+    {{- if or (and $auth $auth.enabled) $files }}
     files:
-      {{- if .Values.k0s.auth.enabled }}
-      - content: |
-          {{- with .Values.k0s.auth.config }}
-          {{- toYaml . | nindent 14 }}
-          {{- end }}
-        permissions: "0644"
-        path: /etc/k0s/auth/auth-config.yaml
-      {{- end }}
+    {{- if and $auth $auth.enabled }}
+    - content: | {{ toYaml $auth.config | nindent 8 }}
+      permissions: "0644"
+      path: /etc/k0s/auth/auth-config.yaml
+    {{- end }}
+    {{- range $file := $files }}
+    - {{ toYaml $file | trim | nindent 6 }}
+    {{- end }}
+    {{- end }}
     k0s:
       apiVersion: k0s.k0sproject.io/v1beta1
       kind: ClusterConfig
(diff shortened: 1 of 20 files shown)


templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
```diff
@@ -15,15 +15,32 @@ spec:
...
+    files:
+    - content: |
+      {{- with .Values.k0s.auth.config }}
+      {{- toYaml . | nindent 8 }}
       {{- end }}
+      permissions: "0644"
+      path: /etc/k0s/auth/auth-config.yaml
+    {{- else if and (.Values.k0s.files) (not .Values.k0s.auth) }}
+    files: {{ toYaml .Values.k0s.files | nindent 4 }}
+    {{- end }}
```
eromanova, line 43: Can this block be simplified? Currently, I find it a little bit excessive. Something like:
```
{{- $auth := .Values.k0s.auth }}
{{- $files := .Values.k0s.files }}
{{- if or (and $auth $auth.enabled) $files }}
files:
{{- if and $auth $auth.enabled }}
  - content: |
{{ toYaml $auth.config | nindent 8 }}
    permissions: "0644"
    path: /etc/k0s/auth/auth-config.yaml
{{- end }}
{{- range $file := $files }}
  - {{ toYaml $file | trim | nindent 4 }}
{{- end }}
{{- end }}
```
Or does it not render properly this way?

templates/cluster/aws-standalone-cp/values.yaml
```diff
@@ -1,77 +1,66 @@
...
+    baseOS: "" # @schema description: The name of the base operating system to use for image lookup the AMI is not set; type: string
+  uncompressedUserData: false # @schema description: Specify whether the user data is gzip-compressed before it is sent to ec2 instance. Cloud-init has built-in support for gzip-compressed user data. User data stored in aws secret manager is always gzip-compressed; type: boolean
+  nonRootVolumes: [] # @schema title: Non-root storage volumes; description: Configuration options for the non root storage volumes, format: https://pkg.go.dev/sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2#AWSMachineSpec.NonRootVolumes; type: array; item: object
 
 # K0s parameters
-k0s:
-  version: v1.32.3+k0s.0
-  api:
-    extraArgs: {}
-  auth:
```
eromanova, line 57: Can we keep the auth block in values? Why was it removed?

PR #1580, opened 2025-05-28
feat: Mount registry certificate secret in CAPI operator

### LINKED ISSUE #1571: [enh] Support custom CA certificates for pulling provider components in CAPI operator
When the CAPI operator attempts to fetch provider components from a private registry secured with a custom CA, it fails with the error:
```
x509: certificate signed by unknown authority.
```

Currently, the CAPI operator API does not expose a way to reference a secret containing the CA certificate. As a workaround, the certificate must be mounted into the CAPI operator deployment via a secret, and the `SSL_CERT_FILE` environment variable must be set to point to the mounted certificate path.

eromanova:
If `RegistryCertSecret` is set, add a new volume and volumeMount referencing the certificate from that secret to the CAPI operator's values.

This allows the CAPI operator to fetch provider components from the private registry signed by an unknown authority.

Also, refactor the component values overrides (all internal logic from `enableAdditionalComponents` was moved to the HelmRelease level without overwriting the Management spec, all component values overrides were moved to a single function).

Fixes #1571

### DIFF
--- a/internal/controller/management_controller.go
+++ b/internal/controller/management_controller.go
@@ -725,25 +725,33 @@ func (r *ManagementReconciler) getComponentValues(ctx context.Context, name stri
 			},
 		}
 
+		capiOperatorValues := make(map[string]any)
 		if r.Config != nil {
 			if err := certmanager.VerifyAPI(ctx, r.Config, r.SystemNamespace); err != nil {
 				return nil, fmt.Errorf("failed to check if cert-manager API is installed: %w", err)
 			}
 			l.Info("Cert manager is installed, enabling additional components")
 			componentValues["admissionWebhook"] = map[string]any{"enabled": true}
-			componentValues["cluster-api-operator"] = map[string]any{"enabled": true}
 			componentValues["velero"] = map[string]any{"enabled": true}
+			capiOperatorValues = map[string]any{"enabled": true}
 		}
 
 		if r.RegistryCertSecret != "" {
-			v, ok := currentValues["cluster-api-operator"].(map[string]any)
-			if !ok {
-				return nil, fmt.Errorf("failed to cast 'cluster-api-operator' (type %T) to map[string]any", currentValues["cluster-api-operator"])
+			v := make(map[string]any)
+			if currentValues != nil {
+				if raw, ok := currentValues["cluster-api-operator"]; ok {
+					var castOk bool
+					if v, castOk = raw.(map[string]any); !castOk {
+						return nil, fmt.Errorf("failed to cast 'cluster-api-operator' (type %T) to map[string]any", raw)
+					}
+				}
 			}
-			processCAPIOperatorCertVolumeMounts(v, r.RegistryCertSecret)
-		}
 
-	case "projectsveltos":
+			capiOperatorValues = chartutil.CoalesceTables(capiOperatorValues, processCAPIOperatorCertVolumeMounts(v, r.RegistryCertSecret))
+		}
+		componentValues["cluster-api-operator"] = capiOperatorValues
+
+	case kcm.ProviderSveltosName:
 		componentValues = map[string]any{
 			"projectsveltos": map[string]any{
 				"registerMgmtClusterJob": map[string]any{
@@ -768,14 +776,14 @@ func (r *ManagementReconciler) getComponentValues(ctx context.Context, name stri
 
 	var merged chartutil.Values
 	// for projectsveltos, we want new values to override values provided in Management spec
-	if name == "projectsveltos" {
+	if name == kcm.ProviderSveltosName {
 		merged = chartutil.CoalesceTables(componentValues, currentValues)
 	} else {
 		merged = chartutil.CoalesceTables(currentValues, componentValues)
 	}
 	raw, err := json.Marshal(merged)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to marshal values for %s component: %w", name, err)
 	}
 	return &apiextensionsv1.JSON{Raw: raw}, nil
 }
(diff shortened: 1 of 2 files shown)


internal/controller/management_controller.go
```diff
@@ -957,6 +961,63 @@ func (r *ManagementReconciler) enableAdditionalComponents(ctx context.Context, m
...
+				map[string]any{
+					"key":  "tls.crt",
+					"path": "tls.crt",
+				},
+			},
+		},
+	}
+	certVolumes := []any{webhookCertVolume, registryCertVolume}
+	if existing, ok := capiOperatorValues["volumes"].([]any); ok {
+		capiOperatorValues["volumes"] = append(existing, certVolumes...)
```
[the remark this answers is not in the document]
eromanova: yes, of course, there’s no point in overriding the Management spec. I’ll rework that part.

internal/controller/management_controller.go
```diff
@@ -707,90 +702,85 @@ type component struct {
...
+			if !ok {
+				return nil, fmt.Errorf("failed to cast 'cluster-api-operator' (type %T) to map[string]any", currentValues["cluster-api-operator"])
+			}
+			processCAPIOperatorCertVolumeMounts(v, r.RegistryCertSecret)
+		}
 
-	raw, err := json.Marshal(chartutil.CoalesceTables(values, defaultValues))
-	if err != nil {
-		return nil, err
+	case "projectsveltos":
```
wahabmk, line 746:
```suggestion
	case kcm.ProviderSveltosName:
```

internal/controller/management_controller.go
```diff
@@ -707,90 +702,85 @@ type component struct {
...
+		componentValues = chartutil.CoalesceTables(componentValues, globalValues)
 	}
 
-	globalValues := map[string]any{
-		"global": map[string]any{
-			"registry": registry,
-		},
+	var merged chartutil.Values
+	// for projectsveltos, we want new values to override values provided in Management spec
+	if name == "projectsveltos" {
```
wahabmk, line 771:
```suggestion
	if name == kcm.ProviderSveltosName {
```

wahabmk (approved):
Just minor suggestions

internal/controller/management_controller.go
```diff
@@ -841,120 +831,80 @@ func (r *ManagementReconciler) getWrappedComponents(mgmt *kcm.Management, releas
...
-				admissionWebhookValues["enabled"] = true
-			} else {
-				l.Info("KCM admission webhook is disabled")
-			}
-		}
+	registryCertMount := map[string]any{
+		"mountPath": "/etc/ssl/certs/registry-ca.pem",
+		"name":      registryCertSecret,
+		"readOnly":  true,
+		"subPath":   "tls.crt",
```
zerospiel: Mostly a question: I'm a little confused why we are "forced" here to use the `subPath` property? The volume [won't receive any updates](https://kubernetes.io/docs/concepts/storage/volumes/#secret) in case of updates to the `Secret`. Can we get rid of the `subPath`, either renaming the `path` of the corresponding `volume` or omitting the `registry-ca.pem` if that would work?

Besides, the `readOnly` property is not explicitly necessary (but might be good for readability, so it is up to the implementer).

internal/controller/management_controller.go
```diff
@@ -707,90 +702,85 @@ type component struct {
...
+	if name == "projectsveltos" {
+		merged = chartutil.CoalesceTables(componentValues, currentValues)
+	} else {
+		merged = chartutil.CoalesceTables(currentValues, componentValues)
 	}
-
-	raw, err := json.Marshal(chartutil.CoalesceTables(values, globalValues))
+	raw, err := json.Marshal(merged)
 	if err != nil {
 		return nil, err
```
zerospiel: nit: could wrap error with the `name` of the given component here and in the beginning to enhance debugging if any

zerospiel:
Nit-picking: we have to name PRs and commit msgs with `feat` rather than `enh(ancement)` (I believe that confusion comes from the issues' names and labels). Let me know what you think

internal/controller/management_controller.go
eromanova: Unfortunately, the volume mount does not work correctly without specifying the subPath. When I attempt to mount the certificate file directly, Kubernetes mounts it as a directory at `/etc/ssl/certs/registry-ca.pem`, resulting in the actual file being placed at `/etc/ssl/certs/registry-ca.pem/registry-ca.pem`. This breaks the expected file path and causes the system to fail to locate the trusted cert correctly.

Using `subPath` ensures that the file is mounted precisely at the desired path without replacing the entire contents of the target directory. The CAPI Operator helm chart does not provide a mechanism to inject certificates as it does not allow us to modify the container spec to include an init container that could run `update-ca-certificates` or copy the cert manually into place. So, `subPath` is currently the only viable method for mounting a single certificate file. As the registry certificate is not expected to be changed regularly, I think we can proceed with this approach.

Regarding the `readOnly` option, I agree it can be safely removed.

zerospiel:
> As far as I know, `goreleaser` can be configured to include PRs with any tags, so we could support a combination like enh, enhancement, and feat. But I think it would be beneficial to establish a clear and consistent tagging convention across the project.

That's right. My point is related exclusively to today's stand, no more. In general, yes, we could configure anything in a way we'd like

internal/controller/management_controller.go
zerospiel: Ok got it. Thanks for the elaboration. Then we have to leave it as it is right now without changes, and then document this behavior, and (optionally) make a PR to the upstream to amend the chart's spec

internal/controller/management_controller.go
```diff
@@ -707,90 +702,93 @@ type component struct {
...
-}
-
-// TODO: Refactor this kludge. Generalize default values passing.
-func applyGlobalRegistry(config *apiextensionsv1.JSON, registry string) (*apiextensionsv1.JSON, error) {
-	values := chartutil.Values{}
-	if config != nil && config.Raw != nil {
-		err := json.Unmarshal(config.Raw, &values)
-		if err != nil {
-			return nil, err
+	if r.GlobalRegistry != "" {
```
zerospiel, line 768: I reckon that the `GlobalK0sURL` field is missing

zerospiel: Ah, sorry, it is supposed for the `cld` controller, resolving

PR #1581, opened 2025-05-28
fix: pass cluster name explicitly to ccm

aglarendil:


kylewuolle (dismissed):

zerospiel (changes requested):
Please do all the required steps described in the PR template you've removed before posting the PR: bump versions, regenerate, bump the according dev configs etc

templates/cluster/openstack-standalone-cp/templates/_helpers.tpl
```diff
@@ -14,6 +14,14 @@
     {{- include "cluster.name" . }}-cp
 {{- end }}
 
+{{- define "ccm-cluster.name -}}
+    {{- if $.Values.ccmCanViewAllProjects }}
+    {{- printf "%s--%s" {{ .Release.Namespace }} {{- include "cluster.name" . }} }}
+    {{ else }}
+    {{- include "cluster.name" . }}
+    {{- end }}
+{{- end }}
```
zerospiel: The shortcut is unused. Was it intended to be used within the `templates/cluster/openstack-standalone-cp/templates/k0scontrolplane.yaml` instead of the current `cluster.name`?

templates/cluster/openstack-standalone-cp/values.yaml
```diff
@@ -13,6 +13,7 @@ clusterNetwork:
 clusterLabels: {}
 clusterAnnotations: {}
 
+ccmCanViewAllProjects: false
```
zerospiel: The value is unused. The same question as in the previous comment: was it supposed to be used by the `templates/cluster/openstack-standalone-cp/templates/k0scontrolplane.yaml`?

aglarendil: thanks for noticing. I forgot this one line, yes. updated

templates/cluster/openstack-standalone-cp/templates/_helpers.tpl
aglarendil: thanks for noticing. I forgot this one line, yes. updated

templates/cluster/openstack-standalone-cp/values.yaml
zerospiel: I now realize: does it make sense to set the field to `null` instead of `false` (and change the shortcut accordingly if that makes sense)? With the current implementation, the `values.cluster.name` is always being added to the CCM values (with the default value something like `namespace-clustername` will be added). Is it expected to have that `namespace-clustername` always set? 

To sum up, is it expected always to have some value set?

aglarendil: It is the other way around, the default value falls back to the shortest variant and yields only cluster name. It is important to have unique cluster names set, otherwise it may lead to conflicts for ccm and break running clusters.

 If the credentials however allow cross-project access, then this flag must be flipped to true, and then we add one more dimension to uniqueness such as namespace name.

templates/provider/kcm-templates/Chart.yaml
```diff
@@ -13,4 +13,4 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.0
+version: 1.0.1
```
zerospiel: This change should not be presented. Only the corresponding cluster template version is to be bumped. This is the RC of failing unit tests

templates/cluster/openstack-standalone-cp/values.yaml
zerospiel: Right, accidentally flipped the conditional. No further questions in this regard

templates/cluster/openstack-standalone-cp/templates/_helpers.tpl
```diff
@@ -14,6 +14,14 @@
     {{- include "cluster.name" . }}-cp
 {{- end }}
 
+{{- define "ccm-cluster.name -}}
```
zerospiel: Please check out the failed jobs and fix the errors.

In this case:

```text
parse error at (openstack-standalone-cp/templates/_helpers.tpl:17): unterminated quoted string
```

zerospiel (changes requested):

templates/cluster/openstack-standalone-cp/templates/_helpers.tpl
```diff
@@ -14,6 +14,14 @@
     {{- include "cluster.name" . }}-cp
 {{- end }}
 
+{{- define "ccm-cluster.name" -}}
+    {{- if $.Values.ccmCanViewAllProjects }}
+    {{- printf "%s--%s" {{ .Release.Namespace }} {{- include "cluster.name" . }} }}
```
zerospiel:
```suggestion
    {{- printf "%s-%s" .Release.Namespace (include "cluster.name" .) }}
```

templates/cluster/openstack-standalone-cp/templates/_helpers.tpl
```diff
@@ -14,6 +14,14 @@
     {{- include "cluster.name" . }}-cp
 {{- end }}
 
+{{- define "ccm-cluster.name" -}}
+    {{- if $.Values.ccmCanViewAllProjects }}
+    {{- printf "%s--%s" .Release.Namespace (include "cluster.name" .) }}
```
zerospiel, line 19: Is a double dash expected, e.g. `default--openstack-standalone-cp`?

aglarendil: yep, double dash here is a conversion of "/" to avoid issues when openstack will create objects including it

zerospiel (approved):

PR #1584, opened 2025-05-30
fix: Ensure services are removed before ClusterDeployment removal

### LINKED ISSUE #942: [bug] <your bug title here>
The k0rdent cluster template can include services (e.g., ingress-nginx) that triggers the creation of external resources, such as load balancers and public IPs. These resources are not automatically cleaned up when the cluster is deleted, leading to potential resource exhaustion and unnecessary costs in cloud environments. Users may lack the permissions or awareness to manually clean up these dangling resources, resulting in long-term issues.

To Reproduce:
- Create a cluster using the k0rdent cluster template with ingress-nginx enabled.
- Install a service e.g. ingress-nginx on the the cluster
- Delete the cluster without explicitly removing the ingress-nginx service or associated resources.
- Check for remaining resources (e.g., load balancers or public IPs) in the cloud environment.
(Easy to see with OpenStack provider for example - it leaves behind a load-balancer and associated public IP used for ingress-nginx)

Expected behavior:
When a cluster is deleted, all associated resources (e.g., load balancers, public IPs) created by the services in the cluster should also be removed automatically, ensuring a clean teardown without leaving dangling resources.

Additional context:
This issue becomes critical in environments with frequent cluster creation and teardown, as it may lead to:

- Resource exhaustion due to leftover resources.
- Increased cloud costs from unused, chargeable resources.
- Administrative overhead for manually identifying and removing dangling resources.

A potential solution involves enhancing the Cluster API (CAPI) providers to handle resource cleanup more effectively. For example:
- Implementing a unique UUID tag for each cluster, applied to all associated resources.
- Allowing the CAPI provider to query cloud providers for resources tagged with the cluster's UUID during deletion and ensuring their removal.

This approach could improve lifecycle management and prevent resource leaks in high-turnover cluster environments.

kylewuolle:
Before deleting the HelmRelease when reconciling a deleted ClusterDeployment we need to check to ensure that any services have been uninstalled because otherwise the control plane on the child cluster will be removed and the uninstall of the resources created by the service will remain. For example the ingress-nginx creates a load balancer that needs to be removed prior to deleting the child cluster.

Fixes #942


internal/controller/clusterdeployment_controller.go
```diff
@@ -847,6 +847,31 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, cd *k
...
+			sveltosv1beta1.ClusterNameLabel: cd.Name,
+		},
+	}
+
+	clusterSummaryList := &sveltosv1beta1.ClusterSummaryList{}
+	if err := r.Client.List(ctx, clusterSummaryList, listOptions...); client.IgnoreNotFound(err) != nil {
+		return ctrl.Result{}, err
+	}
+
+	if len(clusterSummaryList.Items) > 0 {
```
zerospiel: This line could be dropped: if a slice/array is empty, no loop cycles occur

internal/controller/clusterdeployment_controller.go
```diff
@@ -847,6 +847,31 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, cd *k
...
+		for _, cs := range clusterSummaryList.Items {
+			for _, helmReleaseSummary := range cs.Status.HelmReleaseSummaries {
+				if helmReleaseSummary.Status == sveltosv1beta1.HelmChartStatusManaging {
+					return ctrl.Result{}, fmt.Errorf("service %s needs to be removed prior to uninstall of helm release %s",
```
zerospiel: Isn't it better in that case to log the message and requeue with the default duration, instead of firing the errors? The result will be the same, but without errors, the logs will be clearer, and (probably) fewer "reconcile cycles" will occur

internal/controller/clusterdeployment_controller.go
```diff
@@ -847,6 +847,31 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, cd *k
 		return ctrl.Result{}, err
 	}
 
+	// Verify that any service templates which have been installed are removed prior to deleting the helm release
+	// otherwise the k8s control plane will potentially be deleted prior to cleaning up resources
+	listOptions := []client.ListOption{
+		client.InNamespace(cd.Namespace),
+		client.MatchingLabels{
+			sveltosv1beta1.ClusterNameLabel: cd.Name,
```
wahabmk, line 844: The clusterdeployment's name == profile's name but it is != clustersummary's name. E.g:
```sh
➜  ~ kkcm get clusterdeployment
NAME                   READY   SERVICES   TEMPLATE                        MESSAGES          AGE
wali-openstack-dev-1   True    1/1        openstack-standalone-cp-1-0-0   Object is ready   21m
➜  ~ kkcm get profile                                  
NAME                   AGE
wali-openstack-dev-1   21m
➜  ~ kkcm get clustersummary
NAME                                                AGE
p--wali-openstack-dev-1-capi-wali-openstack-dev-1   17m
```

The following function can be used to get clustersummary's name from profile's name:
https://github.com/k0rdent/kcm/blob/5dcf7a608743df102765eed3772b658a454fb98f/internal/controller/multiclusterservice_controller.go#L465

kylewuolle: This is different though. It's listing by cluster name label and not cluster summary name. The addon controller uses this approach as well : https://github.com/projectsveltos/addon-controller/blob/01d8651dbfb08d7ff60c2c0a93235a152e1dc71b/controllers/utils.go#L145

wahabmk: Ohk. I was thinking of getting the clustersummary by its name. So if we go with the label matching approach, maybe we should use all 3 labels to match as is done by the addon-controller.

PR #1593, opened 2025-06-03
fix: global values for azure,capi,k0smotron providers

### LINKED ISSUE #1587: [bug] Not all images are downloading from the same registry when using global values
When using `globalRegistry` flag some images are still being downloaded from the public repositories.
The issue is with the following images:

`cluster-api-manager` - due to `global.registry` values is not being propagated to the capi template
`azure-service-operator` - insufficient overrides in the deployment spec
`kube-proxy` for k0smotron - insufficient overrides in the deployment spec

a13x5:
Resolves #1587 

Please note that for aso it's required to literally copy the entire container spec in order to change image field. It's not possible to partially change it or change it with `.spec.deployment`, since it only affects capi controller and not other deployments.

I left `--crd-pattern` option empty, but it was intended to be configurable via `ADDITIONAL_ASO_CRDS` parameter. It's not possible to make it work with `manifestPatches`, since what is patched won't be passed through envsubst. This means that  `ADDITIONAL_ASO_CRDS` if specified in `configSecret` will be ignored. 

I believe we can expose this option in values, but it looks confusing.


templates/provider/cluster-api-provider-azure/templates/provider.yaml
```diff
@@ -43,6 +44,175 @@ spec:
...
+                valueFrom:
+                  secretKeyRef:
+                    key: RATE_LIMIT_BUCKET_SIZE
+                    name: aso-controller-settings
+                    optional: true
+              - name: POD_NAMESPACE
+                valueFrom:
+                  fieldRef:
+                    fieldPath: metadata.namespace
+              image: {{ $global.registry }}/k8s/azureserviceoperator:{{ $asoVersion }}
```
eromanova, line 169: All cluster API controller images are located under the `{{ $global.registry }}/capi/<imageName>` path. For consistency, it might be best to follow the same path for the ASO as well. But it's up to you.

eromanova (approved):

templates/provider/cluster-api-provider-azure/templates/provider.yaml
zerospiel: I once had the same question, but I got the impression that it is a good practice to leave the path (and version, if applicable) the same as it is without a custom registry value. Only the registry part should be changed.

zerospiel (commented):
LGTM

templates/provider/cluster-api-provider-azure/templates/provider.yaml
a13x5: I follow certain rules on that (which will be documented separately), but generally yes the subpath left untouched. Unfortunately there are cases where it leads to quite ambiguous paths and even clashes. 
Let's leave it as it is now and discuss it with document and concrete examples later.

PR #1594, opened 2025-06-03
ci: convert AI Review WF to manual

zerospiel:
Converts [AI Review WF](https://github.com/k0rdent/kcm/blob/9ea910470c35a6d5ca3b1edff8cefe9627fe372b/.github/workflows/ai-codereview.yml) to [workflow_dispatch](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch)

### DIFF
--- a/.github/workflows/ai-codereview.yml
+++ b/.github/workflows/ai-codereview.yml
@@ -1,11 +1,7 @@
 name: AI Codereview
 on:
-  pull_request_target:
-    branches:
-      - main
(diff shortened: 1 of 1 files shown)


zerospiel:
There is no possibility to combine the pros from both `pull_request_target` (to use upstream's secrets) and `workflow_dispatch` (to run it manually)

PR #1595, opened 2025-06-03
feat: Add support for mounting registry and k0s URL certificates

### LINKED ISSUE #1563: [enh] Adjust cluster templates to support passing custom CA to k0s components
Update the cluster templates to support injecting a custom CA certificate through a secret reference. 

This secret should be passed to the K0sControlPlane resource via the `.spec.k0sConfigSpec.files.contentFrom.secretRef` field. The CA will then be rendered as a file on the target host to enable trusted communication for k0s components.

eromanova:
Adapt all cluster templates to accept registry and k0s URL secrets and mount them as certificate files. This is necessary to support custom self-signed registries.

Closes #1563

### DIFF
--- a/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
+++ b/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
@@ -15,21 +15,10 @@ spec:
       - --enable-cloud-provider
       - --kubelet-extra-args="--cloud-provider=external"
       - --disable-components=konnectivity-server
-    {{- $certs := dict "registry.crt" $global.registryCertSecret "k0s-url.crt" $global.k0sURLCertSecret }}
     {{- $auth := .Values.k0s.auth }}
     {{- $files := .Values.k0s.files }}
     {{- if or (and $auth $auth.enabled) $files $global.registryCertSecret $global.k0sURLCertSecret }}
     files:
-      {{- range $path, $secret := $certs }}
-      {{- if $secret }}
-      - contentFrom:
-          secretRef:
-            name: {{ $secret }}
-            key: tls.crt
-        permissions: "0664"
-        path: /usr/local/share/ca-certificates/{{ $path }}
-      {{- end }}
-      {{- end }}
     {{- if and $auth $auth.enabled }}
     - content: | {{ toYaml $auth.config | nindent 8 }}
       permissions: "0644"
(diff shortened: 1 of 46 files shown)


templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
```diff
@@ -15,10 +15,21 @@ spec:
...
+      {{- range $path, $secret := $certs }}
+      {{- if $secret }}
+      - contentFrom:
+          secretRef:
+            name: {{ $secret }}
+            key: tls.crt
+        permissions: "0664"
+        path: /usr/local/share/ca-certificates/{{ $path }}
+      {{- end }}
+      {{- end }}
```
zerospiel: This one is tricky to decide its place: if we put it in the first place, potentially this entry could get rewritten (i am not sure though about the order of processing the `files` list). might it be a good idea to put the cert either after the `auth.config` or even push it to the very end?

templates/cluster/azure-standalone-cp/templates/k0scontrolplane.yaml
```diff
@@ -10,6 +10,22 @@ spec:
     {{- if $global.k0sURL }}
     downloadURL: "{{ $global.k0sURL }}/k0s-{{ .Values.k0s.version }}-amd64"
     {{- end }}
+    {{- $certs := dict "registry.crt" $global.registryCertSecret "k0s-url.crt" $global.k0sURLCertSecret }}
```
zerospiel: Minor: Shouldn't the dict be declared only if either of the values below exists? Applicable elsewhere if that even matters (if no, then just resolve the comment)

templates/cluster/remote-cluster/templates/k0smotroncontrolplane.yaml
```diff
@@ -37,20 +37,20 @@ spec:
       {{- if $global.registry }}
       images:
         metricsserver:
-          image: "{{ $global.registry}}/k0sproject/metrics-server"
+          image: "{{ $global.registry }}/k0sproject/metrics-server"
```
zerospiel, line 40: Question: we do not put any `files` on this control plane since the remote-cluster template is basically the hosted type, is that right?

templates/cluster/azure-standalone-cp/templates/k0scontrolplane.yaml
eromanova: Yeah, it makes sense. Thanks. I'll update the PR.

templates/cluster/remote-cluster/templates/k0smotroncontrolplane.yaml
eromanova: Yes, that’s correct. When we deploy a remote cluster, the control plane components run in the management cluster. Just like in our other hosted control plane templates.

templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
eromanova: I’ve placed the certificates at the very end to ensure they take the highest priority.

zerospiel (approved):

PR #1598, opened 2025-06-04
feat: propagate registry-credentials across cld

### LINKED ISSUE #1560: [enh] Propagate registry-cred secret across ClusterDeployments
The KCM controller should propagate a reference to the `registry-cred` secret into the `.spec.config` field of each ClusterDeployment, enabling usage by k0s components (in `.spec.k0sConfigSpec.files.contentFrom.secretRef`)

If the secret is required to reside in the same namespace as the `K0sControlPlane`, the controller should also ensure the secret is copied into the corresponding ClusterDeployment namespace.

zerospiel:
Propagates k0sDownloadURL and Registry certificate Secrets to ClusterDeployments' configs.

Introduces a new condition indicating
whether all predeclared Secrets exist.

Copies the Secrets if required.

Fixes error when some of HelmRepository Secrets
might have been missed.

Checks if the given Secrets (HelmRepository and
Certificate Secrets) exist in the system namespace in each of the required controllers
(templates (all), release, management, clusterdeployment).

Fixes lax Conditions type properties.

Closes #1560

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -89,29 +89,30 @@ func init() {
 
 func main() {
 	var (
-		metricsAddr                string
-		probeAddr                  string
-		secureMetrics              bool
-		enableHTTP2                bool
-		templatesRepoURL           string
-		globalRegistry             string
-		globalK0sURL               string
-		insecureRegistry           bool
-		registryCredentialsSecret  string
-		registryCertSecret         string
-		createManagement           bool
-		createAccessManagement     bool
-		createRelease              bool
-		createTemplates            bool
-		validateClusterUpgradePath bool
-		kcmTemplatesChartName      string
-		enableTelemetry            bool
-		enableWebhook              bool
-		webhookPort                int
-		webhookCertDir             string
-		pprofBindAddress           string
-		leaderElectionNamespace    string
-		enableSveltosCtrl          bool
+		metricsAddr                   string
+		probeAddr                     string
+		secureMetrics                 bool
+		enableHTTP2                   bool
+		templatesRepoURL              string
+		globalRegistry                string
+		globalK0sURL                  string
+		insecureRegistry              bool
+		registryCredentialsSecretName string
+		registryCertSecretName        string
+		k0sURLCertSecretName          string
+		createManagement              bool
+		createAccessManagement        bool
+		createRelease                 bool
+		createTemplates               bool
+		validateClusterUpgradePath    bool
+		kcmTemplatesChartName         string
+		enableTelemetry               bool
+		enableWebhook                 bool
+		webhookPort                   int
+		webhookCertDir                string
+		pprofBindAddress              string
+		leaderElectionNamespace       string
+		enableSveltosCtrl             bool
 	)
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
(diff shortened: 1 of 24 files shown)


cmd/main.go
```diff
@@ -129,10 +130,11 @@ func main() {
...
 		"K0s URL prefix which will be passed directly as global.k0sURL to all ClusterDeployments configs")
-	flag.StringVar(&registryCredentialsSecret, "registry-creds-secret", "",
-		"Secret containing authentication credentials for the registry.")
-	flag.StringVar(&registryCertSecret, "registry-cert-secret", "",
-		"Secret containing a client certificate (`tls.crt`) and private key (`tls.key`) and/or a CA certificate (`ca.crt`).")
+	flag.StringVar(&registryCredentialsSecretName, "registry-creds-secret", "",
+		"Name of a Secret containing authentication credentials for the registry.")
+	flag.StringVar(&registryCertSecretName, "registry-cert-secret-name", "",
+		"Name of a Secret containing a client certificate (`tls.crt`) and private key (`tls.key`) and/or a CA certificate (`ca.crt`).")
+	flag.StringVar(&k0sURLCertSecretName, "k0s-url-creds-secret-name", "", "Name of a Secret containing a client certificate (`tls.crt`) and private key (`tls.key`) and (optionally) a CA certificate (`ca.crt`).")
```
eromanova:
```suggestion
	flag.StringVar(&k0sURLCertSecretName, "k0s-url-cert-secret-name", "", "Name of a Secret containing a client certificate (`tls.crt`) and private key (`tls.key`) and (optionally) a CA certificate (`ca.crt`) for the k0s download URL.")
```

cmd/main.go
```diff
@@ -129,10 +130,11 @@ func main() {
...
 	flag.StringVar(&globalK0sURL, "global-k0s-url", "",
 		"K0s URL prefix which will be passed directly as global.k0sURL to all ClusterDeployments configs")
-	flag.StringVar(&registryCredentialsSecret, "registry-creds-secret", "",
-		"Secret containing authentication credentials for the registry.")
-	flag.StringVar(&registryCertSecret, "registry-cert-secret", "",
-		"Secret containing a client certificate (`tls.crt`) and private key (`tls.key`) and/or a CA certificate (`ca.crt`).")
+	flag.StringVar(&registryCredentialsSecretName, "registry-creds-secret", "",
+		"Name of a Secret containing authentication credentials for the registry.")
+	flag.StringVar(&registryCertSecretName, "registry-cert-secret-name", "",
+		"Name of a Secret containing a client certificate (`tls.crt`) and private key (`tls.key`) and/or a CA certificate (`ca.crt`).")
```
eromanova:
```suggestion
		"Name of a Secret containing a client certificate (`tls.crt`) and private key (`tls.key`) and/or a CA certificate (`ca.crt`) for the registry endpoint.")
```

templates/provider/kcm/templates/deployment.yaml
```diff
@@ -28,8 +28,11 @@ spec:
         {{- if .Values.controller.registryCredsSecret }}
         - --registry-creds-secret={{ .Values.controller.registryCredsSecret }}
         {{- end }}
+        {{- if .Values.controller.k0sURLCertSecret }}
+        - --k0s-url-creds-secret-name={{ .Values.controller.k0sURLCertSecret }}
```
eromanova:
```suggestion
        - --k0s-url-cert-secret-name={{ .Values.controller.k0sURLCertSecret }}
```

internal/controller/release_controller.go
```diff
@@ -115,6 +117,11 @@ func (r *ReleaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 		}()
 	}
 
+	if _, err := utils.SetPredeclaredSecretsCondition(ctx, r.Client, release, record.Warnf, r.SystemNamespace, r.DefaultRegistryConfig.CertSecretName, r.DefaultRegistryConfig.CredentialsSecretName); err != nil {
```
eromanova: Could you please elaborate on why this condition is necessary on the release object? The release object contains the subcomponents declaration and does not relate to the provided configuration. This condition should be in the Management object, right?

zerospiel: The condition is set (if at least one of the arguments is not empty) on the following resources:

- `releases`
- `managements`
- `clusterdeployments`

Regarding the `releases` themselves, it is virtually required because the [controller creates](https://github.com/k0rdent/kcm/blob/3b366dfc298440b03bbf2545c41c1040d5fbca65/internal/controller/release_controller.go#L300) a `HelmRepository` referencing these names. I could move the check under the condition when it is the initial installation, which should be more correct in that regard.

Let me know what you think

zerospiel: I would rephrase: I would rather introduce the basic check (like in the `templates_controller`), rather than with the condition. Probably this should be less confusing

zerospiel: I've rewritten that part, so from now on, only the existence of the secrets is being checked exclusively during the initial installation

eromanova (approved):

PR #1609, opened 2025-06-06
feat: add management cluster heartbeat to telemetry

briantd:
This PR adds a management cluster heartbeat to the product telemetry.

The new heartbeat includes:
* Node-level insights into capacity
* Cluster-level insights into kcm Templates and Objects
* Ability to recognize Nvidia-based labels to identify GPU Capacity

### DIFF
--- a/internal/telemetry/decode_k8s_distribution.go
+++ b/internal/telemetry/decode_k8s_distribution.go
@@ -1,0 +1,51 @@
+// Copyright 2024
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package telemetry
+
+import (
+	"errors"
+	"strings"
+)
+
+// DecodeDistribution takes a Kubernetes version string and returns the distribution name (if matched) or "unknown"
+// See https://github.com/lensapp/lens/blob/release/v4.1/src/main/cluster-detectors/distribution-detector.ts for the original implementation
+func DecodeDistribution(k8sVersion string) (string, error) {
+	if k8sVersion == "" {
+		return "", errors.New("kubernetes version cannot be empty")
+	}
+
+	switch {
+	case strings.Contains(strings.ToLower(k8sVersion), "aliyun"):
+		return "alibaba", nil
+	case strings.Contains(strings.ToLower(k8sVersion), "eks"):
+		return "eks", nil
+	case strings.Contains(k8sVersion, "gke"):
+		return "gke", nil
+	case strings.Contains(strings.ToLower(k8sVersion), "iks"):
+		return "iks", nil
+	case strings.Contains(k8sVersion, "+k3s"):
+		return "k3s", nil
+	case strings.Contains(k8sVersion, "rancher"):
+		return "rke", nil
+	case strings.Contains(k8sVersion, "k0s"):
+		return "k0s", nil
+	case strings.Contains(k8sVersion, "vmware"):
+		return "vmware", nil
+	case strings.Contains(k8sVersion, "-CCE"):
+		return "huawei", nil
+	default:
+		return "unknown", nil
+	}
+}
--- a/internal/telemetry/tracker.go
+++ b/internal/telemetry/tracker.go
@@ -18,12 +18,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 type Tracker struct {
@@ -33,6 +35,13 @@ type Tracker struct {
 }
 
 const interval = 24 * time.Hour
+
+// Map to serve as blacklist filter for system namespaces
+var systemNamespaces = map[string]bool{
+	"kube-system":     true,
+	"kube-public":     true,
+	"kube-node-lease": true,
+}
 
 func (t *Tracker) Start(ctx context.Context) error {
 	timer := time.NewTimer(0)
@@ -50,26 +59,39 @@ func (t *Tracker) Start(ctx context.Context) error {
 func (t *Tracker) Tick(ctx context.Context) {
 	l := log.FromContext(ctx).WithName("telemetry tracker")
 
+	// Cluster deployment heartbeat; One event per cluster deployment
 	logger := l.WithValues("event", clusterDeploymentHeartbeatEvent)
 	err := t.trackClusterDeploymentHeartbeat(ctx)
 	if err != nil {
-		logger.Error(err, "failed to track an event")
+		logger.Error(err, "failed to track cluster deployment heartbeat")
 	} else {
-		logger.Info("successfully tracked an event")
+		logger.Info("successfully tracked cluster deployment heartbeat")
+	}
+
+	// Management cluster heartbeat
+	logger = l.WithValues("event", managementClusterHeartbeatEvent)
+	err = t.trackManagementClusterHeartbeat(ctx)
+	if err != nil {
+		logger.Error(err, "failed to track management cluster heartbeat")
+	} else {
+		logger.Info("successfully tracked management cluster heartbeat")
 	}
 }
 
 func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
+	// Get the management cluster
 	mgmt := &kcmv1.Management{}
 	if err := t.Get(ctx, client.ObjectKey{Name: kcmv1.ManagementName}, mgmt); err != nil {
 		return err
 	}
 
+	// Get all cluster templates
 	templatesList := &kcmv1.ClusterTemplateList{}
 	if err := t.List(ctx, templatesList, client.InNamespace(t.SystemNamespace)); err != nil {
 		return err
 	}
 
+	// Create a map to of templates to decode the ClusterDeployment.Spec.Template value
 	templates := make(map[string]kcmv1.ClusterTemplate)
 	for _, template := range templatesList.Items {
 		templates[template.Name] = template
@@ -81,6 +103,7 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
 		return err
 	}
 
+	// Iterate over all cluster deployments and emit a heartbeat for each
 	for _, clusterDeployment := range clusterDeployments.Items {
 		template := templates[clusterDeployment.Spec.Template]
 		// TODO: get k0s cluster ID once it's exposed in k0smotron API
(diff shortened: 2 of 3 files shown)


internal/telemetry/decode_k8s_distribution.go
```diff
@@ -0,0 +1,51 @@
...
+	case strings.Contains(k8sVersion, "rancher"):
+		return "rke", nil
+	case strings.Contains(k8sVersion, "k0s"):
+		return "k0s", nil
+	case strings.Contains(k8sVersion, "vmware"):
+		return "vmware", nil
+	case strings.Contains(k8sVersion, "-CCE"):
+		return "huawei", nil
+	default:
+		return "unknown", nil
```
zerospiel, line 49: is something is missed, then we will never know the actual value, hence i'd stick to checking what is after the `+` sign

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+		"k8s":       k8sProps,
+		"kcm":       kcmProps,
+		"nodes":     nodeMetrics,
+	}
+
+	// Add clusterID, defined as the UUID of the kube-system namespace with a prefix
+	kubeSystemNamespace := &corev1.Namespace{}
+	if err := t.Get(ctx, client.ObjectKey{Name: "kube-system"}, kubeSystemNamespace); err == nil {
+		telemetryProps["clusterID"] = "kube-system:" + string(kubeSystemNamespace.UID)
+	}
```
zerospiel, line 477: that does not make any sense, rewriting the correct value with something unrelatedf. what was the idea behind it?

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+	serviceCount := 0
+	if err := t.List(ctx, services); err == nil {
+		for _, service := range services.Items {
+			// Skip system services
+			if systemNamespaces[service.Namespace] {
+				continue
+			}
+			serviceCount++
+		}
+	}
```
zerospiel, line 423: what is the idea of collecting the services? was it supposed to be user services? it is being done on the per-clusterdeployment basis counting actual user services in use.

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+		gpuNvidiaCapacity := node.Status.Capacity[corev1.ResourceName("nvidia.com/gpu")]
+		gpuNvidiaCapacityShared := node.Status.Capacity[corev1.ResourceName("nvidia.com/gpu.shared")]
+		gpuNvidiaAlloc := node.Status.Allocatable[corev1.ResourceName("nvidia.com/gpu")]
+		gpuNvidiaAllocShared := node.Status.Allocatable[corev1.ResourceName("nvidia.com/gpu.shared")]
+
+		// Labels provide actual values for GPU counts, types, memory, and sharing configuration
+		gpuNvidiaCount := node.Labels["nvidia.com/gpu.count"]
+		gpuNvidiaProduct := node.Labels["nvidia.com/gpu.product"]
+		gpuNvidiaMemory := node.Labels["nvidia.com/gpu.memory"]
+		gpuNvidiaSharing := node.Labels["nvidia.com/gpu.sharing"]
```
zerospiel, line 267: what are the chances somebody puts workloads on a mgmt cluster? what is the purpose of collecting these metrics from a mgmt cluster?

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+	totalPodsAllocGPU := 0
+
+	// Count non-system pods per node
+	nodePodCounts, err := countNonSystemPodsOnNodes(ctx, k8sClient)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Iterate through nodes and build a node-level metrics map
+	for i, node := range nodes.Items {
```
zerospiel, line 236: That will work only on a small number of nodes; otherwise, either OOM or overusage of resources to collect metrics once in a tick. Such a core's objects must be collected and aggregated in batches

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+	// Get all pods in the cluster -- these will be filtered to just those assigned to the node
+	pods := &corev1.PodList{}
+	if err := k8sClient.List(ctx, pods); err != nil {
+		return nil, err
+	}
+
+	nodePodMetrics := make(map[string]map[string]int)
+
+	// Count pods
+	for _, pod := range pods.Items {
```
zerospiel, line 147: ditto about batching and aggregation

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+
+		// Pods requesting GPU -- need to check resources associated with Pod containeres
+		podRequestGPU := false
+		podAllocGPU := false
+		for _, container := range pod.Spec.Containers {
+			if container.Resources.Requests != nil {
+				for resourceName := range container.Resources.Requests {
+					if isGPUResource(resourceName) {
+						podRequestGPU = true
+						break
```
zerospiel, line 186: that breaks the inner loop only

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+				}
+			}
+		}
+
+		if podRequestGPU {
+			nodePodMetrics[pod.Spec.NodeName]["nodePodGpuRequestCount"]++
+		}
+		if podAllocGPU {
+			nodePodMetrics[pod.Spec.NodeName]["nodePodGpuAllocCount"]++
+		}
```
zerospiel, line 205: same question about chances that there are gpu nodes and workloads requiring gpus on a mgmt cluster

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+					if isGPUResource(resourceName) {
+						podRequestGPU = true
+						break
+					}
+				}
+			}
+			if container.Resources.Limits != nil {
+				for resourceName := range container.Resources.Limits {
+					if isGPUResource(resourceName) {
+						podAllocGPU = true
```
zerospiel, line 193: not sure what allocation has to do with `limits`, could you elaborate?

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+		gpuNvidiaSharing := node.Labels["nvidia.com/gpu.sharing"]
+
+		// Node GPU cores and vendor
+		gpuCount := int64(0)
+		gpuType := "none"
+
+		// Scan resources for strings suggesting a GPU
+		for resourceName, quantity := range node.Status.Capacity {
+			if isGPUResource(resourceName) {
+				gpuCount += quantity.Value()
```
zerospiel, line 276: i see the inconsistencies, we count gpu in bytes, memory and storage in GiB. is it expected?

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+				switch {
+				case strings.Contains(string(resourceName), "nvidia"):
+					gpuType = "nvidia"
+				case strings.Contains(string(resourceName), "amd"):
+					gpuType = "amd"
+				default:
+					gpuType = "unspecified"
+				}
+			}
+		}
```
zerospiel, line 288: could be simplified to a simple check of a certain key `amd.com/gpu` or `nvidia.com/

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+
+	return nodeMetrics, clusterMetrics, nil
+}
+
+func (t *Tracker) trackManagementClusterHeartbeat(ctx context.Context) error {
+	// Get the management cluster
+	mgmt := &kcmv1.Management{}
+	if err := t.Get(ctx, client.ObjectKey{Name: kcmv1.ManagementName}, mgmt); err != nil {
+		return err
+	}
```
zerospiel, line 368: duplicates the work of the child heartbeat

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+	pvs := &corev1.PersistentVolumeList{}
+	pvCount := 0
+	if err := t.List(ctx, pvs); err == nil {
+		pvCount = len(pvs.Items)
+	}
+
+	// Get ClusterDeployment count
+	clusterDeployments := &kcmv1.ClusterDeploymentList{}
+	clusterDeploymentCount := 0
+	if err := t.List(ctx, clusterDeployments); err == nil {
```
zerospiel, line 380: duplicates the work of the child heartbeat

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+	providerTemplates := &kcmv1.ProviderTemplateList{}
+	providerTemplateCount := 0
+	if err := t.List(ctx, providerTemplates); err == nil {
+		providerTemplateCount = len(providerTemplates.Items)
+	}
+
+	// Get ClusterTemplate count
+	clusterTemplates := &kcmv1.ClusterTemplateList{}
+	clusterTemplateCount := 0
+	if err := t.List(ctx, clusterTemplates); err == nil {
```
zerospiel, line 401: duplicates the work of the child heartbeat

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+		"podsAllocGPU":     clusterMetrics["podsAllocGPU"],
+	}
+
+	kcmProps := map[string]any{
+		"credentials":        credentialCount,
+		"clusterDeployments": clusterDeploymentCount,
+		"providerTemplates":  providerTemplateCount,
+		"clusterTemplates":   clusterTemplateCount,
+		"serviceTemplates":   serviceTemplateCount,
+		"services":           serviceCount,
```
zerospiel, line 462: what is relation between core's `Service` objects and `kcm`?

internal/telemetry/tracker.go
```diff
@@ -96,8 +119,369 @@ func (t *Tracker) trackClusterDeploymentHeartbeat(ctx context.Context) error {
...
+	versionStr, ok := version.(string)
+	if !ok {
+		return errors.New("kubernetes version is not a string")
+	}
+	dist, err := DecodeDistribution(versionStr)
+	if err != nil {
+		return err
+	}
+
+	k8sProps := map[string]any{
```
zerospiel, line 443: `clusterMetrics` map could be just extended and put as a result, without allocating a new object

zerospiel (changes requested):
I have several concerns regarding metrics collection in terms of compute resources. Along with that, I have doubts regarding certain metrics being collected. So beforehand, I would appreciate to see the whole list of metrics to be collected, because some of them do not make sense to me since they are workloads-related which is unlikely situation on a management cluster

internal/telemetry/decode_k8s_distribution.go
briantd: This function is intended to provide a distribution if recognized, otherwise return unknown. The raw version string is also included in the telemetry payload. Are you recommending that this logic be moved to the reporting layer?

internal/telemetry/tracker.go
briantd: @zerospiel -- What does the "ditto" apply to? Maybe the order I'm viewing the comments is different than the order you left them.

internal/telemetry/tracker.go
briantd: Thanks, that is intended -- however your comment revealed the problem that the 2 aggregation lines should be inside the outer for loop.

The idea was iterate over all the containers. If the container has a resource request spec'd with a GPU-related string, then increment the GPU request count for the POD. I have to iterate through all the resource requests for that container to determine whether that's the case.

Likewise, if the container has a resource limit spec'd with a GPU-related string, then increment the GPU limit count for the POD.

internal/telemetry/tracker.go
briantd: You're right, using the term "allocated" isn't quite right here. I thought that there would be a reason to compare requested GPU cores vs limited GPU cores; however after re-reading the docs I conclued that unless Multi-Instance GPU (MIG) is used the value for requests and limits will NOT be different. I'll remove the alloc part.

If MIG is used, dox read like it's possible to ask for the slice that was actually allocated.

The initial spec was focused on counting PODs that utilized GPUs w/o actually counting the cores used.

internal/telemetry/tracker.go
briantd: From an expected use case perspective, it doesn't seem likely that someone would run a workload requiring a GPU on the management cluster. But if someone did set things up that way, we'd like to know about -- that indicates the product is being used in unexpected and interesting ways.

internal/telemetry/tracker.go
briantd: This is related to my comment here --> https://github.com/k0rdent/kcm/pull/1609/files#r2270900654. Undoubtedly we'd recommend NOT running workloads on a Management Cluster, but there's nothing preventing that from happening right? If people use the product in that way, it'd be enlightening to see.

internal/telemetry/tracker.go
briantd: Is there guidance on the maximum # of nodes to process in one loop? Also, by batches are you suggesting processing say 10 nodes each tick? What's the right way to store the results so the next tick can pick up where the previous left off? Are rare edge-cases like the list of nodes changing going to be a problem?

internal/telemetry/tracker.go
briantd: I can see how this might grab different units if there's GPUs, Slices, and Shared GPUs. I'm not expecting resource types that are determined to be isGPUResource would report Memory or Storage.

If the string matches "nvidia.com/gpu" or "amd.com/gpu" then I should treat it as a core. However if it's one of the other cases like "nvidia.com/mig-" that's going to be a GPU slice, and finally ".shared".

internal/telemetry/tracker.go
briantd: So something like:
```
if strings.Contains(string(resourceName), "nvidia.com/") {
    gpuType = "nvidia"
} else if strings.Contains(string(resourceName), "amd.com/gpu") {
    gpuType = "amd"
 ...
```

internal/telemetry/tracker.go
briantd: We'll want to know the Cluster-level properties for the management cluster -- at least as far as controllers, workers, cores, memory, storage, etc...

Were you thinking that we can create a routine that can be called for both the management cluster heartbeat and the child cluster heartbeats?

internal/telemetry/tracker.go
briantd: The idea of counting user services is that's another measure of cluster utilization.

internal/telemetry/tracker.go
briantd: I'm ok with that -- I thought this would make the assembled payload code easier to read since I'm nesting those values under the "k8s" properties.

internal/telemetry/tracker.go
briantd: Yes, I agree that's misleading. The services Count is supposed to be UserServices running as opposed to services specifically managed by sveltos.

internal/telemetry/tracker.go
briantd: So overwriting it needlessly was an oversight. However, the value used in the overwrite is intentional. There was a convention established for other products that the cluster ID would be based on the kube-system Namespace UID. This is the case for MKE, k0s, and Lens.

The mgmt.UID is unique to k0rdent and won't tie to any other service running on the management cluster. Presumably if someone installs MKE on the Managment Cluster and sends MKE-specific telemetry, that product won't see mgmt.UID and we won't be able to join the telemetry.

zerospiel:
The PR is now too far from the `main` branch; there is no point in rebasing since the whole approach has changed. We will add management telemetry as our next step on the next iteration, and hence integrate it directly into the current approach

PR #1624, opened 2025-06-10
feat: Adapt openstack template to mount CA cert

### LINKED ISSUE #1608: [enh] Adapt openstack standalone template to support mounting secret with CA cert
Currently, when creating a cluster in OpenStack with an OS cloud that uses a self-signed certificate, the deployment fails to work even if the `ca-file` is provided via the `cloud.conf` secret. This is because the Cloud Controller Manager (CCM) and CSI components do not mount the secret, and therefore cannot access the certificate.

Our OpenStack standalone template should be updated to support specifying a secret mount path, allowing the CCM and CSI charts to mount the provided secret as a volume.

eromanova:
Also includes:
* Adapt dev credentials with the CA cert secret
* Support to generate openstack-standalone-cp helm chart schema

Fixes #1608

### DIFF
--- a/templates/cluster/openstack-standalone-cp/values.yaml
+++ b/templates/cluster/openstack-standalone-cp/values.yaml
@@ -1,82 +1,95 @@
-controlPlaneNumber: 3
-workersNumber: 2
+# Schema generation:
+# $ helm plugin install https://github.com/losisin/helm-values-schema-json.git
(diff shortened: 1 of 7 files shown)


templates/cluster/openstack-standalone-cp/values.yaml
```diff
@@ -1,82 +1,95 @@
...
+        name: "default" # @schema description: Name of the security group to filter by; type: string
+        description: "" # @schema description: Description for filtering; type: string
+        projectID: "" # @schema description: Optional: project ID for filtering; type: string
 
-k0s:
-  version: v1.32.5+k0s.1
-  api:
-    extraArgs: {}
+# K0s parameters
+k0s: # @schema description: K0s parameters; type: object; additionalProperties: true
```
zerospiel:
```suggestion
k0s: # @schema description: K0s parameters; type: object
```

zerospiel (dismissed):
LGTM apart from one small piece with regards to the json schema

zerospiel (approved):

Issue #1601, opened 2025-06-05
[enhancement] Add support for 'valuesFiles' in ServiceTemplate 'spec.helm.chartSource' parameters


dis-xcom:
**Is your enhancement / feature request related to a problem? Please describe.**
When working with helm charts for ServiceTemplates, it is possible to specify multiple values files , when helm chart is placed in a HelmRepository, for example:

```
kind: ServiceTemplate
spec:
  helm:
    chartSpec:
...
      sourceRef:
        kind: HelmRepository
        name: ...
      valuesFiles:
      - values.yaml
      - values-common.yaml
...
```

But when helm charts are placed in a git repository, "spec.helm.chartSpec" cannot be used because of issues with GitRepository. The only solution is to use "spec.helm.chartSource".
The problem is that "spec.helm.chartSource" doesn't support "valuesFiles" parameter.

**Describe the solution you'd like**
It would be good to extend "spec.helm.chartSource" in ServiceTemplate, so it could merge valuesFiles for helm charts, like it is done for "spec.helm.chartSpec", for example:

```
apiVersion: k0rdent.mirantis.com/v1beta1
kind: ServiceTemplate
metadata:
  name: foo
  namespace: foo-ns
spec:
  helm:
    chartSource:
      remoteSourceSpec:
        git:
          <GitRepository.spec from fluxcd/source-controller/api/v1>
      path: <path-to-chart-in-repo>

      # Example of the new parameter 'valuesFiles' below:
      valuesFiles:
      - values.yaml
      - values-common.yaml
```

**Describe alternatives you've considered**
Make "spec.helm.chartSpec" to work with GitRepository , if there will be any possibility to make it working in future.

PR #1631, opened 2025-06-11
feat: Support GitRepository for HelmCharts

dis-xcom:
feat: Add support for GitRepository via fluxcd source.

GitRepository is supported in projectsveltos since v0.51.1:
[(feat) HelmCharts: integration with Flux sources](https://github.com/projectsveltos/addon-controller/commit/dad4b69cdef867c66395fe54b40ec786c858c38d)

Tested by deploying ClusterDeployment services, where ServiceTemplates
are based on a private GitRepository.

Fixes #https://github.com/k0rdent/kcm/issues/1601


zerospiel:
the commit-msg does not follow the required convention, so it'd be skipped from the release notes

the corresponding enforcement checks will be implemented soon

internal/sveltos/profile.go
```diff
@@ -200,69 +201,103 @@ func helmChartFromSpecOrRef(
 		return helmChart, fmt.Errorf("failed to get HelmChart %s referenced by ServiceTemplate %s: %w", chartRef.String(), templateRef.String(), err)
 	}
 
-	repo := &sourcev1.HelmRepository{}
+	chartVersion := chart.Spec.Version
+	if chart.Status.Artifact != nil && chart.Status.Artifact.Revision != "" {
+		if _, err := semver.NewVersion(chart.Status.Artifact.Revision); err == nil {
+			// Try to get the HelmChart version from status.artifact.revision
+			// It contains the valid chart version for charts from a GitRepository
+			chartVersion = chart.Status.Artifact.Revision
```
wahabmk, line 209: Why do we need to get the chart's version from status when we can already get it from `chart.Spec.Version` above?

BROngineer: I would say that getting chart version from status is more reliable approach, espesially if we're talking about chart backed by git repo

dis-xcom: In case of GitRepository source, `chart.Spec.Version` cannot be specified, and contains `*` by default, for example: 
```
spec:
  helm:
    chartSpec:
      chart: charts/cilium
      interval: 10m0s
      reconcileStrategy: Revision
      sourceRef:
        kind: GitRepository
        name: box-helm-lib-dis
      valuesFiles:
      - charts/cilium/values.yaml
      - charts/cilium/values-common.yaml
      version: '*'
```

zerospiel:
> the commit-msg does not follow the required convention, so it'd be skipped from the release notes
> 
> the corresponding enforcement checks will be implemented soon

Just to ensure we are all on par: the `commit-msg` has nothing to do with the PR title or description, hence the current commit-msg still violates the guidelines and will be excluded from the Release notes

PR #1639, opened 2025-06-11
ci: introduce goreleaser

### LINKED ISSUE #1603: [ci] Introduce goreleaser to generate releases
Introduce `goreleaser` to:

- streamline the approach to generate artifacts
- automate release generation

zerospiel:
Introduced goreleaser to:

* build ARM64v8/ARMv7/AMD64v1 images
* build manifests
* create formatted releases

Updated WFs with caches.

Goreleaser-featured Release WF decreased running time from ca. 1h to ca. 7m (x10 times).

Overall performance increased ca. x2 times.

Temp removed CI WF from Release WF due to unsuitable order and jobs in the former for the latter WF.

Simplified if-conds in the `push-public` step of the CI WF.

Several minor explicit settings for the Stale WF.

Added new build variables for the sake of readability and reproducible builds (if any).

Removed unused leftovers `mkdocs.yaml` and `internal/projectroot` package.

* [Release job in the fork](https://github.com/zerospiel/hmc/actions/runs/15584690346/job/43887894532)
* [Merge to `main` job in the fork](https://github.com/zerospiel/hmc/actions/runs/15585062466/job/43889110908)
* [Fake E2E test job in the fork](https://github.com/zerospiel/hmc/actions/runs/15584644140) (actual tests have been skipped due to the upstream settings and have nothing to do with the build/push images step)

Attached is the page with a release preview, along with the list of artifacts.

The `metadata.json` has the following contents:

```json
{"project_name":"kcm","tag":"v1.0.1-rc1","previous_tag":"v1.0.0","version":"1.0.1-rc1","commit":"ce3fb5f42dd40003bc7b1ad6977a1f9fce14e2d2","date":"2025-06-11T12:21:29.980191698Z","runtime":{"goos":"linux","goarch":"amd64"}}
```

Closes #1603 

screencapture-github-zerospiel-hmc-releases-2025-06-11-16_02_57


Makefile
```diff
@@ -129,7 +129,7 @@ test: generate-all envtest tidy external-crd ## Run tests.
 # Utilize Kind or modify the e2e tests to load the image locally, enabling
 # compatibility with other vendors.
 .PHONY: test-e2e
-test-e2e: cli-install ## Run the e2e tests using a Kind k8s instance as the management cluster.
+test-e2e: ## Run the e2e tests using a Kind k8s instance as the management cluster.
```
eromanova, line 132: Don't we need the `cli-install` runner before running test-e2e?

.goreleaser.yaml
```diff
@@ -0,0 +1,248 @@
...
+
+  # Upload metadata.json and artifacts.json to the release as well.
+  include_meta: true
+
+dockers:
+  - id: linux-amd64
+    goos: linux
+    goarch: amd64
+    image_templates:
+      - "{{ .Env.REGISTRY }}/{{ .Env.IMAGE_NAME }}:{{ .Env.VERSION }}-amd64"
```
eromanova, line 180: With this change, we build the controller image with an arch suffix (`-amd64`, `-arm64v8` in the tag, right? But in the release job, we still use 
```
run: VERSION="${{ steps.version.outputs.version }}" make kcm-chart-release
```
so, in the kcm helm chart, we'll have the default controller image defined without any arch in tag `ghcr.io/k0rdent/kcm/controller:${{ VERSION }}`. Will this image be present in the artifactory?
Or am I misunderstanding something?

.github/workflows/build_test.yml
```diff
@@ -138,6 +161,7 @@ jobs:
   push-public:
     name: Build and Push Artifacts When Required
     runs-on: ubuntu-latest
+    if: ${{ needs.authorize.outputs.public_repo == 'true' }}
```
eromanova: I'm afraid this won't work. GitHub may not yet have the `needs.authorize.outputs.public_repo` available at the time it decides whether to run the job, even if it depends on the authorize. Thus, this job will never run. That's why it was decided to place this condition at the steps level.

Makefile
zerospiel: It is useless in that context, downloading too much redundant stuff. Each of the dependencies' target has its own dependencies that installs the required bins in return

.goreleaser.yaml
zerospiel: It will, thanks to docker manifests. The umbrella image (the one without architecture postfix) will have 3 images in it: amd64v1, arm64v8, and armv7

Effectively, it is all done within the `docker_manifests` section (there is a bug, though, which I've already fixed in the upstream; however, the new version is not yet published, hence it is now not possible to build images for only one arch)

zerospiel: Ah, that said, each build is virtually 4 images in the registry: tagged per each platform and the umbrella's one for which the manifest is being generated and pushed along with the images

eromanova: I got it. Thanks for the explanation.

.github/workflows/build_test.yml
zerospiel: Reverted, thanks

.goreleaser.yaml
```diff
@@ -0,0 +1,248 @@
...
+# vim: set ts=2 sw=2 tw=0 fo=cnqoj
+
+version: 2
+
+project_name: kcm
+
+before:
+  hooks:
+    - make tidy
+    - sh -c "if [ \"${SKIP_SCM_RELEASE}\" != "true" ]; then make kcm-dist-release && mv ./dist/install.yaml ./install.yaml.out && rmdir ./dist/; fi"
```
eromanova, line 11: Since we've deprecated installation using install.yaml file, I guess we can remove all these hooks, all occurrences of install.yaml, and remove it from artifacts.
But it can be addressed separately.

zerospiel: Better to address it separately, either with a new `goreleaser` version with the bug fixed, or just in another PR, and drop everything related to the file (targets, files, these new lines, etc.)

PR #1650, opened 2025-06-18
style: auto-update all configs

### LINKED ISSUE #1574: [style] Simplify the PR template enforcing checks automatically
Some of the steps that are covered [here](https://github.com/k0rdent/kcm/blob/main/.github/PULL_REQUEST_TEMPLATE) should be moved into CI as separate Github workflows (preflight) that block other long-running workflows and each preflight workflow should give readable error and clear instructions how to fix issues.

Current approach with multiple actions items packed into plain-text list is sub-optimal as it gives requirements that needs to be followed but does not enforce them, leaving enforcement to human reviewer.

zerospiel:
Auto-update all the template configs:

* kcm-templates/release.yaml
* config/dev
* automatically bump templates versions

Enforce commit-msg and PR title checks on CI

Simplify the PR Template

Bump Env Test K8s version to 1.33

Closes #1574


internal/controller/suite_test.go
```diff
@@ -110,7 +110,7 @@ var _ = BeforeSuite(func() {
 		// Note that you must have the required binaries setup under the bin directory to perform
 		// the tests directly. When we run make test it will be setup and used automatically.
 		BinaryAssetsDirectory: filepath.Join("..", "..", "bin", "k8s",
-			fmt.Sprintf("1.32.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
+			fmt.Sprintf("1.33.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
s3rj1k, line 113: os.GetEnv ?

zerospiel: There is the explicit comment that this field is for running out of `make`, hence no hard requirement for the envvar

PR #1653, opened 2025-06-19
fix(cld): flapping delete conditions

### LINKED ISSUE #1547: [bug] Events missing despite controller logs showing they occurred
**Describe the bug**
Sometimes, some of the k0rdent events are not being created, even though the controller logs indicate that the events occurred.

**Steps to reproduce**
1. Deploy k0rdent.
2. Create a ClusterDeployment.
3. Watch for created events and KCM controller logs. The controller claims event creation, but no event appears.

**Expected behavior**
All of the reported events are created.

zerospiel:
Record Delete events on setting DeletionTimestamp or sending a Delete request so the root object still exists at the moment

Fixes #1547

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -932,10 +932,9 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, cd *k
 		if err := r.Client.Update(ctx, cd); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to update clusterDeployment %s: %w", client.ObjectKeyFromObject(cd), err)
 		}
+		r.eventf(cd, "SuccessfulDelete", "ClusterDeployment has been deleted")
 	}
-	r.eventf(cd, "ClusterDeleted", "Cluster %s has been deleted", client.ObjectKeyFromObject(cd))
 
-	r.eventf(cd, "SuccessfulDelete", "ClusterDeployment has been deleted")
 	l.Info("ClusterDeployment deleted")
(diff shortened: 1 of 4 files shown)


internal/controller/clusterdeployment_controller.go
```diff
@@ -900,13 +900,13 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, cd *k
 		}
 
 		l.Info("HelmRelease still exists, retrying")
+		r.eventf(cd, "HelmReleaseDeleted", "HelmRelease %s has been deleted", client.ObjectKeyFromObject(cd))
```
eromanova: Could you please elaborate on why you moved the `HelmReleaseDeleted` condition here? The `cd` (root) object still exists, so the created event should still be processed correctly. When triggered at line 909, the HelmRelease object has already been deleted, while with your change, the event might be created earlier, potentially before the deletion is complete.

zerospiel: You are correct, I was struggling with that too, and decided to align the event with the same "timing" as for the other delete events, at the moment of sending `delete` request (or rather removing a finalizer). I do not mind to move the event back if you insist

eromanova: I was a little bit confused because we log:
```l.Info("HelmRelease still exists, retrying")```
and the next line, we create an event about HelmRelease deletion. If you're ok to move it back, let's do it. Other code LGTM.

zerospiel: Got it back!

Issue #1568, opened 2025-05-27
[bug] `Profile` (almost) always mutates without actual changes


zerospiel:
**Describe the bug**
According to the logs (attached), a `Profile` object (almost) always is being mutated, having no actual changes.

I reckon the same behavior applies to a `ClusterProfile` object.

**To Reproduce**
1. `make dev-apply`
2. Create any `ClusterDeployment`
3. Wait for the full reconciliation (not necessarily until it is ready, but rather until there is nothing to do for the `kcm-controller-manager`)

**Expected behavior**
`Profile` (if applicable `ClusterProfile`) mutates only when actual changes have been applied.

[support-bundle-2025-05-27T18_01_42.tar.gz](https://github.com/user-attachments/files/20462723/support-bundle-2025-05-27T18_01_42.tar.gz)

Issue #1586, opened 2025-05-30
[bug] AWS credentials access issue between namespaces


sk4lf:
k0rdent v1.0.0. I have AWS creds stored in **kcm-system** namespace and child related objects in **default** namespace: helm repos, cluster template, service templates, secrets to services and cluster deployment. AWS creds are propagated to **default** NS via AccessManagement. Cluster deployment contain "propagateCredentials: false" to satisfy Sveltos.

Once I create a clusterdeployment in **default** namespace, **capa** spins the objects in AWS and gives turn to Sveltos and it starts complaining that it doesn't have access to credential secret:

`I0530 15:09:16.223141       1 clustersummary_controller.go:131] "Reconciling" controller="clustersummary" controllerGroup="config.projectsveltos.io" controllerKind="ClusterSummary" ClusterSummary="default/p--sk4lf-ai-test-capi-sk4lf-ai-test" namespace="default" name="p--sk4lf-ai-test-capi-sk4lf-ai-test" reconcileID="18df781a-3455-4e1a-a601-5244f9e6cef1"
I0530 15:09:16.223463       1 clustersummary_controller.go:326] "Reconciling ClusterSummary" controller="clustersummary" controllerGroup="config.projectsveltos.io" controllerKind="ClusterSummary" ClusterSummary="default/p--sk4lf-ai-test-capi-sk4lf-ai-test" namespace="default" name="p--sk4lf-ai-test-capi-sk4lf-ai-test" reconcileID="18df781a-3455-4e1a-a601-5244f9e6cef1"
E0530 15:09:16.352020       1 clustersummary_controller.go:390] "failed to deploy" err=<
	deploying resources failed: secrets "aws-cluster-identity-secret" not found
	deploying helm charts failed: secrets "aws-cluster-identity-secret" not found
 > controller="clustersummary" controllerGroup="config.projectsveltos.io" controllerKind="ClusterSummary" ClusterSummary="default/p--sk4lf-ai-test-capi-sk4lf-ai-test" namespace="default" name="p--sk4lf-ai-test-capi-sk4lf-ai-test" reconcileID="18df781a-3455-4e1a-a601-5244f9e6cef1"`

But once I apply the same clusterdeployment once again, it start seeing the creds and proceed with services deployment. Without this clusterdeployment reapply it will stay erroring.

As far as I can see, Sveltos tries to find creds in child home NS instead of kcm-system.

**To Reproduce**
1. create AWS creds in kcm-system NS
[...]

PR #1670, opened 2025-06-20
feat: ServiceSet type and built-in ProjectSveltos provider

BROngineer:
This PR extends API with new type `ServiceSet`. This type, along with already introduced `StateManagementProvider` allows implementation of provider-agnostic processing for KSM.
Aside from that it allows to implement:
- dependencies management between services defined in single `ClusterDeployment` or `MultiClusterService`
- proper sequential upgrades of installed services.

Some notes regarding removed functions.

`../internal/utils/validation/cd.go`

'ClusterDeployCrossNamespaceServicesRefs' where removed because:
- updated 'ServiceSpec' does not contain svletos-specific types, thus there might be no 'TemplateResourceRefs' field
- updated 'Service' struct contains 'ValuesFrom' field which does not contain 'Namespace', hence no need in cross-namespace reference validation

Everything related to ProjectSveltos was removed from both `clusterdeployment_controller.go` and `multiclusterservice_controller.go`. Aside from that conditions related to SveltosCluster won't be reflected in `ClusterDeployment` status.

**[TBD] Incomplete changes**:

- [x] need to add collecting of statuses of deployed services
- [x] need to re-enable self-management for mothership cluster
- [x] need to add default `StateManagementProvider` creation if flag `enableSveltosCtrl` is set to true (or add another flag to enable default sveltos provider)
- [x] need to add tests for new controllers
- [x] need to align existing tests with changes

**Which issue(s) this PR fixes**
Fixes: #1562

### DIFF
--- a/templates/provider/kcm-templates/files/release.yaml
+++ b/templates/provider/kcm-templates/files/release.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   version: 1.3.0
   kcm:
-    template: kcm-1-3-0
+    template: kcm-1-3-1
   capi:
     template: cluster-api-1-0-5
   providers:
--- a/templates/provider/kcm-templates/files/templates/kcm.yaml
+++ b/templates/provider/kcm-templates/files/templates/kcm.yaml
@@ -1,14 +1,14 @@
 apiVersion: k0rdent.mirantis.com/v1beta1
 kind: ProviderTemplate
 metadata:
-  name: kcm-1-3-0
+  name: kcm-1-3-1
   annotations:
     helm.sh/resource-policy: keep
 spec:
   helm:
     chartSpec:
       chart: kcm
-      version: 1.3.0
+      version: 1.3.1
       interval: 10m0s
       sourceRef:
         kind: HelmRepository
--- a/templates/provider/kcm/Chart.yaml
+++ b/templates/provider/kcm/Chart.yaml
@@ -13,7 +13,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.3.0
+version: 1.3.1
 dependencies:
   - name: flux2
     version: 2.16.0


api/v1beta1/serviceset_types.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Namespace is the namespace where the service is deployed. If the ServiceTemplate
+	// is backed by Helm chart, then the namespace is the namespace where the Helm release is deployed.
+	Namespace string `json:"namespace"`
+
+	// Template is the name of the ServiceTemplate to use to deploy the service.
+	Template string `json:"template"`
+
+	// Values is the values to pass to the ServiceTemplate.
+	Values string `json:"values,omitempty"`
```
zerospiel, line 139: is that the suitable type for the values to be presented? how would it look if one is trying to declare something comprehensive, e.g.:

```
values:
  foo: "string"
  bar: false
  baz:
   nested:
     float: 2.0
     int: 1
     str: strval
```

upd. it seems the question is invalid, because it has been in use in such a manner for ages, but i'd appreciate how it works

api/v1beta1/serviceset_types.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Provider is the definition of the provider to use to deploy services defined in the ServiceSet.
+	Provider ProviderSpec `json:"provider"`
+
+	// Services is the list of services to deploy.
+	Services []ServiceWithValues `json:"services,omitempty"`
+}
+
+// ProviderSpec contains all the spec related to the state management provider.
+type ProviderSpec struct {
```
zerospiel: probably nit-picking: won't it be confused (at least from a dev standpoint but we also should consider an end-user) to have a new type named `ProviderSpec`, already having `Providers []string`, `Provider struct{}`? The question arose because I noticed the change in logs of the mgmt ctrl from `Provider` to `ProviderSpec`

That being said, shouldn't we rename the type to something like `StateManagementProviderSpec`?

cmd/main.go
```diff
@@ -390,12 +388,25 @@ func main() {
...
+		utilruntime.Must(addoncontrollerv1beta1.AddToScheme(scheme))
+		utilruntime.Must(libsveltosv1beta1.AddToScheme(scheme))
+
 		if err = (&sveltos.ClusterReconciler{
 			Client: mgr.GetClient(),
 		}).SetupWithManager(mgr); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "SveltosCluster")
 			os.Exit(1)
 		}
+		if err = (&sveltos.ServiceSetReconciler{
```
zerospiel, line 484: Isn't it confusing, having `enable-stuck-sveltos-ctrl` knob controlling the whole sveltos provider integration? Shouldn't we introduce a new object, something like:

```yaml
# kcm_values.yaml

# the root of the values
sveltos: # sveltos_provider? sveltos_integration?
  enabled: true
  update-stuck-tokens: false
```

test/objects/clusterdeployment/clusterdeployment.go
```diff
@@ -89,11 +88,12 @@ func WithServiceSpec(serviceSpec kcmv1.ServiceSpec) Opt {
 	}
 }
 
-func WithTemplateResourceRefs(templRefs []addoncontrollerv1beta1.TemplateResourceRef) Opt {
-	return func(p *kcmv1.ClusterDeployment) {
-		p.Spec.ServiceSpec.TemplateResourceRefs = append(p.Spec.ServiceSpec.TemplateResourceRefs, templRefs...)
-	}
-}
+// fixme: this is not used
```
zerospiel: I reckon we could just remove it safely

templates/provider/kcm/values.yaml
```diff
@@ -27,7 +27,7 @@ controller:
   affinity: {} # @schema type: object; description: Affinity rules for pod scheduling
   tolerations: [] # @schema type: array; description: Tolerations to allow the pod to schedule on tainted nodes
   validateClusterUpgradePath: true # @schema type: boolean; description: Specifies whether the ClusterDeployment upgrade path should be validated
-  enableSveltosExpiredCtrl: false # @schema type: boolean; description: Enables SveltosCluster controller, updating stuck (expired) sveltos management cluster kubeconfig tokens
+  enableSveltosExpiredCtrl: true # @schema type: boolean; description: Enables SveltosCluster controller, updating stuck (expired) sveltos management cluster kubeconfig tokens
```
zerospiel: ditto as for the `cmd/main.go`. And I would consider refraining from changing the default value, since it is not bw compatible in terms of expectations

internal/serviceset/builder.go
```diff
@@ -0,0 +1,150 @@
...
+		// if the ServiceSet also has labels that don't match the selector, then this ServiceSet
+		// won't be reconciled
+		if b.ServiceSet.Labels == nil {
+			return labels
+		}
+		for k, v := range labels {
+			b.ServiceSet.Labels[k] = v
+		}
+		return b.ServiceSet.Labels
+	}()
```
zerospiel:
```suggestion
		// we'll overwrite ServiceSet labels with those matching the StateManagementProvider selector
		// if the ServiceSet also has labels that don't match the selector, then this ServiceSet
		// won't be reconciled
	if b.ServiceSet.Labels == nil {
		b.ServiceSet.Labels = labels
	} else {
		maps.Copy(b.ServiceSet.Labels, labels)
	}
```

internal/serviceset/builder.go
```diff
@@ -0,0 +1,150 @@
...
+func extractRequiredLabels(selector *metav1.LabelSelector) (map[string]string, error) {
+	if selector == nil {
+		return nil, errors.New("selector cannot be nil")
+	}
+
+	result := make(map[string]string)
+
+	for k, v := range selector.MatchLabels {
+		result[k] = v
+	}
```
zerospiel:
```suggestion
	maps.Copy(result, selector.MatchLabels)
```

api/v1beta1/serviceset_types.go
```diff
@@ -0,0 +1,235 @@
...
+	// Name is the name of the service. If the ServiceTemplate is backed by Helm chart,
+	// then the name is the name of the Helm release.
+	Name string `json:"name"`
+
+	// Namespace is the namespace where the service is deployed. If the ServiceTemplate
+	// is backed by Helm chart, then the namespace is the namespace where the Helm release is deployed.
+	Namespace string `json:"namespace"`
+
+	// Template is the name of the ServiceTemplate to use to deploy the service.
+	Template string `json:"template"`
```
zerospiel, line 136: These fields look like a new `ServiceTemplateReference` type (IMO, it might be more convenient for a reader/user, like in this [example](https://github.com/kubernetes/api/blob/release-1.17/admissionregistration/v1/types.go#L533))

Valid for both the `spec` and the `status`

internal/controller/management_controller.go
```diff
@@ -273,7 +273,7 @@ func (r *ManagementReconciler) reconcileManagementComponents(ctx context.Context
 		}
 
 		if err := r.checkProviderStatus(ctx, component); err != nil {
-			l.Info("Provider is not yet ready", "template", component.Template, "err", err)
+			l.Info("ProviderSpec is not yet ready", "template", component.Template, "err", err)
```
zerospiel: Is it indeed the new type? The question is applicable to the rest of the changed lines

internal/controller/statemanagementprovider/statemanagementprovider_controller.go
```diff
@@ -97,17 +99,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
 		smp.Status.Ready = !slices.ContainsFunc(smp.Status.Conditions, func(c metav1.Condition) bool {
 			return c.Status == metav1.ConditionFalse || c.Status == metav1.ConditionUnknown
 		})
-		err = errors.Join(err, r.Status().Update(ctx, smp))
+		updated, updateErr := r.statusUpdate(ctx, smpCopy, smp)
+		err = errors.Join(err, updateErr)
 		l.Info("StateManagementProvider reconciled", "duration", time.Since(start))
+		if !updated && err == nil {
+			result.RequeueAfter = requeueInterval
+		}
```
zerospiel, line 127: why do we always need to put the same event in the queue if the object has been reconciled (considered as finished)?

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -0,0 +1,902 @@
...
+	if !serviceSet.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(ctx, serviceSet)
+	}
+
+	if !controllerutil.ContainsFinalizer(serviceSet, kcmv1.ServiceSetFinalizer) {
+		controllerutil.AddFinalizer(serviceSet, kcmv1.ServiceSetFinalizer)
+		if err = r.Update(ctx, serviceSet); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: r.requeueInterval}, nil
```
zerospiel: why do we need to wait an extra 10 seconds after setting the finalizer?

internal/serviceset/processor.go
```diff
@@ -0,0 +1,87 @@
...
+		// on next reconciliation we'll attempt to update it if needed
+		if apierrors.IsAlreadyExists(err) {
+			return nil
+		}
+		l.V(1).Info("Successfully created ServiceSet", "namespaced_name", serviceSetObjectKey)
+		return nil
+	case kcmv1.ServiceSetOperationUpdate:
+		l.V(1).Info("updating ServiceSet", "namespaced_name", serviceSetObjectKey)
+		err := p.Update(ctx, serviceSet)
+		// we'll requeue if ServiceSet update fails due to a conflict
```
zerospiel, line 75: i couldn't find any indication of requeue. Could you elaborate on this?

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -0,0 +1,902 @@
...
+				for _, serviceSet := range serviceSets.Items {
+					requests = append(requests, ctrl.Request{
+						NamespacedName: types.NamespacedName{
+							Name:      serviceSet.Name,
+							Namespace: serviceSet.Namespace,
+						},
+					})
+				}
+				return requests
+			}()
```
zerospiel:
```suggestion
			requests := make([]ctrl.Request, 0, len(serviceSets.Items))
			for _, serviceSet := range serviceSets.Items {
				requests = append(requests, ctrl.Request{
					NamespacedName: types.NamespacedName{
						Name:      serviceSet.Name,
						Namespace: serviceSet.Namespace,
					},
				})
			}
			return requests
```

I do not understand why do we need the closure here

internal/controller/clusterdeployment_controller.go
```diff
@@ -1263,6 +987,22 @@ func (r *ClusterDeploymentReconciler) handleCertificateSecrets(ctx context.Conte
...
 }
 
+func (r *ClusterDeploymentReconciler) collectServicesStatuses(ctx context.Context, cd *kcmv1.ClusterDeployment) ([]kcmv1.ServiceState, error) {
+	selector := fields.OneTermEqualSelector(kcmv1.ServiceSetClusterIndexKey, cd.Name)
+	aggregatedServiceStatuses := make([]kcmv1.ServiceState, 0)
+	serviceSets := new(kcmv1.ServiceSetList)
+	if err := r.Client.List(ctx, serviceSets, client.InNamespace(cd.Namespace), client.MatchingFieldsSelector{Selector: selector}); err != nil {
+		return aggregatedServiceStatuses, fmt.Errorf("failed to list ServiceSets: %w", err)
+	}
+
```
```suggestion
	serviceSets := new(kcmv1.ServiceSetList)
	if err := r.Client.List(ctx, serviceSets, client.InNamespace(cd.Namespace), client.MatchingFields{kcmv1.ServiceSetClusterIndexKey: cd.Name}); err != nil {
		return nil, fmt.Errorf("failed to list ServiceSets: %w", err)
	}

	aggregatedServiceStatuses := make([]kcmv1.ServiceState, 0, len(serviceSets.Items)) // not ideal but at least some pre-alloc
```

internal/controller/clusterdeployment_controller.go
```diff
@@ -1380,6 +1166,9 @@ func (r *ClusterDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 				return req
 			}),
+		).
+		Watches(&kcmv1.ServiceSet{},
+			handler.EnqueueRequestForOwner(mgr.GetScheme(), mgr.GetRESTMapper(), &kcmv1.ClusterDeployment{}, handler.OnlyControllerOwner()),
```
```suggestion
		).
                Owns(&kcmv1.ServiceSet{})
```

aren't these equal?

internal/controller/multiclusterservice_controller.go
```diff
@@ -131,168 +97,59 @@ func (r *MultiClusterServiceReconciler) reconcileUpdate(ctx context.Context, mcs
 		// Requeuing to make sure that ClusterProfile is reconciled in subsequent runs.
 		// Without the requeue, we would be depending on an external re-trigger after
 		// the 1st run for the ClusterProfile object to be reconciled.
-		return ctrl.Result{Requeue: true}, nil
+		return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, nil
```
zerospiel, line 99: why do we need to wait extra seconds after setting the finalizer?

internal/controller/multiclusterservice_controller.go
```diff
@@ -131,168 +97,59 @@ func (r *MultiClusterServiceReconciler) reconcileUpdate(ctx context.Context, mcs
...
 	}
 
 	if updated, err := utils.AddKCMComponentLabel(ctx, r.Client, mcs); updated || err != nil {
 		if err != nil {
 			l.Error(err, "adding component label")
 		}
-		return ctrl.Result{Requeue: true}, err // generation has not changed, need explicit requeue
+		return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, err // generation has not changed, need explicit requeue
```
zerospiel, line 106: ditto about waiting

internal/controller/clusterdeployment_controller.go
zerospiel: and a follow-up question: shouldn't we narrow the number of events to trigger on (e.g. drop `create` / `generic`?

I am afraid of a situation in which we could enter a storm of reconciliation loops: cd -> ss -> cd; mcs -> ss -> cd -> ss; and etc.

i am trying to understand the following: `cld` basically owner (with ctrl ref!) of all potenial `servicesets`, right? `mcs` reconciler indirectly produces `servicesets` (via the `cld` selected by the `mcs`'s `selector`), then it triggers `cld` controller since it is the owner of an incoming `serviceset`, which produces potentially a lot of redundant operations (some IO, mostly net), just to understand whether we should/could update services status (which obviosly won't be the last loop in case of the `create` event).

that said, i am not sure if i have the full context. i am asking to think about all of the possible scenarios in which events might be produced and determine (if any) cycle loops (which we do want to avoid!)

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -0,0 +1,902 @@
...
+
+	return []addoncontrollerv1beta1.PolicyRef{
+		{
+			Kind:           "ConfigMap",
+			Namespace:      cred.Spec.IdentityRef.Namespace,
+			Name:           cred.Spec.IdentityRef.Name + "-resource-template",
+			DeploymentType: addoncontrollerv1beta1.DeploymentTypeRemote,
+		},
+	}
+}
```
zerospiel: not related to the code: could you mark both these functions similar to `TODO: refactor, see: #1586`, because of the #1586 itself. we have to refactor/reconsider the approach, avoiding hardcode and pushing users to create obscure empty `configmaps`, which confuses users

i am asking because of the code rearrangement, just not to forget it

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -0,0 +1,902 @@
...
+			DeploymentType: addoncontrollerv1beta1.DeploymentType(tmpl.Spec.Resources.DeploymentType),
+		}
+
+		policyRefs = append(policyRefs, policyRef)
+	}
+	return policyRefs, nil
+}
+
+// serviceTemplateObjectFromService returns the [github.com/K0rdent/kcm/api/v1beta1.ServiceTemplate]
+// object found either by direct reference or in [github.com/K0rdent/kcm/api/v1beta1.ServiceTemplateChain] by defined version.
```
zerospiel, line 895: the `servicetemplatechain` part is not yet implemented, is that right?

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -0,0 +1,902 @@
...
+		}
+		if _, ok := deployedServicesMap[key]; ok {
+			continue
+		}
+		serviceSet.Status.Services = append(serviceSet.Status.Services, kcmv1.ServiceState{
+			Name:                    service.Name,
+			Namespace:               service.Namespace,
+			Template:                service.Template,
+			State:                   kcmv1.ServiceStateNotDeployed,
+			LastStateTransitionTime: ptr.To(metav1.NewTime(now())),
```
zerospiel: nit: we have `internal/utils/pointer` package with the close functionality, i encourage to use it (to avoid this `k8s.io/utils/ptr` small dependency, which i removed twice at this point)

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -0,0 +1,902 @@
...
+	// if IgnoreNotFound returns non-nil error, this means that
+	// an error occurred while trying to request kube-apiserver
+	if client.IgnoreNotFound(err) != nil {
+		return fmt.Errorf("failed to get Profile: %w", err)
+	}
+
+	switch {
+	// we already excluded all errors except NotFound
+	// hence if the error is not nil, it means that the object was not found
+	case err != nil:
```
zerospiel, line 409: this `switch case` resolves #1568, doesn't it?

could you also elaborate, whether the `equality` addresses an issue with different pointers (deeply nested) but if the actual data after dereference is equal?

eg https://play.golang.org/p/g-xI-Tp4Vdy

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -0,0 +1,902 @@
...
+
+	if !controllerutil.ContainsFinalizer(serviceSet, kcmv1.ServiceSetFinalizer) {
+		controllerutil.AddFinalizer(serviceSet, kcmv1.ServiceSetFinalizer)
+		if err = r.Update(ctx, serviceSet); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: r.requeueInterval}, nil
+	}
+
+	smp := new(kcmv1.StateManagementProvider)
```
zerospiel, line 131: could you please elaborate (remind) on this one: so when and what creates a `smp` in this context? basically, what is the WF? with the `servicesets` it's clear to me, that it should not be created by an operator/human because it's created virtually by the `cld` ctrl

api/v1beta1/serviceset_types.go
```diff
@@ -0,0 +1,235 @@
...
+	// ServiceStateFailed is the state when the Service is failed
+	ServiceStateFailed = "Failed"
+	// ServiceStateNotDeployed is the state when the Service is not deployed
+	ServiceStateNotDeployed = "Pending"
+	// ServiceStateProvisioning is the state when the Service is being provisioned
+	ServiceStateProvisioning = "Provisioning"
+	// ServiceStateDeleting is the state when the Service is being deleted
+	ServiceStateDeleting = "Deleting"
+
+	DriftIgnorePatch = `- op: add
```
zerospiel: `SveltosDriftIgnorePatch`?

zerospiel (commented):
I left a couple of nitpicky comments, several generic questions (cycle loops, workflows, etc), a couple of suggestions. In general LGTM.

There are several envtest tests failing (without details):

```text
Summarizing 3 Failures:
  [FAIL] ClusterDeployment Controller When reconciling a resource [It] should reconcile ClusterDeployment in dry-run mode
  ./internal/controller/clusterdeployment_controller_test.go:352
  [FAIL] ClusterDeployment Controller When reconciling a resource [It] should reconcile ClusterDeployment with AWS credentials
  ./internal/controller/clusterdeployment_controller_test.go:476
  [FAIL] MultiClusterService Controller When reconciling a resource [It] should successfully reconcile the resource
  ./internal/controller/multiclusterservice_controller_test.go:274
```

And some unit tests for the admission wh

cmd/main.go
BROngineer: I thought about it. I concluded, that it makes no sense to enable stuck-sveltos-ctrl in case sveltos is not enabled. On the other hand it might make no sense to use sveltos, but do not use sveltos token renewal controller. Therefore, I think all sveltos related stuff could be potentially enabled by single flag.
In general, I still have no strong opinion on this topic and I was going to discuss it with the team, so for now I just kept the single flag.

api/v1beta1/serviceset_types.go
BROngineer: Totally makes sense. However, this field mirrors the same field in `cld/mcs.spec.serviceSpec.services[]`, so we'll need to either change the `cld/mcs` API or keep it as-is (I'd vote for changing API with conversion).
AFAIU, `values` field in `cld/mcs` was defined as a string simply because corresponding field in sveltos [cluster]profile is also defined as a string.

api/v1beta1/serviceset_types.go
BROngineer: Makes sense, sure.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: correct

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: 😄 sure.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: Yes, it should solve that issue. Regarding deeply nested different pointers - equality dereferences and compares values rather than pointers. However I'll include this check to test cases.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: If we'll keep sveltos bundled (I assume we will), then `StateManagementProvider` object should be created by management or release controller (I didn't check yet, what exact controller should do that). In general, the logic is the following: if we have sveltos enabled, then we need to install sveltos itself and create corresponding SMP object.

internal/controller/management_controller.go
BROngineer: It was renamed by mistake. Will be reverted.

internal/controller/statemanagementprovider/statemanagementprovider_controller.go
BROngineer: After successful reconciliation referred controllers might become broken. In this case reconciliation of the serviceSets and produced provider-specific objects will stuck and there will be no easy way to understand why the objects are created but not reconciled.
Therefore it will make sense to continuously requeue SMP object to keep tracking referred objects' state.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: Nice catch. I have no answer 😄 will fix.

UPD: IIRC controller-runtime will anyways requeue serviceset on metadata update, hence excplicit requeue is redundant here.

internal/controller/clusterdeployment_controller.go
BROngineer: Agree. I'll optimize it to reduce the number of reconciliations. Nice catch! Thanks

internal/controller/multiclusterservice_controller.go
BROngineer: `Requeue` field is deprecated. Hence changed it to requeueInterval. Maybe will be better to just set 1 millisecond

internal/controller/multiclusterservice_controller.go
BROngineer: `Requeue` field is deprecated. Hence changed it to requeueInterval. Maybe will be better to just set 1 millisecond

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: Will do, sure

internal/utils/validation/cd.go
```diff
@@ -16,53 +16,15 @@ package validation
...
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
 )
 
-// ClusterDeployCrossNamespaceServicesRefs validates that the service and templates references of the given [github.com/K0rdent/kcm/api/v1beta1.ClusterDeployment]
-// reference all objects only in the obj's namespace.
-func ClusterDeployCrossNamespaceServicesRefs(ctx context.Context, cd *kcmv1.ClusterDeployment) (errs error) {
```
wahabmk, line 32: Why remove this validation? The ClusterDeployment is namespace scoped so should refer to only ServiceTemplates within its namespace right? 

The AccessManagement object is used to propagate Cluster/ServiceTemplates across namespaces [as mentioned](https://docs.k0rdent.io/v1.1.1/admin/?h=accessmanagement#the-process:~:text=kcm%2Dcontroller%2Dmanager%20generates%20an%20empty%20AccessManagement%20object) in the k0rdent docs. The removal of this validation would break this model, because now ServiceTemplates from another namespace can also be used with ClusterDeployment.

cmd/main.go
BROngineer: Added separate flag.

api/v1beta1/serviceset_types.go
BROngineer: It's not a reference to service template. It's name and namespace of the application to be deployed (if it's helm, then it will be release name and namespace) and the template name

DinaBelova:
this PR was not landed by Jul 25th. I believe that to make sure we can cut the 1.2.0 release by EOM (Jul 31st) and not introduce possible regressions on the currently tested rc, we should not block the 1.2.0 on this PR landing.

api/v1beta1/multiclusterservice_types.go
```diff
@@ -99,53 +97,18 @@ type Service struct {
...
 	Disable bool `json:"disable,omitempty"`
 }
 
 // ServiceSpec contains all the spec related to deployment of services.
 type ServiceSpec struct {
-	// +kubebuilder:default:=Continuous
-	// +kubebuilder:validation:Enum:=OneTime;Continuous;ContinuousWithDriftDetection;DryRun
-
-	// SyncMode specifies how services are synced in the target cluster.
-	SyncMode string `json:"syncMode,omitempty"`
```
eromanova, line 113: Just curious: removing these fields from the API would introduce a breaking change. What happens if an existing object already has these fields set and k0rdent is upgraded? Would the user be required to take additional manual steps to update the object after the upgrade? If so, it should be properly documented.

cmd/main.go
```diff
@@ -315,112 +394,98 @@ func main() {
...
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up health check")
-		os.Exit(1)
-	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
-		setupLog.Error(err, "unable to set up ready check")
-		os.Exit(1)
-	}
+		currentPodName := os.Getenv("POD_NAME")
```
eromanova: This code relies on the `POD_NAME` environment variable, but it isn’t currently defined in the Deployment of the KCM helm chart. As far as I know, this variable is not exposed automatically and must be configured explicitly.

api/v1beta1/serviceset_types.go
wahabmk: > AFAIU, values field in cld/mcs was defined as a string simply because corresponding field in sveltos [cluster]profile is also defined as a string.

Correct. Sveltos can also use templates in the string so we decided to keep it a string as well.

api/v1beta1/multiclusterservice_types.go
BROngineer: Yep, should be in documentation for sure. The thing is that at the moment we expose sveltos-related fields in our own types and I want to get rid of this. Maybe it would be even better just to mark such fields as deprecated.

cmd/main.go
BROngineer: Nice catch! Thanks! I was sure I've added it to the chart.

api/v1beta1/multiclusterservice_types.go
wahabmk: So `SyncMode` and other Sveltos related fields will now be in `.spec.serviceSpec.provider.config`?

api/v1beta1/multiclusterservice_types.go
```diff
@@ -174,7 +137,7 @@ type ServiceStatus struct {
 // MultiClusterServiceStatus defines the observed state of MultiClusterService.
 type MultiClusterServiceStatus struct {
 	// Services contains details for the state of services.
-	Services []ServiceStatus `json:"services,omitempty"`
+	Services []ServiceState `json:"services,omitempty"`
```
wahabmk, line 201: Just curious, is there a standard naming convention due to which you changed from `*Status` to `*State`?

api/v1beta1/serviceset_types.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// Version is the version of the Service
+	Version string `json:"version"`
+
+	// State is the state of the Service
+	// +kubebuilder:validation:Enum=Deployed;Provisioning;Failed;Pending;Deleting
+	State string `json:"state"`
+
+	// FailureMessage is the reason why the Service failed to deploy
+	FailureMessage string `json:"failureMessage,omitempty"`
```
wahabmk, line 215: Would this suggestion be more flexible? This way this field would not just be tied to error messages.

```suggestion
	// Message is additional information or error regarding the service's state.
	Message string `json:"message,omitempty"`
```

api/v1beta1/multiclusterservice_types.go
BROngineer: Correct. The intention is to keep provider-specific fields there and pass config to provider types in the serviceset reconciler when producing provider's objects. Thus, cld/mcs won't depend on provider types.

api/v1beta1/serviceset_types.go
```diff
@@ -0,0 +1,244 @@
...
+	Provider ProviderState `json:"provider,omitempty"`
+}
+
+// ProviderState is the state of the provider
+type ProviderState struct {
+	// Ready is true if the provider is ready
+	Ready bool `json:"ready,omitempty"`
+
+	// Suspended is true if the provider is suspended
+	Suspended bool `json:"suspended,omitempty"`
```
wahabmk, line 185: Why do we need the extra suspended field? If `suspended=true` then does that not automatically imply `ready=false`?

internal/serviceset/util.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+}
+
+// needsUpdate checks if the ServiceSet needs to be updated based on the ClusterDeployment spec.
+// It first compares the ServiceSet's provider configuration with the ClusterDeployment's service provider configuration.
+// Then it compares the ServiceSet's observed services' state with its desired state, and after that it compares
+// the ServiceSet's observed services' state with ClusterDeployment's desired services state.
+func needsUpdate(serviceSet *kcmv1.ServiceSet, providerSpec kcmv1.StateManagementProviderConfig, services []kcmv1.Service) bool {
+	// we'll need to update provider configuration if it was changed.
+	if !equality.Semantic.DeepEqual(providerSpec, serviceSet.Spec.Provider) {
```
wahabmk, line 241: I have a few questions regarding this flow, which will also help me understand it better.
 
- Let's say we have a ClusterDeployment with `.spec.serviceSpec.provider.name=projectsveltos`. So we have ClusterDeployment -> ServiceSet -> Profile objects already created.
- Now let's say the `.spec.serviceSpec.provider.name` for the ClusterDeployment changes to `flux`.

What would happen in this scenario?

1. Would the ServiceSet's `.spec.provider.name` get updated to `flux` as a result? 
2. If yes then what happens to the sveltos Profile object already created?
3. Should we even allow a change in the provider in an already existing ClusterDeployment/MultiClusterService? If we do then I think we would need to consider:
    - 3a. Removing the services deployed by the previous provider so that the new provider can deploy them without conflicts. In this case, there would be a downtime.
    - 3b. OR make it so that the new provider somehow adopts the services from previous provider. Which could be extremely hard.

So maybe the `.spec.ServiceSpec.provider.name` field should be immutable?

wahabmk (commented):
Still reviewing but wanted to leave my comments on the portions reviewed so far.

internal/serviceset/util.go
```diff
@@ -0,0 +1,296 @@
...
+		}] = svc.TemplateChain
+	}
+	for _, svc := range deployedServices {
+		chain, ok := chainMap[client.ObjectKey{
+			Namespace: svc.Namespace,
+			Name:      svc.Name,
+		}]
+		if !ok {
+			chain = ""
+		}
```
wahabmk: `chain` should automatically be empty if `!ok`:
```suggestion
chain := chainMap[client.ObjectKey{
			Namespace: svc.Namespace,
			Name:      svc.Name,
		}]
```

internal/serviceset/util.go
```diff
@@ -0,0 +1,296 @@
...
+	c client.Client,
+	services []kcmv1.Service,
+	namespace string,
+) ([]kcmv1.ServiceUpgradePaths, error) {
+	var errs error
+	servicesUpgradePaths := make([]kcmv1.ServiceUpgradePaths, 0, len(services))
+	for _, svc := range services {
+		serviceNamespace := svc.Namespace
+		if serviceNamespace == "" {
+			serviceNamespace = metav1.NamespaceDefault
```
wahabmk, line 69: Let's change the following comment as now the service will be installed in `default` namespace:

https://github.com/k0rdent/kcm/blob/59695cd0438640332214c1c55293d690c2facc96/api/v1beta1/multiclusterservice_types.go#L98-L100

internal/serviceset/builder.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, req := range requirements {
+		switch req.Operator() {
+		case selection.Equals, selection.DoubleEquals:
+			values := req.Values()
+			if values.Len() == 1 {
+				result[req.Key()] = values.List()[0]
+			}
+		case selection.In:
+			// for 'In' with single value, we can extract it, for multiple values
+			// we'll set the first one
```
wahabmk, line 145: Could you kindly explain the reasoning behind this? Will help me better understand this function.

internal/serviceset/builder.go
```diff
@@ -0,0 +1,140 @@
...
+		maps.Copy(b.ServiceSet.Labels, labels)
+	}
+
+	b.ServiceSet.Spec = kcmv1.ServiceSetSpec{
+		Cluster:  b.ClusterDeployment.Name,
+		Provider: b.ClusterDeployment.Spec.ServiceSpec.Provider,
+		Services: b.ServicesToDeploy,
+	}
+	if b.MultiClusterService != nil {
+		b.ServiceSet.Spec.Provider = b.MultiClusterService.Spec.ServiceSpec.Provider
```
wahabmk: I have similar questions about different providers here as well:

1. Let's say ClusterDeployment `cdA` has `.spec.serviceSpec.provider.name=projectsveltos` and creates a ServiceSet `cdA`.
2. Now a MultiClusterServices `mcsA` with `.spec.serviceSpec.provider.name=flux` matches this ClusterDeployment and creates a new ServiceSet `cdA-ac9d`.
3. In this scenario, we will now have 2 providers deploying services onto the same cluster.
4. So how will conflicts be resolved?
5. Also we won't be able to handle conflicts with priority as sveltos will not have any idea of what priority mechanism is implemented by flux.

Maybe we should restrict MultiClusterService matching to ClusterDeployments that have the same provider?

Now that we have ServiceSet, maybe we should also consider implementing priority on the ServiceSet itself? So that we can have control over which ServiceSet gets to deploy services instead of offloading it to the underlying provider. This could be done later on as an enhancement though.

internal/serviceset/builder.go
```diff
@@ -0,0 +1,140 @@
...
+// WithServicesToDeploy sets the list of services to deploy.
+func (b *Builder) WithServicesToDeploy(servicesToDeploy []kcmv1.ServiceWithValues) *Builder {
+	b.ServicesToDeploy = servicesToDeploy
+	return b
+}
+
+// Build constructs and returns a ServiceSet object based on the builder's parameters or returns an error if invalid.
+func (b *Builder) Build() (*kcmv1.ServiceSet, error) {
+	ownerReference := metav1.NewControllerRef(b.ClusterDeployment, kcmv1.GroupVersion.WithKind(kcmv1.ClusterDeploymentKind))
+	b.ServiceSet.OwnerReferences = []metav1.OwnerReference{*ownerReference}
```
wahabmk, line 83: If ServiceSet can only have ClusterDeployment as its owner then is the following scenario true?

1. ClusterDeployment `cdA` creates and owns ServiceSet `cdA`.
2. MultiClusterService `mcsA` matches ClusterDeployment `cdA` and thus creates another ServiceSet `cdA-ac9d`.
3. The ServiceSet `cdA-ac9d` is owned by ClusterDeployment `cdA`.
4. Now user deletes MultiClusterService `mcsA` and as a result the ServiceSet `cdA-ac9d` is NOT deleted because it is not owned by `mcsA`.

If yes then we will have orphaned ServiceSets once the MultiClusterService due to which they were created is deleted.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -0,0 +1,1135 @@
...
+	if err = r.Get(ctx, client.ObjectKey{Name: serviceSet.Spec.Provider.Name}, smp); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	continueReconciliation, err := labelsMatchSelector(serviceSet.Labels, smp.Spec.Selector)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to check ServiceSet labels: %w", err)
+	}
+	if !continueReconciliation {
+		l.V(1).Info("ServiceSet labels do not match provider selector, skipping")
```
wahabmk, line 141: For my own understanding, could you kindly share a scenario where this could happen?

cmd/main.go
BROngineer: That was stale thing from initial implementation. Removed

api/v1beta1/serviceset_types.go
BROngineer: I do not think so. Provider might be ready when all dependencies are ok, but we want to suspend it for a while for any reason. That does not mean, that it becomes not ready

internal/serviceset/util.go
BROngineer: Made `.provider.name` immutable

internal/serviceset/builder.go
BROngineer: Yeah. The intention here is quite simple. Lets say there is multiple values, something like:

```
    matchExpressions:
    - key: environment
      operator: In
      values:
      - production
      - staging
```

since we're building a map in this function we, obviously, can't add two entries with similar key there. So we'll add at least the first value.

internal/serviceset/builder.go
BROngineer: First, clusterDeployment might have no serviceSpec defined. Second, we might want to use different providers for clusterDeployment and multiClusterService. Third - it's user's responsibility to track what is being deployed and how. I mean, we won't prevent user to deploy flux and use it to deploy anything to clusters without this "provider" feature, right? And we don't care if the user will create conflict between stuff deployed by sveltos and by any other CD solution. So, I see no reason to try to implement anything like this.

internal/serviceset/builder.go
BROngineer: There is finalizer on MultiClusterService which guarantees that serviceSets spawned by the MCS will be removed on its removal

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: When the serviceSet created manually, for instance. Or when labels changed.

api/v1beta1/multiclusterservice_types.go
BROngineer: I've restored these fields and added moving them to `.spec.serviceSpec.provider.config`

internal/controller/management_controller.go
```diff
@@ -495,6 +501,100 @@ func (r *ManagementReconciler) ensureAccessManagement(ctx context.Context, mgmt
...
+					Kind:       mgmt.Kind,
+					Name:       mgmt.Name,
+					UID:        mgmt.UID,
+				},
+			},
+		},
+		Spec: kcmv1.StateManagementProviderSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					utils.DefaultStateManagementProviderSelectorKey: utils.DefaultStateManagementProviderSelectorValue,
```
eromanova, line 542: Should this be `currentKcmName` instead of utils.DefaultStateManagementProviderSelectorValue? I’m wondering if it would still work when KCM is deployed with a custom name, since in that case the Deployment name wouldn’t be `kcm-controller-manager`.

BROngineer: Tbh, it doesn't matter. There might be any value. Labels from this selector will be copied to serviceSet. During serviceSet reconciliation there is a check whether serviceSet labels match selector defined in stateManagementProvider, thats it.

internal/controller/multiclusterservice_controller.go
```diff
@@ -83,6 +87,15 @@ func (r *MultiClusterServiceReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, nil
 	}
 
+	// due to API change we'll need to update existing objects
+	if len(mcs.Spec.ServiceSpec.Services) != 0 && mcs.Spec.ServiceSpec.Provider.Name == "" {
+		mcs.Spec.ServiceSpec, err = updateServiceSpecWithProvider(mcs.Spec.ServiceSpec)
```
eromanova: Controllers should generally never update `.spec`. The spec represents the user’s desired state, and only the user should modify it.

If the controller changes .spec, it can lead to a reconciliation fight.

Can we do the following instead:
1. If provider.Config is empty in the ServiceSpec:
* Build the provider configuration based on the deprecated fields.
* Use this derived config when building the ServiceSet spec.
* Do not update the ClusterDeployment or MultiClusterService spec.

2.	If `provider.Config` is present: use the provided `provider.Config` as-is.

3. Add warnings to kcm controller that deprecated fields are in use, suggesting moving them to `provider.Config`.
?

BROngineer: Fixed

kylewuolle (dismissed):

internal/serviceset/util.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+}
+
+// needsUpdate checks if the ServiceSet needs to be updated based on the ClusterDeployment spec.
+// It first compares the ServiceSet's provider configuration with the ClusterDeployment's service provider configuration.
+// Then it compares the ServiceSet's observed services' state with its desired state, and after that it compares
+// the ServiceSet's observed services' state with ClusterDeployment's desired services state.
+func needsUpdate(serviceSet *kcmv1.ServiceSet, providerSpec kcmv1.StateManagementProviderConfig, services []kcmv1.Service) bool {
+	// we'll need to update provider configuration if it was changed.
+	if !equality.Semantic.DeepEqual(providerSpec, serviceSet.Spec.Provider) {
```
wahabmk, line 241: We don't need this check anymore ince `.spec.serviceSpec.provider.name` is immutable now.

wahabmk (approved):
Nice work!

PR #1757, opened 2025-07-16
chore(deps): bump projectsveltos version to v0.57.2

BROngineer:
This PR updates projectsveltos version
- imported packages to v0.57.2
- installed chart to v0.57.2
- addon-controller to v0.57.2.fix

Update of addon-controller is required to implement custom CA feature for Helm registries.

### DIFF
--- a/templates/provider/projectsveltos/values.schema.json
+++ b/templates/provider/projectsveltos/values.schema.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "description": "A KCM provider projectsveltos template",
+  "type": "object",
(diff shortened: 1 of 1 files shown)


templates/provider/projectsveltos/values.schema.json
a13x5, line 1: It's not needed. All values are external anyway + it should be removed when proper sveltos release will be released

BROngineer: without it ci fails: https://github.com/k0rdent/kcm/actions/runs/16326874743/job/46119201373

a13x5 (approved):
LGTM

Issue #1575, opened 2025-05-28
[bug] IPAM broken on ARM64


randybias:
**Describe the bug**
Both of the IPAM controllers are broken on ARM64 systems, but for slightly different reasons.

**To Reproduce**
1. Spin up k0rdent on any ARM64 system and watch the two IPAM controllers go into CrashLoopBackOff State
2. Error is: "exec format error" for both of them

**Expected behavior**
Controllers should start cleanly.

**Additional context**
The in-cluster IPAM is known broken on ARM64 since February; see [this issue in the CAPI IPAM issues](https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/issues/314).

The infloblox image we use is NOT multi-arch as seen in the manifest file if you check it:

`docker manifest inspect ghcr.io/telekom/cluster-api-ipam-provider-infoblox:v0.1.0-alpha.8
`

k0rdent must support both x86 and ARM64 cleanly.

kylewuolle:
I've created a PR upstream to perform multi-arch builds https://github.com/telekom/cluster-api-ipam-provider-infoblox/pull/93. I will be submitting a PR soon that will allow disabling of provider templates.

PR #1784, opened 2025-07-24
test: Support to run e2e tests on arm64 architecture

### LINKED ISSUE #1576: [ci] The build and test system needs to test on ARM64
It's a hard requirement that we support the ARM64 processor; however, we clearly aren't testing that for releases as the most recent 1.0.0 release has IPAM failing on ARM64: https://github.com/k0rdent/kcm/issues/1575

Make sure we test on x86 AND ARM for all major releases

None.  Just fix it.

**Additional context**
This one should be easy.

eromanova:

1. Adds support for testing K0rdent on ubuntu arm64.
2. Enables testing of cluster deployment on arm64 across the following providers: aws, eks, azure, aks, gcp, gke.
3. Temporarily removes Infoblox from the list of deployed providers in e2e tests, pending resolution of #1575
4. Cleans up a redundant workaround previously used for AWS EKS testing.

**Which issue(s) this PR fixes** _(optional, `Fixes #123`)_: Fixes #1576

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -28,6 +28,7 @@ jobs:
     outputs:
       config: ${{ steps.get_config.outputs.config }}
       public_repo: ${{ steps.get_config.outputs.public_repo }}
+      runner: ${{ steps.get_config.outputs.runner }}
     steps:
       - name: Get Testing Configuration
         id: get_config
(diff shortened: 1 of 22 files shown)


.github/workflows/build_test.yml
```diff
@@ -43,7 +44,23 @@ jobs:
             echo "Invalid testing configuration provided"
             exit 1
           fi
-          echo "config=$(echo -n "$config" | base64 -w 0)" >> "$GITHUB_OUTPUT"
+
+          arch=$(echo -n "$config" | yq -r '.architecture // "amd64"')
```
zerospiel, line 49: This comment relates to the line nr 237 (`SKIP_ARM: true`): shouldn't we build the arch images depending on the given arch (output of this job) in the hosted case?

test/e2e/clusterdeployment/aws/aws.go
```diff
@@ -40,6 +41,19 @@ func CheckEnv() {
 	})
 }
 
+func PopulateEnvVars(architecture config.Architecture) {
+	GinkgoHelper()
+
+	switch architecture {
+	case config.ArchitectureAmd64:
+		GinkgoT().Setenv(clusterdeployment.EnvVarAWSAMIID, "") // If unset, the default AMI lookup value defined in the template will be used
```
zerospiel, line 49: Should we make the new envs mandatory, namely, put them into the `CheckEnv` functions of a corresponding provider? This will ensure fail-fast on a local machine e2e tests, as well as potential misuse of test templates. The question is applicable to all of the providers

But maybe it does not make sense since these env variables are not to be customized from a user's inputs

test/e2e/config/config.go
```diff
@@ -144,6 +156,12 @@ func (c *ProviderTestingConfig) String() string {
 	return string(prettyConfig)
 }
 
+func (c *ClusterTestingConfig) SetDefaultArchitecture() {
+	if c.Architecture == "" {
+		c.Architecture = ArchitectureAmd64
```
zerospiel: For the hosted case, wouldn't it be better to inherit "parent"'s architecture instead if nothing is set?

zerospiel (commented):
In general looks good to me, I put several questions in comments

zerospiel (approved):

zerospiel (approved):

PR #1802, opened 2025-07-28
feat: ability to disable provider templates

### LINKED ISSUE #1575: [bug] IPAM broken on ARM64

kylewuolle:
# Description
This PR allows users to disable provider templates so that they are not installed. This is useful for situations where the given provider has no implementation for a specific architecture like ARM64.

Fixes #1575

### DIFF
--- a/hack/update-dev-configs.bash
+++ b/hack/update-dev-configs.bash
@@ -27,12 +27,23 @@ UNTRACKED_CHANGED=$(git ls-files --others --exclude-standard "$TEMPLATE_DIR")
 ALL_CHANGED=$(echo -e "$COMMITTED_CHANGED\n$TRACKED_CHANGED\n$UNTRACKED_CHANGED" \
   | sort -u | grep -E '\.ya?ml$' || true)
 
+escape_helm() {
+  sed -i 's/\({{.*}}\)/temp: "\1"/' $1
+}
+
+unescape_helm() {
+  sed -i 's/temp: "\(.*\)"/\1/' $1
+}
+
 for file in $ALL_CHANGED; do
   [[ -f "$file" ]] || continue
 
+  escape_helm $file
   kind=$(${YQ} e '.kind' "$file")
-  [[ "$kind" != "ClusterTemplate" ]] && continue
-
+  if [[ "$kind" != "ClusterTemplate" ]]; then
+    unescape_helm $file
+    continue
+  fi
   template_name=$(${YQ} e '.metadata.name' "$file")
   template_type=$(${YQ} e '.spec.helm.chartSpec.chart' "$file")
 
--- a/internal/controller/release_controller.go
+++ b/internal/controller/release_controller.go
@@ -153,11 +153,21 @@ func (r *ReleaseReconciler) validateProviderTemplates(ctx context.Context, relea
 		return err
 	}
 	validTemplates := make(map[string]bool)
+	var filteredExpectedTemplates []string
 	for _, t := range providerTemplates.Items {
-		validTemplates[t.Name] = t.Status.ObservedGeneration == t.Generation && t.Status.Valid
+		if !t.Status.Disabled {
+			for _, expected := range expectedTemplates {
+				if expected != t.Name {
+					filteredExpectedTemplates = append(filteredExpectedTemplates, t.Name)
+				}
+			}
+			filteredExpectedTemplates = append(filteredExpectedTemplates, t.Name)
+		}
+		validTemplates[t.Name] = t.Status.Disabled || t.Status.ObservedGeneration == t.Generation && t.Status.Valid
 	}
+
 	invalidTemplates := []string{}
-	for _, t := range expectedTemplates {
+	for _, t := range filteredExpectedTemplates {
 		if !validTemplates[t] {
 			invalidTemplates = append(invalidTemplates, t)
 		}
--- a/internal/controller/template_controller.go
+++ b/internal/controller/template_controller.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
@@ -205,8 +206,17 @@ func (r *TemplateReconciler) ReconcileTemplate(ctx context.Context, template tem
 		}
 	}
 
-	helmSpec := template.GetHelmSpec()
 	status := template.GetCommonStatus()
+	status.Disabled = false
+	if val, ok := template.GetAnnotations()[kcmv1.ProvidertemplateDisabledAnnoation]; ok {
+		if disabled, _ := strconv.ParseBool(val); disabled {
+			l.Info("ProviderTemplate is disabled", "Name", template.GetName())
+			status.Disabled = true
+			return ctrl.Result{}, r.updateStatus(ctx, template, "")
+		}
+	}
+
+	helmSpec := template.GetHelmSpec()
 	var (
 		err     error
 		hcChart *sourcev1.HelmChart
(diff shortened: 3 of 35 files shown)


cmd/main.go
```diff
@@ -114,6 +114,7 @@ func main() {
 		pprofBindAddress              string
 		leaderElectionNamespace       string
 		enableSveltosCtrl             bool
+		disabledProviderTemplates     comaSeperated
```
s3rj1k: IMO: this is so bad UX having a separate args flag for that

kylewuolle: any ideas on an alternative?

s3rj1k: - annotation of some object
- label on some object
- Dedicated ConfigMap with a list of disabled providers

Personally I would prefer annotations

BROngineer: Agree with @s3rj1k. I believe we don't want to restart the controller to enable or disable provider template.

kylewuolle: Sounds good. Maybe annotate the provider template object itself then?

s3rj1k: > Maybe annotate the provider template object itself then

If that works out, can be used together with Helm values to disable providers during first install

templates/provider/kcm/values.yaml
```diff
@@ -1,6 +1,6 @@
 nameOverride: ""
 fullnameOverride: ""
-
+disabledProviderTemplates: [""]
```
zerospiel: Could you add the schema desc?

internal/controller/template_controller.go
```diff
@@ -194,6 +192,11 @@ type templateCommon interface {
 func (r *TemplateReconciler) ReconcileTemplate(ctx context.Context, template templateCommon) (ctrl.Result, error) {
 	l := ctrl.LoggerFrom(ctx)
 
+	if slices.Contains(r.DisabledProviderTemplates, template.GetName()) {
+		l.Info("template is disabled", "name", template.GetName())
```
zerospiel: `name` key is a) duplicated; b) the logger already has the same value in the same key

internal/controller/template_controller.go
```diff
@@ -48,14 +48,12 @@ import (
...
 	client.Client
-
-	downloadHelmChartFunc func(context.Context, *sourcev1.Artifact) (*chart.Chart, error)
-
-	SystemNamespace       string
-	DefaultRegistryConfig helm.DefaultRegistryConfig
-	CreateManagement      bool
-
-	defaultRequeueTime time.Duration
+	downloadHelmChartFunc     func(context.Context, *sourcev1.Artifact) (*chart.Chart, error)
```
zerospiel: nit: the idea of the separation of the field blocks is to improve readability (a little), if that's feasible, i'd rather keep line breaks as they were

internal/controller/release_controller.go
```diff
@@ -57,17 +58,14 @@ import (
...
-
-	KCMTemplatesChartName string
-	SystemNamespace       string
-
-	DefaultRegistryConfig helm.DefaultRegistryConfig
-
-	CreateManagement bool
-	CreateRelease    bool
-	CreateTemplates  bool
+	Config                    *rest.Config
```
zerospiel: nit: ditto about line breaks

internal/controller/release_controller.go
```diff
@@ -156,8 +154,10 @@ func (r *ReleaseReconciler) validateProviderTemplates(ctx context.Context, relea
 	}
 	invalidTemplates := []string{}
 	for _, t := range expectedTemplates {
-		if !validTemplates[t] {
-			invalidTemplates = append(invalidTemplates, t)
+		if !slices.Contains(r.DisabledProviderTemplates, t) {
+			if !validTemplates[t] {
+				invalidTemplates = append(invalidTemplates, t)
+			}
```
zerospiel:
```suggestion
		if !validTemplates[t] && !slices.Contains(r.DisabledProviderTemplates, t) {
			invalidTemplates = append(invalidTemplates, t)
```

better use short circuit evaluation and within one statement

cmd/main.go
```diff
@@ -470,3 +473,14 @@ func setupWebhooks(mgr ctrl.Manager, currentNamespace string, validateClusterUpg
 	}
 	return nil
 }
+
+type comaSeperated []string
```
zerospiel: just wondering: why did you choose that approach instead of, for example, a plain implementation of the `flag.Value` interface? both are seem the same, the latter is more popular, i'd say (and more flexible)

zerospiel (commented):
Looks good, I also would agree with the previous CR that adding either an annotation or a label would provide more flexibility. Though in this particular case, I'd stick to a label since it is up to a human operator to set it, not to a machine.

cmd/main.go
kylewuolle: Thanks for the suggestion. I've updated the PR to use the annotation approach.

hack/update-dev-configs.bash
```diff
@@ -27,12 +27,23 @@ UNTRACKED_CHANGED=$(git ls-files --others --exclude-standard "$TEMPLATE_DIR")
 ALL_CHANGED=$(echo -e "$COMMITTED_CHANGED\n$TRACKED_CHANGED\n$UNTRACKED_CHANGED" \
   | sort -u | grep -E '\.ya?ml$' || true)
 
+escape_helm() {
+  sed -i 's/\({{.*}}\)/temp: "\1"/' $1
```
s3rj1k, line 31: MacOS tested ?

s3rj1k: same for other `sed`s

kylewuolle:
> @kylewuolle @BROngineer I just want to clarify the intentions behind this change. Currently, providers are enabled or disabled by adding or removing them from the Management object. Do we really want to introduce another mechanism for managing this behavior?
> 
> For your specific case with the ipam-provider: right now, if you want to disable it, you remove it once from the Management object, and it will never be installed again unless you add it back.
> 
> After the changes introduced in this PR, the following situations could occur:
> 
> * You have a ProviderTemplate that is requested in Management → you manually disable it with the annotation → it is still expected from the Management object’s perspective, but removed from the cluster → you get another KCM release and upgrade to it → the newly created ProviderTemplate does not have the annotation, so the provider gets deployed again.
> * You have a ProviderTemplate disabled with the annotation → you add it to the Management object → admission does not prevent it, but nothing gets deployed.
> 
> I also left some in-line comments that are related. Please take a look.

OK maybe that will suffice then. So for disabling providers we just instruct users to remove the provider in the management object. Does that work for you @s3rj1k

s3rj1k:
> OK maybe that will suffice then. So for disabling providers we just instruct users to remove the provider in the management object. Does that work for you @s3rj1k

We already talked about that multiple times with @Kshatrix and other members of the team.

If we keep it this way and we have broken ARM images for example we have very bad UX.

If someone does not want all providers to be in Release and K0rdent still forces that person to install all of them and after install manually remove some of them instead of giving this option from helm directly, still bad UX.

PR #1810, opened 2025-07-29
Fix: Catch errors not found errors and remind users to label the template

vtrenton:
Customer ran into an issue while creating a service template but did not know about the required `k0rdent.mirantis.com/managed: "true"` label. This PR captures the "Not found" and makes a note to tell the user to add a label. This makes the issue easier to diagnose for users.

### DIFF
--- a/internal/controller/template_controller.go
+++ b/internal/controller/template_controller.go
@@ -359,6 +359,11 @@ func fillStatusWithProviders(template templateCommon, helmChart *chart.Chart) er
 func (r *TemplateReconciler) updateStatus(ctx context.Context, template templateCommon, validationError string) error {
 	status := template.GetCommonStatus()
 	status.ObservedGeneration = template.GetGeneration()
+
+	if validationError != "" && strings.Contains(validationError, "failed to get source:") && strings.Contains(validationError, "not found") {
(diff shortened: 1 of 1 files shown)


internal/controller/template_controller.go
```diff
@@ -359,6 +359,11 @@ func fillStatusWithProviders(template templateCommon, helmChart *chart.Chart) er
 func (r *TemplateReconciler) updateStatus(ctx context.Context, template templateCommon, validationError string) error {
 	status := template.GetCommonStatus()
 	status.ObservedGeneration = template.GetGeneration()
+
+	if validationError != "" && strings.Contains(validationError, "failed to get source:") && strings.Contains(validationError, "not found") {
```
zerospiel, line 363: Please check whether the error is at least `apierrors.IsNotFound()`. I also suggest putting the check in the place where the actual `"github.com/fluxcd/source-controller/api/v1".HelmChart` is being retrieved, and check its conditions on that particular `"github.com/fluxcd/source-controller/api/v1".FetchFailedCondition` and its reason set to `SourceUnavailable` along with that check of the underlying error. That would be much better than checking for arbitrary errors that contain another arbitrary string that could change at any moment.

zerospiel (changes requested):
Thanks for the PR. The general approach to probing arbitrary errors should be mended. And please fix the CI

PR #1832, opened 2025-08-05
fix: redundant servicetemplates fetching

zerospiel:
Remove excessive fetching of ServiceTemplate objects during getting a referenced one from a Service

### DIFF
--- a/internal/sveltos/profile.go
+++ b/internal/sveltos/profile.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"slices"
 	"strings"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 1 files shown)


internal/sveltos/profile.go
```diff
@@ -523,33 +524,25 @@ func serviceTemplateObjectFromService(
...
-			return nil, fmt.Errorf("ServiceTemplate %s is not supported by ServiceTemplateChain %s", svc.Template, key)
-		}
+	if !templateChain.Status.Valid {
+		return nil, fmt.Errorf("the ServiceTemplateChain %s is invalid with the error: %s", key, templateChain.Status.ValidationError)
+	}
+
+	if slices.ContainsFunc(templateChain.Spec.SupportedTemplates, func(st kcmv1.SupportedTemplate) bool {
+		return st.Name == svc.Template
+	}) {
+		return template, nil
```
wahabmk, line 544: In the original, before returning, the matching ServiceTemplate is fetched to make sure it exists. Are we sure we don't need that check?

zerospiel: it is the same as we'd fetched in the very beginning of the function

zerospiel: namely (in the prior code):

```go
if supportedTemplate.Name != svc.Template {
    continue
}
// supportedTemplae.NAme == svc.Template
// hence 
templateKey := client.ObjectKey{Name: supportedTemplate.Name, Namespace: namespace}
// is the same as
templateKey := client.ObjectKey{Name: svc.Template, Namespace: namespace}
// which we already fetched

wahabmk: The start of the function was hidden so I missed that. All good!

PR #1840, opened 2025-08-07
 fix: Continue e2e tests after cluster deployment failure

### LINKED ISSUE #1814: [bug][test] Ensure provider e2e tests continue after partial cluster deployment failure
When running e2e tests with complex configurations (e.g., multiple clusters for a single provider), the testing process currently halts if any cluster fails to deploy. For example, with the following configuration:

```
aws:
  - hosted: {}
  - architecture: arm64
    hosted: {}
  - template: aws-eks-1-0-3
```

If the first cluster fails to deploy (due to an error), the entire e2e test suite for that provider stops, and the remaining configurations are not tested. This prevents validation of the other setups, even if they could succeed independently.

eromanova:
    * Enable ContinueOnFailure for ordered Ginkgo containers to allow test execution to proceed after individual failures
    * Swap for loop and It blocks to ensure dynamic It block generation to isolate each cluster deployment test
    * Add GinkgoRecover() to properly handle panics triggered by Fail() within It blocks
    * Move config parsing before RunSpecs() to ensure configuration is available when test cases are defined
    * Refactor config initialization logic and defaults population
    * Remove redundant empty config checks
    * Additionally, fixes AWS env vars population for arm64 in case of hosted deployment

Fixes #1814


test/e2e/config/config.go
```diff
@@ -85,6 +82,10 @@ type ClusterTestingConfig struct {
 }
 
 func Parse() error {
+	if len(configBytes) == 0 {
+		Initialize()
```
zerospiel, line 86: with that change, if config is not provided, that means the whole providers would run? if so, then i suggest to amend the CI jobs accordingly, otherwise they will be skipped

eromanova: By default, the `config.yaml` file is populated with the following config https://github.com/k0rdent/kcm/blob/main/test/e2e/config/config.yaml
```
aws: []
``` 
This triggers deployment of the default AWS cluster only.
If a testing configuration isn’t explicitly provided in the `authorize` job, the `config.yaml` file won’t be overwritten (if the comment to the authorize job is empty) https://github.com/k0rdent/kcm/blob/main/.github/workflows/build_test.yml#L337, and the default configuration will be used. But, if the file is completely empty (which is only possible during manual e2e test runs, not in our current CI), the `Initialize()` function will automatically populate it with all available providers.
This logic remains unchanged in this PR.
Do you think we should change this behavior so that an empty `config.yaml` skips provider testing entirely?

zerospiel: Might be put in the `docs/dev.md` but not mandatory within the PR. Thanks for the clarification, no further actions required

zerospiel (approved):

PR #1845, opened 2025-08-08
feat: introduce local telemetry collector

zerospiel:
TBD


internal/telemetry/collector/local.go
```diff
@@ -0,0 +1,560 @@
...
+	default:
+		return prefix + "other"
+	}
+}
+
+func bucketProvider(p string) string {
+	// (bootstrap|control-plane|infrastructure)-<provider>
+	// NOTE: very opinionated metric, let us reflect whether we actually need it
+	return "providers:" + p
+}
```
zerospiel: I'd rather drop this metric: it does not provide much value, having a potentially large cardinality (a dozen), increasing the final size of the file

internal/telemetry/collector/local.go
```diff
@@ -0,0 +1,560 @@
...
+		return prefix + "11-20"
+	case n <= 50:
+		return prefix + "21-50"
+	default:
+		return prefix + "51+"
+	}
+}
+
+func bucketTemplate(name string) string {
+	return "template:" + name
```
zerospiel: Another metric with a potentially large cardinality (uncontrolled in this case). How frequently is a cluster deployment to change its template?

templates/provider/kcm/templates/deployment.yaml
```diff
@@ -19,6 +19,22 @@ spec:
       annotations:
         kubectl.kubernetes.io/default-container: manager
     spec:
+      {{ if and .Values.controller .Values.controller.telemetry (eq .Values.controller.telemetry.mode "local") .Values.controller.telemetry.baseDir }}
```
eromanova: nit:
This block is duplicated multiple times. Can it be defined at the top of the file, for example:
```
{{- $isLocalTelemetry := and .Values.controller .Values.controller.telemetry (eq .Values.controller.telemetry.mode "local") .Values.controller.telemetry.baseDir -}}
```
and reused later as `{{- if $isLocalTelemetry }}`?

internal/telemetry/collector/local.go
```diff
@@ -0,0 +1,560 @@
...
+		return prefix + "256-512"
+	case bytes <= 1024<<30:
+		return prefix + "512-1024"
+	default:
+		return prefix + "1024+"
+	}
+}
+
+func bucketGPUGiUsed(typ string, bytes uint64) string {
+	prefix := fmt.Sprintf("gpu.%s.gibibytes:", typ)
```
zerospiel: I wonder if that is expected to have reqs in GiB rather than in MiB. What would be a better option in terms of the requests? The same question is applicable for the capacity (but I reckon GiB is okay in case of the capacity)

internal/telemetry/collector/local.go
```diff
@@ -0,0 +1,560 @@
...
+	case n == 0:
+		return prefix + "0"
+	case n <= 4:
+		return prefix + "1-4"
+	case n <= 16:
+		return prefix + "5-16"
+	case n <= 64:
+		return prefix + "17-64"
+	case n <= 256:
+		return prefix + "64-256"
```
eromanova:
```suggestion
		return prefix + "65-256"
```

templates/provider/kcm/templates/deployment.yaml
zerospiel: [Done](https://github.com/k0rdent/kcm/compare/422b405bee6ab4a935e32bc4c57cb23ff7ab2f77..9869ec47c692fa71cb8265533f5d2b72e9b8055e), thanks

internal/telemetry/collector/local.go
zerospiel: [Done](https://github.com/k0rdent/kcm/compare/422b405bee6ab4a935e32bc4c57cb23ff7ab2f77..9869ec47c692fa71cb8265533f5d2b72e9b8055e) (and expanded), thanks

internal/telemetry/collector/local.go
zerospiel: Removed the metric

internal/telemetry/collector/local.go
zerospiel: Decided to leave the metric

internal/telemetry/collector/local.go
zerospiel: Changed from `gibi` to `mebi` with the `[2MiB,16GiB]` bound

internal/telemetry/collector/local.go
zerospiel: The approach has been reworked to automate ranges creation

eromanova (approved):

PR #1847, opened 2025-08-11
fix: Mount CA secret to flux before installing kcm-templates

### LINKED ISSUE #1841: [bug] KCM installation fails with custom registry that uses redirects: tls cert valid for a, not b
KCM installation fails when using a custom registry with a custom CA and when the registry uses redirects due to a known bug in flux, see:
* [fluxcd/flux2#4838](https://github.com/fluxcd/flux2/issues/4838)
* [fluxcd/source-controller#510](https://github.com/fluxcd/source-controller/issues/510)

Setting `certSecretRef` in the HelmRepository is not sufficient when the registry redirects to a different hostname. Flux controllers (helm and source) only use the provided CA certificate, and do not support redirects that present a different certificate chain (e.g., cert for a different domain):

_Note: registry URLs have been redacted_
```
kubectl -n k0rdent get helmchart kcm-1-1-0-rc5-tpl -o yaml
apiVersion: source.toolkit.fluxcd.io/v1
kind: HelmChart
metadata:
  name: kcm-1-1-0-rc5-tpl
  namespace: k0rdent
spec:
  chart: kcm-templates
  interval: 10m0s
  reconcileStrategy: ChartVersion
  sourceRef:
    kind: HelmRepository
    name: kcm-templates
  version: 1.1.0-rc5
status:
  conditions:
  - lastTransitionTime: "2025-08-06T14:07:44Z"
    message: building artifact
    observedGeneration: 1
    reason: ProgressingWithRetry
    status: "True"
    type: Reconciling
  - lastTransitionTime: "2025-08-06T14:07:44Z"
    message: 'chart pull error: failed to download chart for remote reference: failed
      to get ''oci://<registryA>/charts/kcm-templates:1.1.0-rc5'':
      failed to copy: httpReadSeeker: failed open: failed to do request: Get "https://<registryA>/docker/registry/v2/blobs/sha256/f6/f631fb2e19bd64728400ae37c0d8ffbd0d789bc01d95ea680d99d809bacb920c":
      tls: failed to verify certificate: x509: certificate is valid for *.<registryB>,
      not <registryA>'
    observedGeneration: 1
    reason: ChartPullError
    status: "False"
    type: Ready
```

Even though the certificate is valid for `<registryB>`, the controllers expect it to match `<registryA>`, and the verification fails.

The regular helm CLI works fine with the same setup:
```
$ helm pull --ca-file ca.pem oci://registry.k8s.pac.dockerps.io/mke4/charts/kcm-templates --version 1.1.0-rc5
Pulled: registry.k8s.pac.dockerps.io/mke4/charts/kcm-templates:1.1.0-rc5
```

When HelmRepository is not configured with `spec.certSecretRef`, then k0rdent installation is also working correctly, even if using the same setup with redirections.

**Workaround:**

1. Remove `spec.certSecretRef` from the default kcm HelmRepository
2. (optional) Mount the CA cert as a volume directly into both helm and source controllers (required only if the registry CA secret chain is not trusted by the management cluster nodes). Can be achieved with the following kcm values:
```
flux2:
  helmController:
    volumeMounts:
      - name: registry-cert
        mountPath: /etc/ssl/certs/registry-ca.pem
        subPath: registry-ca.pem
    volumes:
      - name: registry-cert
        secret:
          defaultMode: 420
          secretName: registry-cert  # name of the secret containing the CA certificate under the ca.crt key
          items:
            - key: ca.crt
              path: registry-ca.pem
  sourceController:
    volumeMounts:
      - name: registry-cert
        mountPath: /etc/ssl/certs/registry-ca.pem
        subPath: registry-ca.pem
    volumes:
      - name: registry-cert
        secret:
          defaultMode: 420
          secretName: registry-cert  # same secret as above
          items:
            - key: ca.crt
              path: registry-ca.pem
```

eromanova:
This is the fix for the Flux issue to ensure that the registry CA certificate is mounted into the Flux source controller deployment before attempting to pull `kcm-templates` chart from a private registry. Without the certificate in place early, the `kcm-templates` installation fails on:
```
  - lastTransitionTime: "2025-08-08T21:08:06Z"
    message: 'HelmChart ''k0rdent/kcm-1-1-0-rc14-tpl'' is not ready: chart pull error:
      failed to download chart for remote reference: failed to get ''oci://registry.mke4dev.miratnis.com/mke/charts/kcm-templates:1.1.0-rc14'':
      failed to do request: Head "https://registry.mke4dev.miratnis.com/v2/mke/charts/kcm-templates/manifests/1.1.0-rc14":
      tls: failed to verify certificate: x509: certificate signed by unknown authority'
```
This patch should only run during the initial deployment, before the kcm-templates chart is installed.

Also removes volume overrides for the helm controller, since they are only required for the source controller.

**Which issue(s) this PR fixes** _(optional, `Fixes #123`)_
Fixes #1841

### DIFF
--- a/internal/controller/release_controller.go
+++ b/internal/controller/release_controller.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
@@ -29,6 +30,8 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/storage/driver"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -36,6 +39,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -314,6 +318,13 @@ func (r *ReleaseReconciler) reconcileKCMTemplates(ctx context.Context, releaseNa
 			return false, fmt.Errorf("some of the predeclared Secrets (%v) are missing (%v) in the %s namespace", helmRepositorySecrets, missingSecrets, r.SystemNamespace)
 		}
 
+		if r.DefaultRegistryConfig.CertSecretName != "" {
+			err = r.patchFluxWithRegistryCASecret(ctx)
+			if err != nil {
+				return false, fmt.Errorf("failed to patch flux components with registry CA secret volume: %w", err)
+			}
+		}
+
 		releaseName, err = utils.ReleaseNameFromVersion(build.Version)
 		if err != nil {
 			return false, fmt.Errorf("failed to get Release name from version %q: %w", build.Version, err)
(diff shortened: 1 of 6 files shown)


templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -332,6 +332,8 @@ rules:
   - deployments
   verbs:
   - list
+  - watch
```
zerospiel, line 334: nit: comments on the 328-329 LOC are no longer valid

internal/controller/release_controller.go
```diff
@@ -399,6 +410,76 @@ func (r *ReleaseReconciler) reconcileKCMTemplates(ctx context.Context, releaseNa
...
+		return err
+	}
+
+	managerIdx := -1
+	for i, c := range deployment.Spec.Template.Spec.Containers {
+		if c.Name == managerContainerName {
+			managerIdx = i
+			break
+		}
+	}
```
zerospiel:
```suggestion
	managerIdx := slices.IndexFunc(deployment.Spec.Template.Spec.Containers, func(c corev1.Container) bool {
		return c.Name == managerContainerName
	})
```

internal/controller/release_controller.go
```diff
@@ -399,6 +410,76 @@ func (r *ReleaseReconciler) reconcileKCMTemplates(ctx context.Context, releaseNa
...
+// Applies only to the initial deployment to add the registry
+// CA certificate to flux components before installing kcm-templates HelmChart
+func (r *ReleaseReconciler) patchFluxWithRegistryCASecret(ctx context.Context) error {
+	const (
+		deploymentName       = "source-controller"
+		caCertVolumeName     = "registry-cert"
+		caCertFileName       = "registry-ca.pem"
+		managerContainerName = "manager"
+	)
+	mode := int32(420)
```
zerospiel: nit: can we switch to the in-place `DefaultMode: "github.com/K0rdent/kcm/internal/utils/pointer".To(int32(420))`? the reason is that the var is defined at the beginning of the method and used only after 40 LOC

PR #1857, opened 2025-08-14
feat: telemetry local storage configuration

### LINKED ISSUE #1854: [telemetry] local storage configuration
provide more flexibility to configure telemetry-related storage (type of pv, fs type, sizes, dirs, retantion (?), and etc)

zerospiel:
Rework telemetry block, placing it on top of the values.yaml.

Expose storage configuration options for the telemetry local mode, if the defaults are not suitable for a user.

Add basic helm unittest files.

Closes #1854


templates/provider/kcm/templates/_helpers.tpl
```diff
@@ -119,15 +119,24 @@ kcm-webhook
 {{- end -}}
 
 {{/*
-The name of the telemetry PV name
+The name of the telemetry PVC name
```
eromanova:
```suggestion
The name of the telemetry PVC
```

templates/provider/kcm/templates/_helpers.tpl
```diff
@@ -119,15 +119,24 @@ kcm-webhook
...
+{{- if and (eq $vol.source "existing") $vol.existingClaim -}}
+{{- $vol.existingClaim -}}
+{{- else -}}
+{{- printf "%s-telemetry" (include "kcm.fullname" .) -}}
+{{- end -}}
+{{- end -}}
 
 {{/*
-The name of the telemetry PVC name
+The name of the telemetry PV name
```
eromanova:
```suggestion
The name of the telemetry PV
```

templates/provider/kcm/templates/deployment.yaml
```diff
@@ -65,10 +68,12 @@ spec:
...
-        {{- range $key, $value := .Values.controller.telemetry }}
-        {{- if not (eq (printf "%s" $value) "") }}
-        - --telemetry-{{ $key }}={{ $value }}
-        {{- end }}
+        - --telemetry-mode={{ $tel.mode }}
+        - --telemetry-concurrency={{ $tel.concurrency }}
+        - --telemetry-interval={{ $tel.interval }}
+        - --telemetry-jitter={{ $tel.jitter }}
+        {{- if $isLocalTelemetry }}
+        - --telemetry-baseDir={{ $baseDir }}
```
eromanova: Since we’ve moved away from the `--telemetry-{{ $key }}` approach anyway, it might be better to rename this parameter to `telemetry-base-dir` to avoid mixing camelCase with dash delimiters. But it's up to you.

templates/provider/kcm/values.yaml
```diff
@@ -33,13 +33,39 @@ controller:
...
+
+      existingClaim: "" # @schema description: Name of the claim to be used, valid if source of the storage is to existing; type: string
+
+      hostPath: # @schema type: object; description: Settings of the hostPath PV and PVC source if source of the storage is set to hostPath; additionalProperties: false
+        name: "" # @schema type: string; description: Name of the PV to be created, defaults to <fullname>-telemetry-data
+        reclaimPolicy: Retain # @schema type: string; enum: [Retain, Delete]; description: Policy for end-of-life maintenance of the volume
+        storageClassName: "manual" # @schema type: string; description: Name of the StorageClass to match PV and PVC
+        accessModes: ["ReadWriteOnce"] # @schema item: string; itemEnum: [ReadWriteOnce, ReadWriteOncePod, ReadOnlyMany, ReadWriteMany]; description: Ways the volume can be mounted, RWOP is not supported with the default Deployment settings
+        volumeMode: Filesystem # @schema type: string; enum: []; description: Defines if a volume is intended to be used with a formatted filesystem or to remain in raw block state
```
eromanova: `enum: []`
Do we allow any value here?

zerospiel: According to tests: yes! But anyway, I've added `filesystem, block` now, I've just forgotten to add this in the first place, thanks for noticing

eromanova (approved):

PR #1877, opened 2025-08-25
fix: set kcm version before making the release

eromanova:
We already run `set-kcm-version` as part of the `kcm-chart-release` makefile target, but it happens too late, after the `release.yaml` file for assets has already been generated. As a result, `release.yaml` may contain an incorrect version.

To avoid this, we need to run `make set-kcm-version` before making the release.


zerospiel (changes requested):
This won't work due to clean git state, and goreleaser will fail

zerospiel (approved):
LGTM

PR #1882, opened 2025-08-25
feat: observability of events and metrics for ipam

### LINKED ISSUE #1676: [enh] enhance IPAM controllers observability
Need to enhance IPAM-related controllers' observability by:
- emitting events when applicable
- adding conditions to IPAM-related types to make it possible to track phases of reconciliation cycle
- consider to add metrics to be collected by monitoring software

kylewuolle:
This PR adds events and metric for cluster IPAM. 

Fixes #1676

### DIFF
--- a/api/v1beta1/clusteripamclaim_types.go
+++ b/api/v1beta1/clusteripamclaim_types.go
@@ -26,8 +26,14 @@ const (
 	// ClusterIPAMClaimKind Denotes the clusteripamclaim resource Kind.
 	ClusterIPAMClaimKind = "ClusterIPAMClaim"
 
+	// ClusterIPAMKind Denotes the clusteripam resource Kind.
+	ClusterIPAMKind = "ClusterIPAM"
+
 	// InvalidClaimConditionType Denotes that the claim is invalid
 	InvalidClaimConditionType = "InvalidClaimCondition"
+
+	// IPAMProviderConditionType Denotes the status of the IPAM provider
+	IPAMProviderConditionType = "IPAMProviderConditionType"
 
 	// InClusterProviderName denotes the In-Cluster CAPI IPAM provider name
 	InClusterProviderName = "in-cluster"
--- a/internal/controller/ipam/clusteripam_controller.go
+++ b/internal/controller/ipam/clusteripam_controller.go
@@ -20,6 +20,8 @@ import (
 	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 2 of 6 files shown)


internal/telemetry/event.go
```diff
@@ -16,21 +16,32 @@ package telemetry
...
+
+func TrackClusterDeploymentCreate(id, clusterDeploymentID, template string, dryRun bool) error {
+	return trackEvent(id, "cluster-deployment-create", map[string]any{
+		"clusterDeploymentID": clusterDeploymentID,
+		"template":            template,
+		"dryRun":              dryRun,
+	})
+}
+
+func TrackClusterIPAMCreate(clusterIPAMId types.UID, cluster, ipamProvider string) error {
```
zerospiel:
```suggestion
func TrackClusterIPAMCreate(clusterIPAMId string, cluster, ipamProvider string) error {
```

nit-picking: better cask during call to align with another public func

internal/controller/ipam/clusteripamclaim_controller.go
```diff
@@ -82,6 +91,13 @@ func (r *ClusterIPAMClaimReconciler) createOrUpdateClusterIPAM(ctx context.Conte
 		clusterIPAM.Spec = clusterIPAMSpec
 		return nil
 	})
+
+	if err := telemetry.TrackClusterIPAMCreate(clusterIPAM.UID, clusterIPAMClaim.Spec.Cluster, clusterIPAMClaim.Spec.Provider); err != nil {
```
zerospiel: this will always send events regardless of whether it was actually created or not, which might cause an undefined and unnecessary increase in costs

internal/metrics/metrics.go
```diff
@@ -32,73 +32,94 @@ const (
...
+}
+
+func setGaugeAndLog(ctx context.Context, gauge *prometheus.GaugeVec, labels prometheus.Labels, active bool, logMsg string) { //nolint:revive
+	value := 0.0
+	if active {
 		value = 1
 	}
+	gauge.With(labels).Set(value)
+
+	ctrl.LoggerFrom(ctx).V(1).Info(logMsg, append(labelMapToSlice(labels), "value", value)...)
```
zerospiel: to avoid calling `labelMapToSlice` when it is not necessary, it's better to check whether the level is enabled:

```o
l := ctrl.LoggerFrom(ctx)
if l.V(1).Enabled() { // log }
```

internal/controller/ipam/clusteripam_controller.go
```diff
@@ -58,9 +60,19 @@ func (r *ClusterIPAMReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				Status:             metav1.ConditionUnknown,
+				Reason:             kcmv1.FailedReason,
+				Message:            fmt.Sprintf("ClusterIPAMClaim provider processing failed: %v", err),
+				ObservedGeneration: clusterIPAMClaim.Generation,
+			})
+
 		return ctrl.Result{}, fmt.Errorf("failed to create provider specific data for ClusterIPAM %s/%s: %w", clusterIPAM.Namespace, clusterIPAM.Name, err)
 	}
 
+	apimeta.RemoveStatusCondition(&clusterIPAMClaim.Status.Conditions, kcmv1.IPAMProviderConditionType)
```
zerospiel: do i understand it correctly that the condition behaves like `error or nothing`? if so, then i reckon the name should be changed to something similar to `IPAMProviderConditionError` 

otherwise, the condition must be set (if changed) in case of no prior error

api/v1beta1/clusteripamclaim_types.go
```diff
@@ -26,9 +26,15 @@ const (
...
 	ClusterIPAMClaimKind = "ClusterIPAMClaim"
 
+	// ClusterIPAMKind Denotes the clusteripam resource Kind.
+	ClusterIPAMKind = "ClusterIPAM"
+
 	// InvalidClaimConditionType Denotes that the claim is invalid
 	InvalidClaimConditionType = "InvalidClaimCondition"
 
+	// IPAMProviderConditionType Denotes the status of the IPAM provider
+	IPAMProviderConditionType = "IPAMProviderConditionType"
```
zerospiel: i suppose, the `*Type` suffix is redundant in the name

zerospiel (changes requested):
I have my concerns in regards to uncontrolled events sent to segment.io. Otherwise minor comments/questions

internal/controller/ipam/clusteripamclaim_controller.go
zerospiel: besides, the event won't be accurate in that matter since update or noop will be considered as the creation

PR #1887, opened 2025-08-26
fix: moved network configuration for standalone charts into values.yaml

### LINKED ISSUE #1873: [bug] Network Configuration in Standalone Cluster Charts is hardcoded
**Describe the bug**
In several of the Cluster charts the Network configuration is hardcoded in the template itself and cannot be modified via the values.yaml. For users building their own ClusterTemplates this makes finding the configuration challenging.

This seems to be an issue on the "standalone" charts:
https://github.com/k0rdent/kcm/blob/efb50cde93392bf627e907f49fb6fc39ee98ea27/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml#L64
https://github.com/k0rdent/kcm/blob/efb50cde93392bf627e907f49fb6fc39ee98ea27/templates/cluster/vsphere-standalone-cp/templates/k0scontrolplane.yaml#L56

**To Reproduce**
1. Copy chart from github
2. Attempt to modify network settings such as change CNI or set Calico settings.
3. realize this can only be done via the template Object directly meaning digging through templates instead of using values.yaml

**Expected behavior**
Default values are OK, but this needs to be exposed via the values.yaml

vtrenton:
Allows for more simple network configuration ergonomics when  building custom ClusterTemplate charts. Configuring networks is an extremely common use case. So we should make this as simple as possible.

Fixes #1873

### DIFF
--- a/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
+++ b/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
@@ -62,9 +62,7 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
         network:
-          provider: calico
-          calico:
-            mode: ipip
+          {{- toYaml .Values.k0s.network | nindent 10 }}
         {{- if $global.registry }}
         images:
           metricsserver:
--- a/config/dev/aws-clusterdeployment.yaml
+++ b/config/dev/aws-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: aws-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
-  template: aws-standalone-cp-1-0-12
+  template: aws-standalone-cp-1-0-13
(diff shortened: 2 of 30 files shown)


templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
```diff
@@ -62,9 +62,7 @@ spec:
             {{- toYaml . | nindent 12 }}
           {{- end }}
         network:
-          provider: calico
-          calico:
-            mode: ipip
+          {{- toYaml .Values.k0s.network | nindent 10 }}
```
zerospiel, line 65: nit-picking: could you replace elsewhere with (please pay attention to dashes `-`)

```
        {{- with .Values.k0s.network }}
        network: {{ toYaml . | nindent 10 }}
        {{- end }}
```

almost the same but ensures no `network` block will be empty if `null` is assigned to the block

zerospiel: i will follow up

PR #1895, opened 2025-08-27
fix: network cfg in hosted tpls; fix standalone

### LINKED ISSUE #1873: [bug] Network Configuration in Standalone Cluster Charts is hardcoded

zerospiel:
Adds configuration of k0s network to all of the hosted templates.

Fixes potentially dangerous configuration in standalone templates introduced by #1887.

Closes #1873

### DIFF
--- a/templates/cluster/aws-hosted-cp/values.yaml
+++ b/templates/cluster/aws-hosted-cp/values.yaml
@@ -60,3 +60,7 @@ k0s: # @schema description: K0s parameters; type: object
   workerArgs: [] # @schema description: Args specifies extra arguments to be passed to k0s worker. See: https://docs.k0sproject.io/stable/worker-node-config/ See: https://docs.k0sproject.io/stable/cli/k0s_worker/; type: array; item: string; uniqueItems: true
   api: # @schema description: Kubernetes API server parameters; type: object
     extraArgs: {} # @schema description: Map of key-values (strings) for any extra arguments to pass down to Kubernetes api-server process; type: object; additionalProperties: true
+  network: # @schema description: Kubernetes cluster network configuration; type: object; additionalProperties: true
+    provider: calico # @schema description: Network provider to use; type: string; examples: [kuberouter, calico, custom]; minLength: 1
+    calico: # @schema description: Calico-specific configuration; type: object; additionalProperties: true
+      mode: ipip # @schema description: Calico network mode; type: string; examples: [ipip, vxlan, never]; minLength: 1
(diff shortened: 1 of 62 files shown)


templates/cluster/aws-hosted-cp/values.yaml
```diff
@@ -60,3 +60,7 @@ k0s: # @schema description: K0s parameters; type: object
   workerArgs: [] # @schema description: Args specifies extra arguments to be passed to k0s worker. See: https://docs.k0sproject.io/stable/worker-node-config/ See: https://docs.k0sproject.io/stable/cli/k0s_worker/; type: array; item: string; uniqueItems: true
   api: # @schema description: Kubernetes API server parameters; type: object
     extraArgs: {} # @schema description: Map of key-values (strings) for any extra arguments to pass down to Kubernetes api-server process; type: object; additionalProperties: true
+  network: # @schema description: Kubernetes cluster network configuration; type: object; additionalProperties: true
+    provider: calico # @schema description: Network provider to use; type: string; examples: [calico, cilium, flannel, kube-router]; minLength: 1
```
eromanova: nit: Maybe better to change these examples to the correct ones from k0s `network.provider` enum settings:
https://github.com/k0sproject/k0s/blob/release-1.32/pkg/apis/k0s/v1beta1/network.go#L55

`kuberouter;calico;custom`

zerospiel: Done, thanks for noticing!

eromanova (approved):

Issue #1793, opened 2025-07-28
[bug] Openstack cluster templates don't support Credentials


Kshatrix:
Openstack cluster templates don't support Credentials. (they use `identityRef.name` in the chart values when `clusterIdentity.name` is injected from the Credential object)

bnallapeta:
@Kshatrix 

Issue: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2386 
Proposal PR: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2628 

Waiting for the CAPO community to respond on this.

eromanova:
Hi @bnallapeta. Thanks for pointing out the upstream issue, but it’s slightly different in this context.

In k0rdent, we have a credential reference propagation mechanism: the value of `spec.identityRef.name` from the referenced Credential object is automatically propagated to `spec.config.clusterIdentity.name` in the ClusterDeployment by the KCM controller.

But, in our OpenStack templates, we currently ignore `spec.config.clusterIdentity.name`, which forces users to manually specify the credential secret as `identityRef.name`. This is the core issue here since the whole Credential approach is being ignored. We should instead switch to using `clusterIdentity` as designed.

But you also raised an important point: the lack of cross-namespace support for credential secrets. This breaks the expected propagation workflow when using the AccessManagement object for OpenStack credentials. Until the upstream issue is addressed, we should clearly document it.

PR #1898, opened 2025-08-27
Add missing identityRef.name field to OpenStack values files

jlnhnng:
Adds the missing identityRef.name field to both OpenStack values files used by the standalone and hosted control-plane templates.
Files updated:
- templates/cluster/openstack-standalone-cp/values.yaml
- templates/cluster/openstack-hosted-cp/values.yaml

The templates reference .Values.identityRef.name in OpenStackMachineTemplate but the values files lacked this key, causing validation/render errors.


zerospiel:
Please use the `clusterIdentity.name`. This particular field you're trying to add back was a bug https://github.com/k0rdent/kcm/issues/1793

jlnhnng:
The current OpenStack cluster templates still use `identityRef.name` in OpenStackMachineTemplate. This needs to be fixed, right? Currently a provisioning of a cluster fails without the `identityRef.name`.

zerospiel:
In any case, introducing the removed fields back is incorrect, and targeting new templates to the `main` branch also delays the delivery

Issue #1927, opened 2025-08-29
[bug] multiclusterservice cleanup


BROngineer:
Services are not cleaned up when they are delivered via MultiClusterService and ClusterDeployment's labels does not match selector anymore.

PR #1943, opened 2025-09-02
fix: multiclusterservice cleanup

BROngineer:
This PR fixes a bug when services are not removed from cluster when labels on corresponding `ClusterDeployment` are no longer match selector in MultiClusterService

Fixes: #1927

### DIFF
--- a/internal/controller/multiclusterservice_controller.go
+++ b/internal/controller/multiclusterservice_controller.go
@@ -27,7 +27,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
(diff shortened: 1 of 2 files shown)


internal/controller/multiclusterservice_controller.go
```diff
@@ -463,3 +467,48 @@ func (r *MultiClusterServiceReconciler) createOrUpdateServiceSet(
...
+
+	var errs error
+	for _, serviceSet := range serviceSets.Items {
+		// this will happen in case the corresponding ClusterDeployment was deleted,
+		// which triggered ServiceSet deletion as
+		if !serviceSet.DeletionTimestamp.IsZero() {
+			continue
+		}
+
+		// this is a self-management ServiceSet, skipping
```
zerospiel, line 492: do i understand correctly that service sets without an explicit cluster set cannot be deleted? if so, why do such sets even have finalizers if they are unmanaged?

BROngineer: No, they can. When the owner (mcs) is being deleted such serviceSets got deleted too.

Issue #1914, opened 2025-08-28
[bug] "Reconciling MultiClusterService" loop stops, pod restart is required


denis-ryzhkov:
* This bug happens with a specific setup below, despite the previous fix https://github.com/k0rdent/kcm/pull/1899
* Steps to reproduce:
  ```
  kind create cluster -n k0rdent

  helm install -n kcm-system --create-namespace kcm \
  oci://ghcr.io/k0rdent/kcm/charts/kcm --version 1.3.0

  kubectl get mgmt -A
  # Wait for READY: True

  helm upgrade -i --reset-values --wait \
    --create-namespace -n kof kof-operators \
    oci://ghcr.io/k0rdent/kof/charts/kof-operators --version 1.3.0-rc2

  helm upgrade -i --reset-values --wait -n kof kof-mothership \
    --set kcm.installTemplates=true \
    oci://ghcr.io/k0rdent/kof/charts/kof-mothership --version 1.3.0-rc2

  kubectl get svctmpl -A
  # Wait for all VALID: true

  helm upgrade -i --reset-values --wait -n kof kof-regional \
    oci://ghcr.io/k0rdent/kof/charts/kof-regional --version 1.3.0-rc2

  kubectl logs -n kcm-system deploy/kcm-controller-manager -f | grep regional
  ```
* Expected: `Reconciling MultiClusterService` should be logged every 10 seconds without end.
* Actual: just 3 lines are logged and then the loop stops:
  ```
  {"level":"info","ts":"2025-08-28T16:04:41Z","msg":"Reconciling MultiClusterService","controller":"multiclusterservice","controllerGroup":"k0rdent.mirantis.com","controllerKind":"MultiClusterService","MultiClusterService":{"name":"kof-regional-cluster"},"namespace":"","name":"kof-regional-cluster","reconcileID":"96087d00-a870-4f11-b53c-a3be78ed3048"}

  {"level":"info","ts":"2025-08-28T16:04:41Z","msg":"Reconciling MultiClusterService","controller":"multiclusterservice","controllerGroup":"k0rdent.mirantis.com","controllerKind":"MultiClusterService","MultiClusterService":{"name":"kof-regional-cluster"},"namespace":"","name":"kof-regional-cluster","reconcileID":"74315a33-9d9b-4977-bf94-2a7dad4a1544"}

[...]

PR #1946, opened 2025-09-02
fix: enable multiclusterservice requeue on status update

BROngineer:
This PR adds `MultiClusterService` requeue on status update

Fixes: #1914

### DIFF
--- a/internal/controller/multiclusterservice_controller.go
+++ b/internal/controller/multiclusterservice_controller.go
@@ -158,7 +158,7 @@ func (r *MultiClusterServiceReconciler) reconcileUpdate(ctx context.Context, mcs
 }
 
 // updateStatus check whether status needs to be updated, if so updates the status for the MultiClusterService object
-// and returns flag whether status was updated and an error.
+// and returns a flag whether requeue should happen and an error.
 func (r *MultiClusterServiceReconciler) updateStatus(ctx context.Context, oldObj, newObj *kcmv1.MultiClusterService) (bool, error) {
 	// we'll requeue if no changes were applied to keep tracking ClusterDeployments
 	// which were created or updated.


internal/controller/multiclusterservice_controller.go
```diff
@@ -158,16 +157,24 @@ func (r *MultiClusterServiceReconciler) reconcileUpdate(ctx context.Context, mcs
...
 
-// updateStatus updates the status for the MultiClusterService object.
-func (r *MultiClusterServiceReconciler) updateStatus(ctx context.Context, mcs *kcmv1.MultiClusterService) error {
-	mcs.Status.ObservedGeneration = mcs.Generation
-	mcs.Status.Conditions = updateStatusConditions(mcs.Status.Conditions)
-
-	if err := r.Client.Status().Update(ctx, mcs); err != nil {
-		return fmt.Errorf("failed to update status for MultiClusterService %s/%s: %w", mcs.Namespace, mcs.Name, err)
+// updateStatus check whether status needs to be updated, if so updates the status for the MultiClusterService object
+// and returns flag whether status was updated and an error.
```
zerospiel:
```suggestion
// and returns a flag whether requeue should happen and an error.
```

internal/controller/multiclusterservice_controller.go
```diff
@@ -112,10 +112,9 @@ func (r *MultiClusterServiceReconciler) reconcileUpdate(ctx context.Context, mcs
 		// we need to explicitly requeue MultiClusterService object,
 		// otherwise we'll miss if some ClusterDeployment will be updated
 		// with matching labels.
-		if equality.Semantic.DeepEqual(clone.Status, mcs.Status) {
+		var requeue bool
+		if requeue, err = r.updateStatus(ctx, clone, mcs); requeue {
```
zerospiel, line 116: so we almost always requeue every 10 seconds? Won't this clog the queue? I'd suggest going with a simple poller, but i am not sure whether it's worth it to implement it to address the issue (namely, might be an overkill solution)

BROngineer: I'd consider implementing poller later. Need to check whether it will be applicable in terms of other requirements (like cleaning up services on cluster deployments which labels does not match selector anymore).

Issue #1951, opened 2025-09-02
[bug] inconsistent services status reporting


BROngineer:
Services which are being deployed, but not ready yet are not reflected in ClusterDeployment listing, despite they are reflected with state "Provisioning" in ClusterDeployment `.status.services` field.

PR #1952, opened 2025-09-03
fix: collecting ksm service statuses

BROngineer:
This PR fixes inconsistency in services status reflected in ClusterDeployment status.

Fixes: #1951

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -554,8 +554,6 @@ func (r *ClusterDeploymentReconciler) updateServices(ctx context.Context, cd *kc
 	upgradePaths, err = serviceset.ServicesUpgradePaths(ctx, r.Client, cd.Spec.ServiceSpec.Services, cd.Namespace)
 	cd.Status.ServicesUpgradePaths = upgradePaths
 	errs = errors.Join(errs, err)
-
-	errs = errors.Join(errs, r.setServicesCondition(ctx, cd))
 	return errs
 }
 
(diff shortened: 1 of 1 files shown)


internal/controller/clusterdeployment_controller.go
```diff
@@ -555,13 +555,63 @@ func (r *ClusterDeploymentReconciler) updateServices(ctx context.Context, cd *kc
...
+	}
+
+	c.Message = fmt.Sprintf("%d/%d", readyServices, totalServices)
+	apimeta.SetStatusCondition(&cd.Status.Conditions, c)
+	return nil
+}
+
 // updateStatus updates the status for the ClusterDeployment object.
 func (r *ClusterDeploymentReconciler) updateStatus(ctx context.Context, cd *kcmv1.ClusterDeployment, template *kcmv1.ClusterTemplate) error {
-	apimeta.SetStatusCondition(cd.GetConditions(), getServicesReadinessCondition(cd.Status.Services, len(cd.Spec.ServiceSpec.Services)))
```
eromanova, line 563: Looks like `getServicesReadinessCondition` is now unused.

internal/controller/clusterdeployment_controller.go
```diff
@@ -555,13 +555,63 @@ func (r *ClusterDeploymentReconciler) updateServices(ctx context.Context, cd *kc
 	cd.Status.ServicesUpgradePaths = upgradePaths
 	errs = errors.Join(errs, err)
 
+	errs = errors.Join(errs, r.setServicesCondition(ctx, cd))
```
eromanova: Wouldn’t it be better to keep the service condition update in the updateStatus function, as it was before? Since `updateStatus` always runs in defer, it guarantees execution, whereas `updateServices` may be skipped if an error occurs earlier in the controller.

internal/controller/clusterdeployment_controller.go
BROngineer: Yes, it was deleted.

eromanova: Sorry, I'm blind.

PR #1958, opened 2025-09-03
feat: add Region CRD and controller

### LINKED ISSUE #1902: Add Region CRD
TBD

### LINKED ISSUE #1904: Extract regional components to a separate helm chart
All regional core components (velero, cert-manager, CAPI operator) should be extracted from KCM into a separate Helm chart, so they can be installed and managed independently on regional clusters.

This helm chart will be a dependency of the main `kcm` helm chart.

### LINKED ISSUE #1905: Implement region controller
TBD

eromanova:
This PR adds a new Region CRD and basic code for the region controller. 
Region is a cluster-scoped object that contains all the information to adopt an existing cluster as a KCM regional cluster using kubeConfig or clusterDeployment reference.
 
In this basic implementation, the region controller supports KCM regional components installation and removal.
Removing the Region object removes all regional components from the referenced cluster.

It's not yet supported to deploy clusters or credentials in a specific region. This will be done in follow-up PRs.

### Region CRD example

```
kind: Region
metadata:
  name: ekaz-rgnl
spec:
  kubeConfig:
    key: value
    name: docker-ekaz-kubeconfig
  providers:
  - name: cluster-api-provider-k0sproject-k0smotron
  - name: cluster-api-provider-aws
  - name: projectsveltos
status:
  availableProviders:
  - bootstrap-k0sproject-k0smotron
  - control-plane-k0sproject-k0smotron
  - infrastructure-aws
  - infrastructure-internal
  - infrastructure-k0sproject-k0smotron
  capiContracts:
    bootstrap-k0sproject-k0smotron:
      v1beta1: v1beta1
    control-plane-k0sproject-k0smotron:
      v1beta1: v1beta1
    infrastructure-aws:
      v1beta1: v1beta1_v1beta2
    infrastructure-k0sproject-k0smotron:
      v1beta1: v1beta1
  components:
    capi:
      success: true
      template: cluster-api-1-0-5
    cluster-api-provider-aws:
      exposedProviders:
      - infrastructure-aws
      success: true
      template: cluster-api-provider-aws-1-0-4
    cluster-api-provider-k0sproject-k0smotron:
      exposedProviders:
      - bootstrap-k0sproject-k0smotron
      - control-plane-k0sproject-k0smotron
      - infrastructure-k0sproject-k0smotron
      success: true
      template: cluster-api-provider-k0sproject-k0smotron-1-0-7
    kcm-regional:
      success: true
      template: kcm-regional-1-0-0
    projectsveltos:
      success: true
      template: projectsveltos-0-57-2
  conditions:
  - lastTransitionTime: "2025-09-03T13:21:31Z"
    message: All components are successfully installed
    observedGeneration: 2
    reason: AllComponentsHealthy
    status: "True"
    type: Ready
```

### Additional changes:
1. Moved all components reconciliation and cleanup code out of the management controller to reuse it in the region controller.
2. Created a new `kcm-regional` template that contains all regional components (cert manager, velero, capi operator, ProviderInterface CRD). `kcm-regional` is now a dependency of the kcm helm chart.
3. Moved scheme generation to a separate package.

Fixes #1902 
Fixes #1904 
Fixes #1905


Makefile
```diff
@@ -181,7 +189,7 @@ $(IMAGES_PACKAGE_DIR): | $(LOCALBIN)
 TEMPLATE_FOLDERS = $(patsubst $(TEMPLATES_DIR)/%,%,$(wildcard $(TEMPLATES_DIR)/*))
 
 .PHONY: helm-package
-helm-package: $(CHARTS_PACKAGE_DIR) $(EXTENSION_CHARTS_PACKAGE_DIR) helm
+helm-package: $(CHARTS_PACKAGE_DIR) $(EXTENSION_CHARTS_PACKAGE_DIR) helm lint-charts
```
zerospiel:
```suggestion
helm-package: $(CHARTS_PACKAGE_DIR) $(EXTENSION_CHARTS_PACKAGE_DIR) helm
```

The target is already being called implicitly as a dependency of `package-chart-%`

api/v1beta1/region_types.go
```diff
@@ -0,0 +1,141 @@
...
+// RegionSpec defines the desired state of Region
+type RegionSpec struct {
+	// KubeConfig references the Secret containing the kubeconfig
+	// of the cluster being onboarded as a regional cluster.
+	// The Secret must reside in the system namespace.
+	KubeConfig *fluxmeta.SecretKeyReference `json:"kubeConfig,omitempty"`
+
+	// ClusterDeployment is the reference to the existing ClusterDeployment object
+	// to be onboarded as a regional cluster.
+	ClusterDeployment *corev1.ObjectReference `json:"clusterDeployment,omitempty"`
```
zerospiel: isn't it better to either create our own structure for reference of `cld` or simplify with just a namespaced name? i believe the latter should suffice

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -115,6 +116,7 @@ rules:
   - k0rdent.mirantis.com
   resources:
   - managements/status
+  - regions/status
```
zerospiel, line 132: shouldn't the `finalizers` "subresource" be added above as well?

zerospiel (commented):
This is a partial review; I haven't reviewed the main logic yet. Additional question: could we introduce new controllers in their own packages within `internal/controller`? Otherwise, the `controller` package would become unmanageable someday

internal/controller/region/controller.go
```diff
@@ -0,0 +1,398 @@
...
+	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+	"github.com/K0rdent/kcm/internal/controller/components"
+	"github.com/K0rdent/kcm/internal/record"
+	"github.com/K0rdent/kcm/internal/utils"
+	"github.com/K0rdent/kcm/internal/utils/ratelimit"
+	schemeutil "github.com/K0rdent/kcm/internal/utils/scheme"
+)
+
+// RegionReconciler reconciles a Region object
+type RegionReconciler struct {
```
zerospiel: style: to avoid `region.RegionReconciler` one could rename `RegionReconciler` to `Reconciler` but it's up to the implementer

internal/controller/region/controller.go
```diff
@@ -0,0 +1,398 @@
...
+	if err := r.MgmtClient.Get(ctx, secretObj, secret); err != nil {
+		return nil, fmt.Errorf("failed to get Secret with kubeconfig: %w", err)
+	}
+
+	kubeconfigBytes, ok := secret.Data[kubeconfigKey]
+	if !ok {
+		return nil, fmt.Errorf("kubeconfig from Secret %s is empty", secretObj)
+	}
+	return kubeconfigBytes, nil
+}
```
zerospiel: probably nit-picking: there is `getChildClient` in the `internal/telemetry/collector/client.go` that potentially could be reused (with a proper rename and changing signature also to return the bytes) since the function virtually does the same. Up to the implementer

internal/controller/region/controller.go
```diff
@@ -0,0 +1,398 @@
...
+	DynamicClient          *dynamic.DynamicClient
+	SystemNamespace        string
+	GlobalRegistry         string
+	RegistryCertSecretName string // Name of a Secret with Registry Root CA with ca.crt key; used by RegionReconciler
+
+	DefaultHelmTimeout time.Duration
+	defaultRequeueTime time.Duration
+
+	CreateAccessManagement bool
+	IsDisabledValidationWH bool // is webhook disabled set via the controller flags
```
zerospiel: these both look unused

internal/controller/region/controller.go
```diff
@@ -0,0 +1,398 @@
...
+	if err := r.MgmtClient.List(ctx, &hrList, listOpts...); err != nil {
+		return false, fmt.Errorf("failed to list %s: %w", helmcontrollerv2.GroupVersion.WithKind(helmcontrollerv2.HelmReleaseKind), err)
+	}
+
+	// We should ensure the removal order according to helm release dependencies
+	dependents := make(map[string]map[string]struct{})
+	for _, hr := range hrList.Items {
+		if dependents[hr.Name] == nil {
+			dependents[hr.Name] = make(map[string]struct{})
+		}
```
zerospiel: looks redundant (map creation is done right below, having no such `hr.Name` keys)

internal/controller/region/controller.go
```diff
@@ -0,0 +1,398 @@
...
+
+		l.V(1).Info("Deleting HelmRelease", "name", hr.Name)
+		r.eventf(region, "RemovingComponent", "Removing %s HelmRelease", hr.Name)
+
+		hrCopy := &helmcontrollerv2.HelmRelease{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      hr.Name,
+				Namespace: r.SystemNamespace,
+			},
+		}
```
zerospiel: why can't just `hr` variable be used in the delete call?

internal/controller/region/controller.go
```diff
@@ -0,0 +1,398 @@
...
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *RegionReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.defaultRequeueTime = 10 * time.Second
+
+	return ctrl.NewControllerManagedBy(mgr).
+		WithOptions(controller.TypedOptions[ctrl.Request]{
+			RateLimiter: ratelimit.DefaultFastSlow(),
+		}).
+		For(&kcmv1.Region{}).
```
zerospiel: generic question: are we okay with events on status updates? meaning, won't we do redundant loops on the event of the status change?

internal/helm/release.go
```diff
@@ -62,6 +65,9 @@ func ReconcileHelmRelease(ctx context.Context,
 			hr.Labels = make(map[string]string)
 		}
 		hr.Labels[kcmv1.KCMManagedLabelKey] = kcmv1.KCMManagedLabelValue
+		for k, v := range opts.Labels {
+			hr.Labels[k] = v
+		}
```
zerospiel:
```suggestion
		maps.Copy(hr.Labels, opts.Labels)
```

api/v1beta1/region_types.go
```diff
@@ -0,0 +1,146 @@
...
+const (
+	RegionKind = "Region"
+
+	RegionFinalizer = "k0rdent.mirantis.com/region"
+
+	KCMRegionLabelKey = "k0rdent.mirantis.com/region"
+)
+
+// +kubebuilder:validation:XValidation:rule="has(self.kubeConfig) != has(self.clusterDeployment)",message="exactly one of kubeConfig or clusterDeployment must be set"
+
```
zerospiel, line 31: Not true, the godoc's description would be then showed with this comment which we want to avoid

internal/controller/region/controller.go
eromanova: Sure, this was unintentionally left in from the previous implementation

internal/controller/region/controller.go
eromanova: Sure, this was unintentionally left in from the previous implementation

internal/utils/scheme/scheme.go
```diff
@@ -0,0 +1,107 @@
...
+		infobloxv1alpha1.AddToScheme,
+	} {
+		if err := f(s); err != nil {
+			return nil, fmt.Errorf("failed to add to scheme: %w", err)
+		}
+	}
+	return s, nil
+}
+
+func GetChildScheme() (*runtime.Scheme, error) {
```
zerospiel: this might be a bit confusing, because i would say, that a child cluster could be flexible and the scheme might vary depending on a specific case. That's being said, I'd put it back to the `telemetry/collector` package

PR #1966, opened 2025-09-05
chore: add make dev-upgrade target to upgrade dev kcm

### LINKED ISSUE #1700: Add `make upgrade` for KOF CI to test against KCM `main` safely
* On release of 1.1.0 KOF faced a blocker with `k0rdent/catalog` expecting KCM API `v1alpha1` already deleted in KCM - https://github.com/k0rdent/catalog/issues/468#issuecomment-3004582639
* @va4st suggested to update KOF CI to test against KCM `main` "so issues like `v1alpha1` api drop will be captured right after it will appear."
* @gmlexx suggested to ask KCM to add `make upgrade` we could use in KOF CI to avoid blocking KOF PR merges because there are some manual upgrade steps needed in KCM.

eromanova:
This PR introduces a `dev-upgrade` Makefile target that performs an upgrade of the dev environment (originally set up with `make dev-apply`).

Example workflow: upgrading from `v1.3.1` to the current main:
```
git checkout v1.3.1
make dev-apply
kubectl wait management kcm --for=condition=Ready=True --timeout 10m
git checkout main
make dev-upgrade
```

Fixes #1700

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -445,6 +445,22 @@ dev-gcp-creds: envsubst
 .PHONY: dev-apply
 dev-apply: kind-deploy registry-deploy dev-push dev-deploy dev-templates dev-release ## Apply the development environment by deploying the kind cluster, local registry and the KCM helm chart.
 
+.PHONY: dev-upgrade
+dev-upgrade: generate-all dev-push dev-templates ## Upgrade dev environment and wait until Management is upgraded to the new Release and ready
+	@echo "Applying new Release object: kcm-$(FQDN_VERSION)"
+	@@$(YQ) e ".spec.version = \"${VERSION}\" | .metadata.name = \"kcm-$(FQDN_VERSION)\"" $(PROVIDER_TEMPLATES_DIR)/kcm-templates/files/release.yaml | $(KUBECTL) apply -f -
+	@echo "Waiting for Release kcm-$(FQDN_VERSION) to become Ready..."
+	@$(KUBECTL) wait release "kcm-$(FQDN_VERSION)" --for='jsonpath={.status.ready}=true' --timeout 5m
+	@echo "Patching Management object to use Release: kcm-$(FQDN_VERSION)"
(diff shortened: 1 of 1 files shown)


s3rj1k:
Simple `kubectl wait --for=condition=Ready=True management/kcm` does not work?

Makefile
```diff
@@ -445,6 +445,22 @@ dev-gcp-creds: envsubst
...
+.PHONY: dev-upgrade
+dev-upgrade: generate-all dev-push dev-templates ## Upgrade dev environment and wait until Management is upgraded to the new Release and ready
+	@echo "Applying new Release object: kcm-$(FQDN_VERSION)"
+	@@$(YQ) e ".spec.version = \"${VERSION}\" | .metadata.name = \"kcm-$(FQDN_VERSION)\"" $(PROVIDER_TEMPLATES_DIR)/kcm-templates/files/release.yaml | $(KUBECTL) apply -f -
+	@echo "Waiting for Release kcm-$(FQDN_VERSION) to become Ready..."
+	@$(KUBECTL) wait release "kcm-$(FQDN_VERSION)" --for='jsonpath={.status.ready}=true' --timeout 5m
+	@echo "Patching Management object to use Release: kcm-$(FQDN_VERSION)"
+	@$(KUBECTL) patch management kcm --type=merge -p '{"spec":{"release":"kcm-$(FQDN_VERSION)"}}'
+	@echo "Sleeping 30s to allow Management status to update.."
+	@sleep 30
```
zerospiel, line 457: nit: seems redundant considering the conditional wait

eromanova: The issue is that there’s a point in time when the Management object already reports the new release in its status, but the readiness condition hasn’t been updated yet (it still shows `Ready` from the previous state). As a result, the `dev-upgrade` target finishes too early without actually waiting for readiness.

This is likely a kcm-controller issue, since the new release shouldn’t be reflected in the status until all the components are updated.

zerospiel: Got it. Then please create a `Bug`-typed issue, otherwise LGTM

zerospiel (approved):

Issue #1967, opened 2025-09-05
[bug] Service update in CD not reflected in ServiceSet


wahabmk:
**Describe the bug**

**To Reproduce**
1. <step 1>
2. <step 2>
3. <step 3>

**Expected behavior**

**Additional context**

wahabmk:
Also verify the same behavior for MultiClusterService.

PR #1968, opened 2025-09-08
feat: implementation & validation of service dependencies

wahabmk:
# Description

1. Add `.serviceSpec.services[].dependsOn` field through which users can specify which other services a particular service depends on. If service B depends on A, then KSM will not attempt to deploy B until A has been successfully deployed.
2. Add webhook validation for depends on which verifies that:
    2a. All services being depended on are defined in `.serviceSpec.services[]` themselves.
    2b. There is no cyclic dependency.

# Scenarios Validated

For these scenarios:

- A and B are services.
- B -> A means B depends on A.

## For ClusterDeployment

**1. CD has A, then it is edited to have a newer version of A.**
❌ What should have happened is that A should have been upgraded to the newer version of A, but what is actually happening is that CD spec shows A is updated to newer version but the ServiceSet spec still shows older version of A. This is actually a pre-existing bug. https://github.com/k0rdent/kcm/issues/1967

**2. CD has B -> A, then MCD with higher priority with newer version of B is created.**
✅ Everything remains same and only B is upgraded to newer version.

**3. CD has A, then MCS with higher priority with B is created, then CD is edited to have B -> A with newer version of B.**
✅ Since CD has lower priority, B is not upgraded.
**3a. Now make CD higher priority**
✅ Now B is upgraded to newer version.
**3b. Instead of doing 3a, delete the MCS instead.**
✅ Again B is upgraded to newer version as there is no higher priority MCS anymore.

**4. CD has B, then B -> A is added.**
✅ First B is removed, then A is deployed, then B is deployed again.

**5. CD has B, then MCS with higher priority with same version of B is created, then CD is edited to have B -> A with newer version of B.**
✅ First B is deployed, then as MCS has higher priority it will take over B by deleting and then re-deploying it, then A is deployed (although required a re-trigger of reconciler, figure out why) & B wasn't deleted and re-deployed as in scenario 4 because this time MCS with higher priority was already managing it.

## For MultiClusterService

**1. MCS matching CD has A, then it is upgraded to newer version of A.**
❌ https://github.com/k0rdent/kcm/issues/1967

**2. MCS matching CD has B -> A.**
✅ First A is deployed on CD then B.

**3. CD already has A, then MCS with lower priority with B -> A matching CD is created.**
✅ Since CD already has A, MCS simply deploys B and does not attempt to take over A by re-deploying it because it has lower priority.

**4. CD already has A, then MCS with higher priority with B -> A matching CD is created.**
✅ Since CD already has A, MCS deploys both its A and B at the same time. MCS deploys A again (while CD's A is deleted) because it has higher priority. Strictly speaking MCS should deploy A before B (not at the same time), but the reason it does so is because there is already an A deployed on CD so technically B can be immediately deployed. https://github.com/k0rdent/kcm/issues/1970 has been created to consider priorities as well when resolving dependencies.


internal/serviceset/util_test.go
```diff
@@ -222,3 +228,135 @@ func Test_ServicesToDeploy(t *testing.T) {
...
+	for _, tc := range []struct {
+		testName        string
+		cdNamespace     string
+		cdName          string
+		desiredServices []kcmv1.Service
+		objects         []client.Object
+		expected        []kcmv1.Service
+	}{
+		{
+			testName:    "test0",
```
zerospiel: style: names could've been more descriptive, I suppose

unrelated: not sure whether the `TODO` by the end of the list is still applicable

internal/webhook/multiclusterservice_webhook.go
```diff
@@ -72,6 +80,14 @@ func (v *MultiClusterServiceValidator) ValidateUpdate(ctx context.Context, _, ne
 		return nil, fmt.Errorf("%s: %w", invalidMultiClusterServiceMsg, err)
 	}
 
+	if err := validation.ValidateServiceDependency(mcs.Spec.ServiceSpec.Services); err != nil {
```
zerospiel: are these validations handled in case of disabled validations? e.g. if I had disabled validations, and something is off with the services (in terms of dependency cycles), should we continue reconciliation loops (will that even work), or should we stop reconciliations (set condition/event and stop events) until the spec is fixed? If the latter, then I suppose this is yet to be done

internal/webhook/clusterdeployment_webhook.go
```diff
@@ -128,6 +136,14 @@ func (v *ClusterDeploymentValidator) ValidateUpdate(ctx context.Context, oldObj,
 		return nil, fmt.Errorf("%s: %w", invalidClusterDeploymentMsg, err)
 	}
 
+	if err := validation.ValidateServiceDependency(newClusterDeployment.Spec.ServiceSpec.Services); err != nil {
+		return nil, fmt.Errorf("%s: %w", invalidMultiClusterServiceMsg, err)
+	}
+
+	if err := validation.ValidateServiceDependencyCycle(newClusterDeployment.Spec.ServiceSpec.Services); err != nil {
```
zerospiel: are these validations handled in case of disabled validations? e.g. if I had disabled validations, and something is off with the services (in terms of dependency cycles), should we continue reconciliation loops (will that even work), or should we stop reconciliations (set condition/event and stop events) until the spec is fixed? If the latter, then I suppose this is yet to be done

internal/utils/validation/services_test.go
```diff
@@ -0,0 +1,227 @@
...
+)
+
+func TestValidateServiceDependency(t *testing.T) {
+	for _, tc := range []struct {
+		name        string
+		services    []kcmv1.Service
+		expectedErr string
+	}{
+		{
+			name: "test1",
```
zerospiel: Ditto with the cases naming here and below

zerospiel (commented):
LGTM, minor comments on tcases names and the question regarding the turned off validation

Issue #1923, opened 2025-08-29
[enhancement] allow to define helm options for service backed by Helm chart


BROngineer:
ProjectSveltos, as a default CD driver, allows to define helm arguments which will be applied during Helm chart deployment. We need to add support for it.

**DoD**:
- [ ] it is possible to pass helm options to underlying sveltos Profile (ref: https://github.com/projectsveltos/addon-controller/blob/f68eeddc284cc5ce49e22f9d29cc37780d5dc1be/api/v1beta1/spec.go#L150)
- [ ] scenario covered in unit/integration tests
- [ ] documentation updated

PR #1969, opened 2025-09-08
feat: add helm options to service templates

kylewuolle:
This PR adds helm options to the `ServiceTemplate` CRD as well as to the `ClusterDeployment`/`MultiClusterService` CRDs. In the case where the `ServiceTemplate` and  `ClusterDeployment`/`MultiClusterService` both have helm options the ones from the  `ClusterDeployment`/`MultiClusterService` CRD are used/merged together. 

#1923


api/v1beta1/multiclusterservice_types.go
```diff
@@ -104,6 +107,80 @@ type Service struct {
...
+
+	// +optional
+
+	// Replaces if set indicates to replace an older release with this one
+	Replace *bool `json:"replace,omitempty"`
+
+	// +optional
+
+	// Labels that would be added to release metadata.
+	Labels *map[string]string `json:"labels,omitempty"`
```
zerospiel, line 176: Why should this field hold a reference to a reference?

api/v1beta1/multiclusterservice_types.go
```diff
@@ -104,6 +107,80 @@ type Service struct {
...
+
+	// +optional
+
+	// SkipSchemaValidation determines if JSON schema validation is disabled.
+	SkipSchemaValidation *bool `json:"skipSchemaValidation,omitempty"`
+
+	// +optional
+
+	// Replaces if set indicates to replace an older release with this one
+	Replace *bool `json:"replace,omitempty"`
```
zerospiel, line 171: Do we strictly have 3 states to distinguish from each other? If yes, what are they? As far as my understanding goes, there are only 2 states (true/false) without `null`, whereas the latter is considered as `false` (according to the `convertHelmOptions`)

api/v1beta1/multiclusterservice_types.go
```diff
@@ -104,6 +107,80 @@ type Service struct {
...
+
+	// +optional
+
+	// Labels that would be added to release metadata.
+	Labels *map[string]string `json:"labels,omitempty"`
+
+	// +optional
+
+	// Description is the description of an helm operation
+	Description *string `json:"description,omitempty"`
```
zerospiel, line 181: Same here: why do we have to hold a reference to a reference? According to the `convertHelmOptions`, this is also not a hard requirement, considering this struct a copy of the upstream (as of now)

api/v1beta1/multiclusterservice_types.go
```diff
@@ -80,6 +80,9 @@ type Service struct {
 	// The string type is used in order to allow for templating.
 	Values string `json:"values,omitempty"`
 
+	// HelmOptions are the options to be passed to the provider for helm installation or updates
+	HelmOptions ServiceHelmOptions `json:"helmOptions,omitempty"`
```
zerospiel: On the other hand, making a reference here makes sense (easily distinguish "empty" from "not set" and effectively decreasing the size of the struct ~2 times)

api/v1beta1/serviceset_types.go
```diff
@@ -124,6 +124,9 @@ type StateManagementProviderConfig struct {
 }
 
 type ServiceWithValues struct {
+	// HelmOptions are the options to be passed to the provider for helm installation or updates
+	HelmOptions ServiceHelmOptions `json:"helmOptions,omitempty"`
```
zerospiel: On the other hand, making a reference here makes sense (easily distinguish "empty" from "not set" and effectively decreasing the size of the struct ~2 times)

api/v1beta1/multiclusterservice_types.go
kylewuolle: The idea is to allow a user to override existing values. The HelmOptions in Sveltos have some defaults already and by making these pointers we make sure the defaults are used unless overridden. We're setting up a hierarchy of settings here: Sveltos->Global `HelmOptions` on `ServiceTemplate`->HelmOptions on the `Service` instance

zerospiel: Thanks for the explanation. I still don't quite get it, and I'm not in the loop, hence let's consider the example with the `replace` field:
1. So there is `replace` set to `true` in a sveltos' object (for instance)
2. A user of `servicetemplate` hasn't set it, hence it is `false`
3. We consider `servicetemplate` options authoritative, right?
4. Due to 3, after the merge, the options should then hold `false`, which is expected, right?

So if the example is correct, then I still do not see the hard requirement for the `bool`-typed fields.

I think that also works if in sveltos we have `false` (or unset, which is also `false`), and the user has set `true` in a `servicetemplate`, we also should get `true` after the merge, effectively transmitting that to sveltos

zerospiel: Oh, right after the message above, I realized that:
1. my example actually works only if the value has been actually set to `false`
2. The example won't work if "unset"

That being said, I understand the purpose.

I would suggest setting up the default value `false` just so the object has the value set after schema validation, that way we can ditch handling the references.

api/v1beta1/multiclusterservice_types.go
zerospiel: @kylewuolle please once again do not resolve comments until they are actually resolved. Why do we need the reference here? I understand that highly likely this is the same case as with a book typed field (but I cannot prove it at the moment, could you check just in case?). I mean cannot it just be defaulted to an empty value? In that case we don't have to handle references. Anyway, it's up to you @BROngineer and @wahabmk  since it is your API. But until it's in our repo, could you folks take in accounting our comments regarding the API just in case?

api/v1beta1/multiclusterservice_types.go
kylewuolle: The issue is say if you have a ServiceTemplate like this:

```
kind: ServiceTemplate
metadata:
  annotations:
    helm.sh/resource-policy: keep
    meta.helm.sh/release-name: ingress-nginx
    meta.helm.sh/release-namespace: kcm-system
  labels:
  name: ingress-nginx-4-11-4
  namespace: kcm-system
spec:
  helm:
    chartSpec:
      chart: ingress-nginx
      interval: 10m0s
      reconcileStrategy: ChartVersion
      sourceRef:
        kind: HelmRepository
        name: k0rdent-catalog
      version: 4.11.3
  helmOptions:
    labels:
       test: "label"
    enableClientCache: true
    dependencyUpdate: true
```
Then you create a ClusterDeployment like this:

```
kind: ClusterDeployment
metadata:
  name: test
  namespace: kcm-system
spec:
  config:
    controlPlane:
      instanceType: t3.small
      rootVolumeSize: 40
    controlPlaneNumber: 1
    publicIP: false
    region: us-east-1
    worker:
      amiID: null
      instanceType: t3.small
      rootVolumeSize: 40
    workersNumber: 1
  ipamClaim: {}
  propagateCredentials: true
  serviceSpec:
    provider:
      name: ksm-projectsveltos
    services:
    - name: managed-ingress-nginx
      namespace: default
      template: ingress-nginx-4-12-1
      enableClientCache: false
  template: aws-standalone-cp-1-0-14
```

If we don't use pointers in the HelmOptions fields we'll get all of the fields filled in even when not specified and we'll have no idea what to overwrite. In essence the ServiceTemplate fields will always be overwritten if you specify any helm option fields at all in ClusterDeployment/services.

Issue #1839, opened 2025-08-07
Make templates jsonschema available through API


Kshatrix:
Right now, you have to download the chart associated with the template and extract the JSON schema from it. To make it more easily accessible, we need to store it in a ConfigMap and reference it in the template’s status.

PR #1972, opened 2025-09-10
feat: add JSON Schema configmap for templates if available

Kshatrix:
Added the `schemaConfigMapName` field to the template’s status and enabled the generation of the ConfigMap. If the referenced Helm chart includes values.schema.json (as it normally should), the schema will be available in the ConfigMap.

Resolves: #1839

### DIFF
--- a/internal/controller/template_controller.go
+++ b/internal/controller/template_controller.go
@@ -26,6 +26,7 @@ import (
 	helmcontrollerv2 "github.com/fluxcd/helm-controller/api/v2"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	"helm.sh/helm/v3/pkg/chart"
+	corev1 "k8s.io/api/core/v1"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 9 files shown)


internal/controller/template_controller.go
```diff
@@ -301,11 +306,60 @@ func (r *TemplateReconciler) ReconcileTemplate(ctx context.Context, template tem
...
+func (r *TemplateReconciler) ensureSchemaConfigmap(ctx context.Context, template templateCommon, helmChart *chart.Chart) error {
+	if helmChart.Schema == nil {
+		return nil
+	}
+
+	ownerRef := metav1.NewControllerRef(template, template.GetObjectKind().GroupVersionKind())
+	schemaConfigMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            generateSchemaConfigMapName(template),
+			Namespace:       template.GetNamespace(),
```
eromanova: ProviderTemplates are cluster-scoped. So, the Namespace will be empty. In that case, should we use the system namespace?

internal/controller/template_controller.go
```diff
@@ -301,11 +306,64 @@ func (r *TemplateReconciler) ReconcileTemplate(ctx context.Context, template tem
...
+	case kcmv1.ProviderTemplateKind:
+		templateTypePrefix = "pt"
+	case kcmv1.ServiceTemplateKind:
+		templateTypePrefix = "st"
+	}
+	return fmt.Sprintf("schema-%s-%s", templateTypePrefix, template.GetName())
+}
+
+func (r *TemplateReconciler) ensureSchemaConfigmap(ctx context.Context, template templateCommon, helmChart *chart.Chart) error {
+	if helmChart.Schema == nil {
```
zerospiel:
```suggestion
	if len(helmChart.Schema) == 0 {
```

nit: the array might be non-nil though empty

eromanova (approved):

PR #1973, opened 2025-09-10
TBD

zerospiel:


internal/controller/clusterdeployment_controller.go
```diff
@@ -719,6 +744,84 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, cd *k
...
+	if !kube.IsAPIServerReady(ctx, restCfg, readinessTimeout) {
+		// server is not ready, nothing to do
+		return false, nil
+	}
+
+	const deletionTimeout = 10 * time.Second // magic number
+	eg, gctx := errgroup.WithContext(ctx)
+	now := time.Now()
+	eg.Go(func() error {
+		exluded := []string{metav1.NamespaceSystem, metav1.NamespaceDefault}
```
eromanova: Why are these namespaces excluded? For instance, if I deploy a service with a LoadBalancer in the default namespace, it won’t be cleaned up. Is it expected?

internal/controller/clusterdeployment_controller.go
```diff
@@ -719,6 +744,84 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, cd *k
...
+		exist, err := r.existsAnyExludingNamespaces(gctx, cl, &corev1.PersistentVolumeClaimList{}, exluded)
```
eromanova:
```suggestion
		exist, err := r.existsAnyExludingNamespaces(gctx, cl, &corev1.ServiceList{}, exluded)
```
The error message points to Services while `&corev1.PersistentVolumeClaimList{}` is used as a parameter.

internal/controller/clusterdeployment_controller.go
```diff
@@ -719,6 +744,84 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, cd *k
...
+			return true, nil // requeue
+		}
+
+		return false, err // already wrapped
+	}
+
+	return false, nil
+}
+
+func (*ClusterDeploymentReconciler) existsAnyExludingNamespaces(ctx context.Context, c client.Client, list client.ObjectList, excludeNS []string) (bool, error) {
```
eromanova:
```suggestion
func (*ClusterDeploymentReconciler) existsAnyExcludingNamespaces(ctx context.Context, c client.Client, list client.ObjectList, excludeNS []string) (bool, error) {
```

internal/controller/clusterdeployment_controller.go
zerospiel: The only reason is not to drop `kubernetes` `Service`. Since I use `deletecollection` I have decided to go this way. If that is not the best approach, I could use the good old `delete` for each item from a list. What would you suggest?

internal/controller/clusterdeployment_controller.go
zerospiel: Fixed

internal/controller/clusterdeployment_controller.go
zerospiel: Fixed, thanks

internal/controller/clusterdeployment_controller.go
eromanova: Can we restrict cleanup to only LoadBalancer type Services across all namespaces? I guess these are the only Services that can leave orphaned cloud load balancer resources.

zerospiel: Good idea, it seems I have to change the function into 2 separate functions in that way. I will try to handle it. Thanks for the suggestion

zerospiel: Addressed in [this commit](https://github.com/k0rdent/kcm/commit/95367dcf89e2e3cad762f40dcb40efeb129f61c1) ([diff](https://github.com/k0rdent/kcm/compare/65ab2846e360d1a993e2cd37a62e809dc580767b..95367dcf89e2e3cad762f40dcb40efeb129f61c1))

PR #1980, opened 2025-09-11
feat: introduce Region field for Credential

### LINKED ISSUE #1903: Adapt existing API with regions
1. ClusterDeployment spec should be extended to support the `region` field
2. Credential spec should be extended to support `region` field

### LINKED ISSUE #1979: Adapt kcm controller code to use the regional client
Adapt kcm controller code to use the regional client when accessing regional cluster resources (SveltosCluster, Cluster, etc).

eromanova:
* Add `spec.region` to Credential
* Adapt ClusterDeployment reconciler to use regional client
* Adapt Credential reconciler to use regional client

Fixes #1903
Partially fixes #1979

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -882,7 +882,7 @@ func (*ClusterDeploymentReconciler) existsAnyExcludingNamespaces(ctx context.Con
 	return kube.ExistsAny(ctx, c, list, client.MatchingFieldsSelector{Selector: sel})
 }
 
-func getProviderGVKs(ctx context.Context, rgnClient client.Client, name string) []schema.GroupVersionKind {
+func (*ClusterDeploymentReconciler) getProviderGVKs(ctx context.Context, rgnClient client.Client, name string) []schema.GroupVersionKind {
 	providerInterfaces := &kcmv1.ProviderInterfaceList{}
 
 	if err := rgnClient.List(ctx, providerInterfaces,
@@ -917,12 +917,12 @@ func (r *ClusterDeploymentReconciler) releaseProviderCluster(ctx context.Context
 
 	// Associate the provider with it's GVK
 	for _, provider := range providers {
-		gvks := getProviderGVKs(ctx, scope.rgnClient, provider)
+		gvks := r.getProviderGVKs(ctx, scope.rgnClient, provider)
 		if len(gvks) == 0 {
 			continue
 		}
 
-		cluster, err := getProviderCluster(ctx, scope.rgnClient, cd.Namespace, cd.Name, gvks...)
+		cluster, err := r.getProviderCluster(ctx, scope.rgnClient, cd.Namespace, cd.Name, gvks...)
 		if err != nil {
 			if !errors.Is(err, errClusterNotFound) {
 				return err
@@ -930,13 +930,13 @@ func (r *ClusterDeploymentReconciler) releaseProviderCluster(ctx context.Context
 			return nil
 		}
 
-		found, err := clusterCAPIMachinesExist(ctx, scope.rgnClient, cd.Namespace, cluster.Name)
+		found, err := r.clusterCAPIMachinesExist(ctx, scope.rgnClient, cd.Namespace, cluster.Name)
 		if err != nil {
 			continue
 		}
 
 		if !found {
-			finalizersUpdated, err := removeClusterFinalizer(ctx, scope.rgnClient, cluster)
+			finalizersUpdated, err := r.removeClusterFinalizer(ctx, scope.rgnClient, cluster)
 			if finalizersUpdated {
 				r.eventf(cd, "ClusterDeleted", "Cluster %s has been deleted", client.ObjectKeyFromObject(cd))
 			}
@@ -972,7 +972,7 @@ func (r *ClusterDeploymentReconciler) getInfraProvidersNames(ctx context.Context
 }
 
 // getProviderCluster fetches a first provider Cluster from the given list of GVKs.
-func getProviderCluster(ctx context.Context, rgnClient client.Client, namespace, name string, gvks ...schema.GroupVersionKind) (*metav1.PartialObjectMetadata, error) {
+func (*ClusterDeploymentReconciler) getProviderCluster(ctx context.Context, rgnClient client.Client, namespace, name string, gvks ...schema.GroupVersionKind) (*metav1.PartialObjectMetadata, error) {
 	for _, gvk := range gvks {
 		itemsList := &metav1.PartialObjectMetadataList{}
 		itemsList.SetGroupVersionKind(gvk)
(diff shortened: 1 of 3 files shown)


internal/controller/region/client.go
```diff
@@ -0,0 +1,87 @@
...
+	fluxmeta "github.com/fluxcd/pkg/apis/meta"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+	"github.com/K0rdent/kcm/internal/utils/kube"
+	schemeutil "github.com/K0rdent/kcm/internal/utils/scheme"
+)
+
+func GetKubeConfigSecretRef(region *kcmv1.Region) (*fluxmeta.SecretKeyReference, error) {
```
zerospiel, line 33: style: functions comments are very welcome

internal/controller/clusterdeployment_controller.go
```diff
@@ -1144,7 +1187,7 @@ func (r *ClusterDeploymentReconciler) handleCertificateSecrets(ctx context.Conte
 
 	l.V(1).Info("Copying certificate secrets from the system namespace to the ClusterDeployment namespace")
 	for _, secretName := range secretsToHandle {
-		if err := utils.CopySecret(ctx, r.Client, client.ObjectKey{Namespace: r.SystemNamespace, Name: secretName}, cd.Namespace); err != nil {
+		if err := utils.CopySecret(ctx, r.rgnClient, client.ObjectKey{Namespace: r.SystemNamespace, Name: secretName}, cd.Namespace); err != nil {
```
zerospiel: Would appreciate elaboration on why we use the regional client here, it's not obvious at first glance. Is it because these objects aren't k0rdent CRs and hence must be on the regional?

internal/controller/clusterdeployment_controller.go
```diff
@@ -833,7 +876,7 @@ func (*ClusterDeploymentReconciler) existsAnyExcludingNamespaces(ctx context.Con
 func (r *ClusterDeploymentReconciler) getProviderGVKs(ctx context.Context, name string) []schema.GroupVersionKind {
 	providerInterfaces := &kcmv1.ProviderInterfaceList{}
 
-	if err := r.Client.List(ctx, providerInterfaces,
+	if err := r.rgnClient.List(ctx, providerInterfaces,
```
zerospiel: Would appreciate elaboration on why we use the regional client here, it's not obvious at first glance

internal/controller/clusterdeployment_controller.go
```diff
@@ -761,7 +799,12 @@ func (r *ClusterDeploymentReconciler) deleteChildResources(ctx context.Context,
 	factory, restCfg := kube.DefaultClientFactoryWithRestConfig()
 
 	secretRef := client.ObjectKeyFromObject(cd)
-	cl, err := kube.GetChildClient(ctx, r.Client, secretRef, "value", r.Client.Scheme(), factory)
+	childScheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(childScheme); err != nil {
+		return false, fmt.Errorf("failed to add corev1 to child scheme: %w", err)
+	}
+
+	cl, err := kube.GetChildClient(ctx, r.rgnClient, secretRef, "value", childScheme, factory)
```
zerospiel: nit (up to the implementer): we could reuse the scheme from either the mgmt or the regional clients, and just drop the comment that we only need `corev1` which is already presented there. That way we avoid this extra step. Though, for the sake of clarity, the current block is 100% correct

internal/controller/clusterdeployment_controller.go
```diff
@@ -306,24 +311,14 @@ func (r *ClusterDeploymentReconciler) updateCluster(ctx context.Context, cd *kcm
...
-			UID:        cd.UID,
-		},
-		ChartRef: clusterTpl.Status.ChartRef,
-		Timeout:  r.DefaultHelmTimeout,
-	}
-	if clusterTpl.Spec.Helm.ChartSpec != nil {
-		hrReconcileOpts.ReconcileInterval = &clusterTpl.Spec.Helm.ChartSpec.Interval.Duration
+	hrReconcileOpts, err := r.getHelmReleaseReconcileOpts(ctx, cd, clusterTpl)
+	if err != nil {
+		return ctrl.Result{}, err
```
zerospiel: style: will it be okay to wrap an error?

internal/controller/clusterdeployment_controller.go
```diff
@@ -113,13 +113,29 @@ func (r *ClusterDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		err = fmt.Errorf("failed to get Credential %s: %w", credKey, err)
+		if r.setCondition(clusterDeployment, kcmv1.CredentialReadyCondition, err) {
+			r.warnf(clusterDeployment, "CredentialError", err.Error())
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to get Credential %s: %w", credKey, err)
+	}
+	r.region = cred.Spec.Region
+
+	var err error
+	if r.rgnClient, err = region.GetClientFromRegionName(ctx, r.MgmtClient, r.SystemNamespace, cred.Spec.Region); err != nil {
```
zerospiel: question: I wasn't sure how to deduplicate the work, but it might be worth it to save not only the region's name but rather the whole object? At the moment we `get` the `Region` object here and within the `getKubeConfigSecretRef` method on line `#395`. I understand that this might not be the best idea, considering the object might change during the process. So just consider this as an open conversation. Is it worth it? May the `Region` change its `spec`?

internal/controller/clusterdeployment_controller.go
eromanova: ProviderInterfaces at the moment are the only objects from the k0rdent API that became a part of the regional API (it's a part of provider templates and provider templates are installed on regional clusters). That's why we should use the regional client to access them.

internal/controller/clusterdeployment_controller.go
eromanova: When a cluster is created in a specific region, the `k0sURLCertificate` and `registryCertificate` secrets must exist in the corresponding regional clusters, since CAPI objects (e.g., `K0smotronControlPlane`, `K0sControlPlane`) are created there and depend on these secrets.
It's not implemented yet, but these secrets should be copied to all regions (if configured).

internal/controller/clusterdeployment_controller.go
eromanova: > May the `Region` change its `spec`?

It’s worth to forbid changes to the kubeconfig secret reference in the `Region` spec, while still allowing updates to the list of providers.
I’ll check if this can be refactored. Thanks.

zerospiel: If that won't be like just a couple LOC, then I do not think such an optimization is worth it

PR #1996, opened 2025-09-15
fix: drop `regional` section from Release

eromanova:
Introducing the `spec.regional` field in the Release object caused upgrade issues, since the updated Release CRD must be applied before applying the new Release object with this field. In the previous implementation, this made the upgrade path 1.3.1 → 1.4.0 unavailable.

As a workaround, the `spec.regional` field was removed, and the `kcm-regional` Helm chart version is now tied directly to the `kcm` version.

Additional changes:
1. Restart kcm controller before waiting for Management readiness in the `make dev-upgrade` target.

Related to #1903


zerospiel (approved):

zerospiel:
> These versions are bumped automatically before the release

I might've missed it, but I am not aware of this automation, just in case. Would appreciate the link to the automation.

eromanova:
@zerospiel It isn’t handled automatically in the code but before a release, the `kcm-chart-release` Makefile target is executed: https://github.com/k0rdent/kcm/blob/main/.github/workflows/release.yml#L89. This triggers the `set-kcm-version` https://github.com/k0rdent/kcm/blob/main/Makefile#L104, which updates the versions of the `kcm`, `kcm-regional`, and `kcm-templates` Helm charts to match the current tag.

PR #2001, opened 2025-09-17
feat: deprecate the Provider Interface controller

### LINKED ISSUE #1978: Deprecate provider interface controller
ProviderInterface is part of the regional API, while the KCM controller, which includes the provider interface reconciler, is deployed only on management clusters. As a result, ProviderInterfaces deployed on regional clusters remain unreconciled.
Since the ProviderInterface status is not meaningful, the code should be refactored to avoid relying on it.

eromanova:
1. Deprecate the Provider Interface controller and the ProviderInterface status, since it's no longer in use.
2. Refactor the mechanism for getting cluster GVKs during provider cluster deletion.

Fixes #1978


internal/providerinterface/util.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"slices"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+)
+
+const fluxHelmLabelKey = "helm.toolkit.fluxcd.io/name"
```
zerospiel: `
```suggestion
const fluxHelmLabelKey = kcmv1.FluxHelmChartNameKey
```

But might be removed

PR #2002, opened 2025-09-17
feat: rework ClusterTemplate valid status check

### LINKED ISSUE #1977: Rework ClusterTemplate valid status check
Currently, a ClusterTemplate is considered `valid` if the referenced HelmChart exists and the management cluster has all required providers with the proper compatibility attributes exposed. This validity check is then used to validate the creation of a ClusterDeployment.
With the introduction of regional clusters, a template may be valid when the Cluster is deployed in the management cluster, but invalid if the cluster is deployed in a regional cluster.

To address this, the validation logic should be reworked:
* A template should be marked as valid based solely on the existence and validity of the HelmChart.
* Provider availability and compatibility attribute checks should be refactored to account for regional clusters.

eromanova:
Currently, a ClusterTemplate is considered valid if the referenced HelmChart exists and the management cluster has all required providers with the proper compatibility attributes exposed. This validity check is then used to validate the creation of a ClusterDeployment.
With the introduction of regional clusters, a template may be valid when the Cluster is deployed in the management cluster, but invalid if the cluster is deployed in a regional cluster.

This PR includes:
1. Don't store the compatibility attributes and exposed providers validation result in the ClusterTemplate status.
2. Instead, check compatibility attributes and exposed providers when necessary while validating.

Fixes #1977

### DIFF
--- a/test/objects/region/region.go
+++ b/test/objects/region/region.go
@@ -26,7 +26,7 @@ const (
 
 type Opt func(region *kcmv1.Region)
 
-func NewRegion(opts ...Opt) *kcmv1.Region {
+func New(opts ...Opt) *kcmv1.Region {
 	p := &kcmv1.Region{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       kcmv1.RegionKind,
--- a/internal/webhook/clusterdeployment_webhook_test.go
+++ b/internal/webhook/clusterdeployment_webhook_test.go
@@ -60,7 +60,7 @@ var (
 		}),
 	)
 
-	rgn = region.NewRegion(
+	rgn = region.New(
 		region.WithName(testRegionName),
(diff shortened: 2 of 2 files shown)


test/objects/region/region.go
```diff
@@ -0,0 +1,69 @@
...
+	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+)
+
+const (
+	DefaultName = "rgn"
+)
+
+type Opt func(region *kcmv1.Region)
+
+func NewRegion(opts ...Opt) *kcmv1.Region {
```
zerospiel: ultra nit: just `New` (`region.NewRegion` -> `region.New`)

zerospiel (approved):
LGTM but should be rebased in order to avoid duplication in the `internal/utils/validation/util.go`

zerospiel (approved):

PR #2009, opened 2025-09-17
feat: MultiClusterService dependencies validation

wahabmk:

### DIFF
--- a/api/v1beta1/multiclusterservice_types.go
+++ b/api/v1beta1/multiclusterservice_types.go
@@ -279,6 +279,7 @@ type MultiClusterServiceSpec struct {
 	ClusterSelector metav1.LabelSelector `json:"clusterSelector,omitempty"`
 	// ServiceSpec is spec related to deployment of services.
 	ServiceSpec ServiceSpec `json:"serviceSpec,omitempty"`
+	DependsOn   []string    `json:"dependsOn,omitempty"`
 }
 
 // ServiceStatus contains details for the state of services.
--- a/internal/utils/validation/multiclusterservice.go
+++ b/internal/utils/validation/multiclusterservice.go
@@ -1,0 +1,78 @@
+package validation
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ValidateMCSDependencyOverall calls all of the functions
+// related to MultiClusterService dependency validation one by one.
+func ValidateMCSDependencyOverall(ctx context.Context, c client.Client, mcs *kcmv1.MultiClusterService) error {
+	mcsList := new(kcmv1.MultiClusterServiceList)
+	if err := c.List(ctx, mcsList); err != nil {
+		return fmt.Errorf("failed to list MultiClusterServices: %w", err)
+	}
+
+	if err := validateMCSDependency(mcs, mcsList); err != nil {
+		return fmt.Errorf("failed service dependency validation: %w", err)
+	}
+
+	if err := validateMCSDependencyCycle(mcs, mcsList); err != nil {
+		return fmt.Errorf("failed service dependency cycle validation: %w", err)
+	}
+
+	return nil
+}
+
+// validateMCSDependency validates if all dependencies of a MultiClusterService already exist.
+func validateMCSDependency(mcs *kcmv1.MultiClusterService, mcsList *kcmv1.MultiClusterServiceList) error {
+	if mcs == nil || len(mcs.Spec.DependsOn) == 0 {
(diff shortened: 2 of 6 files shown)


internal/util/validation/services.go
```diff
@@ -157,24 +157,27 @@ func ValidateServiceDependencyOverall(services []kcmv1.Service) error {
 	return nil
 }
 
-// validateServiceDependency validates is all dependencies of services has been defined as well.
+// validateServiceDependency validates if all dependencies of services have been defined as well.
 func validateServiceDependency(services []kcmv1.Service) error {
 	if len(services) == 0 {
```
kylewuolle, line 170: Might not be worth it but you could probably combine cyclic dependency checking with this check for service definition.

wahabmk: I thought about that but then decided to keep them separate mainly for ease of writing tests.

wahabmk:
e2e tests will be added in https://github.com/k0rdent/kcm/issues/1992.

kylewuolle (approved):

api/v1beta1/multiclusterservice_types.go
```diff
@@ -55,6 +55,9 @@ const (
 
 	// ServicesDependencyValidationCondition defines the condition of services' dependencies.
 	ServicesDependencyValidationCondition = "ServicesDependencyValidation"
+
+	// MultiClusterDependencyValidationCondition defines the condition of MultiClusterService dependencies.
```
eromanova: nit:
```suggestion
	// MultiClusterServiceDependencyValidationCondition defines the condition of MultiClusterService dependencies.
```

internal/controller/multiclusterservice_controller.go
```diff
@@ -560,17 +591,73 @@ func (r *MultiClusterServiceReconciler) cleanup(ctx context.Context, mcs *kcmv1.
...
+		for _, svc := range sset.Status.Services {
+			deployed := 0
+			if _, ok := svcToCheck[serviceset.ServiceKey(svc.Namespace, svc.Name)]; ok {
+				if svc.State == kcmv1.ServiceStateDeployed {
+					deployed++
+				}
+			}
+
+			if deployed != len(depMCS.Spec.ServiceSpec.Services) {
+				errs = errors.Join(errs, fmt.Errorf("all services managed by ServiceSet %s owned by MultiClusterService %s not yet deployed", ssetKey, client.ObjectKeyFromObject(mcs)))
```
eromanova: nit:
Maybe rephrase this error message to be something like:
```
not all services in ServiceSet %s (owned by MultiClusterService %s) are deployed: %d of %d ready
```
to be more descriptive?

internal/controller/multiclusterservice_controller.go
```diff
@@ -325,6 +354,19 @@ func (r *MultiClusterServiceReconciler) reconcileDelete(ctx context.Context, mcs
 		}
 	}()
 
+	l.Info("Validating MultiClusterService dependencies for delete")
+	if err := validationutil.ValidateMCSDelete(ctx, r.Client, mcs); err != nil {
```
eromanova, line 358: Maybe we should add this and the other validation rules to the admission webhook as well?

zerospiel: As far as I remember, the VWhC has been deleted by #2071

kylewuolle (dismissed):

internal/controller/multiclusterservice_controller.go
```diff
@@ -325,6 +354,19 @@ func (r *MultiClusterServiceReconciler) reconcileDelete(ctx context.Context, mcs
...
+		if r.setCondition(mcs, kcmv1.MultiClusterServiceDependencyValidationCondition, err) {
+			record.Warnf(mcs, mcs.Generation, kcmv1.MultiClusterServiceDependencyValidationCondition, err.Error())
+		}
+		l.Error(err, "failed validation for MultiClusterService deletion, will retrigger")
+		// Will retrigger this error because we want this MCS to be deleted once:
+		// 1. Either the MCS this one depends on is deleted.
+		// 2. Or the dependency is removed.
+		return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, err
```
BROngineer: Returning result with requeue and non-nil error makes no sense: an object will be anyways enqueued for reconciliation in case of reconcile returns an error. Hence, either return an error, so an object will be reconciled with configured rate limit or return result with requeue if we want explicit interval.

BROngineer (commented):
Overall lgtm with one comment.

internal/controller/multiclusterservice_controller.go
wahabmk: Yes, this was copied over so I forgot to change it. Thanks for pointing out!

BROngineer (approved):

PR #2011, opened 2025-09-17
feat: adapt services reconciliation to regional clusters

eromanova:
* Use the regional client to access sveltos objects.
* Do not allow removal of ClusterDeployment until all ServiceSets are removed.
* Since kcm controller can't watch regional sveltos objects (profile, clustersummary), requeue the reconciler until the service is deployed.
* Refactor scheme getter functions.

Relates to #1979

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -49,11 +49,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+	"github.com/K0rdent/kcm/internal/controller/region"
 	"github.com/K0rdent/kcm/internal/record"
 	"github.com/K0rdent/kcm/internal/serviceset"
 	"github.com/K0rdent/kcm/internal/utils"
 	"github.com/K0rdent/kcm/internal/utils/pointer"
 	"github.com/K0rdent/kcm/internal/utils/ratelimit"
+	schemeutil "github.com/K0rdent/kcm/internal/utils/scheme"
 )
 
 const (
@@ -97,6 +99,7 @@ type ServiceSetReconciler struct {
 
 	timeFunc func() time.Time
 
+	SystemNamespace string
 	// AdapterName is the name of the workload running the controller
 	// effectively this name is used to identify adapter in the
 	// [github.com/k0rdent/kcm/api/v1beta1.StateManagementProvider] spec.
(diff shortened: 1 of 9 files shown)


internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -626,7 +652,8 @@ func (r *ServiceSetReconciler) collectServiceStatuses(ctx context.Context, servi
 	})
 
 	l.Info("Collecting Service statuses completed", "duration", time.Since(start))
-	return nil
+	requeue = !serviceSet.Status.Deployed
```
zerospiel, line 626: Having this, do we even need the watchers then? For the `clustersummaries` and `profiles` I mean

internal/controller/clusterdeployment_controller.go
```diff
@@ -815,6 +828,23 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, mgmt
...
+	serviceSets := &kcmv1.ServiceSetList{}
+	err := r.MgmtClient.List(ctx, serviceSets, client.MatchingFields{kcmv1.OwnerRefIndexKey: cd.Name})
+	if err != nil {
+		return false, err
+	}
+	if len(serviceSets.Items) == 0 {
+		return false, nil
+	}
+	for _, ss := range serviceSets.Items {
+		if err := r.MgmtClient.Delete(ctx, &ss); err != nil {
```
zerospiel, line 858: optional: we could've used `deletecollection` call, combining with the field selector by the owner reference

zerospiel (dismissed):
LGTM. 1 question, and 1 optional-to-implement suggestion (does not change much)

internal/controller/clusterdeployment_controller.go
eromanova: You mean `DeleteAllOf`, right? I did try it, but the call fails when a field indexer is used. From what I’ve found, that seems to be expected behavior: https://github.com/kubernetes-sigs/controller-runtime/issues/1956 (simple Delete works pretty good). So for now, I’ve stuck with the original approach.

zerospiel: YEah, `deleteallof`. If I understand it correctly (pseudocode):

```
sel := selector.Everything()
sel.AddOneTermSelector("metadata.owners")
c.DeleteAllOf(ctx, &kcmv1.ServiceSet{}, &client.DeleteAllOfOptions{ListOptions: client.ListOptions{FieldSelector: sel}})
```

that should've worked. But whatever

Please feel free to resolve

eromanova: Great. Let me try it. Thanks for the suggestion.

eromanova: I'll refactor this part in one of the next PR. Resolving.

zerospiel (dismissed):

kylewuolle (approved):

PR #2014, opened 2025-09-17
fix: reflect deletion status in Region conditions

eromanova:
Now, when the Region object is removed, the Ready condition is updated with the actual status of HelmReleases removal

Related task #1979

### DIFF
--- a/internal/controller/region/controller.go
+++ b/internal/controller/region/controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Kshatrix (approved):

internal/controller/region/controller.go
```diff
@@ -273,9 +276,17 @@ func (r *Reconciler) removeHelmReleases(ctx context.Context, region *kcmv1.Regio
...
 		l.Info("Waiting for all HelmReleases to be deleted before removing finalizer")
-		return true, errs
+		meta.SetStatusCondition(&region.Status.Conditions, metav1.Condition{
+			Type:               kcmv1.ReadyCondition,
+			ObservedGeneration: region.Generation,
+			Status:             metav1.ConditionFalse,
+			Reason:             kcmv1.NotAllComponentsHealthyReason,
+			Message:            fmt.Sprintf("Waiting for all HelmReleases to be deleted: %s", hrNames),
+		})
+		return true, nil
```
zerospiel, line 289: the condition will hang if it was set and then all of the releases are actually gone. is it okay, considering this step is the last before deletion from the cache and the server?

eromanova: It’s not critical since the Region object is removed almost immediately, but for consistency, I’ll clarify this in a future PR. Thanks.

Issue #2023, opened 2025-09-19
[bug] ClusterTemplate is invalid: configmap RBAC issue


eromanova:
**Describe the bug**

When the ClusterTemplate is created in a non-system namespace, it remains invalid with the following error:

```
    valid: false
    validationError: 'failed to create or update schema ConfigMap: configmaps is forbidden:
      User "system:serviceaccount:kcm-system:kcm-controller-manager" cannot create
      resource "configmaps" in API group "" in the namespace "ekaz"'
```

<details>
<summary>ClusterTemplate</summary>

```
apiVersion: v1
items:
- apiVersion: k0rdent.mirantis.com/v1beta1
  kind: ClusterTemplate
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"k0rdent.mirantis.com/v1beta1","kind":"ClusterTemplate","metadata":{"annotations":{},"name":"openstack-standalone-cp-1-0-16","namespace":"ekaz"},"spec":{"helm":{"chartSpec":{"chart":"openstack-standalone-cp","interval":"10m0s","reconcileStrategy":"ChartVersion","sourceRef":{"kind":"HelmRepository","name":"kcm-templates"},"version":"1.0.16"}}}}
    creationTimestamp: "2025-09-19T09:15:19Z"
    generation: 1
    labels:
      k0rdent.mirantis.com/component: kcm
    name: openstack-standalone-cp-1-0-16
    namespace: ekaz
    resourceVersion: "13335"
    uid: f0516734-0fdb-43f3-9a85-a231b876cc26
  spec:
    helm:
      chartSpec:
        chart: openstack-standalone-cp
        interval: 10m0s
        reconcileStrategy: ChartVersion
        sourceRef:
          kind: HelmRepository
          name: kcm-templates
        version: 1.0.16
  status:
    chartRef:
      kind: HelmChart
      name: openstack-standalone-cp-1-0-16
      namespace: ekaz
    chartVersion: 1.0.16
    config:
      apiServerLoadBalancer:
        enabled: true
      bastion:
        enabled: false
        spec:
          flavor: ""
          image:
            filter:
              name: ""
              tags: []
          providerID: ""
          sshKeyName: ""
      ccmCanViewAllProjects: false
      ccmEnableUniqueClusterName: false
      ccmRegional: true
      clusterAnnotations: {}
      clusterIdentity:
        name: ""
      clusterLabels: {}
      clusterNetwork:
[...]

PR #2024, opened 2025-09-19
fix: add configmap rbac permissions for controller

Kshatrix:
Fixes: #2023

### DIFF
--- a/templates/provider/kcm/templates/rbac/controller/roles.yaml
+++ b/templates/provider/kcm/templates/rbac/controller/roles.yaml
@@ -296,6 +296,7 @@ rules:
   - ""
   resources:
   - secrets
+  - configmaps
   verbs:
   - get
   - list


templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -296,6 +296,7 @@ rules:
   - ""
   resources:
   - secrets
+  - configmaps
```
zerospiel, line 298: just in case: `configmaps` is already presented, and I'd make a dedicated section for it because the comment `# copy secrets between namespaces` does not make any sense then

Kshatrix: the comment is not really meaningful, so I've removed it. duplicated permissions for configmaps (and also for secrets) have been removed as well.

zerospiel (approved):

eromanova (approved):

PR #2028, opened 2025-09-22
ci: add kcm upgrade check github workflow

gmlexx:
Ensuring that release upgrade works


a13x5 (changes requested):
Usage of dev targets for testing in main and especially release triggers has a lot of potential to have CI failures without actual issues. 
Specifically this workflow has trigger on the release and expects that artifacts will appear somehow for testing (this will not happen since on the time this workflow will run no artifact will be built).

This type of checks should be a part of e2e testing suite.

PR #2030, opened 2025-09-22
fix: increase default helm timeout for dev setup

eromanova:
Increase the default Helm install/upgrade timeout to 20 minutes to account for slow network.

### DIFF
--- a/config/dev/kcm_values.yaml
+++ b/config/dev/kcm_values.yaml
@@ -2,6 +2,7 @@ image:
   repository: localhost/kcm/controller
   tag: latest
 controller:
+  defaultHelmTimeout: 20m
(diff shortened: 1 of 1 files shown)


zerospiel (approved):
All good, just wondering what was the reason for the change?

zerospiel:
> The Kof team runs k0rdent/kcm + kof upgrade tests in their CI, but they fail with a `context deadline exceeded` error in Helm, even though the same process works locally. I thought increasing the timeout would help with this issue, but we'll need to reverify it.

Okay, got it. I was under the impression that this parameter is configurable and might be changed on the fly. Anyway, thanks for the clarification

PR #2039, opened 2025-09-24
chore: use dev segment token for ci builds

Kshatrix:


.github/workflows/build_test.yml
```diff
@@ -235,6 +235,7 @@ jobs:
           VERSION: ${{ needs.lint-test.outputs.version }}
           SKIP_SCM_RELEASE: true
           SKIP_ARM: ${{ needs.authorize.outputs.skip_arm }}
+          SEGMENT_TOKEN: ${{ secrets.SEGMENT_TOKEN_DEV }}
```
eromanova, line 239: This step runs only when the public repository is used, which currently happens only when the hosted cluster deployment is triggered.
If we want dev `SEGMENT_TOKEN` to be available in other tests (ex. controller-only runs or local runs without hosted clusters, when we don't push the kcm controller image to the public CI repo), this environment variable must also be added to all `test-e2e` executions.

PR #2047, opened 2025-09-25
fix(backups): collect extra labels for regions

zerospiel:


zerospiel:
Not needed

PR #2061, opened 2025-10-02
fix: cert-manager dependent regional components disabling

### LINKED ISSUE #2060: [bug] Cannot disable cert-manager-dependent subcomponents using legacy values
**Describe the bug**
When all the regional kcm components (`velero`, `cert-manager`, `capi operator`) were moved into a subchart to support installation on regional clusters, the `kcm-regional` Helm chart became a dependency of the main kcm chart.
As a result, configuration for regional components now needs to be specified under the regional section (e.g. `regional.velero`, `regional.capi-operator`, etc.).

For backward compatibility, the old top-level values still work for configuration but disabling `velero` and `capi-operator` (which depend on cert-manager) no longer works using the old values syntax.

Workaround:

Explicitly disable the subcomponent under the regional section. For example, to disable velero:

```
regional:
  velero:
    enabled: false
```

eromanova:
This PR restores support for disabling regional subcomponents that depend on the cert manager (velero, capi-operator).

Fixes #2060

### DIFF
--- a/internal/controller/components/values.go
+++ b/internal/controller/components/values.go
@@ -162,8 +162,13 @@ func getRegionalComponentValues(
 		l.Info("Waiting for Cert manager API before enabling additional components")
 	} else {
 		l.Info("Cert manager is installed, enabling additional components")
-		veleroValues["enabled"] = true
-		capiOperatorValues["enabled"] = true
+		// enabling components unless explicitly disabled in values
+		if !componentDisabled(veleroValues) {
+			veleroValues["enabled"] = true
+		}
+		if !componentDisabled(capiOperatorValues) {
+			capiOperatorValues["enabled"] = true
+		}
 	}
 
 	if opts.RegistryCertSecretName != "" {
@@ -176,6 +181,16 @@ func getRegionalComponentValues(
 	regionalValues["velero"] = veleroValues
 
 	return regionalValues, nil
+}
+
+func componentDisabled(values chartutil.Values) bool {
+	if values == nil {
+		return false
+	}
+	if enabled, ok := values["enabled"].(bool); ok {
+		return !enabled
+	}
+	return false
 }
 
 // getCurrentValuesForKey looks up a key in currentValues and returns it as map[string]any.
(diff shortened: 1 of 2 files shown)


internal/controller/components/values.go
```diff
@@ -178,6 +183,18 @@ func getRegionalComponentValues(
...
 
+func componentDisabled(values chartutil.Values) bool {
+	if values == nil {
+		return false
+	}
+	if enabledVal, exists := values["enabled"]; exists {
+		if enabled, ok := enabledVal.(bool); ok {
+			return !enabled
+		}
+	}
```
zerospiel:
```suggestion
	if enabled, ok := values["enabled"].(bool); ok {
		return !enabled
	}
```

nit: probably couldve been simplified but the explicit version is also good enough

internal/controller/components/values.go
```diff
@@ -162,8 +162,13 @@ func getRegionalComponentValues(
...
 		l.Info("Cert manager is installed, enabling additional components")
-		veleroValues["enabled"] = true
-		capiOperatorValues["enabled"] = true
+		// enabling components unless explicitly disabled in values
+		if !componentDisabled(veleroValues) {
+			veleroValues["enabled"] = true
+		}
+		if !componentDisabled(capiOperatorValues) {
+			capiOperatorValues["enabled"] = true
+		}
```
zerospiel, line 171: do i understand it correctly, that before introducing regions we had these checks but they got lost?

eromanova: We didn't have exactly these checks since we didn't need them.
Values from the Management spec always have higher precedence over the values generated in the kcm controller https://github.com/k0rdent/kcm/blob/v1.3.0/internal/controller/management_controller.go#L897. Before introducing regions, if `velero.enabled` was set to `false`, it had never been overwritten by the controller.

Now that legacy values are merged with the new ones, these checks are necessary because the controller dynamically computes the values for regional components.

zerospiel: Indeed, i've mixed them up with the previous flux/capi-op valus processing. Thanks for the clarification

zerospiel (approved):

Issue #1455, opened 2025-04-28
[enh] ServiceTemplate versioning and sequential upgrades


BROngineer:
Goals:
- allow sequential upgrades for services defined in ClusterDeployment/MultiClusterService

PR #2062, opened 2025-10-02
feat: implement sequential upgrade

kylewuolle:
This PR implements sequential upgrades for services.
https://github.com/k0rdent/kcm/issues/1455

### DIFF
--- a/api/v1beta1/serviceset_types.go
+++ b/api/v1beta1/serviceset_types.go
@@ -93,6 +93,41 @@ const (
 	ServiceSetOperationNone   ServiceSetOperation = "none"
 )
 
+/*
+// +kubebuilder:object:generate=false
+// +k8s:deepcopy-gen=false
+type VersionedService interface {
+	GetVersion() string
+	SetVersion(string) VersionedService
+	GetTemplate() string
+}
+
+func (s Service) GetVersion() string {
+	return s.Version
+}
+
+func (s Service) SetVersion(v string) VersionedService {
+	s.Version = v
+	return s
+}
+
+func (s Service) GetTemplate() string {
+	return s.Template
+}
+
+func (s ServiceWithValues) GetVersion() string {
+	return s.Version
+}
+
+func (s ServiceWithValues) SetVersion(v string) ServiceWithValues {
+	s.Version = v
+	return s
+}
+
+func (s ServiceWithValues) GetTemplate() string {
+	return s.Template
+}
+*/
 // ServiceSetSpec defines the desired state of ServiceSet
 type ServiceSetSpec struct {
 	// Cluster is the name of the ClusterDeployment
@@ -131,6 +166,8 @@ type ServiceWithValues struct {
 	// then the name is the name of the Helm release.
 	Name string `json:"name"`
 
+	Version string `json:"version"`
+
 	// Namespace is the namespace where the service is deployed. If the ServiceTemplate
 	// is backed by Helm chart, then the namespace is the namespace where the Helm release is deployed.
 	Namespace string `json:"namespace"`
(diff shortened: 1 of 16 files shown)


api/v1beta1/serviceset_types.go
```diff
@@ -93,6 +93,41 @@ const (
...
+
+func (s ServiceWithValues) SetVersion(v string) ServiceWithValues {
+	s.Version = v
+	return s
+}
+
+func (s ServiceWithValues) GetTemplate() string {
+	return s.Template
+}
+*/
```
BROngineer: Leaving a comment to not to forget to cleanup

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -186,6 +186,10 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		if !equality.Semantic.DeepEqual(clone.Status, serviceSet.Status) {
 			err = errors.Join(err, r.Status().Update(ctx, serviceSet))
 		}
+
+		if !equality.Semantic.DeepEqual(clone.Spec.Services, serviceSet.Spec.Services) {
+			err = errors.Join(err, r.Update(ctx, serviceSet))
+		}
```
BROngineer, line 180: We should not update object's spec while reconciling that object

api/v1beta1/serviceset_types.go
```diff
@@ -133,6 +133,8 @@ type ServiceWithValues struct {
 	// then the name is the name of the Helm release.
 	Name string `json:"name"`
 
+	Version string `json:"version"`
```
BROngineer: Looks like this field is required. How it's handled for the case of kcm upgrade? Would it make sense to make this field optional?

internal/serviceset/util.go
```diff
@@ -57,6 +57,63 @@ func ObjectKey(systemNamespace string, cd *kcmv1.ClusterDeployment, mcs *kcmv1.M
 	}
 }
 
+func ResolveServiceVersions(ctx context.Context, c client.Client, namespace string, services any) error {
```
BROngineer, line 60: Would it worth to introduce small interface rather than use `any` type? It seems that in the following functions being called in this one, we only need to be able to get the version and template name from the service/serviceWithValues and to set the version.

internal/serviceset/util.go
```diff
@@ -57,6 +57,63 @@ func ObjectKey(systemNamespace string, cd *kcmv1.ClusterDeployment, mcs *kcmv1.M
...
+			template := kcmv1.ServiceTemplate{}
+			if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: svc.Template}, &template); err != nil {
+				return fmt.Errorf("failed to fetch template %s: %w", svc.Template, err)
+			}
+
+			version := template.Spec.Version
+			if version == "" && template.Spec.Helm != nil && template.Spec.Helm.ChartSpec != nil {
+				version = template.Spec.Helm.ChartSpec.Version
+			}
+			svc.Version = version
```
BROngineer, line 93: How would we act in case version is empty but the template is not based on Helm? Shouldn't we fallback to current behaviour, when the template name is treated as the template version?
Also, IIRC we do not force user to use template.spec.version field, hence more likely it will be empty.

internal/serviceset/util.go
```diff
@@ -57,6 +57,63 @@ func ObjectKey(systemNamespace string, cd *kcmv1.ClusterDeployment, mcs *kcmv1.M
...
+			template := kcmv1.ServiceTemplate{}
+			if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: svc.Template}, &template); err != nil {
+				return fmt.Errorf("failed to fetch Template %s: %w", svc.Template, err)
+			}
+
+			version := template.Spec.Version
+			if version == "" && template.Spec.Helm != nil && template.Spec.Helm.ChartSpec != nil {
+				version = template.Spec.Helm.ChartSpec.Version
+			}
+			svc.Version = version
```
BROngineer: Same as in above

BROngineer (commented):
@kylewuolle overall looks good. Just need to clarify few points.

internal/serviceset/util.go
kylewuolle: Problem is that causes issues with kubebuilder. I went down that path and it didn't work out!

PR #2063, opened 2025-10-02
feat: region validation

### LINKED ISSUE #1907: Region-specific admission validation
TBD

eromanova:
This PR adds validation rules for the Region object, both to the admission webhook and to the controller when the admission webhook is disabled.

Partially fixes #1907

### DIFF
--- a/internal/util/validation/mgmt.go
+++ b/internal/util/validation/mgmt.go
@@ -25,7 +25,7 @@ import (
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
 )
 
-type ParentCluster interface {
+type ComponentsManager interface {
 	client.Object
 
 	Components() kcmv1.ComponentsCommonSpec
@@ -36,11 +36,11 @@ var ErrProviderIsNotReady = errors.New("provider is not yet ready")
 
 // GetIncompatibleContracts validates if all of the providers specified in the given [github.com/K0rdent/kcm/api/v1beta1.Management]
 // or [github.com/K0rdent/kcm/api/v1beta1.Region] have compatible CAPI [contract versions].
-// Returns [ErrProviderIsNotReady] is the corresponding [github.com/K0rdent/kcm/api/v1beta1.ProviderTemplate]
+// Returns [ErrProviderIsNotReady] if the corresponding [github.com/K0rdent/kcm/api/v1beta1.ProviderTemplate]
 // is not yet ready and the validation cannot proceed further.
 //
 // [contract versions]: https://cluster-api.sigs.k8s.io/developer/providers/contracts
-func GetIncompatibleContracts(ctx context.Context, cl client.Client, release *kcmv1.Release, obj ParentCluster) (string, error) {
+func GetIncompatibleContracts(ctx context.Context, cl client.Client, release *kcmv1.Release, obj ComponentsManager) (string, error) {
 	capiTplName := release.Spec.CAPI.Template
 	if obj.Components().Core != nil && obj.Components().Core.CAPI.Template != "" {
 		capiTplName = obj.Components().Core.CAPI.Template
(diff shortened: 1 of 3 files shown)


internal/util/validation/mgmt.go
```diff
@@ -25,18 +25,25 @@ import (
...
+	Components() kcmv1.ComponentsCommonSpec
+}
+
 // ErrProviderIsNotReady signals if the corresponding [github.com/K0rdent/kcm/api/v1beta1.ProviderTemplate] is not yet ready.
 var ErrProviderIsNotReady = errors.New("provider is not yet ready")
 
 // GetIncompatibleContracts validates if all of the providers specified in the given [github.com/K0rdent/kcm/api/v1beta1.Management]
-// have compatible CAPI [contract versions]. Returns [ErrProviderIsNotReady] is the corresponding [github.com/K0rdent/kcm/api/v1beta1.ProviderTemplate]
+// or [github.com/K0rdent/kcm/api/v1beta1.Region] have compatible CAPI [contract versions].
+// Returns [ErrProviderIsNotReady] is the corresponding [github.com/K0rdent/kcm/api/v1beta1.ProviderTemplate]
```
zerospiel:
```suggestion
// Returns [ErrProviderIsNotReady] if the corresponding [github.com/K0rdent/kcm/api/v1beta1.ProviderTemplate]
```

internal/util/validation/provider_template.go
```diff
@@ -23,13 +23,25 @@ import (
...
+	Regions map[string]bool
+	// ProviderContracts is the list of supported Cluster API contract versions,
+	// e.g. infrastructure-aws: []{v1alpha3, v1alpha4, v1beta1}
+	// [contract versions]: https://cluster-api.sigs.k8s.io/developer/providers/contracts
+	ProviderContracts []string
+}
+
 // GetInUseProvidersWithContracts constructs a map based on the given [github.com/K0rdent/kcm/api/v1beta1.ProviderTemplate]
 // where keys are a provider name in the format (bootstrap|control-plane|infrastructure)-<name> and values are the
 // corresponding CAPI [contract versions], e.g. infrastructure-aws: []{v1alpha3, v1alpha4, v1beta1}
```
zerospiel: nit: the description is not accurate with the introduced change

internal/util/validation/provider_template.go
```diff
@@ -23,13 +23,25 @@ import (
...
 
+// InUseProviderParams describes in which regions this provider is currently in use
+// and which Cluster API contract versions it supports
+type InUseProviderParams struct {
+	// Regions maps region names to a boolean indicating whether
+	// this provider is in use in that region
+	Regions map[string]bool
+	// ProviderContracts is the list of supported Cluster API contract versions,
+	// e.g. infrastructure-aws: []{v1alpha3, v1alpha4, v1beta1}
+	// [contract versions]: https://cluster-api.sigs.k8s.io/developer/providers/contracts
```
zerospiel:
```suggestion
```

internal/util/validation/provider_template.go
```diff
@@ -52,8 +66,39 @@ func GetInUseProvidersWithContracts(ctx context.Context, cl client.Client, pTpl
...
+				regions[cld.Status.Region] = true
+			}
+			providerContracts = append(providerContracts, cltpl.Status.ProviderContracts[providerName])
 		}
+		inUseProviders[providerName] = InUseProviderParams{Regions: regions, ProviderContracts: providerContracts}
 	}
 	return inUseProviders, nil
 }
+
+func GetInUseProvidersWithContractsForParent(ctx context.Context, cl client.Client, pTpl *kcmv1.ProviderTemplate, obj ParentCluster) (map[string][]string, error) {
```
zerospiel: i do not quite understand the naming, what does the `parent` mean in particular? initially i had an impression that this is all about validating from some dependent object but the `parent` is actually either `region` or `mgmt` (if i understood it correctly). and that confuses a lot

from the usage perspective, all of the objects being validated are on the mgmt cluster, and that is the reason why i am confused with the names

besides, it looks like we could just rename the previous function (maybe just make it private and taht's all) and leave the names as they are

internal/util/validation/mgmt.go
```diff
@@ -25,18 +25,25 @@ import (
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
 )
 
+type ParentCluster interface {
```
zerospiel: WRT my concern regarding the naming: what does a type implementing the object do? are the implementers just `mgmt` and `region` or are there other types?

internal/util/validation/provider_template.go
eromanova: In this context, `parent` refers to the abstract object responsible for managing providers and components. It is always either a Management or a Region object. I agree that `parent` may not be the clearest term, so I’ll reword it for clarity.

internal/util/validation/mgmt.go
eromanova: No, only `mgmt` and `region`. Agree that it does not sound correct. Need to reword.

PR #2066, opened 2025-10-06
fix: update Credential status when Region is not ready

### LINKED ISSUE #2056: [bug][nit] Regional credential still valid after corresponding Region object removal
**Describe the bug**
Credential, created for region remains valid and available for usage after Region object is removed.

**To Reproduce**
1. Deploy any cluster 
2. Create Region referencing that cluster
3. Create any credential set (for aws  - secret, awssi in regional and credential in mgmt)
4. Remove Region
5. Check credential

**Expected behavior**
Credential should not valid anymore

**Additional context**
Credential no more valid only if the corresponding ex-regional cluster is removed.

eromanova:
Currently, when the Region is removed, the Credential object remains valid. This PR fixes it and sets the up-to-date status to the regional Credential object.

Fixes #2056

### DIFF
--- a/api/v1beta1/indexers.go
+++ b/api/v1beta1/indexers.go
@@ -397,11 +397,11 @@ func ExtractServiceSetProvider(o client.Object) []string {
 const CredentialRegionIndexKey = "spec.region"
 
 func setupCredentialRegionIndexer(ctx context.Context, mgr ctrl.Manager) error {
-	return mgr.GetFieldIndexer().IndexField(ctx, &Credential{}, CredentialRegionIndexKey, ExtractCredentialRegion)
+	return mgr.GetFieldIndexer().IndexField(ctx, &Credential{}, CredentialRegionIndexKey, extractCredentialRegion)
 }
 
-// ExtractCredentialRegion returns the region name from [Credential] object.
-func ExtractCredentialRegion(o client.Object) []string {
+// extractCredentialRegion returns the region name from [Credential] object.
+func extractCredentialRegion(o client.Object) []string {
 	cred, ok := o.(*Credential)
 	if !ok {
 		return nil
--- a/internal/controller/credential_controller.go
+++ b/internal/controller/credential_controller.go
@@ -94,7 +94,7 @@ func (r *CredentialReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 				ObservedGeneration: cred.Generation,
 				Message:            failedMsg,
 			})
-			return ctrl.Result{}, errors.New(failedMsg)
+			return ctrl.Result{}, client.IgnoreNotFound(err)
 		}
 
 		rgnClient, _, err = kube.GetRegionalClient(ctx, r.MgmtClient, r.SystemNamespace, rgn, schemeutil.GetRegionalScheme)


internal/controller/credential_controller.go
```diff
@@ -72,16 +76,38 @@ func (r *CredentialReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		}
+		if failedMsg != "" {
+			apimeta.SetStatusCondition(cred.GetConditions(), metav1.Condition{
+				Type:               kcmv1.CredentialReadyCondition,
+				Status:             metav1.ConditionFalse,
+				Reason:             kcmv1.FailedReason,
+				ObservedGeneration: cred.Generation,
+				Message:            failedMsg,
+			})
+			return ctrl.Result{}, errors.New(failedMsg)
```
zerospiel: do we actually have to requeue here with an error? Maybe it makes sense to return with an error only if there was a problem getting the `regions` object, otherwise just update the status and "done" (empty result with no error). wdyt?

api/v1beta1/indexers.go
```diff
@@ -387,3 +390,21 @@ func ExtractServiceSetProvider(o client.Object) []string {
...
+
+// CredentialRegionIndexKey indexer field name to extract region name from [Credential] object.
+const CredentialRegionIndexKey = "spec.region"
+
+func setupCredentialRegionIndexer(ctx context.Context, mgr ctrl.Manager) error {
+	return mgr.GetFieldIndexer().IndexField(ctx, &Credential{}, CredentialRegionIndexKey, ExtractCredentialRegion)
+}
+
+// ExtractCredentialRegion returns the region name from [Credential] object.
+func ExtractCredentialRegion(o client.Object) []string {
```
zerospiel: nit: i do not see the reason for the function to be public (there are no tests yet for this function to be called outside the package)

internal/controller/credential_controller.go
eromanova: You're right. It doesn’t make much sense to keep requeuing until the region exists.

internal/controller/credential_controller.go
```diff
@@ -72,16 +76,38 @@ func (r *CredentialReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-		})
-		return ctrl.Result{}, err
+	rgnClient := r.MgmtClient
+	if cred.Spec.Region != "" {
+		var failedMsg string
+		rgn := &kcmv1.Region{}
+		if err = r.MgmtClient.Get(ctx, client.ObjectKey{Name: cred.Spec.Region}, rgn); err != nil {
+			failedMsg = fmt.Sprintf("Failed to get Region %s: %v", cred.Spec.Region, err)
+		}
+		if rgn != nil && !rgn.DeletionTimestamp.IsZero() {
```
zerospiel:
```suggestion
		if !rgn.DeletionTimestamp.IsZero() {
```

nit: `v := &<>{}` and `v := new(<>)` already allocate an object in a span in the memory area

eromanova: Correct. Thank you.

zerospiel (approved):

PR #2071, opened 2025-10-09
chore: remove multiclusterservice validation webhook

BROngineer:
This PR removes MultiClusterService validation webhook. The intention here is to not to block automations where MultiClusterService object could be created along with ServiceTemplate objects. When using webhook objects creation will fail due to ServiceTemplate objects won't be in Ready state immediately, hence it will be impossible to create MultiClusterService object.
The validation is moved to MultiClusterService controller in this PR, so that MultiClusterService object will be reconciled again when defined ServiceTemplate objects become Ready.

### DIFF
--- a/internal/controller/multiclusterservice_controller.go
+++ b/internal/controller/multiclusterservice_controller.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


internal/controller/multiclusterservice_controller.go
```diff
@@ -119,14 +120,12 @@ func (r *MultiClusterServiceReconciler) reconcileUpdate(ctx context.Context, mcs
...
-		r.setCondition(mcs, kcmv1.ServicesDependencyValidationCondition, err)
-		if err != nil {
-			l.Error(err, "failed to validate service dependencies, will not retrigger this error")
-			return ctrl.Result{}, nil
-		}
+	l.Info("Validating service dependencies")
+	err = r.validateMultiClusterService(ctx, l, mcs)
+	r.setCondition(mcs, kcmv1.ServicesDependencyValidationCondition, err)
+	if err != nil {
+		// we won't retrigger reconciliation on validation error
```
zerospiel, line 126: nit: maybe then revert the log msg just in case it'd be visible in the logs and not only as a condition?

internal/controller/multiclusterservice_controller.go
```diff
@@ -174,6 +173,18 @@ func (r *MultiClusterServiceReconciler) reconcileUpdate(ctx context.Context, mcs
 	return result, servicesErr
 }
 
+func (r *MultiClusterServiceReconciler) validateMultiClusterService(ctx context.Context, l logr.Logger, mcs *kcmv1.MultiClusterService) error {
```
zerospiel:
```suggestion
func (r *MultiClusterServiceReconciler) validateMultiClusterService(ctx context.Context, mcs *kcmv1.MultiClusterService) error {
```

logger is in the context, no need to pass it as an argument
Besides, it is sufficient to have only one log entry, where it was before the change (see another comment)

PR #2073, opened 2025-10-09
feat(restore): support regions restoration

### LINKED ISSUE #2044: [doc] region backup-restore flow

zerospiel:
Adds logic to the cld controller to skip reconciliation until the
related Region has the Pause annotation.

Adds regions watcher to trigger on the Pause annotation deletion.

Reverts changes in the backup controller to fail if a regional backup
fails to be created.

Creates backups for all of the regions, setting errors to the according
regional statuses.

Fixes the restoration path to correctly restore the latest backup
statuses for (not yet) existing region objects.

Closes #2044

### DIFF
--- a/internal/controller/backup/single.go
+++ b/internal/controller/backup/single.go
@@ -32,7 +32,7 @@ import (
 func (r *Reconciler) createAllSingleBackups(ctx context.Context, s *scope) (ctrl.Result, error) {
 	mgmtBackup := s.mgmtBackup
 	now := time.Now().UTC()
-	l := ctrl.LoggerFrom(ctx).V(1)
+	l := ctrl.LoggerFrom(ctx)
 
 	// ensure RegionsLastBackups is initialized
 	if mgmtBackup.Status.RegionsLastBackups == nil {


internal/controller/clusterdeployment_controller.go
```diff
@@ -201,6 +201,32 @@ func (r *ClusterDeploymentReconciler) reconcileUpdate(ctx context.Context, scope
...
+		}
+
+		if apimeta.SetStatusCondition(&cd.Status.Conditions, metav1.Condition{
+			Type:               kcmv1.PausedCondition,
+			Status:             metav1.ConditionFalse,
+			Reason:             kcmv1.NotPausedReason,
+			Message:            "",
+			ObservedGeneration: cd.Generation,
+		}) {
+			r.eventf(cd, "RegionUnpaused", "Region %s is has been unpaused", scope.region.Name)
```
eromanova:
```suggestion
			r.eventf(cd, "RegionUnpaused", "Region %s has been unpaused", scope.region.Name)
```

zerospiel: Done, thanks

PR #2075, opened 2025-10-10
feat: credential cluster identity distribution

### LINKED ISSUE #1906: Credential distribution to regional clusters
TBD

eromanova:
This PR introduces a credential distribution system that propagates:
- Regional credentials (with `spec.region`): by copying cluster identities and their referenced resources to regional clusters across multiple namespaces.
- Management credentials (without `spec.region`): by distributing credentials managed by the Access Management system across multiple namespaces within the management cluster.

**Changes**:
1. ProviderInterface
- Adds a new field: `spec.clusterIdentities`, which replaces the deprecated `spec.clusterIdentityKinds`.
- The deprecated field is still supported for backward compatibility.
- spec.clusterIdentities now defines all provider Cluster Identities, including their GVKs and referenced objects.

2. Credential Controller
- Enhanced to fetch all Cluster Identities and their referenced resources from the management cluster.
- Copies these identities and references to the management cluster’s credential namespace (for management credentials), or the corresponding regional cluster’s credential namespace (for regional credentials).

**Flow Example**:
1. The user creates two Credential resources (`test1` and `test2`) and their associated identities and references in the system namespace
2. `test1` is a management credential (no `spec.region`), while test2 is a regional credential (has spec.region)
3. The user configures an AccessManagement object to distribute both credentials to namespaces `ns1` and `ns2`:
```
spec:
  accessRules:
    - targetNamespaces:
        list:
        - ns1
        - ns2
      credentials:
      - test1
      - test2
```
4. The Access Management controller copies credentials `test1` and `test2` to the target namespaces (`ns1`, `ns2`) within the management cluster
5. The Credential controller retrieves all Cluster Identities and references defined in the ProviderInterface and creates them in the appropriate cluster:
- identities for Credential `test1` will be created in the management cluster in `ns1` and `ns2` namespaces
- identities for Credential `test2` will be created in the regional cluster in `ns1` and `ns2` namespaces

**Limitations**
For regional credential identity distribution to work, the provider must be enabled in the management cluster. The corresponding identity CRDs must be installed, otherwise, the distribution will be skipped.

Fixes #1906

### DIFF
--- a/internal/controller/credential_controller.go
+++ b/internal/controller/credential_controller.go
@@ -31,7 +31,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
@@ -79,23 +78,22 @@ func (r *CredentialReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 func (r *CredentialReconciler) update(ctx context.Context, cred *kcmv1.Credential) (ctrl.Result, error) {
 	l := ctrl.LoggerFrom(ctx)
 
-	var err error
 	if controllerutil.AddFinalizer(cred, kcmv1.CredentialFinalizer) {
-		if err = r.MgmtClient.Update(ctx, cred); err != nil {
+		if err := r.MgmtClient.Update(ctx, cred); err != nil {
 			l.Error(err, "failed to update Credential finalizers")
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{RequeueAfter: r.syncPeriod}, nil
 	}
 
-	var updated bool
-	if updated, err = labelsutil.AddKCMComponentLabel(ctx, r.MgmtClient, cred); updated || err != nil {
+	if updated, err := labelsutil.AddKCMComponentLabel(ctx, r.MgmtClient, cred); updated || err != nil {
 		if err != nil {
 			l.Error(err, "adding component label")
 		}
 		return ctrl.Result{}, err
 	}
 
+	var err error
 	defer func() {
 		r.setReadyCondition(cred, err)
 		err = errors.Join(err, r.updateStatus(ctx, cred))
@@ -174,7 +172,7 @@ func (r *CredentialReconciler) update(ctx context.Context, cred *kcmv1.Credentia
 }
 
 func (r *CredentialReconciler) delete(ctx context.Context, cred *kcmv1.Credential) (ctrl.Result, error) {
-	l := log.FromContext(ctx)
+	l := ctrl.LoggerFrom(ctx)
 
 	rgnClient, err := kubeutil.GetRegionalClientByRegionName(ctx, r.MgmtClient, r.SystemNamespace, cred.Spec.Region, schemeutil.GetRegionalScheme)
 	if err == nil {
--- a/templates/provider/cluster-api-provider-openstack/templates/interface.yaml
+++ b/templates/provider/cluster-api-provider-openstack/templates/interface.yaml
@@ -14,7 +14,11 @@ spec:
       version: v1
       kind: Secret
     - group: infrastructure.cluster.x-k8s.io
-      version: v1alpha1 # Custom Cluster Identity kind for OpenStack. See: https://github.com/Mirantis/cluster-api-provider-openstack/pull/7
+      # Custom Cluster Identity kind for OpenStack. See: https://github.com/Mirantis/cluster-api-provider-openstack/pull/7
+      # WARNING: It uses the old v1alpha1 API version, which may complicate handling if/when multiple API versions
+      # are supported.
+      # Details: https://github.com/Mirantis/cluster-api-provider-openstack/pull/7#pullrequestreview-3335972221
+      version: v1alpha1
       kind: OpenStackClusterIdentity
       references:
         - group: ""
--- a/templates/provider/cluster-api-provider-aws/templates/rbac.yaml
+++ b/templates/provider/cluster-api-provider-aws/templates/rbac.yaml
@@ -24,6 +24,7 @@ rules:
     verbs:
       - create
       - delete
+      - patch
       - update
       - get
       - list
--- a/templates/provider/cluster-api-provider-azure/templates/rbac.yaml
+++ b/templates/provider/cluster-api-provider-azure/templates/rbac.yaml
@@ -22,6 +22,7 @@ rules:
     verbs:
       - create
       - delete
+      - patch
       - update
       - get
       - list
--- a/templates/provider/cluster-api-provider-openstack/templates/rbac.yaml
+++ b/templates/provider/cluster-api-provider-openstack/templates/rbac.yaml
@@ -21,6 +21,7 @@ rules:
     verbs:
       - create
       - delete
+      - patch
       - update
       - get
       - list
--- a/templates/provider/cluster-api-provider-vsphere/templates/rbac.yaml
+++ b/templates/provider/cluster-api-provider-vsphere/templates/rbac.yaml
@@ -21,6 +21,7 @@ rules:
     verbs:
       - create
       - delete
+      - patch
       - update
       - get
       - list


internal/controller/credential/clusteridentities.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	if shouldDelete {
+		if err := rgnClient.Delete(ctx, clIdty); client.IgnoreNotFound(err) != nil {
+			return fmt.Errorf("failed to delete Cluster Identity object of Kind %s: %s: %w", clIdty.GetKind(), objectKeyUnstructured(clIdty), err)
+		}
+		return nil
+	}
+
+	clIdty.SetLabels(labels)
+	if err := rgnClient.Update(ctx, clIdty); client.IgnoreNotFound(err) != nil {
```
zerospiel: should we `patch` here instead of `update`?

api/v1beta1/providerinterface_types.go
```diff
@@ -42,7 +42,32 @@ type ProviderInterfaceSpec struct {
...
+	References []ClusterIdentityReference `json:"references,omitempty"`
+}
+
+// ClusterIdentityReference defines how to locate and resolve a referenced object
+// associated with a ClusterIdentity
+type ClusterIdentityReference struct {
+	GroupVersionKind `json:",inline"`
+	// NameFieldPath specifies the field path in the ClusterIdentity object where the name of
+	// the referenced object can be found
+	NameFieldPath string `json:"nameFieldPath"`
```
zerospiel, line 71: I believe we have to give examples (via `kubebuilder:example`) here and/or provide schema validation, otherwise we might end up with either `.path.subpath` or `path.subpath`, whereas the former will produce then an empty first element after the `strings.Split` call

Applicable for both fields

internal/controller/credential/clusteridentities.go
```diff
@@ -0,0 +1,332 @@
...
+	var (
+		errs error
+
+		result          = []*unstructured.Unstructured{clIdty}
+		sourceNamespace = systemNamespace
+		targetNamespace = cred.Namespace
+	)
+
+	for _, reference := range ci.References {
+		name, nameFound, err := unstructured.NestedString(clIdty.Object, strings.Split(reference.NameFieldPath, ".")...)
```
zerospiel, line 158: wrt the comment about validation of the first dot `<.>` char, here we might get `["", "path", "subpath"]`, and get nothing

internal/controller/credential/clusteridentities.go
```diff
@@ -0,0 +1,332 @@
...
+				cred.Spec.IdentityRef.Kind, objectKey(cred.Spec.IdentityRef.Namespace, clIdty.GetName()), reference.NameFieldPath))
+			continue
+		}
+
+		// when the namespaceFieldPath is unspecified, the provider requires the identity reference object
+		// to exist in the system namespace
+		var namespacePath []string
+
+		if reference.NamespaceFieldPath != "" {
+			namespacePath = strings.Split(reference.NamespaceFieldPath, ".")
```
zerospiel, line 175: ditto about the potential empty entry

internal/controller/credential/clusteridentities.go
```diff
@@ -0,0 +1,332 @@
...
+		clIdtyRef.SetNamespace(sourceNamespace)
+
+		if err := mgmtClient.Get(ctx, client.ObjectKey{Name: name, Namespace: sourceNamespace}, clIdtyRef); err != nil {
+			errs = errors.Join(errs, fmt.Errorf("failed to get ClusterIdentity reference object of Kind=%s %s: %w",
+				cred.Spec.IdentityRef.Kind, objectKey(sourceNamespace, name), err))
+			continue
+		}
+
+		// set the namespace of Credential on the distributed identity object
+		if reference.NamespaceFieldPath != "" && len(namespacePath) > 0 {
```
zerospiel: nit: the second condition `len(namespacePath) > 0` is redundant here (if the ns field path ref is not empty, the `namespacePath` won't be empty)

internal/controller/credential/clusteridentities.go
```diff
@@ -0,0 +1,332 @@
...
+			if err = unstructured.SetNestedField(clIdty.Object, targetNamespace, namespacePath...); err != nil {
```
zerospiel, line 201:
```suggestion
			if err = unstructured.SetNestedField(clIdtyRef.Object, targetNamespace, namespacePath...); err != nil {
```

shouldn't we set the desired object reference instead?

api/v1beta1/credential_types.go
```diff
@@ -24,6 +24,15 @@ const (
...
 	CredentialReadyCondition = "CredentialReady"
+
+	// CredentialLabelKeyPrefix is a label key prefix applied to all ClusterIdentity objects and their references.
+	// Each managed ClusterIdentity will have this label set in format of:
+	// k0rdent.mirantis.com/credential.<cred-namespace>.<cred-name>: true
+	// Which means that this ClusterIdentity is managed by the Credential `cred-namespace/cred-name`.
+	// One ClusterIdentity can be managed by multiple Credential objects.
+	CredentialLabelKeyPrefix = "k0rdent.mirantis.com/credential"
+
+	CredentialFinalizer = "k0rdent.mirantis.com/credential"
```
zerospiel, line 35: unrelated to the PR: some day (maybe with `v1beta2`) we have to migrate from the current type of `idendityRef` to something that CAPI has: instead of the `APIVersion` use the proper `APIGroup` (so the idea is to drop the `version` field)

internal/controller/credential/clusteridentities.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	return result, errs
+}
+
+func ensureClusterIdentityObject(ctx context.Context, rgnClient client.Client, cred *kcmv1.Credential, obj *unstructured.Unstructured) error {
+	l := ctrl.LoggerFrom(ctx)
+
+	existingObj := &unstructured.Unstructured{}
+	existingObj.SetAPIVersion(obj.GetAPIVersion())
+	existingObj.SetKind(obj.GetKind())
```
zerospiel:
```suggestion
    existingObj.SetGroupVersionKind(obj.GroupVersionKind())
```

nit: interchangeable though looks cleaner (imo)

internal/controller/credential/clusteridentities.go
eromanova: I used Update because we’re working with unstructured objects and need to modify both labels and the object’s content (which may reside under `data`, `spec`, or other sections). Since the exact field paths aren’t known, I'm not sure if Patch can be used here.

internal/controller/credential/clusteridentities.go
eromanova: It makes sense to add a validation rule for it. Thanks.

internal/controller/credential/clusteridentities.go
zerospiel: But in this particular function (and in the caller) there are no changes apart from `labels` or did I overlook something?

api/v1beta1/credential_types.go
zerospiel: Since it was just an observation, I resolve it

internal/controller/credential/clusteridentities.go
eromanova: No, we need to update the namespace of the ClusterIdentity reference (e.g., a Secret) in the Cluster Identity object (e.g., AzureClusterIdentity) to match the new namespace when the reference is copied.

zerospiel: Sorry, I still do not quite get it: so we got the source namespace from a cluster identity reference (`credentials.spec.identityRef`), then we change its namespace path to the target namespace (which is creds's namespace) and after we neither update this field on the API server nor use this field. What have I missed?

internal/controller/credential/clusteridentities.go
eromanova: Oh, correct. I messed it up with `ensureClusterIdentityObject`. Sorry. Will update this part.

internal/controller/credential/clusteridentities.go
eromanova: Sure, let me clarify this part because this may be confusing. Credential references the Cluster Identity object, and Cluster Identity itself can reference multiple objects (Secrets, etc.), for example, AzureClusterIdentity references the Secret. When I use the Access Management to copy the Credential across namespaces, I expect AzureClusterIdentity and its referenced Secret to exist in the system namespace. The access management controller copies the Credential to the target namespace, and the Cluster Identity distribution system should collect  all the required identity objects from the system namespace and copy them to the target namespace. But we should also update the references. Now, the new AzureClusterIdentity object should reference the Secret from the target namespace (that's exactly what this line does). The `namespaceFieldPath` is the path to the secret namespace in AzureClusterIdentity. We update the Secret namespace a little bit earlier (line 190: `clIdtyRef.SetNamespace(sourceNamespace)`).

internal/controller/credential_controller.go
```diff
@@ -65,14 +68,36 @@ func (r *CredentialReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if controllerutil.AddFinalizer(cred, kcmv1.CredentialFinalizer) {
+		if err = r.MgmtClient.Update(ctx, cred); err != nil {
+			l.Error(err, "failed to update Credential finalizers")
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: r.syncPeriod}, nil
+	}
+
+	var updated bool
+	if updated, err = labelsutil.AddKCMComponentLabel(ctx, r.MgmtClient, cred); updated || err != nil {
```
zerospiel: I didn't quite get the necessity of this `err` variable here and reassigning it globally, whereas we return immediately. I do not think setting transient errors to the condition's status is worth it. Won't it be easier to:

```go
	if controllerutil.AddFinalizer(cred, kcmv1.CredentialFinalizer) {
		if err := r.MgmtClient.Update(ctx, cred); err != nil {
			l.Error(err, "failed to update Credential finalizers")
			return ctrl.Result{}, err
		}
		return ctrl.Result{RequeueAfter: r.syncPeriod}, nil
	}

	if updated, err := labelsutil.AddKCMComponentLabel(ctx, r.MgmtClient, cred); updated || err != nil {
		if err != nil {
			l.Error(err, "adding component label")
		}
		return ctrl.Result{}, err
	}

	var err error
	defer func() {
		r.setReadyCondition(cred, err)
		err = errors.Join(err, r.updateStatus(ctx, cred))
	}()
```

internal/controller/credential_controller.go
```diff
@@ -151,6 +173,43 @@ func (r *CredentialReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	return ctrl.Result{RequeueAfter: r.syncPeriod}, nil
 }
 
+func (r *CredentialReconciler) delete(ctx context.Context, cred *kcmv1.Credential) (ctrl.Result, error) {
+	l := log.FromContext(ctx)
```
zerospiel:
```suggestion
	l := ctrl.LoggerFrom(ctx)
```

internal/controller/credential/clusteridentities.go
zerospiel: Thanks for the elaboration, the problem initially was that i haven't kept in my the `result = []*unstructured.Unstructured{clIdty}` line where the refernce is being put to the result initially

templates/provider/cluster-api-provider-openstack/templates/interface.yaml
```diff
@@ -9,7 +9,17 @@ spec:
...
       kind: OpenStackCluster
-  clusterIdentityKinds:
-    - Secret
-    - OpenStackClusterIdentity
+  clusterIdentities:
+    - group: ""
+      version: v1
+      kind: Secret
+    - group: infrastructure.cluster.x-k8s.io
+      version: v1alpha1 # Custom Cluster Identity kind for OpenStack. See: https://github.com/Mirantis/cluster-api-provider-openstack/pull/7
```
zerospiel: There should be a gigantic `TODO` or `NOTE` or `WARN` because it does not make much sense to have such a drift between the versions. We can't drop this identity as of now, can we?

internal/controller/credential/clusteridentities.go
```diff
@@ -0,0 +1,335 @@
...
+	if shouldDelete {
+		if err := rgnClient.Delete(ctx, clIdty); client.IgnoreNotFound(err) != nil {
+			return fmt.Errorf("failed to delete Cluster Identity object of Kind %s: %s: %w", clIdty.GetKind(), objectKeyUnstructured(clIdty), err)
+		}
+		return nil
+	}
+	original := clIdty.DeepCopy()
+	clIdty.SetLabels(labels)
+
+	if err := rgnClient.Patch(ctx, clIdty, client.MergeFrom(original)); client.IgnoreNotFound(err) != nil {
```
zerospiel, line 312: should we add `patch` verb to the corresponding RBAC lists?

templates/provider/cluster-api-provider-openstack/templates/interface.yaml
zerospiel: Maybe having this comment suffices, so do as you think would be better in this particular case and feel free to resolve the comment

eromanova: If we drop this identity from the ProviderInterface, we disable the usage of OpenStackClusterIdentity completely (there is a validation https://github.com/k0rdent/kcm/blob/main/internal/webhook/clusterdeployment_webhook.go#L84). Right, we can't drop it.

internal/controller/credential_controller.go
```diff
@@ -174,15 +172,17 @@ func (r *CredentialReconciler) update(ctx context.Context, cred *kcmv1.Credentia
...
-	}
-
-	if err := credential.ReleaseClusterIdentities(ctx, rgnClient, cred); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to release all Cluster Identities for Credential %s: %w", client.ObjectKeyFromObject(cred), err)
+	if err == nil {
+		if err := credential.ReleaseClusterIdentities(ctx, rgnClient, cred); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to release all Cluster Identities for Credential %s: %w", client.ObjectKeyFromObject(cred), err)
+		}
+	} else {
+		// if the region is inaccessible, we can’t release ClusterIdentities remove the finalizer to proceed with cleanup
```
zerospiel, line 183: just wondering: is it okay then to leave "orphan" objects? i see there are 2 options:

1. the current one, leaving orphan objects
2. introduce timeouts, where we increase the complexity: add some timestamp to the status/annotation like `firstSeenOnDelete` or `firstDeletionTry` or whatever, and if quite some time have passed since that (and if it's not empty), then okay, remove the finalizer, otherwise retry

I can't suggest which one is better because I am not in the loop of the distribution; maybe the second option is too complex and it is not worth implementing.

eromanova: I think that’s a good idea. The region might be temporarily inaccessible, but once it becomes reachable again, we can proceed with the identity objects cleanup. I'll think about the best approach. Thanks.

eromanova: Will be addressed in one of the upcoming PRs

Issue #1921, opened 2025-08-29
[enhancement] allow pausing of service deployment for specific clusters


BROngineer:
Allow skipping services (all services defined in `.spec.serviceSpec.services[]`) deployment for ClusterDeployments.

Depends on: 
- https://github.com/projectsveltos/addon-controller/issues/1358
- https://github.com/k0rdent/kcm/issues/1930

**DoD**:
- [ ] it is possible to pause reconciliation of the ServiceSet (and underlying Profile)
- [ ] scenario is covered in e2e tests
- [ ] documentation updated

PR #2084, opened 2025-10-14
feat: pause reconciliation of sveltos profiles via serviceset annotation

kylewuolle:
This pr allows pausing of sveltos profile / clusterprofile by setting an annotation on the serviceset

https://github.com/k0rdent/kcm/issues/1921

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -403,7 +403,7 @@ func (*ServiceSetReconciler) createOrUpdateProfile(ctx context.Context, rgnClien
 		return fmt.Errorf("failed to get Profile: %w", err)
 	}
 
-	externalPaused, internalPaused := handlePauseAnnotations(&profile.ObjectMeta, serviceSet)
+	annotationsUpdated := handlePauseAnnotations(&profile.ObjectMeta, serviceSet)
 
 	switch {
 	// we already excluded all errors except NotFound
@@ -421,7 +421,7 @@ func (*ServiceSetReconciler) createOrUpdateProfile(ctx context.Context, rgnClien
 		}
 	// if profile spec is not equal to the spec we just created,
 	// we need to update it
-	case externalPaused != internalPaused || !equality.Semantic.DeepEqual(profile.Spec, *spec):
+	case annotationsUpdated || !equality.Semantic.DeepEqual(profile.Spec, *spec):
 		profile.OwnerReferences = []metav1.OwnerReference{*ownerReference}
 		profile.Spec = *spec
 		if err = rgnClient.Update(ctx, profile); err != nil {
@@ -443,7 +443,7 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
 		return fmt.Errorf("failed to get Profile: %w", err)
 	}
 
-	externalPaused, internalPaused := handlePauseAnnotations(&profile.ObjectMeta, serviceSet)
+	annotationsUpdated := handlePauseAnnotations(&profile.ObjectMeta, serviceSet)
 
 	switch {
 	// we already excluded all errors except NotFound
@@ -460,7 +460,7 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
 		}
 	// if profile spec is not equal to the spec we just created,
 	// we need to update it
-	case externalPaused != internalPaused || !equality.Semantic.DeepEqual(profile.Spec, *spec):
+	case annotationsUpdated || !equality.Semantic.DeepEqual(profile.Spec, *spec):
 		profile.OwnerReferences = []metav1.OwnerReference{*ownerReference}
 		profile.Spec = *spec
 		if err = rgnClient.Update(ctx, profile); err != nil {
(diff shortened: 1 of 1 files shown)


config/dev/kcm_values.yaml
```diff
@@ -15,3 +15,6 @@ controller:
 telemetry:
   mode: disabled
   interval: 1m
+regional:
+  velero:
+    enabled: false
```
zerospiel: I vote for this to be added! We neither validate it in our tests, nor is it being used in common development. @eromanova wdyt?

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -403,6 +403,26 @@ func (*ServiceSetReconciler) createOrUpdateProfile(ctx context.Context, rgnClien
...
+
+	_, pauseAnnotationExists := annotations[addoncontrollerv1beta1.ProfilePausedAnnotation]
+
+	if paused {
+		annotations[addoncontrollerv1beta1.ProfilePausedAnnotation] = "true"
+	} else if pauseAnnotationExists {
+		delete(annotations, addoncontrollerv1beta1.ProfilePausedAnnotation)
+	}
+
+	profile.SetAnnotations(annotations)
```
zerospiel: Consider refactoring to a single function:
```go
func handlePauseAnnotations(profile *addoncontrollerv1beta1.ClusterProfile, serviceSet *kcmv1.ServiceSet) (internallyPaused, externallyPaused bool) {
	profileAnnotations := profile.GetAnnotations()
	if profileAnnotations == nil {
		profileAnnotations = make(map[string]string)
	}

	_, internallyPaused = serviceSet.GetAnnotations()[kcmv1.ServiceSetPausedAnnotation]
	_, externallyPaused = profileAnnotations[addoncontrollerv1beta1.ProfilePausedAnnotation]

	if internallyPaused {
		profileAnnotations[addoncontrollerv1beta1.ProfilePausedAnnotation] = "true"
	} else if externallyPaused {
		delete(profileAnnotations, addoncontrollerv1beta1.ProfilePausedAnnotation)
	}

	profile.SetAnnotations(profileAnnotations)

	return internallyPaused, externallyPaused
}
```

Consider calling `make()` on map. No need for the null map check. Consider renaming variables

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -419,7 +439,7 @@ func (*ServiceSetReconciler) createOrUpdateProfile(ctx context.Context, rgnClien
 		}
 	// if profile spec is not equal to the spec we just created,
 	// we need to update it
-	case !equality.Semantic.DeepEqual(profile.Spec, *spec):
+	case !equality.Semantic.DeepEqual(profile.Spec, *spec) || pauseAnnotationExists && !paused:
```
zerospiel:
```suggestion
	case pauseAnnotationExists && !paused || !equality.Semantic.DeepEqual(profile.Spec, *spec):
```

Please use short-circuit evaluation

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -441,6 +461,26 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
...
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+
+	paused := false
+	if serviceSet.GetAnnotations() != nil {
+		_, paused = serviceSet.GetAnnotations()[kcmv1.ServiceSetPausedAnnotation]
+	}
+
+	_, pauseAnnotationExists := annotations[addoncontrollerv1beta1.ProfilePausedAnnotation]
```
zerospiel: don't we have to use the same check for the `externallyPaused && !internallyPaused` in order to update the `ClusterProfile` here, in the `createOrUpdateClusterProfile` method?

zerospiel (dismissed):

internal/controller/adapters/sveltos/serviceset_controller.go
kylewuolle: Refactored thanks!

internal/controller/adapters/sveltos/serviceset_controller.go
kylewuolle: Good point I've added that!

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -456,7 +460,7 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
 		}
 	// if profile spec is not equal to the spec we just created,
 	// we need to update it
-	case !equality.Semantic.DeepEqual(profile.Spec, *spec):
+	case externalPaused && !internalPaused || !equality.Semantic.DeepEqual(profile.Spec, *spec):
```
zerospiel: now i have a question, appreciate if you could clarify: so we actually update a profile if it has an anno and a service set has NO anno, right? but it seems counterintuitive, i would say we have to update the profile if our service set has an anno -> we set an anno to the profile -> it needs to be updated. or am i missing something?

besides, what if a profile has been manually paused by an admin, and he hasn't set an anno to the service set (say, on purpose), in that matter the current code will just delete the annotation from the profile

if my thoughts are correct, i would modify the function to just return a boolean-typed value that indicates whether a profile should or should not be updated

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -466,6 +470,26 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
...
 }
 
+func handlePauseAnnotations(profile *metav1.ObjectMeta, serviceSet *kcmv1.ServiceSet) (externalPaused, internalPaused bool) {
+	profileAnnotations := profile.GetAnnotations()
+	if profileAnnotations == nil {
+		profileAnnotations = make(map[string]string)
+	}
+
+	internalPaused = serviceSet.GetAnnotations() != nil && serviceSet.GetAnnotations()[kcmv1.ServiceSetPausedAnnotation] != ""
+	externalPaused = profileAnnotations[addoncontrollerv1beta1.ProfilePausedAnnotation] != ""
```
zerospiel:
```suggestion
	_, internalPaused = serviceSet.GetAnnotations()[kcmv1.ServiceSetPausedAnnotation]
	_, externalPaused = profileAnnotations[addoncontrollerv1beta1.ProfilePausedAnnotation]
```

otherwise, it works incorrectly when either of annos has an empty value

and there is no need to check if the map is `nil` during the read

config/dev/kcm_values.yaml
zerospiel: okay, seems there is no conclusion, so let's revert it

internal/controller/adapters/sveltos/serviceset_controller.go
kylewuolle: We have to handle the unpausing too here. I've changed the code to just check for externalPaused != internalPaused. For the case where an admin may add the annotation to the Profile I think they would have to use a self managed one in that case or only set the pause annotation on ServiceSet and leave the Profile alone.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -466,6 +470,26 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
...
+	switch {
+	case internalPaused:
+		profileAnnotations[addoncontrollerv1beta1.ProfilePausedAnnotation] = "true"
+	case externalPaused:
+		delete(profileAnnotations, addoncontrollerv1beta1.ProfilePausedAnnotation)
+	}
+
+	profile.SetAnnotations(profileAnnotations)
+	return externalPaused, internalPaused
+}
```
zerospiel:
```suggestion
func handlePauseAnnotations(profile *metav1.ObjectMeta, serviceSet *kcmv1.ServiceSet) (annoUpdated bool) {
	profileAnnotations := profile.GetAnnotations()
	if profileAnnotations == nil {
		profileAnnotations = make(map[string]string)
	}

	_, internalPaused := serviceSet.GetAnnotations()[kcmv1.ServiceSetPausedAnnotation]
	_, externalPaused := profileAnnotations[addoncontrollerv1beta1.ProfilePausedAnnotation]
	
	if internalPaused == externalPaused {
            return false
        }

	switch {
	case internalPaused:
		profileAnnotations[addoncontrollerv1beta1.ProfilePausedAnnotation] = "true"
	case externalPaused:
		delete(profileAnnotations, addoncontrollerv1beta1.ProfilePausedAnnotation)
	}

	profile.SetAnnotations(profileAnnotations)
	return true
}
```

would it be easier to return a single value that answers `should we update the given profile?`

BROngineer: Agree. @kylewuolle please update this function. Aside from that lgtm.

PR #2096, opened 2025-10-19
feat: support ClusterDeployment reference in the Region spec

eromanova:

This PR includes:
1. Adds support for referencing an existing ClusterDeployment in the Region object.
2. Copies the ClusterDeployment’s kubeconfig to the system namespace to install regional cluster components.
3. Fixes the owner reference of the kubeconfig Secret copied to the ClusterDeployment namespace.

Related task #1905

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -436,9 +436,16 @@ func (*ClusterDeploymentReconciler) setStatusRegion(scope clusterScope, cd *kcmv
 	}
 }
 
+// copyRegionalKubeConfigSecret copies the Regional cluster kubeconfig to the namespace of the ClusterDeployment
+// when the kubeconfig secret reference is defined in the Region.
+// To deploy the cluster template into the Regional cluster, its kubeconfig must be present
+// in the same namespace as the ClusterDeployment.
 func (r *ClusterDeploymentReconciler) copyRegionalKubeConfigSecret(ctx context.Context, scope clusterScope) error {
 	cd := scope.cd
-	if scope.cred.Spec.Region == "" || scope.region == nil || cd.Namespace == r.SystemNamespace {
+
+	// nothing to copy when the region does not have kubeConfig reference defined or the namespace equals
+	// the system namespace.
+	if scope.cred.Spec.Region == "" || scope.region == nil || scope.region.Spec.KubeConfig == nil || cd.Namespace == r.SystemNamespace {
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


internal/controller/region/controller.go
```diff
@@ -204,6 +214,38 @@ func (r *Reconciler) update(ctx context.Context, rgnlClient client.Client, restC
...
+	// the system namespace.
+	if region == nil || region.Spec.ClusterDeployment == nil || region.Spec.ClusterDeployment.Namespace == r.SystemNamespace {
+		return sourceKubeConfigSecretRef, nil
+	}
+
+	if sourceKubeConfigSecretRef == nil {
+		return nil, errors.New("source kubeconfig secret reference is nil")
+	}
+
+	nameOverride := region.Spec.ClusterDeployment.Namespace + "." + sourceKubeConfigSecretRef.Name
```
zerospiel, line 232: quick question: why should we change the original name?

internal/controller/clusterdeployment_controller.go
```diff
@@ -436,17 +436,33 @@ func (*ClusterDeploymentReconciler) setStatusRegion(scope clusterScope, cd *kcmv
...
 	kubeConfigSecretRef, err := kubeutil.GetRegionalKubeconfigSecretRef(scope.region)
 	if err != nil {
 		return fmt.Errorf("failed to get kubeconfig secret reference: %w", err)
 	}
-	return kubeutil.CopySecret(ctx, r.MgmtClient, r.MgmtClient, client.ObjectKey{Namespace: r.SystemNamespace, Name: kubeConfigSecretRef.Name}, cd.Namespace, scope.region, nil)
+	return kubeutil.CopySecret(
+		ctx,
+		r.MgmtClient,
+		r.MgmtClient,
+		client.ObjectKey{Namespace: r.SystemNamespace, Name: kubeConfigSecretRef.Name},
```
zerospiel, line 460: i just want to elaborate on the approach:
- cld controller expects the kubeconfig reference to exist in the system namespace (why?) on the mgmt cluster (target is the same as the source)
- cld controller copies the reference from the system to a target (cld's) namespace on the mgmt cluster (target is the same as the source)
- in the meantime, the region controller copies the reference from the cld's namespace to the system namespce on the mgmt cluster (target is the same as the source) but changes the name along the way

Isn't this kind of a cycle (system -> target -> system with changed name)?

internal/controller/region/controller.go
eromanova: This is needed to ensure the uniqueness of the secret that we copy. There can be situations when the ClusterDeployment with the same name is deployed in different namespaces, and then both are registered as a regional cluster.

internal/controller/clusterdeployment_controller.go
eromanova: There will be no cycle since `spec.clusterDeployment` and `spec.kubeConfig` are mutually exclusive in the Region object. When we reference ClusterDeployment in the Region object, we need the kubeconfig to exist in the system namespace to install regional components in it. When we reference the kubeconfig secret in the Region object, we expect this secret to exist in the system namespace. But we need to copy it to the namespace where the ClusterDeployment (deployed in this region) will be created, since flux requires the kubeconfig secret to exist in the same namespace as the HelmRelease.

PR #2111, opened 2025-10-21
fix: templates: aws-csi image paths; azure required parameters

### LINKED ISSUE #2087: [bug] kcm allows creation of cluster without vmSize specified on azure standalone
**Describe the bug**
When creating a cluster from the standalone k0s chart "vmSize" (which specifies the machine type) is allowed to be set to an empty string and leaves a cryptic error. 
```
  failureMessage: 'failed to get VM SKU  in compute api: reconcile error that cannot
    be recovered occurred: resource sku with name '''' and category ''virtualMachines''
    not found in location ''westus''. Object will not be requeued'
  failureReason: InvalidConfiguration
```

**To Reproduce**
1. install k0rdent
2. create a clusterdeployment but DO NOT specify the vmSize field
3. Cluster will attempt to create but fail with a cryptic error

**Expected behavior**
The clusterdeployment object should propegate a more useful message or an admission webhook should require that field be set if it is not specified

a13x5:
Resolves #2087

Adds required fields for Azure templates as well as actual required parameters on the chart level. This is needed to have clear error, since CAPZ won't check if vmsize or subscription ID is correct. Thus the cause cluster creation error is unclear to the user and if fails much later.

Fixed some some typos in aws-csi images paths

### DIFF
--- a/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
+++ b/templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
@@ -135,7 +135,7 @@ spec:
                         repository: {{ $global.registry }}/kubernetes-csi/external-attacher
                     snapshotter:
                       image:
-                        repository: {{ $global.registry }}kubernetes-csi/external-snapshotter/csi-snapshotter
+                        repository: {{ $global.registry }}/external-snapshotter/csi-snapshotter
                     livenessProbe:
                       image:
                         repository: {{ $global.registry }}/kubernetes-csi/livenessprobe
--- a/config/dev/aws-clusterdeployment.yaml
+++ b/config/dev/aws-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: aws-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
-  template: aws-standalone-cp-1-0-14
+  template: aws-standalone-cp-1-0-15
   credential: aws-cluster-identity-cred
   config:
     clusterLabels: {}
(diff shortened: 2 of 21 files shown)


templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
```diff
@@ -135,7 +135,7 @@ spec:
                         repository: {{ $global.registry }}/kubernetes-csi/external-attacher
                     snapshotter:
                       image:
-                        repository: {{ $global.registry }}kubernetes-csi/external-snapshotter/csi-snapshotter
+                        repository: {{ $global.registry }}/external-snapshotter/csi-snapshotter
```
eromanova, line 138: Just curious, should we update the path to the `csi-snapshotter` image in azure standalone and hosted templates as well? https://github.com/k0rdent/kcm/blob/v1.4.0/templates/cluster/azure-standalone-cp/templates/k0scontrolplane.yaml#L131

a13x5: Those are different images, here the csi-snapshotter will expand to:

`public.ecr.aws/eks-distro/kubernetes-csi/external-snapshotter/csi-snapshotter:v8.0.1-eks-1-30-10` -> `global.registry/external-snapshotter/csi-snapshotter:v8.0.1-eks-1-30-10`

On azure it will be:

`mcr.microsoft.com/oss/kubernetes-csi/csi-snapshotter:v8.0.1` -> `global.registry/kubernetes-csi/csi-snapshotter:v8.0.1`

And as weird as it is even though both are based on upstrean csi-snapshotter v8.0.1 the contents of the images certainly contains vendor-specific patches.

PR #2113, opened 2025-10-22
feat: regional telemetry collection

### LINKED ISSUE #1909: Adapt telemetry functionality for regional support
1. Update the controller to use the regional client when accessing objects from the regional API.
2. Validate telemetry to ensure it functions correctly for regional clusters.

zerospiel:
Migrates telemetry into a separated binary.

Migrates telemetry from kcm to kcm-regional chart, introduces
several new settings for the dedicated telemetry controller.

Amends goreleaser settings according to the new binary.

Amends makefile targets according to the new binary.

Removes legacy telemetry events from the main controllers binary.

Closes #1909

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -510,13 +510,21 @@ dev-gcp-creds: envsubst
 dev-apply: kind-deploy registry-deploy dev-push dev-deploy dev-templates dev-release ## Apply the development environment by deploying the kind cluster, local registry and the KCM helm chart.
 
 .PHONY: dev-upgrade
-dev-upgrade: generate-all dev-push dev-templates ## Upgrade dev environment and wait until Management is upgraded to the new Release and ready
+dev-upgrade: VALUES_FILE ?= config/dev/kcm_values.yaml
+dev-upgrade: yq generate-all dev-push dev-templates ## Upgrade dev environment and wait until Management is upgraded to the new Release and ready
 	@echo "Applying new Release object: kcm-$(FQDN_VERSION)"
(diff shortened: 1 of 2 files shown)


cmd/telemetry/main.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+var (
+	scheme   = schemeutil.MustGetManagementScheme()
+	setupLog = ctrl.Log.WithName("setup")
+)
+
+func main() {
+	for _, f := range []func(*runtime.Scheme) error{
+		addoncontrollerv1beta1.AddToScheme,
```
eromanova, line 44: `addoncontrollerv1beta1` scheme is already a part of the management scheme returned from `MustGetManagementScheme()`.

cmd/telemetry/main.go
```diff
@@ -0,0 +1,132 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+
+	"github.com/K0rdent/kcm/internal/telemetry"
+	schemeutil "github.com/K0rdent/kcm/internal/util/scheme"
+)
+
+var (
+	scheme   = schemeutil.MustGetManagementScheme()
```
eromanova: Since the telemetry controller can run in a regional cluster, should the scheme also be configured as regional, or will the current configuration function properly?

cmd/telemetry/main.go
zerospiel: Not quite, it is added on the management only if sveltos controller knob is turned on (which is true by default)

eromanova (approved):

Issue #1961, opened 2025-09-04
[bug] ServiceSet not reflecting ClusterDeployment


wahabmk:
# Description

**To Reproduce**
1. Create ClusterDeployment and deploy 2 instances of ingress-nginx service via [ingress-nginx ServiceTemplate](https://catalog.k0rdent.io/latest/apps/ingress-nginx/).
2. Wait for the ClusterDeployment to reconcile and show error in status.
3. Now remove either one or both of the services (it won't reflect in the ServiceSet).
4. Now add some other new service (again it won't reflect in the ServiceSet).

So Even if another service is added to the ClusterDeployment, it is not longer reflected in the ServiceSet and it becomes impossible to add or remove any services.

**Expected behavior**

After removing the services in step 3:

- The services should be removed from the ServiceSet and the target cluster.
- Any new service added should also be reflected in the ServiceSet and installed on the target cluster.

**Additional context**
KCM Version: v1.3.1

# Example

Created a ClusterDeployment with the following spec in which ingress-nginx is installed twice in different namespaces:
```sh
apiVersion: k0rdent.mirantis.com/v1beta1
kind: ClusterDeployment
metadata:
  name: wali-dev-1
  namespace: kcm-system
  labels:
    owner: dev-team
spec:
  template: aws-standalone-cp-1-0-14
  . . .
  serviceSpec:
    provider:
      name: ksm-projectsveltos
      config:
        # priority: 1000
        continueOnError: true
        stopOnConflict: false
    services:
      - template: ingress-nginx-4-11-5
        name: nginx
        namespace: nginx
      - template: ingress-nginx-4-11-5
        name: nginx2
        namespace: nginx2
      - template: hello-world
```

The status shows that there is failure which is expected:
```sh
➜  ~ kubectl -n kcm-system get clusterdeployment wali-dev-1 -o yaml
apiVersion: k0rdent.mirantis.com/v1beta1
kind: ClusterDeployment
. . .
spec:
  . . .
  serviceSpec:
    continueOnError: false
    priority: 100
    provider:
      config:
        continueOnError: true
        stopOnConflict: false
      name: ksm-projectsveltos
    services:
    - name: nginx
      namespace: nginx
      template: ingress-nginx-4-11-5
    - name: nginx2
      namespace: nginx2
      template: ingress-nginx-4-11-5
    stopOnConflict: false
    syncMode: Continuous
status:
  . . .
  services:
  - failureMessage: |
[...]

PR #2142, opened 2025-10-29
fix: ServiceSet update bug if status isn't Deployed

wahabmk:
# Description

Fixes https://github.com/k0rdent/kcm/issues/1961. To see detailed description of this bug see the comments below.

### DIFF
--- a/internal/serviceset/util.go
+++ b/internal/serviceset/util.go
@@ -17,8 +17,10 @@ package serviceset
 import (
 	"context"
 	"crypto/sha256"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"slices"
 
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -345,8 +347,8 @@ func GetServiceSetWithOperation(
 	}
 
 	serviceSetRequired := len(operationReq.Services) > 0 || operationReq.PropagateCredentials
-	switch {
-	case err != nil:
+
+	if err != nil {
 		if serviceSetRequired {
 			l.V(1).Info("Pending services to deploy, ServiceSet does not exist", "operation", kcmv1.ServiceSetOperationCreate)
 			serviceSet.SetName(operationReq.ObjectKey.Name)
@@ -355,16 +357,25 @@ func GetServiceSetWithOperation(
 		}
 		l.V(1).Info("No services to deploy, ServiceSet does not exist", "operation", kcmv1.ServiceSetOperationNone)
 		return nil, kcmv1.ServiceSetOperationNone, nil
-	case !serviceSetRequired:
+	}
+
+	if !serviceSetRequired {
 		l.V(1).Info("No services to deploy, ServiceSet exists", "operation", kcmv1.ServiceSetOperationDelete)
 		return serviceSet, kcmv1.ServiceSetOperationDelete, nil
-	case needsUpdate(serviceSet, operationReq.ProviderSpec, operationReq.Services):
+	}
+
+	update, err := needsUpdate(serviceSet, operationReq.ProviderSpec, operationReq.Services)
+	if err != nil {
+		return nil, "", err
+	}
+	if update {
 		l.V(1).Info("Pending services to deploy, ServiceSet exists", "operation", kcmv1.ServiceSetOperationUpdate)
 		return serviceSet, kcmv1.ServiceSetOperationUpdate, nil
-	default:
-		l.V(1).Info("No actions required, ServiceSet exists", "operation", kcmv1.ServiceSetOperationNone)
-		return serviceSet, kcmv1.ServiceSetOperationNone, nil
 	}
+
+	// default case
+	l.V(1).Info("No actions required, ServiceSet exists", "operation", kcmv1.ServiceSetOperationNone)
+	return serviceSet, kcmv1.ServiceSetOperationNone, nil
 }
 
 // needsUpdate checks if the ServiceSet needs to be updated based on the ClusterDeployment spec.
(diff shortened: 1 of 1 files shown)


internal/serviceset/util.go
```diff
@@ -376,32 +376,8 @@ func needsUpdate(
 	providerSpec kcmv1.StateManagementProviderConfig,
 	services []kcmv1.Service,
 ) bool {
-	// we'll need to update provider configuration if it was changed.
-	if !equality.Semantic.DeepEqual(providerSpec, serviceSet.Spec.Provider) {
-		return true
-	}
```
wahabmk, line 382: Removed this check for the following reasons:

- We don't need it anymore because `provider.name` is [immutable](https://github.com/k0rdent/kcm/blob/eea7692411036fd80418ffecd46656dfcbe60e28/api/v1beta1/serviceset_types.go#L118-L121).
- In some cases the comparison failed even though the spec was same because the JSON string in `Config` was parsed differently, e.g:
```sh
providerSpec.name=ksm-projectsveltos, providerSpec.SelfManagement=true, providerSpec.Config={"syncMode":"Continuous","priority":100}
serviceSet.Spec.Provider.name=ksm-projectsveltos, serviceSet.Spec.Provider.SelfManagement=true, serviceSet.Spec.Provider.Config={"priority":100,"syncMode":"Continuous"}
```
- This check was masking the bug mentioned in [comment](https://github.com/k0rdent/kcm/pull/2142/files#r2471574141) below.

internal/serviceset/util.go
```diff
@@ -412,18 +388,11 @@ func needsUpdate(
 		}
 	}
 
-	// This is comparing the spec to the status of the ServiceSet fetched from kubernetes.
-	// The difference between observed and desired services state means that ServiceSet was not fully
-	// deployed yet. Therefore we won't update ServiceSet until that.
-	if !equality.Semantic.DeepEqual(observedServiceStateMap, desiredServiceStateMap) {
-		return false
-	}
```
wahabmk, line 420: Removing this check for the following reasons:

- If a service deployment fails with `Failed` state, then any update to the services spec in CD/MCS is not reflected in associated ServiceSet as this equality check fails because `Deployed != Failed` (we are setting `Deployed` state in line 403)
- For example consider the following MCS:
```yaml
kind: MultiClusterService
metadata:
  name: mcs3
spec:
  clusterSelector:
    matchLabels:
      k0rdent.mirantis.com/management-cluster: "true"
      sveltos-agent: present
  serviceSpec:
    provider:
      selfManagement: true
    services:
      - template: cert-manager-1-18-2
        name: cert-manager
        namespace: cert-manager
        values: |
          cert-manager:
            crds:
              enabled: false    # so that deployment fails
```
- The  ServiceSet will show that cert-manager service failed because the CRDs don't exist (or delete them manually beforehand to simulate failure):
```yaml
kind: ServiceSet
metadata:
  . . .
spec:
  . . .
  services:
  - name: cert-manager
    namespace: cert-manager
    template: cert-manager-1-18-2
    values: |
      cert-manager:
        crds:
          enabled: false
status:
  conditions:
  - lastTransitionTime: "2025-10-29T02:34:11Z"
    message: Profile is ready
    observedGeneration: 1
    reason: ServiceSetProfileReady
    status: "True"
    type: ServiceSetProfile
  deployed: false
  provider:
    ready: true
  services:
  - failureMessage: context deadline exceeded
    lastStateTransitionTime: "2025-10-29T02:39:31Z"
    name: cert-manager
    namespace: cert-manager
    state: Failed
    template: cert-manager-1-18-2
    type: Helm
    version: cert-manager-1-18-2
```
- Now if we change the service spec for the MCS to:
```yaml
. . .
    services:
      - template: cert-manager-1-18-2
        name: cert-manager
        namespace: cert-manager
        values: |
          cert-manager:
            crds:
              enabled: true  
. . .
```
- ❌ We expect to see ServiceSet updated as well, but in reality the ServiceSet remains at generation 1 with old values.
- If we add another service to the MCS:
```yaml
kind: MultiClusterService
metadata:
  . . .
spec:
  . . .
  serviceSpec:
    . . .
    services:
    - name: cert-manager
      namespace: cert-manager
      template: cert-manager-1-18-2
      values: |
        cert-manager:
          crds:
            enabled: true
    - name: ingress-nginx
      namespace: ingress-nginx
      template: ingress-nginx-4-13-0
status:
  conditions:
  . . .
    type: MultiClusterServiceDependencyValidation
  - lastTransitionTime: "2025-10-29T02:34:11Z"
    message: 0/1
    reason: Failed
    status: "False"
    type: ClusterInReadyState
  - lastTransitionTime: "2025-10-29T02:34:11Z"
    message: 0/1 Clusters are ready.
    reason: Failed
    status: "False"
    type: Ready
  observedGeneration: 5
  servicesUpgradePaths:
  - availableUpgrades:
    - upgradePaths:
      - cert-manager-1-18-2
    name: cert-manager
    namespace: cert-manager
    template: cert-manager-1-18-2
  - availableUpgrades:
    - upgradePaths:
      - ingress-nginx-4-13-0
    name: ingress-nginx
    namespace: ingress-nginx
    template: ingress-nginx-4-13-0
```
- ❌ The expectation is that the new service will get added to the associated ServiceSet, but in reality the ServiceSet is not updated and remains at generation 1.
- Even removing the cert-manager service from MCS does nothing (because the nginx service still exists and the execution still enters [`needsUpdate`](https://github.com/k0rdent/kcm/blob/eea7692411036fd80418ffecd46656dfcbe60e28/internal/serviceset/util.go#L361) case (and returns false due to the equality check failing).
- So the only way to fix this is to remove ALL services from the MCS so that the execution would enter the [delete](https://github.com/k0rdent/kcm/blob/eea7692411036fd80418ffecd46656dfcbe60e28/internal/serviceset/util.go#L358-L360) case instead, which is not really a fix because this deletes the ServiceSet.

wahabmk: We have the following service states:
https://github.com/k0rdent/kcm/blob/eea7692411036fd80418ffecd46656dfcbe60e28/api/v1beta1/serviceset_types.go#L30-L39

So if we still want to keep this check we can do something like:

```
needsUpdate := true

for each svc in ServiceState's spec {
  if that svc's state in ServiceSet's status == "Provisioning" or "Pending" {
    needsUpdate := false
  }
}

returns needsUpdate
```

But there can still be problems with this because:
- We cannot be sure how long the Provisioning or Pending state will last in the future with further development work in either KSM or Sveltos. 
- The deployment of other unrelated services will be blocked even if 1 service is stuck in Provisioning or Pending.

So the safer option to me seems to be to remove this check altogether and let the underlying provisioner (Sveltos) handle it.

zerospiel (dismissed):

internal/serviceset/util.go
wahabmk: Actually, I think we need to at least keep a check for changes in `providerSpec.Config` (even if `providerSpec.Name` is immutable) so that any provider specific config is reflected in the ServiceSet. I'll confirm.

BROngineer (approved):
Thanks @wahabmk . All changes looks reasonable. Running e2e tests.

PR #2155, opened 2025-11-07
chore: fallback to aws when no config provided in provider check

BROngineer:
Fallback to `aws: []` when no config provided for `authorize` job. Empty config makes CI exit with no e2e tests run, due to checks return false on `Check cloud providers presence` and `Check remote presence` steps.

### DIFF
--- a/.github/actions/check-provider-in-config/action.yml
+++ b/.github/actions/check-provider-in-config/action.yml
@@ -32,11 +32,11 @@ runs:
         echo "cloud_present=false" >> "$GITHUB_OUTPUT"
 
         if [ -z "${{ inputs.config }}" ]; then
-          echo "No config provided — skipping checks"
-          exit 0
+          echo "No config provided — fallback to aws"
+          echo "aws: []" > config.yaml
+        else
+          echo "${{ inputs.config }}" | base64 -d > config.yaml
         fi
-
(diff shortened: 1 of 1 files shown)


.github/actions/check-provider-in-config/action.yml
```diff
@@ -32,12 +32,12 @@ runs:
         echo "cloud_present=false" >> "$GITHUB_OUTPUT"
 
         if [ -z "${{ inputs.config }}" ]; then
-          echo "No config provided — skipping checks"
-          exit 0
+          echo "No config provided — fallback to aws"
+          echo "aws: []" > config.yaml
+        else
+          echo "${{ inputs.config }}" | base64 -d > config.yaml
```
zerospiel, line 38: will that work for the onprem providers (only vsphere atm)?

zerospiel: ah, actually it should work, we do not exit but rather pushing the sentinel, and then check against it; thus, the onprem should not be returned

BROngineer: exactly. I just write default "aws: []" to config.yaml if no config provided, otherwise write config from input and proceed with further validation

zerospiel (approved):

PR #2181, opened 2025-11-17
chore: kcm-regional global registry values + imagePullSecrets

### LINKED ISSUE #2129: [enhancement] Support globalRegistry for kcm and kcm-regional images
For the kcm chart `controller.globalRegistry` should be used
For the regional chart it should be `global.registry`

a13x5:
Resolves #2129 

Adds global registry configuration for kcm-regional.

Additionally adds missing imagePullSecrets configuration option for both kcm and kcm-regional charts

### DIFF
--- a/templates/provider/kcm-regional/templates/deployment.yaml
+++ b/templates/provider/kcm-regional/templates/deployment.yaml
@@ -1,3 +1,4 @@
+{{- $global := .Values.global | default dict }}
 {{- $tel := .Values.telemetry -}}
 {{- $telController := .Values.telemetry.controller -}}
 {{- $isTelemetryEnabled := ne $tel.mode "disabled" -}}
@@ -23,7 +24,11 @@ spec:
       {{- if $isLocalTelemetry }}
       initContainers:
         - name: telemetry-data-permissions
-          image: busybox:latest
+          image: {{ if $global.registry }}
+          {{- $global.registry }}/busybox
+          {{- else }}
+          {{- $telController.image.initContainer.repository }}
+          {{- end }}:{{ $telController.image.initContainer.tag }}
           command: ["sh","-c","chown -R 65532:65532 {{ $baseDir }} || true"]
           securityContext:
             runAsUser: 0
@@ -56,7 +61,11 @@ spec:
         - --pprof-bind-address={{ $telController.debug.pprofBindAddress }}
         command:
         - /telemetry
-        image: {{ $telController.image.repository }}:{{ $telController.image.tag | default .Chart.AppVersion }}
+        image: {{ if $global.registry }}
+        {{- $global.registry }}/kcm-telemetry
+        {{- else }}
+        {{- $telController.image.repository }}
+        {{- end }}:{{ $telController.image.tag | default .Chart.AppVersion }}
         imagePullPolicy: {{ $telController.image.pullPolicy }}
         name: telemetry
         ports:
@@ -105,6 +114,9 @@ spec:
         runAsNonRoot: true
       serviceAccountName: {{ include "kcm-regional.fullname" . }}-telemetry
       terminationGracePeriodSeconds: 10
+      {{- if .Values.imagePullSecrets }}
+      imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 8 }}
+      {{- end }}
       {{- if $isLocalTelemetry }}
       volumes:
         - name: telemetry-storage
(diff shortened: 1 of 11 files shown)


templates/provider/kcm-regional/templates/deployment.yaml
```diff
@@ -105,6 +114,11 @@ spec:
         runAsNonRoot: true
       serviceAccountName: {{ include "kcm-regional.fullname" . }}-telemetry
       terminationGracePeriodSeconds: 10
+      {{- if $global.imagePullSecrets }}
+      imagePullSecrets: {{ toYaml $global.imagePullSecrets | nindent 6 }}
```
zerospiel:
```suggestion
      imagePullSecrets: {{ toYaml $global.imagePullSecrets | nindent 8 }}
```

One might add extra tests to the `templates/provider/kcm-regional/tests/telemetry_deployment_test.yaml` to ensure it works as expected with the `helm plugin install https://github.com/helm-unittest/helm-unittest.git` and `helm unittest -f templates/provider/kcm-regional/tests/telemetry_deployment_test.yaml`

templates/provider/kcm-regional/templates/deployment.yaml
```diff
@@ -105,6 +114,11 @@ spec:
...
+      {{- else if .Values.imagePullSecrets }}
+      imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 6 }}
```
zerospiel:
```suggestion
      imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 8 }}
```

templates/provider/kcm-regional/templates/deployment.yaml
```diff
@@ -105,6 +114,11 @@ spec:
         runAsNonRoot: true
       serviceAccountName: {{ include "kcm-regional.fullname" . }}-telemetry
       terminationGracePeriodSeconds: 10
+      {{- if $global.imagePullSecrets }}
```
zerospiel: There is nothing in the code that sets `global.imagePullSecrets`, thus I do not quite get the purpose of this particular `if`-statement

templates/provider/kcm/templates/deployment.yaml
```diff
@@ -107,6 +107,9 @@ spec:
         runAsNonRoot: true
       serviceAccountName: {{ include "kcm.fullname" . }}-controller-manager
       terminationGracePeriodSeconds: 10
+      {{- if .Values.imagePullSecrets }}
+      imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 6 }}
```
zerospiel:
```suggestion
      imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 8 }}
```

templates/provider/kcm-regional/templates/deployment.yaml
a13x5: Several kcm subcharts (e.g. cert-manager) already support this global value. It will be beneficial to support that in `kcm-regional` as well and it may simplify certain scenarios in the future.

s3rj1k: > it may simplify certain scenarios in the future

and break other scenarios 😃

templates/provider/kcm-regional/templates/deployment.yaml
a13x5: Additional indentation here is optional, since `imagePullSecrets` is a list. 
Thus the current
```yaml
      imagePullSecrets:
      - name: test
```
And suggested:
```yaml
      imagePullSecrets:
        - name: test
```

Are both valid yaml. I choose 6 to match the style of other lists in the file (such as `containers`, `args`, etc.).
I will put indent 8 as suggested, but it doesn't really matter IMO

templates/provider/kcm-regional/templates/deployment.yaml
zerospiel: > Several kcm subcharts (e.g. cert-manager) already support this global value

@a13x5  I might have missed something, so could you point out the portion of code that does have such support? At the moment, the `global` values are being set by the `kcm` and there is support only for the fixed set of (sub)values (`registry`, `k0sURL`, `registryCertSecret`, `k0sURLCertSecret`).

So my point is that, despite it being harmless, this particular `{{- if $global.imagePullSecrets }}` line and statement might be confusing for devs/users that there is such support.

Otherwise, the PR looks good to me

a13x5: Sure, if that's confusing it will be implemented as a part of a separate epic then with more detailed explanation. Removed for now.

Issue #2077, opened 2025-10-10
[bug] Self Management of mothership broken


wahabmk:
# Description

While working on #1961,  I discovered that self-management of mothership feature is broken due to 3 distinct bugs:

1. Due to a panic [here](https://github.com/k0rdent/kcm/blob/64225125684ea87da4c74c1c5114762b5572984f/internal/controller/multiclusterservice_controller.go#L485) (cuz CD is nil).
2. When I fixed that, I discovered another panic downstream [here](https://github.com/k0rdent/kcm/blob/64225125684ea87da4c74c1c5114762b5572984f/internal/controller/adapters/sveltos/serviceset_controller.go#L437) (cuz the newly introduced regional client is nil if CD is nil) and even further downstream [here](https://github.com/k0rdent/kcm/blob/64225125684ea87da4c74c1c5114762b5572984f/internal/controller/adapters/sveltos/serviceset_controller.go#L598) when collecting status.
3. When I bypassed bug 2, I hit yet another bug where the ServiceSet is created, however, the ServiceSet's status keeps on showing "Provisioning" for the service which I can verify was successfully deployed and running on the management cluster. This bug is caused [here](https://github.com/k0rdent/kcm/blob/64225125684ea87da4c74c1c5114762b5572984f/internal/controller/adapters/sveltos/serviceset_controller.go#L596-L600) while collecting status because we are fetching status of a Profile, whereas we create a ClusterProfile for self-management.

PR #2182, opened 2025-11-18
test: e2e tests for MCS dependencies on management cluster

wahabmk:
Fixes https://github.com/k0rdent/kcm/issues/1992.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -192,46 +192,6 @@ jobs:
           retention-days: 1
           overwrite: true
 
-  mothership-e2etest:
-    name: E2E Mothership
-    runs-on: ${{ needs.authorize.outputs.runner || 'ubuntu-latest' }}
-    if: ${{ contains( github.event.pull_request.labels.*.name, 'test e2e') }}
-    needs:
-      - lint-test
-      - authorize
-    concurrency:
-      group: controller-${{ github.head_ref || github.run_id }}
-      cancel-in-progress: true
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v5
-        with:
-          fetch-depth: 0
-          ref: ${{ needs.lint-test.outputs.pr_merge_commit }}
-
-      - name: Setup kubectl
-        uses: azure/setup-kubectl@v4
-
-      - name: Setup Go and Cache
-        uses: ./.github/actions/setup-go-cache
-
-      - name: Run E2E tests
-        env:
-          GINKGO_LABEL_FILTER: 'mothership'
-          CLUSTER_DEPLOYMENT_PREFIX: ${{ needs.lint-test.outputs.clusterprefix }}
-          VERSION: ${{ needs.lint-test.outputs.version }}
-          SEGMENT_TOKEN: ${{ secrets.SEGMENT_TOKEN_DEV }}
-        run: make test-e2e
-
-      - name: Archive test results
-        uses: actions/upload-artifact@v5
-        if: always()
-        with:
-          name: support-bundles
-          path: "/home/runner/work/kcm/kcm/support-bundle-*"
-          retention-days: 1
-          overwrite: true
-
   push-public:
     name: Build and Push Artifacts When Required
     runs-on: ${{ needs.authorize.outputs.runner || 'ubuntu-latest' }}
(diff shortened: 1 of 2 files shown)


internal/serviceset/builder.go
```diff
@@ -180,15 +180,22 @@ func ConvertServiceSpecToProviderConfig(serviceSpec kcmv1.ServiceSpec) (kcmv1.St
 	}
 
 	cfg := config{
-		SyncMode:             serviceSpec.SyncMode,
+		//nolint:staticcheck
```
zerospiel: out of curiosity: why these `nolint` statements? the question is also applicable to the `internal/telemetry/collector/local.go`, `internal/telemetry/collector/local_test.go`, `internal/telemetry/collector/segmentio.go`, `internal/telemetry/collector/segmentio_test.go` ? is it because of the properly set the `deprecated` comment? I'd put then the comment like `//nolint:staticcheck // deprecated field usage, should be removed` (or whatever is more suitable)

besides, what we actually should do, e.g. with the `syncMode`, should it be avoided from the usage? if so, then I believe, we could easily drop it from the telemetry collection (at least)

wahabmk:
@zerospiel 
1. Yes looks like you're right, the CI needs to be updated. I wonder if I can add it to the [E2E Controller test](https://github.com/k0rdent/kcm/blob/87c618444d0a8cf482787cf7c81d1094354abdd6/.github/workflows/build_test.yml#L178-L184) as a "Run E2E tests on Management Cluster".

2. I thought [this section](https://github.com/k0rdent/kcm/blob/87c618444d0a8cf482787cf7c81d1094354abdd6/docs/dev.md?plain=1#L223-L229) of the doc would be sufficient to show how to run these tests, i.e:
```sh
GINKGO_LABEL_FILTER=management:mcs make test-e2e
```

3. These tests are run only on the management cluster itself so they are agnostic of any cloud providers.

4. I am not using kyverno chart with in these tests, so unrelated to this PR.

internal/serviceset/builder.go
wahabmk: These nolint statements are there because the linter returns error because these fields are marked as deprecated. I will change these to add comment similar to:
https://github.com/k0rdent/kcm/blob/87c618444d0a8cf482787cf7c81d1094354abdd6/internal/util/validation/cd.go#L78

The deprecated fields like `syncMode` are kept for backward compatibility reasons as [this section](https://docs.k0rdent.io/latest/admin/ksm/ksm-built-in-provider/?h=syncmode#provider-configuration) of the official docs explain.

zerospiel: Okay, so this field could be safely removed from the telemetry collection. I will remove them then, thanks

Feel free to resolve the thread

test/e2e/multiclusterservice/multiclusterservice.go
```diff
@@ -60,14 +63,46 @@ func BuildMultiClusterService(cd *kcmv1.ClusterDeployment, multiClusterServiceTe
 	}
 }
 
-func CreateMultiClusterService(ctx context.Context, cl crclient.Client, mc *kcmv1.MultiClusterService) {
+func CreateMultiClusterService(ctx context.Context, cl client.Client, mcs *kcmv1.MultiClusterService) {
+	Expect(mcs).NotTo(BeNil())
+	kind := mcs.GetObjectKind().GroupVersionKind().Kind
+	Expect(kind).To(Equal(kcmv1.MultiClusterServiceKind))
```
zerospiel: This leads to a failure in the last run

.github/workflows/build_test.yml
```diff
@@ -192,6 +192,46 @@ jobs:
           retention-days: 1
           overwrite: true
 
+  mothership-e2etest:
```
zerospiel: A question: so this is a dedicated job, and it will run unconditionally (just the label required), am i right? i wonder whether we could run this test conditionally depending on the config, as it is right now for the other providers. for instance, what if want to run only vsphere provider tests on our internal runner? if i, once again, understand correctly, there will be 2 runners engaged: the internal one and GH public runners (please correct me if i wrong), but the latter is highly likely is not the desirable state: initially the only intense was to check the vsphere

wahabmk: I think we should run these mothership tests for changes to the controllers because changes that seemed unrelated have broken self-management in the past as detailed in https://github.com/k0rdent/kcm/issues/2077. But since these tests are provider-agnostic, I agree that it seems wasteful to setup a whole new mothership cluster with `make test-e2e` just to run these tests.

So I have made it so that the mothership tests are run with AWS provider tests (as AWS is the default). We just need these tests to run once on the management cluster (shouldn't matter which provider).

I am also open to other suggestions.

BROngineer: Seems reasonable to me

Issue #2126, opened 2025-10-27
[enhancement] Adapt ClusterDeployment controller to handle DataSource


eromanova:
ClusterDeployment controller should be adapted to handle DataSource references:
1. Create ClusterDataSource object if `spec.dataSource` is not empty either in ClusterDeployment or in Credential.
2. Wait until the status of the ClusterDataSource is populated with the `kineDataSourceSecretName` and `caSecret` (done by the external controller).
3. Pass `kineDataSecretName` and `caSecret` to HelmRelease values for each ClusterDeployment with the dataSource reference.

PR #2200, opened 2025-11-25
fix: trigger the deletion of ClusterDataSource

eromanova:
Clusterdeployment controller needs to trigger the ClusterDataSource deletion, otherwise, the ClusterDeployment deletion will be stuck.

**Which issue(s) this PR fixes** _(optional, `Fixes #123`)_: #2126

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -1266,7 +1266,8 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, mgmt 
 	r.eventf(cd, "ServiceSetsDeleted", "ServiceSets for cluster %s have been deleted", client.ObjectKeyFromObject(cd))
 
 	if cdsRequeue, err := r.deleteClusterDataSource(ctx, scope); err != nil {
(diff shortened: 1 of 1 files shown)


internal/controller/clusterdeployment_controller.go
```diff
@@ -1308,6 +1309,16 @@ func (r *ClusterDeploymentReconciler) waitClusterDataSourceDeletion(ctx context.
...
+		cd := scope.cd
+		if err := r.MgmtClient.Delete(ctx, &kcmv1.ClusterDataSource{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cd.Name,
+				Namespace: cd.Namespace,
+			},
+		}); client.IgnoreNotFound(err) != nil {
+			r.setCondition(cd, kcmv1.DeletingCondition, err)
+			return false, err
+		}
```
```suggestion
		if err := r.MgmtClient.Delete(ctx, cds); client.IgnoreNotFound(err) != nil {
			return false, fmt.Errorf("failed to delete ClusterDataSource %s: %w", cdsKey, err)
		}
```

Condition should be set by the caller, and the message should also be amended from `failed to wait ...` to `failed to delete ...`

PR #2201, opened 2025-11-25
fix(templates): migrate ASO objects to v1beta1

### LINKED ISSUE #1808: Update to v1beta1 API in Azure AKS template
Currently, when deploying Azure AKS clusters, there is a warning:

```
I0729 12:55:39.330560   12552 warnings.go:110] "Warning: infrastructure.cluster.x-k8s.io/v1alpha1 AzureASOManagedControlPlane is deprecated. infrastructure.cluster.x-k8s.io/v1beta1 is equivalent and should be used instead."
```

It does not block the deployment at the moment, but we should consider updating the template accordingly.

zerospiel:

Closes #1808

### DIFF
--- a/templates/cluster/azure-aks/templates/cluster.yaml
+++ b/templates/cluster/azure-aks/templates/cluster.yaml
@@ -15,10 +15,10 @@ spec:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   controlPlaneRef:
-    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
+    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
     kind: AzureASOManagedControlPlane
     name: {{ include "cluster.name" .  }}
   infrastructureRef:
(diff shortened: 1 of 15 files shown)


templates/cluster/azure-aks/templates/cluster.yaml
```diff
@@ -15,10 +15,10 @@ spec:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   controlPlaneRef:
-    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
+    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
```
eromanova, line 18: `machinepool-controlplane.yaml` and `machinepool-worker.yaml` files should also be updated with a new infrastructureRef.

zerospiel: Indeed, thanks for noticing

PR #2203, opened 2025-11-27
fix: do not update mgmt release immediately

### LINKED ISSUE #1971: [bug] immediate update of the Management's `status.release` during k0rdent upgrade
**Describe the bug**
During a k0rdent upgrade, the `status.release` field in the Management object is updated almost immediately after `spec.release` changes, before the components have actually been upgraded.

**To Reproduce**
1. Deploy k0rdent v1.2.0
2. Upgrade it to the latest v1.3.1
3. Check the Management object status. The `release` equals v1.3.1 while the components are still in the process of upgrading.

**Expected behavior**
The `status.release` field should only be updated to the new release after all components are fully upgraded.

zerospiel:
The fix does not do much in the sense of the "not immediate update" but gives a more robust status in case of any component errors, providing backward compatibility with the current approach.

The more correct approach would be to set the observed release after all of its components reconcile and be ready, though that does not align well with the current approach and might lead to bad UX where the `.status.release` shows the previous release and the errors from a new one from the `.spec.release`. Thus no changes in this matter.

We could consider changing this in the future, though it should be done carefully since some of the other controllers might rely on the current `managements` status.

Closes #1971

### DIFF
--- a/internal/controller/management_controller.go
+++ b/internal/controller/management_controller.go
@@ -181,9 +181,11 @@ func (r *ManagementReconciler) update(ctx context.Context, management *kcmv1.Man
 	}
 
 	requeue, errs := components.Reconcile(ctx, r.Client, r.Client, management, r.Config, release, opts)
-
+	if errs == nil { // if NO error
+		// update only if we actually observed the new release
+		management.Status.Release = management.Spec.Release
+	}
 	management.Status.ObservedGeneration = management.Generation
-	management.Status.Release = management.Spec.Release
 
 	shouldRequeue, err := r.startDependentControllers(ctx, management)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/controller/management_controller.go
```diff
@@ -181,9 +181,11 @@ func (r *ManagementReconciler) update(ctx context.Context, management *kcmv1.Man
 	}
 
 	requeue, errs := components.Reconcile(ctx, r.Client, r.Client, management, r.Config, release, opts)
-
-	management.Status.ObservedGeneration = management.Generation
-	management.Status.Release = management.Spec.Release
+	if errs == nil { // if NO error
```
eromanova, line 184: Right now `observedGeneration` is only updated when `components.Reconcile()` succeeds. But `observedGeneration` is meant to reflect that the controller has seen and processed the latest Management spec even if reconciliation ended with an error.  For example, when a template is not valid yet, the controller has already processed the new Management spec, but observedGeneration will still be outdated, which may be confusing (someone can assume the controller didn't process the new version, and this error message relates to some old template). Would it make sense to update `observedGeneration` whenever the spec has been fully evaluated, regardless of whether the `components.Reconcile` succeeded?

zerospiel: Makes total sense, will do this

zerospiel: Fixed

eromanova (approved):

Issue #2167, opened 2025-11-11
[bug] MCS not working when `helmOptions.wait` is set to `true`


AndrejsPon00:
**Describe the bug**
If `helmOptions.wait` is set to true for any service in MCS, the MCS does not work properly and becomes stuck with the following status:

```yaml
  - lastTransitionTime: "2025-11-11T09:20:28Z"
     message: 0/1
     reason: Failed
     status: "False"
     type: ClusterInReadyState
  - lastTransitionTime: "2025-11-11T09:20:28Z"
     message: 0/1 Clusters are ready. 
     reason: Failed
     status: "False"
     type: Ready
```

**To Reproduce**
1. Apply the `ServiceTemplate`:
```yaml
kubectl apply -f - <<EOF
apiVersion: k0rdent.mirantis.com/v1beta1
kind: ServiceTemplate
metadata:
  annotations:
    helm.sh/resource-policy: keep
  name: cert-manager-test
  namespace: kcm-system
spec:
  helm:
    chartSpec:
      chart: cert-manager
      interval: 10m0s
      reconcileStrategy: ChartVersion
      sourceRef:
        kind: HelmRepository
        name: cert-manager
      version: v1.16.4
EOF
```
2. Apply the MultiClusterService:
```yaml
kubectl apply -f - <<EOF
apiVersion: k0rdent.mirantis.com/v1beta1
kind: MultiClusterService
metadata:
  name: test-cert-manager-installation
spec:
  clusterSelector:
    matchLabels:
      debug-label: "true"
  serviceSpec:
    services:
      - name: cert-manager
        namespace: kcm-system
        template: cert-manager-test
        helmOptions:
          wait: true
        values: |
          crds:
            enabled: true
EOF
```
3. Deploy a ClusterDeployment with label `debug-label: "true"` on AWS or use an Adopted cluster. (The issue occurs with these types, other cluster types have not been tested)

**Expected behavior**
According to the [API documentation](https://github.com/k0rdent/kcm/blob/main/api/v1beta1/multiclusterservice_types.go#L144) the `wait` option should cause Helm to wait until all Pods, PVCs, Services, and the minimum number of Pods in a Deployment, StatefulSet, or ReplicaSet are in a ready state before marking the release as successful.

KCM version: v1.5.0

PR #2208, opened 2025-11-27
fix: helm options merging fails

kylewuolle:
This PR fixes an issue where the helm options merging fails with an error.

https://github.com/k0rdent/kcm/issues/2167

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller_test.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller_test.go
@@ -15,6 +15,7 @@
 package sveltos
 
 import (
+	"reflect"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -25,6 +26,7 @@ import (
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 	clusterapiv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	. "sigs.k8s.io/controller-runtime/pkg/envtest/komega"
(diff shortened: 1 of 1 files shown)


internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -906,6 +902,23 @@ func helmChartFromSpecOrRef(
...
+	data, err := json.Marshal(src)
+	if err != nil {
+		return fmt.Errorf("failed to marshal HelmOptions: %w", err)
+	}
+
+	if err := json.Unmarshal(data, dst); err != nil {
+		return fmt.Errorf("failed to unmarshal HelmOptions: %w", err)
+	}
+	return nil
+}
```
eromanova: Does this function actually merge `src` into `dst`? Since it marshals src to json and then unmarshals into dst, it looks like it will overwrite `dst` rather than merge with it.
If that behavior is intentional, maybe the name could reflect that it replaces rather than merges, unless I’m missing something?

kylewuolle: Fixed that up. You're right it was just overwriting. The new code does a proper merge and the tests prove it!

BROngineer (dismissed):
lgtm

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -100,7 +100,7 @@ type profileConfig struct {
 type ServiceSetReconciler struct {
 	client.Client
 
-	timeFunc  func() time.Time
+	TimeFunc  func() time.Time
```
zerospiel: Could you please elaborate on the reason why this field has changed to the public?

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -906,6 +900,27 @@ func helmChartFromSpecOrRef(
 	return helmChart, nil
 }
 
+// mergeHelmOptions merges the values from the given source ServiceHelmOptions to the destination ServiceHelmOptions
+func mergeHelmOptions(src, dst *kcmv1.ServiceHelmOptions) {
```
zerospiel, line 905: it'd be great to have a dedicated set of unit tests (ordinary ones, with `func Test_mergeHelmOptions(t *testing.T) {}`, it should be straightforward to implement the table-driven tests) covering all possible cases (the env tests are good but it's harder to cover all of the cases; they are very sound in that matter)

for instance, i do not completely sure it still addresses the problem @eromanova mentioned earlier with the overwriting

internal/controller/adapters/sveltos/serviceset_controller.go
kylewuolle: For testing purposes only. The tests fail unless this function is set and it's only set in setupwithmanager which can't be called in the tests.

internal/controller/adapters/sveltos/serviceset_controller.go
kylewuolle: Ok so export this function just for testing purposes?

internal/controller/adapters/sveltos/serviceset_controller.go
zerospiel: The tests are within the package, so i see no reason for this field to have the public scope

internal/controller/adapters/sveltos/serviceset_controller.go
zerospiel: there is no need for the func to be exported; the tests are within the package

zerospiel: Regarding the initial comment, i have checked the structure more accurately, it consists of pointers, so the main logic should be correct (to my understanding without the tests)

internal/controller/adapters/sveltos/serviceset_controller_test.go
```diff
@@ -15,9 +15,354 @@
...
+				src:  &kcmv1.ServiceHelmOptions{Timeout: testTimeout},
+				dst:  &kcmv1.ServiceHelmOptions{},
+				want: &kcmv1.ServiceHelmOptions{Timeout: testTimeout},
+			},
+		),
+
+		Entry("copy map pointer",
+			tc{
+				name: "copy labels map",
+				src:  &kcmv1.ServiceHelmOptions{Labels: &map[string]string{"env": "prod"}},
```
eromanova, line 208: Question: when labels are provided in both `src` and `dst`, should they be merged or overridden? If they should be merged, I can suggest adding a test case for it (the current test entry verifies that labels are overwritten).

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -906,6 +900,27 @@ func helmChartFromSpecOrRef(
...
+	if src == nil || dst == nil {
```
eromanova, line 906: If `src` is non-nil while `dst` is nil, then logically after merging, `dst` should become equal to `src`. In the current implementation we simply return without modifying `dst`. I just want to confirm that this behavior is intentional.

internal/controller/adapters/sveltos/serviceset_controller_test.go
kylewuolle: good point added that

internal/controller/adapters/sveltos/serviceset_controller.go
kylewuolle: true yeah the source should be used in that case. I've put that logic outside this function for simplicity.

PR #2215, opened 2025-12-01
fix(webhook): validate templates on chains creation

### LINKED ISSUE #402: Validate that all the Templates exist before creating the TemplateChains
We should improve the `ClusterTemplateChain` and `ServiceTemplateChain` validation to verify if all the `ClusterTemplates` and `ServiceTemplates` specified in the chain spec exist.

zerospiel:

Fixes #402

### DIFF
--- a/internal/webhook/templatechain_webhook.go
+++ b/internal/webhook/templatechain_webhook.go
@@ -59,7 +59,7 @@ func (in *ClusterTemplateChainValidator) ValidateCreate(ctx context.Context, obj
 		return warnings, errInvalidTemplateChainSpec
 	}
 
-	if chain.Labels[kcmv1.KCMManagedLabelKey] != kcmv1.KCMManagedLabelValue || chain.Namespace == in.SystemNamespace { // validate only unmanged or system
+	if chain.Labels[kcmv1.KCMManagedLabelKey] != kcmv1.KCMManagedLabelValue || chain.Namespace == in.SystemNamespace { // validate only unmanaged or system
 		if errs := validateChainsTemplates(ctx, in.Client, chain.Namespace, chain.Spec, kcmv1.ClusterTemplateKind); len(errs) > 0 {
 			return nil, apierrors.NewInvalid(chain.GroupVersionKind().GroupKind(), chain.Name, errs)
 		}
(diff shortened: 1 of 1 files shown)


internal/webhook/templatechain_webhook.go
```diff
@@ -56,9 +58,32 @@ func (*ClusterTemplateChainValidator) ValidateCreate(_ context.Context, obj runt
...
 
+	if errs := validateChainsTemplates(ctx, in.Client, chain.Namespace, chain.Spec, kcmv1.ClusterTemplateKind); len(errs) > 0 {
+		return nil, apierrors.NewInvalid(chain.GroupVersionKind().GroupKind(), chain.Name, errs)
+	}
+
 	return nil, nil
 }
 
+func validateChainsTemplates(ctx context.Context, cl client.Client, namespace string, chainSpec kcmv1.TemplateChainSpec, templatesKind string) field.ErrorList {
+	var errs field.ErrorList
```
eromanova, line 72: Since TemplateChains are responsible for templates distribution, we should validate only:
1. The TemplateChain is in the system namespace.
2. Unmanaged TemplateChains (without `k0rdent.mirantis.com/managed: true` label)

Otherwise, the access management template chain distribution will not work (ex. template chain A is configured to be distributed into namespace `ns1`, but the templates defined in the template chain A don't exist in the namespace `ns1` yet).

zerospiel: Nice catch, thanks. I've added the statement to validate only if unmanaged or is in a system namespace

internal/webhook/templatechain_webhook.go
```diff
@@ -56,9 +59,34 @@ func (*ClusterTemplateChainValidator) ValidateCreate(_ context.Context, obj runt
 		return warnings, errInvalidTemplateChainSpec
 	}
 
+	if chain.Labels[kcmv1.KCMManagedLabelKey] != kcmv1.KCMManagedLabelValue || chain.Namespace == in.SystemNamespace { // validate only unmanged or system
```
eromanova:
```suggestion
	if chain.Labels[kcmv1.KCMManagedLabelKey] != kcmv1.KCMManagedLabelValue || chain.Namespace == in.SystemNamespace { // validate only unmanaged or system
```

internal/webhook/templatechain_webhook.go
```diff
@@ -94,6 +123,12 @@ func (*ServiceTemplateChainValidator) ValidateCreate(_ context.Context, obj runt
...
```
eromanova:
```suggestion
	if chain.Labels[kcmv1.KCMManagedLabelKey] != kcmv1.KCMManagedLabelValue || chain.Namespace == in.SystemNamespace { // validate only unmanaged or system
```

internal/webhook/templatechain_webhook.go
zerospiel: Applied both, thanks!

PR #2222, opened 2025-12-03
fix: Improve run-time for mcs mothership e2e tests

wahabmk:
PR to reduce the runtime for MCS e2e tests on the mothership cluster.


zerospiel (dismissed):
LGTM. Much better in terms of time spent (~10 min now vs 30+ mins). I still would advocate for these tests to be:
1. default to `aws` cloud provider as it was before introducing
2. optional to run as it is still ~10 extra minutes and for some folks these tests might be redundant to start (eg. one just bumps a version of a cloud provider)

But I assume this shouldn't be done in this PR

PR #2227, opened 2025-12-08
chore: drop `authURL` from dev openstack cluster config

eromanova:
**What this PR does / why we need it**: This parameter is unused.


bnallapeta:
> @bnallapeta

I suppose this is done so that we provide the auth_url and others all via the clouds.yaml referenced through `identityRef`. Please correct me if I am wrong @eromanova

PR #2233, opened 2025-12-08
fix(cleanup): collect owners and delete

zerospiel:

Provides a dedicated function to delete PVCs:
- remove finalizers and delete all owners of `pods`
- remove finalizers and delete all `pvc` objects

Closes k0rdent/k0rdent-enterprise#208

### DIFF
--- a/internal/util/kube/delete.go
+++ b/internal/util/kube/delete.go
@@ -17,12 +17,18 @@ package kube
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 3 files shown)


internal/util/kube/delete.go
```diff
@@ -107,3 +113,267 @@ func DeleteAllExceptAndWait[T client.Object](
...
+
+			for _, pod := range pods {
+				topOwner, err := findTopLevelAllowedController(ctx, c, &pod.ObjectMeta)
+				if err != nil {
+					return fmt.Errorf("failed to resolve owners for pod %s/%s: %w", pod.Namespace, pod.Name, err)
+				}
+
+				if topOwner == nil {
+					// delete plain pod
+					if err := removeFinalizers(ctx, c, &pod); err != nil {
```
Kshatrix, line 167: finalizers should be removed after deletetion

zerospiel: Addressing in #2238

PR #2259, opened 2025-12-12
feat: registry authentication support

a13x5:
Adds additional flag `image-pull-secret` which is name of the imagePullSecret in `dockerconfigjson` format. This secret is passed to all components of the system as `global.imagePullSecrets[0].name` value.

Some secret creation such as provider values secret require globalRegistry to be set as well.

**Out of scope**: imagePullSecrets pass to child cluster

### DIFF
--- a/internal/controller/components/reconciler.go
+++ b/internal/controller/components/reconciler.go
@@ -515,7 +515,7 @@ func getFalseConditions(gp capioperatorv1.GenericProvider) []string {
 	return messages
 }
 
-func makeProviderSecret(username, password, namespace string, cmp *component, template *kcmv1.ProviderTemplate) (*corev1.Secret, bool, error) {
+func makeProviderSecretData(username, password string, cmp *component, template *kcmv1.ProviderTemplate) (map[string][]byte, bool, error) {
 	commonStatus := template.GetCommonStatus()
 
 	if commonStatus.Config == nil {
@@ -532,7 +532,7 @@ func makeProviderSecret(username, password, namespace string, cmp *component, te
 		return nil, true, nil
 	}
 
-	userConfig := make(map[string]string)
+	userConfig := make(map[string][]byte)
 
 	if cmp.Config != nil {
 		var userProvidedValues map[string]any
@@ -554,27 +554,19 @@ func makeProviderSecret(username, password, namespace string, cmp *component, te
 		return nil, false, err
 	}
 
-	ociCreds := map[string]string{
-		"OCI_USERNAME": username,
-		"OCI_PASSWORD": password,
+	ociCreds := map[string][]byte{
+		"OCI_USERNAME": []byte(username),
+		"OCI_PASSWORD": []byte(password),
 	}
 
 	maps.Copy(defaultConfig, userConfig)
 	maps.Copy(defaultConfig, ociCreds)
 
-	secret := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      getProviderConfigSecretName(cmp.name),
-			Namespace: namespace,
-		},
-		StringData: defaultConfig,
-	}
-
-	return secret, false, nil
+	return defaultConfig, false, nil
 }
 
-func getProviderConfig(vals map[string]any) (map[string]string, error) {
-	providerConfig := make(map[string]string)
+func getProviderConfig(vals map[string]any) (map[string][]byte, error) {
+	providerConfig := make(map[string][]byte)
 	providerConfigValueR, ok := vals["config"]
 	if ok {
 		providerConfigValue, castOk := providerConfigValueR.(map[string]any)
@@ -584,7 +576,7 @@ func getProviderConfig(vals map[string]any) (map[string]string, error) {
 
 		for k := range providerConfigValue {
 			if val, ok := providerConfigValue[k].(string); ok {
-				providerConfig[k] = val
+				providerConfig[k] = []byte(val)
 			}
 		}
 	}
@@ -625,19 +617,35 @@ func reconcileProviderConfigSecret(
 	template *kcmv1.ProviderTemplate,
 ) error {
 	l := ctrl.LoggerFrom(ctx)
-	providerSecret, skipCreation, err := makeProviderSecret(username, password, namespace, cmp, template)
+	providerSecretData, skipCreation, err := makeProviderSecretData(username, password, cmp, template)
 	if err != nil {
-		return fmt.Errorf("failed to generate provider secret: %w", err)
+		return fmt.Errorf("failed to generate provider secret data: %w", err)
 	}
 
 	if !skipCreation {
-		l.Info("Creating provider variable secret for " + cmp.name)
-		if err := rgnlClient.Patch(ctx, providerSecret, client.Apply, client.FieldOwner("mgmt")); client.IgnoreAlreadyExists(err) != nil {
-			return fmt.Errorf("failed to patch provider secret %q: %w", providerSecret.Name, err)
-		}
-
 		if err := modifyConfigSecretValues(cmp); err != nil {
 			return fmt.Errorf("failed to modify configSecret value: %w", err)
+		}
+
+		l.Info("Ensuring provider variable secret for " + cmp.name)
+
+		secretName := getProviderConfigSecretName(cmp.name)
+		providerSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secretName,
+				Namespace: namespace,
+			},
+		}
+
+		op, err := ctrl.CreateOrUpdate(ctx, rgnlClient, providerSecret, func() error {
+			providerSecret.Data = providerSecretData
+			return nil
+		})
+		if err != nil {
+			return fmt.Errorf("failed to create or update provider config secret %s: %w", secretName, err)
+		}
+		if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated {
+			l.Info("Successfully mutated provider config secret", "name", secretName, "operation_result", op)
 		}
 	}
 


internal/controller/components/reconciler.go
```diff
@@ -57,6 +63,9 @@ type ReconcileComponentsOpts struct {
 	// RegistryCertSecretName is the name of the secret with CA certificate of the global registry
 	// to be passed to components values.
 	RegistryCertSecretName string
+	// ImagePullSecretName is the name of the of a Secret containing
```
eromanova: nit: typo
```suggestion
	// ImagePullSecretName is the name of a Secret containing
```

internal/controller/components/reconciler.go
```diff
@@ -156,6 +185,44 @@ func Reconcile(
...
+				errMsg := fmt.Sprintf("failed to generate provider secret: %s", err)
+				updateComponentsStatus(statusAccumulator, component, template, errMsg)
+				errs = errors.Join(errs, errors.New(errMsg))
+				continue
+			}
+
+			if !skipCreation {
+				l.Info("Creating provider variable secret for " + component.name)
+				if err := mgmtClient.Create(ctx, providerSecret); client.IgnoreAlreadyExists(err) != nil {
+					errMsg := fmt.Sprintf("failed to patch provider secret %q for provider %s: %s", providerSecret.Name, component.name, err)
```
eromanova:
```suggestion
					errMsg := fmt.Sprintf("failed to create provider secret %q for provider %s: %s", providerSecret.Name, component.name, err)
```

internal/controller/components/reconciler.go
```diff
@@ -156,6 +185,44 @@ func Reconcile(
...
+			if err != nil {
+				errMsg := fmt.Sprintf("failed to generate provider secret: %s", err)
+				updateComponentsStatus(statusAccumulator, component, template, errMsg)
+				errs = errors.Join(errs, errors.New(errMsg))
+				continue
+			}
+
+			if !skipCreation {
+				l.Info("Creating provider variable secret for " + component.name)
+				if err := mgmtClient.Create(ctx, providerSecret); client.IgnoreAlreadyExists(err) != nil {
```
eromanova: It should be created using the regional client (the secret must exist on the regional cluster to deploy providers on it). In case of the management cluster, `rgnlClient == mgmtClient`.
```suggestion
				if err := rgnlClient.Create(ctx, providerSecret); client.IgnoreAlreadyExists(err) != nil {
```

internal/controller/components/reconciler.go
```diff
@@ -156,6 +185,44 @@ func Reconcile(
...
+					errs = errors.Join(errs, errors.New(errMsg))
+					continue
+				}
+			}
+
+			if component.targetNamespace != "" {
+				if err := kubeutil.CopySecret(
+					ctx,
+					mgmtClient,
+					mgmtClient,
```
eromanova: ditto: the first client is the source (always mgmt) while the second client is the target. In case of the management cluster, `rgnlClient == mgmtClient`.
```suggestion
					rgnlClient,
```

internal/controller/components/reconciler.go
```diff
@@ -460,3 +527,105 @@ func getFalseConditions(gp capioperatorv1.GenericProvider) []string {
...
+		return authConfig.Username, authConfig.Password, nil
+	}
+
+	return "", "", errors.New("unable to identify auth parameters in the dockerconfig")
+}
+
+func makeProviderSecret(username, password, componentName, namespace string, template *kcmv1.ProviderTemplate) (*corev1.Secret, bool, error) {
+	commonStatus := template.GetCommonStatus()
+
+	if commonStatus.Config == nil {
```
eromanova: I may be missing something, but template `status.config` appears to contain only the default Helm chart values, not the overrides supplied by the user. How are user-specified overrides from the Management or Region resource handled in this case?

internal/controller/components/values.go
```diff
@@ -92,26 +92,57 @@ func getComponentValues(
...
+
+			vals["configSecret"] = map[string]any{
+				"create": false,
+				"name":   getProviderConfigSecretName(name),
+			}
+		}
+
+		vals["global"] = global
+
+		componentValues = chartutil.CoalesceTables(componentValues, vals)
```
eromanova, line 142: For all provider components (except sveltos), the values provided in the Management/Region will override the values generated in `getComponentValues` function (see lines 150-154):
```
	// for projectsveltos, we want new values to override values provided in Management spec
	if name == kcmv1.ProviderSveltosName {
		merged = chartutil.CoalesceTables(componentValues, currentValues)
	} else {
		merged = chartutil.CoalesceTables(currentValues, componentValues)
	}
```
So, if the user explicitly sets:
```
configSecret:
  name: abc
```
these values will take precedence over `vals`.

internal/controller/components/reconciler.go
```diff
@@ -460,3 +527,105 @@ func getFalseConditions(gp capioperatorv1.GenericProvider) []string {
...
+	}
+
+	data := secret.Data
+	configstr, ok := data[".dockerconfigjson"]
+	if !ok {
+		return "", "",
+			errors.New("unable to get .dockerconfigjson from imagePullSecret data")
+	}
+
+	var config dockerconfig.ConfigFile
```
zerospiel: nit: i'd vote to copy the struct into the code explicitly (and comment the revision+location it has been copied from) rather than introducing a new (not very pleasant in terms of versioning) dependency

internal/controller/components/reconciler.go
```diff
@@ -114,8 +123,14 @@ func Reconcile(
 			CompatibilityContracts: make(map[string]kcmv1.CompatibilityContracts),
 		}
 		requeue bool
+
+		pullSecret       *corev1.Secret
+		registryUsername string
+		registryPassword string
 	)
 
+	needsPullSecret := opts.GlobalRegistry != "" && opts.ImagePullSecretName != ""
```
zerospiel: I find it hard to understand why the global `imagePullSecrets` only works if the global registry is set. From my perspective, they are not dependent on each other, so I would pull the secret when only the former flag is set

Cross-link with [this comment](https://github.com/k0rdent/kcm/pull/2259#discussion_r2614039387) after I almost conducted the CR.

internal/controller/components/reconciler.go
```diff
@@ -114,8 +123,14 @@ func Reconcile(
 			CompatibilityContracts: make(map[string]kcmv1.CompatibilityContracts),
 		}
 		requeue bool
+
+		pullSecret       *corev1.Secret
```
zerospiel:
```suggestion
		pullSecret       = new(corev1.Secret)
```

this way we won't hit a potential runtime null dereference

and then there is no need for the CL nr. 143 pullSecret = &corev1.Secret{}

internal/controller/components/reconciler.go
zerospiel: i've checked: the whole struct is overkill, we could copy just the required fields

internal/controller/components/reconciler.go
```diff
@@ -460,3 +527,105 @@ func getFalseConditions(gp capioperatorv1.GenericProvider) []string {
...
+				fmt.Errorf("unable to decode auth: %w", err)
+		}
+
+		authSplit := strings.Split(string(auth), ":")
+		if len(authSplit) != 2 {
+			return "", "",
+				errors.New("incorrect \":\" delimeted auth value")
+		}
+
+		return authSplit[0], authSplit[1], nil
```
zerospiel:
```suggestion
		username, password, found := strings.Cut(string(auth), ":")
		if !found {
			return "", "",
				errors.New("incorrect \":\" delimeted auth value")
		}

		return username, password, nil
```

nit: just a little bit clearer, up to the implementer

internal/controller/components/reconciler.go
```diff
@@ -460,3 +527,105 @@ func getFalseConditions(gp capioperatorv1.GenericProvider) []string {
...
+	}
+
+	if _, ok := defaultValues["configSecret"]; !ok {
+		return nil, true, nil
+	}
+
+	providerConfig := make(map[string]string)
+	providerConfigValueR, ok := defaultValues["config"]
+	if ok {
+		providerConfigValue, castOk := providerConfigValueR.(map[string]any)
```
zerospiel, line 572: i'm just wondering: can't we cast to `map[string]string` here? that way we could ditch all these not very convenient further casts. i am not sure, though, whether it will actually work

internal/controller/components/reconciler.go
```diff
@@ -460,3 +527,105 @@ func getFalseConditions(gp capioperatorv1.GenericProvider) []string {
...
+		return "", "",
+			fmt.Errorf("failed to unmarshal dockerconfig: %w", err)
+	}
+
+	auths := config.AuthConfigs
+
+	rs := strings.Split(registry, "/")
+	registryHost := rs[0]
+
+	authConfig, ok := auths[registryHost]
```
zerospiel:
```suggestion
	registryHost := strings.Split(registry, "/")[0]
	authConfig, ok := config.AuthConfigs[registryHost]
```

nit: could be compacted

regarding the registry host extraction: is there any more correct way to get the registry's host? i am asking because in the k0s there were several issues regarding the similar case, understanding where the host is (IINM). [This PR](https://github.com/k0sproject/k0s/pull/6243) with the fix. I might be overcautious in this regard (because we are extracting the host, not the whole registry) but just to double check it

internal/controller/components/reconciler.go
```diff
@@ -156,6 +185,44 @@ func Reconcile(
 			continue
 		}
 
+		if needsPullSecret {
```
zerospiel: style: might be a good idea to extract this code block into a dedicated method/function

internal/controller/components/values.go
```diff
@@ -92,26 +92,57 @@ func getComponentValues(
...
 			},
 		}
+
+		if opts.ImagePullSecretName != "" {
+			imagePatch := `- op: add
+  path: /spec/template/spec/imagePullSecrets
+  value:
+  - name: ` + opts.ImagePullSecretName
+
+			projectsveltos["classifierManager"] = map[string]any{
```
zerospiel, line 110: will this work for all of the images across all of the sveltos deployments?

internal/controller/components/values.go
```diff
@@ -92,26 +92,57 @@ func getComponentValues(
...
+
+		if opts.ImagePullSecretName != "" {
+			global["imagePullSecrets"] = []map[string]any{
+				{
+					"name": opts.ImagePullSecretName,
+				},
+			}
+
+			vals["configSecret"] = map[string]any{
+				"create": false,
```
zerospiel: I can't find this field in the `InfrastructureProvider` spec, could you elaborate where the field comes from?

internal/controller/components/values.go
zerospiel: TBH, I didn't get this. As of my understanding with the new changes, if a user sets the `configSecret`, we:

1. merge `vals` (with the `configSecret`) into the `componentValues`, the latter is the authoritative but we fill this map on our own.
2. merge `componentValues` into `currentValues` (CL 150-154 of the current rev), where the user-set `configSecret` overwrites ours.

After writing this, I realized the problem. Perhaps, it might be a good idea to set the `configSecret` after the merge (if everything is set)

internal/controller/components/values.go
```diff
@@ -92,26 +92,57 @@ func getComponentValues(
...
+			"projectsveltos": projectsveltos,
+		}
 	}
 
 	if opts.GlobalRegistry != "" {
-		globalValues := map[string]any{
-			"global": map[string]any{
-				"registry": opts.GlobalRegistry,
-			},
+		vals := make(map[string]any)
```
zerospiel, line 125: okay, after reviewing the provider charts, i've realized why we hid this setting under the `global.registry` but it still does not make much sense to me since one could just want to set the `imagepullsecrets` without setting the global registry (considering it is the OSS version and it indeed DOES support global variables)

So IMO the global map should be filled if either registry or pullsecrets is set (or both), that will naturally require complicated if-else in the charts but this way the logic won't contain intrinsics requiring from one to delve into the codebase

internal/controller/components/values.go
zerospiel: Not valid considering the `values.yaml`, we have this field.

internal/controller/components/values.go
a13x5: This patch is only needed for sveltos agent deployment (which is hardoded as a literal yaml), the rest of sveltos components use `global.registry` parameter

internal/controller/components/reconciler.go
a13x5: The issue you mention more relevant for image references (which may be "non-canonical", e.g. golang:latest). In our case we are working with the actual hostname (well more like URL without scheme thing). Since hostname itself can't contain any special characters, but dots and minus sign (rfc952) I believe this is pretty safe to assume that this hostname extraction will work good enough.

internal/controller/components/reconciler.go
a13x5: That was the first thing I tried, but it looks like type casting (to map[string]string) doesn't work in this case, hence the need for this ugly iteration.

internal/controller/components/values.go
a13x5: `configSecret` is only used in our provider templates. The projectsveltos template is special and uses upstream chart directly. This parameter will be ignored either way, so I believe we can skip this issue altogether.

I had an idea to enrich values with `configSecret` parameter after provider template values extraction to be sure if this provider supports these values or not, but I believe this is really extra and complicates things without any meaningful benefit.

internal/controller/clusterdeployment_controller.go
```diff
@@ -934,6 +939,31 @@ func (*ClusterDeploymentReconciler) fillDataSourceValues(scope *clusterScope, va
...
 }
 
+func (r *ClusterDeploymentReconciler) fillRegistryAuthValues(ctx context.Context, values map[string]any) error {
+	if r.GlobalRegistry == "" && r.ImagePullSecretName == "" {
+		return nil
+	}
+
+	pullSecret := new(corev1.Secret)
+	if err := r.MgmtClient.Get(ctx, client.ObjectKey{Name: r.ImagePullSecretName, Namespace: r.SystemNamespace}, pullSecret); err != nil {
+		return fmt.Errorf("filed to get ImagePullSecret %q: %w", r.ImagePullSecretName, err)
```
```suggestion
		return fmt.Errorf("failed to get ImagePullSecret %q: %w", r.ImagePullSecretName, err)
```

internal/controller/clusterdeployment_controller.go
```diff
@@ -934,6 +939,31 @@ func (*ClusterDeploymentReconciler) fillDataSourceValues(scope *clusterScope, va
 	values["dataSource"] = val
 }
 
+func (r *ClusterDeploymentReconciler) fillRegistryAuthValues(ctx context.Context, values map[string]any) error {
+	if r.GlobalRegistry == "" && r.ImagePullSecretName == "" {
+		return nil
+	}
+
+	pullSecret := new(corev1.Secret)
+	if err := r.MgmtClient.Get(ctx, client.ObjectKey{Name: r.ImagePullSecretName, Namespace: r.SystemNamespace}, pullSecret); err != nil {
```
zerospiel: if the `globalRegistry` is set but the `imagePullSecretName` isn't, the method will fail, requiring that the user set the latter. Is that something expected?

internal/controller/clusterdeployment_controller.go
```diff
@@ -934,6 +939,31 @@ func (*ClusterDeploymentReconciler) fillDataSourceValues(scope *clusterScope, va
...
+		return fmt.Errorf("filed to get ImagePullSecret %q: %w", r.ImagePullSecretName, err)
+	}
+
+	registryUsername, registryPassword, err := pullsecretutil.GetRegistryCredsFromPullSecret(pullSecret, r.GlobalRegistry)
```
zerospiel: The other way around: if only `imagePullSecretName` is set, that somehow pushes the user to set the `globalRegistry`, otherwise the error, is that expected?

templates/cluster/vsphere-hosted-cp/templates/k0sworkerconfigtemplate.yaml
```diff
@@ -27,6 +27,14 @@ spec:
         {{- end }}
         {{- end }}
         {{- end }}
+        {{- if and $global.registry .Values.registryAuth.username .Values.registryAuth.password }}
```
zerospiel: potential rendering runtime error: if one sets the `registryAuth` to `null`, the render fails. So `if and $global.registry Values.registryAuth .Values.registryAuth.username .Values.registryAuth.password`

applicable elsewhere

might be a good idea to prohibit `null` object on the schema level

templates/cluster/vsphere-hosted-cp/templates/k0smotroncontrolplane.yaml
```diff
@@ -89,6 +89,10 @@ spec:
               {{- if $global.registryCertSecret }}
               caFile: /usr/local/share/ca-certificates/registry/ca.crt
               {{- end }}
+              {{- if and .Values.registryAuth.username .Values.registryAuth.password }}
```
zerospiel: potential rendering runtime error: if one sets the `registryAuth` to `null`, the render fails. So `if and $global.registry Values.registryAuth .Values.registryAuth.username .Values.registryAuth.password`

applicable elsewhere

might be a good idea to prohibit `null` object on the schema level

internal/controller/components/values.go
zerospiel: I would say that is slightly another issue, as I've described above. So a user manually sets `configSecret` in one of our provider templates, then all our code trying to fill out the `configSecret` will be rewritten by the user-set values since the latter is authoritative in this case.

So for instance, a user set

```yaml
configSecret:
  create: true
  name: foobar
```

and we were trying to fill it with:

```yaml
configSecret:
  create: false # differs from the user-set value
  name: foobar # have not changed
```

The result would be:

```
configSecret:
  create: true # user-set value prevails
  name: foobar
```

Which will then cause an error since the secret has been created by our code

a13x5: Made it more coupled with the actual secret creation routine, this eliminates precedence issue and also will only pass these values if we actually about to create the Secret for the supported capi provider and not spam these to all providers.

Makefile
```diff
@@ -131,8 +131,11 @@ capo-orc-fetch: CAPO_DIR := $(PROVIDER_TEMPLATES_DIR)/cluster-api-provider-opens
 capo-orc-fetch: CAPO_ORC_VERSION := 2.1.0
 capo-orc-fetch: CAPO_ORC_TEMPLATE := "$(CAPO_DIR)/templates/orc.yaml"
 capo-orc-fetch:
-	@curl -L --fail -s https://github.com/k-orc/openstack-resource-controller/releases/download/v$(CAPO_ORC_VERSION)/install.yaml | \
-	sed -E 's|(image: )([^\s/]+)(/.*)|\1{{ default "\2" (and .Values.global .Values.global.registry) }}\3|' > $(CAPO_ORC_TEMPLATE); \
+	@curl -L --fail -s https://github.com/k-orc/openstack-resource-controller/releases/download/v$(CAPO_ORC_VERSION)/install.yaml \
+	| sed '1i\{{- $$global := .Values.global | default dict }}' \
+	| sed -E 's|(image: )([^\s/]+)(/.*)|\1{{ default "\2" $$global.registry }}\3|' \
```
eromanova: I see issues running this on mac:
```
sed: 1: "1i\{{- $global := .Valu ...": extra characters after \ at the end of i command
sed: 1: "/serviceAccountName: or ...": extra characters after \ at the end of a command
make: *** [capo-orc-fetch] Error 1
```

zerospiel (approved):

PR #2265, opened 2025-12-15
fix: validate only diff during object updates

### LINKED ISSUE #2033: [bug] Admission should validate the diff, not the resource itself
The admission webhook should only validate the differences between the current and the applied configuration. For example, if a ProviderTemplate has already been added to the Management object, it should not block modifications to the Management object.

eromanova:
This PR fixes the admission webhook behaviour on Update requests so that it validates only diffs rather than the entire object. It's needed to allow valid object updates even if some previously configured field (ex., the ProviderTemplate configuration in the Management object) has since become invalid.

Contains:
1. Validate the Credential during ClusterDeployment update only if `spec.credential` was updated.
2. Validate Services (service templates and dependencies) during ClusterDeployment update only if `spec.serviceSpec.services` was updated.
3. Validate ProviderTemplates (validity, contracts compatibility) during Management/Region update only if the ProviderTemplate was updated.
4. Refactor Management and Region validation functions.
5. Update and adjust unit tests accordingly.

Fixes #2033

### DIFF
--- a/internal/util/validation/mgmt.go
+++ b/internal/util/validation/mgmt.go
@@ -123,7 +123,7 @@ func getIncompatibleContractsForProviderTemplates(
 		}
 
 		if len(pTpl.Status.CAPIContracts) == 0 {
-			return "", nil
+			continue
 		}
 
 		if !pTpl.Status.Valid {
--- a/internal/webhook/management_webhook_test.go
+++ b/internal/webhook/management_webhook_test.go
@@ -107,11 +107,19 @@ func TestManagementValidateUpdate(t *testing.T) {
 		bootstrapK0smotronProvider = "bootstrap-k0sproject-k0smotron"
 		k0smotronTemplateName      = "k0smotron-0-0-7"
 
-		awsProviderTemplateName = "cluster-api-provider-aws-0-0-4"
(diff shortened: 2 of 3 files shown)


internal/util/validation/mgmt.go
```diff
@@ -55,24 +53,77 @@ func GetIncompatibleContracts(ctx context.Context, cl client.Client, release *kc
...
+
+	for _, tplName := range templateNames {
 		pTpl := new(kcmv1.ProviderTemplate)
 		if err := cl.Get(ctx, client.ObjectKey{Name: tplName}, pTpl); err != nil {
 			return "", fmt.Errorf("failed to get ProviderTemplate %s: %w", tplName, err)
 		}
 
 		if len(pTpl.Status.CAPIContracts) == 0 {
-			continue
+			return "", nil
```
zerospiel: why have we changed the logic here? shouldn't we check all of the provider templates?

some unit tests to check the logic would be appreciated because the logic is not a light breeze and changes to it might break in an unexpected way

eromanova: That's definitely a leftover that was added by mistake during refactoring and should be fixed. Thanks, Michael.

zerospiel (approved):

Issue #2179, opened 2025-11-14
[bug] Race condition between ksm-sveltos-adapter and Sveltos controllers causes Profile update failures


PhiRho:
## Description

The `ksm-sveltos-adapter` controller in `kcm-controller-manager` frequently fails to update Profile resources due to resource version conflicts with Sveltos controllers. This prevents ClusterDeployment service value changes from propagating to deployed Helm releases.

## Steps to Reproduce

1. Create a ClusterDeployment with services configured
2. Update service values in ClusterDeployment spec (e.g., add or modify a Helm value field)
3. Observe ServiceSet events for the cluster
4. Note frequent "ServiceSetEnsureProfileFailed" warnings with "object has been modified" errors

## Expected Behavior

When ClusterDeployment service values are updated:
1. ClusterDeployment controller updates ServiceSet spec
2. ServiceSet controller updates Profile spec
3. Sveltos processes Profile and deploys Helm release with new values
4. Updates complete within normal reconciliation period (minutes)

## Actual Behavior

**Symptoms:**
- ServiceSet controller repeatedly fails to update Profile with error: `Operation cannot be fulfilled on profiles.config.projectsveltos.io "cluster-name": the object has been modified; please apply your changes to the latest version and try again`
- Profile generation increments rapidly (observed: 47,000+ over 7 days)
- Service value changes take unpredictable time to propagate (or may not propagate)
- Update success rate ~98% but failures cluster together, causing delays

**Root Cause:**
Race condition between two controllers modifying the same Profile:
- **ksm-sveltos-adapter** (k0rdent): Updates Profile `.spec` with service configurations
- **Sveltos addon-controller**: Updates Profile `.status` with cluster matching info
- Both controllers reconcile frequently and independently
- When both attempt updates concurrently, resource version conflicts occur
- ksm-sveltos-adapter uses naive update without retry logic

## Evidence

```
ServiceSet Events (over 7 day period):
  Type     Reason                          Count   Message
  ----     ------                          -----   -------
  Normal   ServiceSetEnsureProfileSuccess  47714   Successfully ensured Profile
  Warning  ServiceSetEnsureProfileFailed   963     Failed to ensure Profile: object has been modified

Profile Resource:
  metadata.generation: 47158
  (Indicates ~47k spec updates over 7 days = ~6.7 updates/hour average)
```

## Environment

[...]

PR #2270, opened 2025-12-16
fix: profile spec equality check to consider defaults

wahabmk:
# Description

- In the Sveltos ServiceSet controller, we compare the provided Profile/ClusterProfile spec with the existing one in the kube cluster to determine if it needs to be reconciled. 
- The equality check did not take into account the defaults that are already set on any Profile/ClusterProfile object fetched from the kube cluster. So if the user provided spec that had empty fields it wouldn't match with the fetched spec.
- This PR enhances the equality check to consider defaults.

Fixes https://github.com/k0rdent/kcm/issues/2179.

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -61,6 +61,7 @@ import (
 )
 
 const (
+	defaultTier             = 100
 	sveltosDriftIgnorePatch = `- op: add
   path: /metadata/annotations/projectsveltos.io~1driftDetectionIgnore
   value: ok`
@@ -442,8 +443,9 @@ func (*ServiceSetReconciler) createOrUpdateProfile(ctx context.Context, rgnClien
 		if err = rgnClient.Create(ctx, profile); err != nil {
 			return fmt.Errorf("failed to create Profile for ServiceSet %s: %w", serviceSet.Name, err)
 		}
-	// if profile spec is not equal to the spec we just created,
-	// we need to update it
+	// If profile spec is not equal to the spec we just created so
+	// we need to update it. Make sure that the empty values in `spec`
+	// are defaulted otherwise comparison will always return false.
 	case annotationsUpdated || !equality.Semantic.DeepEqual(profile.Spec, *spec):
 		profile.OwnerReferences = []metav1.OwnerReference{*ownerReference}
 		profile.Spec = *spec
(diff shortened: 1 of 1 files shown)


internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -1554,3 +1555,71 @@ func clusterReference(serviceSet *kcmv1.ServiceSet) *corev1.ObjectReference {
 		APIVersion: kcmv1.GroupVersion.WithKind(kcmv1.ClusterDeploymentKind).GroupVersion().String(),
 	}
 }
+
+// isProfileSpecEqual compares provided Sveltos Profile spec.
+//
+// TODO: Maybe there is a better way to do it by fetching the defaults
```
zerospiel: if there are schema-level defaults set, then they are being set after the validation, and at this moment (when the object is already in the server's data store) these defaults are already propagated. that being said, if the `syncMode` has a default `continious`, then the `syncMode` can't have an empty value

though, one should consider to properly set defaults, e.g.

```
	for i := range s1.KustomizationRefs {
		if s1.KustomizationRefs[i].DeploymentType == "" {
			s1.KustomizationRefs[i].DeploymentType = addoncontrollerv1beta1.DeploymentTypeRemote
		}
	}
```

in this code, in this code, clear that the `deploymentType` has the default `remote`, though the parent field holding this field should also have the default value, e.g. an empty array or object

just in case to reference something: https://ahmet.im/blog/crd-generation-pitfalls/#field-defaulting

zerospiel: Having this, feel free to resolve the comment, because I was just saying how one could handle this without writing extra code

In regards to the release itself, I'd ask @eromanova

wahabmk: You are correct that the fields should be defaulted once they are in the server. In the following original line of code that does the comparison:
https://github.com/k0rdent/kcm/blob/ed65e6c8e3fd8e0acae5ee42450fdbf79e142640/internal/controller/adapters/sveltos/serviceset_controller.go#L447

The `profile.Spec` already has the defaults set because it is fetched from the kube cluster. However, `*spec` is constructed using the [profileSpec()](https://github.com/k0rdent/kcm/blob/ed65e6c8e3fd8e0acae5ee42450fdbf79e142640/internal/controller/adapters/sveltos/serviceset_controller.go#L517) func from the ServiceSet where the Sveltos related fields are in [`serviceSet.spec.provider.config`](https://github.com/k0rdent/kcm/blob/ed65e6c8e3fd8e0acae5ee42450fdbf79e142640/api/v1beta1/serviceset_types.go#L116). So that is why `*spec` does not have the defaults applied. 

Perhaps it would be better to set these defaults in the `profileSpec()` function where the spec is being constructed?

zerospiel: Got it, thanks for the elaboration. Deciding whether to move the current changes to this method up to the implementer. At the moment, everything looks good to me

Issue #2278, opened 2025-12-18
[bug] Invalid service namespace (starting with “.”) not rejected - causes cascading errors


izadorozhna:
**Describe the bug**
Service with a dot-prefixed namespace is not validated, causing failures when adding other valid services later:
```
failureMessage: 'query: failed to query with labels: invalid namespace ".": [may not be ''.'']'
```
Reproduced at `k0rdent-enterprise-1-2-0` version.

**To Reproduce**
1. Create a child cluster deployment in k0rdent.
2. Create a HelmRepository and a ServiceTemplate (no matter what repo & service you take).
3. Add a service to the cluster deployment from this ServiceTemplate; the service namespace should start with a dot ("."). For example, this is how I accidentally reproduced the issue:
```
spec:
  serviceSpec:
    services:
    - name: ./charts/prometheus
      namespace: ./charts/prometheus
      template: prometheus-25-8-2
```
4. The addition of this service will fail. After this, remove the invalid service, and try adding a valid service with a valid namespace name. For example, I added a `redis` service:
```
    services:
    - name: redis
      namespace: redis
      template: redis-0-2-0
```

**Expected behavior**
The validators should exist at the controller/admission side. They should prevent the creation of an invalid namespace in the child clusters.  If the user accidentally sets a dot for the namespace name, the validation should check such cases. After this, the user can add other valid services, and the child cluster remains intact.

**Actual result**
Even after removing the invalid service and adding a valid service (e.g. `redis` with a valid namespace name `redis`), the cld was not ready because the valid service could not be ready since the cld DB had the invalid namespace in the past:
```
k get cld -A
kcm-system   ksi-mirantis-test-k0rdent-scp-01      False   0/1        openstack-standalone-cp-1-0-11   0/1 Services are ready.           108m
```
In fact, the valid `redis` service was added successfully at the target child cluster; however, the cld was not ready and had a failure:
```
services:
  - failureMessage: 'query: failed to query with labels: invalid namespace ".": [may
      not be ''.'']'
    lastStateTransitionTime: "2025-12-17T00:54:45Z"
    name: redis
    namespace: redis
    state: Failed
    template: redis-0-2-0
    type: Helm
    version: redis-0-2-0
  servicesUpgradePaths:
  - availableUpgrades:
    - upgradePaths:
      - redis-0-2-0
    name: redis
    namespace: redis
    template: redis-0-2-0
```
As you see, in fact, the redis was added successfully - all redis resources were created at the target cld:
```
[...]

PR #2282, opened 2025-12-18
fix: add namespace validation for services

kylewuolle:
Simple fix to add namespace validation on services within the webhook.
#2278

### DIFF
--- a/internal/util/validation/services.go
+++ b/internal/util/validation/services.go
@@ -19,7 +19,7 @@ import (
 	"errors"
 	"fmt"
 
-	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


internal/util/validation/services.go
```diff
@@ -35,6 +37,12 @@ var (
 func ServicesHaveValidTemplates(ctx context.Context, cl client.Client, services []kcmv1.Service, ns string) error {
 	var errs error
 	for _, svc := range services {
+		if svc.Namespace != "" {
+			for _, msg := range validation.IsDNS1123Label(svc.Namespace) {
```
wahabmk: There is also [`ValidateNamespaceName`](https://github.com/kubernetes/apimachinery/blob/84d740c9e27f3ccc94c8bc4d13f1b17f60f7080b/pkg/api/validation/generic.go#L66-L69) func which is an indirect alias to this one but may be better to use as its name is more indicative that it should be used to verify namespace.

kylewuolle: done!

wahabmk (approved):

Issue #2094, opened 2025-10-16
[enhancement] Separated 'staging' repo path for the artifacts


va4st:
**Is your enhancement / feature request related to a problem? Please describe.**
Right now we have two different repos and behaviours for kcm artifacts:
* The `charts` repo and the `controller` image
* The `charts-ci` repo and the `controller-ci` image.

Currently we have an issue to continuously check the upgrades to CI versions that built after merging (e.g to test 1.4.0 -> 1.4.0-20-g490f0705).
The issue appears because the templates repo for released versions is `oci://ghcr.io/k0rdent/kcm/charts/kcm-templates` and for unreleased (ci) `oci://ghcr.io/k0rdent/kcm/charts-ci/kcm-templates`. 
This leads to errors like 
```bash
{"level":"info","ts":"2025-10-15T16:58:16Z","msg":"HelmRelease is not ready yet","controller":"release","controllerGroup":"k0rdent.mirantis.com","controllerKind":"Release","Release":{"name":"kcm-1-4-0-15-g47cbc954"},"namespace":"","name":"kcm-1-4-0-15-g47cbc954","reconcileID":"0b2132e5-c38d-449b-b888-9d8a1e9973b9","HelmRelease":{"name":"kcm-1-4-0-15-g47cbc954-tpl","namespace":"kcm-system"},"message":"HelmChart 'kcm-system/kcm-1-4-0-15-g47cbc954-tpl' is not ready: chart pull error: failed to download chart for remote reference: failed to get 'oci://ghcr.io/k0rdent/kcm/charts/kcm-templates:1.4.0-15-g47cbc954': ghcr.io/k0rdent/kcm/charts/kcm-templates:1.4.0-15-g47cbc954: not found"
```

This basically means that for successful upgrade the release (released or CI) entities should persist in the same repository.
The same thing applicable to the controller image name (controller for releases, controller-ci for CI versions).

All that things right now stopping us from adding upgrade tests to the CI and continuously monitor the upgrade possibility and status to the latest code.

**Describe the solution you'd like**
The proposal is simple - create the staging repo path and store in that repo both time of artifafcts with the same names (ci and releases).
E.g: the structure for charts should be 
```bash
oci://ghcr.io/k0rdent/kcm/staging/kcm-templates:1.4.0
oci://ghcr.io/k0rdent/kcm/staging/kcm-templates:1.4.0-15-g47cbc954
[...]

PR #2314, opened 2025-12-30
ci: Set staging chart defaults properly

va4st:
This PR introduces configuration abilities for telemetry image repo and templates repo using Makefile to make corresponding charts (in that case - staging one) use valid defaults.

related k0rdent/kcm#2094

### DIFF
--- a/.github/workflows/build-staging.yaml
+++ b/.github/workflows/build-staging.yaml
@@ -70,7 +70,7 @@ jobs:
           IMG_TELEMETRY: ghcr.io/k0rdent/kcm/staging/telemetry:${{ steps.version.outputs.version }}
         run: |
           make set-kcm-repo
-          make set-telemetry-repo
+          make set-regional-telemetry-repo
           make set-templates-repo
           make kcm-chart-release
           make helm-push
--- a/Makefile
+++ b/Makefile
@@ -106,9 +106,9 @@ set-kcm-version: yq
 set-kcm-repo: yq
 	$(YQ) eval '.image.repository = "$(IMG_REPO)"' -i $(PROVIDER_TEMPLATES_DIR)/kcm/values.yaml
 
-.PHONY: set-telemetry-repo
-set-telemetry-repo: yq
-	$(YQ) eval '.regional.telemetry.controller.image.repository = "$(IMG_TELEMETRY_REPO)"' -i $(PROVIDER_TEMPLATES_DIR)/kcm/values.yaml
+.PHONY: set-regional-telemetry-repo
+set-regional-telemetry-repo: yq
+	$(YQ) eval '.telemetry.controller.image.repository = "$(IMG_TELEMETRY_REPO)"' -i $(PROVIDER_TEMPLATES_DIR)/kcm-regional/values.yaml
 
 .PHONY: set-templates-repo
 set-templates-repo: yq


Makefile
```diff
@@ -106,6 +106,14 @@ set-kcm-version: yq
 set-kcm-repo: yq
 	$(YQ) eval '.image.repository = "$(IMG_REPO)"' -i $(PROVIDER_TEMPLATES_DIR)/kcm/values.yaml
 
+.PHONY: set-telemetry-repo
+set-telemetry-repo: yq
+	$(YQ) eval '.regional.telemetry.controller.image.repository = "$(IMG_TELEMETRY_REPO)"' -i $(PROVIDER_TEMPLATES_DIR)/kcm/values.yaml
```
a13x5: It's better to set this in the `kcm-regional` chart directly. Versions of kcm and kcm-regional are now synced + this will potentially have an impact when regional deployed separately.

va4st: agreed, fixed

Makefile
```diff
@@ -106,6 +106,14 @@ set-kcm-version: yq
 set-kcm-repo: yq
 	$(YQ) eval '.image.repository = "$(IMG_REPO)"' -i $(PROVIDER_TEMPLATES_DIR)/kcm/values.yaml
 
+.PHONY: set-regional-telemetry-repo
+set-regional-telemetry-repo: yq
+	$(YQ) eval '.telemetry.controller.image.repository = "$(IMG_TELEMETRY_REPO)"' -i $(PROVIDER_TEMPLATES_DIR)/kcm-regional/values.yaml
```
zerospiel, line 111: shouldn't these variables be set in the workflow?

.github/workflows/build-staging.yaml
```diff
@@ -70,6 +70,8 @@ jobs:
           IMG_TELEMETRY: ghcr.io/k0rdent/kcm/staging/telemetry:${{ steps.version.outputs.version }}
         run: |
           make set-kcm-repo
+          make set-regional-telemetry-repo
```
zerospiel, line 73: shouldn't these targets be used for the post-merge target as well? if so, maybe we can combine them in order not to increase the amount of ambiguous targets in the makefile (that already has dozens of such targets)?

Makefile
va4st: It's dynamically obtained in the very beginning of Makefile, if I got everything correctly:
```bash
IMG_REPO = $(shell echo $(IMG) | cut -d: -f1)
IMG_TELEMETRY_REPO = $(shell echo $(IMG_TELEMETRY) | cut -d: -f1)
```

.github/workflows/build-staging.yaml
va4st: It's question for @a13x5, I believe :) 
I'm actually think about extending set-kcm-repo target, but decided to create a separated ones to not change existing workflows significantly.

Makefile
zerospiel: right, the `REGISTRY_REPO` is set in the WF, and the others are also set thus propagated, i missed it for a moment

.github/workflows/build-staging.yaml
a13x5: This change is only relevant for this workflow, since we already have proper default repository defined in the values.
This workflow however introduces additional repository `ghcr.io/k0rdent/kcm/staging` which should be patched to the regional chart in order for it to work when installing separately (in a region).

I understand the complain, but I believe this yq patch should be clearly separated to this exact workflow. Additional make target looks fine to me. If this is confusing the same step could be done as an in-line script in the same step (instead of make invocation), but it less preferable, since we use make in such way across all other workflows.

zerospiel (approved):

a13x5 (approved):

PR #2323, opened 2026-01-07
feat(controller): add global proxy settings

### LINKED ISSUE #2300: [enhancement] Support global HTTP Proxy configuration
**Context**:
There is no proper way to set up proxy configuration for all components in k0rdent if it’s required, and the user should do that on a per-component basis by passing the specific parameters/custom templates to achieve the goal: use proxy for communications outside of the system. Some customers use a proxy to handle communications to the internet, such as communication to AWS endpoints.

**What to do**:
go runtime uses the standard env variables for proxy configuration (`HTTPS_PROXY`, `HTTP_PROXY`, `NO_PROXY`), so it should be sufficient just to pass those for each controller which is require outside communications.

For the providers, the configuration could be generalized by exposing the manager’s container env via values. For the rest of the components, corresponding values should be used.

The child cluster components proxy configuration and registry communication should be out of scope, since:

* It highly depends on the user requirements (e.g., a cluster created on AWS with an internal proxy enabled will most probably have issues, but a vSphere cluster may need it), so it should be offloaded to the custom cluster template creation, with the description of what should be changed. 
* k0rdent is not in control of the management cluster, so realistically, we can only configure registry communications for Flux and sveltos, which will change the way they download charts.

To simplify UX, the proxy configuration should be globally set in the controller values, so proxy env values are propagated through the system automatically without the need to provide them for each component individually (as it is now).

zerospiel:
Adds standard proxy settings to the kcm chart, propagating the values to all of the CAPI providers, CAPI itself, as well as the kcm-regional chart values (if applicable).

The logic does NOT set the following:
- any variables related to `projectsveltos` "provider"
- any variables or settings related to managed clusters (no propagation to any of `clustertemplates`)

There is no possibility to configure volumes and mounts in any CAPI-Operator `*Provider` object (e.g., `InfrastructureProvider`), so there is no room for the certificates option (yet).

The logic uses the given `Secret` with the proxy data, then it propagates the data to the `kcm-regional` chart, only for the telemetry-collector, only for the `online` use-case, and to all of the available `providertemplates` via the environment variables.

Closes #2300

### DIFF
--- a/templates/provider/cluster-api-provider-infoblox/values.schema.json
+++ b/templates/provider/cluster-api-provider-infoblox/values.schema.json
@@ -32,6 +32,17 @@
     "manager": {
       "description": "Controller manager settings, such as feature gates, verbosity and other options",
       "type": "object"
+    },
+    "proxy": {
+      "description": "Proxy settings of the controller",
+      "type": "object",
+      "properties": {
+        "enabled": {
+          "description": "Enable proxy",
+          "type": "boolean"
+        }
+      },
+      "additionalProperties": false
     }
   }
 }
--- a/templates/provider/cluster-api-provider-infoblox/values.yaml
+++ b/templates/provider/cluster-api-provider-infoblox/values.yaml
@@ -8,3 +8,6 @@ config: {} # @schema type: object; description: Key-value configuration data sto
 manager: {} # @schema type: object; description: Controller manager settings, such as feature gates, verbosity and other options
 
 deployment: {} # @schema type: object; description: Deployment-level overrides for the provider controller, including replicas, containers, and scheduling settings
+
+proxy: # @schema type: object; description: Proxy settings of the controller; additionalProperties: false
+  enabled: true # @schema type: boolean; description: Enable proxy
--- a/templates/provider/cluster-api-provider-openstack/Chart.yaml
+++ b/templates/provider/cluster-api-provider-openstack/Chart.yaml
@@ -13,7 +13,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.14
+version: 1.0.15
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
--- a/templates/provider/kcm-templates/files/release.yaml
+++ b/templates/provider/kcm-templates/files/release.yaml
@@ -22,7 +22,7 @@ spec:
     - name: cluster-api-provider-aws
       template: cluster-api-provider-aws-1-0-12
     - name: cluster-api-provider-openstack
-      template: cluster-api-provider-openstack-1-0-14
+      template: cluster-api-provider-openstack-1-0-15
     - name: cluster-api-provider-docker
       template: cluster-api-provider-docker-1-0-9
     - name: cluster-api-provider-gcp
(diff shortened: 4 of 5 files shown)


internal/telemetry/runner.go
```diff
@@ -62,7 +69,37 @@ func NewRunner(cfg *Config) (*Runner, error) {
 			tz = loc.String()
 		}
 
+		base, ok := http.DefaultTransport.(*http.Transport)
+		if !ok {
+			return nil, errors.New("http.DefaultTransport is not a *http.Transport")
+		}
+		transport := base.Clone()
```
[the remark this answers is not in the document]
zerospiel: Configuration phase, no need to create from scratch, considering we need either the default one or one with the modified `.Proxy` function

templates/provider/cluster-api-provider-k0sproject-k0smotron/templates/_helpers.tpl
```diff
@@ -1,18 +1,83 @@
+{{/*
+Build proxy env vars if global.proxy is set
+*/}}
+{{- define "infrastructureProvider.proxyEnv" -}}
```
eromanova: nit: Since this helper template is shared across all k0smotron providers (infrastructure, bootstrap, controlPlane), It might make sense to give it a more generic name (ex. simply `provider.proxyEnv`).

internal/controller/components/values.go
```diff
@@ -149,13 +157,25 @@ func getComponentValues(
...
 	return &apiextv1.JSON{Raw: raw}, nil
 }
 
+func getProxyConfig() (map[string]string, bool) {
+	secretName, ok := os.LookupEnv("PROXY_SECRET")
+	if !ok || len(secretName) == 0 {
+		return nil, false
+	}
+	return map[string]string{
+		"secretName": secretName,
```
eromanova, line 176: Should the secret containing the proxy settings be automatically propagated to the regional clusters, or is this expected to be a manual step performed by the user?

templates/provider/cluster-api-provider-openstack/templates/_helpers.tpl
```diff
@@ -1,17 +1,83 @@
 {{/*
-Build the default deployment settings when global.registry is set
+Build proxy env vars if global.proxy is set
+*/}}
+{{- define "infrastructureProvider.proxyEnv" -}}
```
eromanova, line 4: Currently, the orc controller does not have proxy settings, should we adapt its manifest as well?

zerospiel: We should, indeed. Thans for noticing! Done

internal/controller/components/values.go
zerospiel: It should be, it goes without saying, thanks for noticing! Done

templates/provider/cluster-api-provider-k0sproject-k0smotron/templates/_helpers.tpl
zerospiel: Done

eromanova (approved):
LGTM. But we still have a problem with the orc component that deploys in the `orc-system` namespace, while we expect the proxy secret to exist in the system namespace. We need a discussion about how to properly address it automatically. For now, I think we can add a note to the documentation.

zerospiel:
> LGTM. But we still have a problem with the orc component that deploys in the `orc-system` namespace, while we expect the proxy secret to exist in the system namespace. We need a discussion about how to properly address it automatically. For now, I think we can add a note to the documentation.

IINM, we suggest the same regarding the `imagePullSecrets` for this case (@a13x5 could you please confirm that: do we ask users to prepare secrets, when the global `imagePullSecrets` is given, and then copy them to the `orc-system` namespace?). If this is the same case, then I agree, that the docs should suffice

PR #2324, opened 2026-01-08
feat: add helm actions and install, upgrade and uninstall options for services

### LINKED ISSUE #2261: [enhancement] extend helm integration with action-specific options
We should be able to specify [action-specific options](https://github.com/projectsveltos/addon-controller/blob/3b772536ea4197648c845a0102fbfa69077b0da2/api/v1beta1/spec.go#L239-L341) for helm charts when deploying services as they are supported by k0rdent default KSM provider (projectsveltos)

kylewuolle:
This PR adds the helmAction and also the install, upgrade and uninstall helm options for services.
 
fixes #2261 and #2366

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -418,41 +418,45 @@ func (r *ServiceSetReconciler) ensureProfile(ctx context.Context, rgnClient clie
 
 func (*ServiceSetReconciler) createOrUpdateProfile(ctx context.Context, rgnClient client.Client, serviceSet *kcmv1.ServiceSet, spec *addoncontrollerv1beta1.Spec) error {
 	ownerReference := metav1.NewControllerRef(serviceSet, kcmv1.GroupVersion.WithKind(kcmv1.ServiceSetKind))
-
-	profile := new(addoncontrollerv1beta1.Profile)
-	key := client.ObjectKeyFromObject(serviceSet)
-	err := rgnClient.Get(ctx, key, profile)
-	// if IgnoreNotFound returns non-nil error, this means that
-	// an error occurred while trying to request kube-apiserver
-	if client.IgnoreNotFound(err) != nil {
-		return fmt.Errorf("failed to get Profile: %w", err)
-	}
-
-	annotationsUpdated := handlePauseAnnotations(&profile.ObjectMeta, serviceSet)
-
-	switch {
-	// we already excluded all errors except NotFound
-	// hence if the error is not nil, it means that the object was not found
-	case err != nil:
-		profile.Name = serviceSet.Name
-		profile.Namespace = serviceSet.Namespace
-		profile.Labels = map[string]string{
-			kcmv1.KCMManagedLabelKey: kcmv1.KCMManagedLabelValue,
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		profile := new(addoncontrollerv1beta1.Profile)
+		key := client.ObjectKeyFromObject(serviceSet)
+		err := rgnClient.Get(ctx, key, profile)
+		// if IgnoreNotFound returns non-nil error, this means that
+		// an error occurred while trying to request kube-apiserver
+		if client.IgnoreNotFound(err) != nil {
+			return fmt.Errorf("failed to get Profile: %w", err)
 		}
-		profile.OwnerReferences = []metav1.OwnerReference{*ownerReference}
-		profile.Spec = *spec
-		if err = rgnClient.Create(ctx, profile); err != nil {
-			return fmt.Errorf("failed to create Profile for ServiceSet %s: %w", serviceSet.Name, err)
+
+		annotationsUpdated := handlePauseAnnotations(&profile.ObjectMeta, serviceSet)
+
+		switch {
+		// we already excluded all errors except NotFound
+		// hence if the error is not nil, it means that the object was not found
+		case err != nil:
+			profile.Name = serviceSet.Name
+			profile.Namespace = serviceSet.Namespace
+			profile.Labels = map[string]string{
+				kcmv1.KCMManagedLabelKey: kcmv1.KCMManagedLabelValue,
+			}
+			profile.OwnerReferences = []metav1.OwnerReference{*ownerReference}
+			profile.Spec = *spec
+			if err = rgnClient.Create(ctx, profile); err != nil {
+				return fmt.Errorf("failed to create Profile for ServiceSet %s: %w", serviceSet.Name, err)
+			}
+		// If profile spec is not equal to the spec we just created so
+		// we need to update it. Make sure that the empty values in `spec`
+		// are defaulted otherwise comparison will always return false.
+		case annotationsUpdated || !equality.Semantic.DeepEqual(profile.Spec, *spec):
+			profile.OwnerReferences = []metav1.OwnerReference{*ownerReference}
+			profile.Spec = *spec
+			return rgnClient.Update(ctx, profile)
 		}
-	// If profile spec is not equal to the spec we just created so
-	// we need to update it. Make sure that the empty values in `spec`
-	// are defaulted otherwise comparison will always return false.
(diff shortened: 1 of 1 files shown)


api/v1beta1/multiclusterservice_types.go
wahabmk, line 206: Should we remove these fields or mark them deprecated because some of them are part of the newly introduced options? Like `Replace` is part of `addoncontrollerv1beta1.HelmInstallOptions`.

api/v1beta1/multiclusterservice_types.go
```diff
@@ -208,6 +213,21 @@ type ServiceHelmOptions struct {
...
+
+	// +optional
+
+	// UpgradeOptions are options specific to helm upgrade
+	UpgradeOptions *addoncontrollerv1beta1.HelmUpgradeOptions `json:"upgradeOptions,omitempty"`
+
+	// +optional
+
+	// InstallOptions are options specific to helm install
+	InstallOptions *addoncontrollerv1beta1.HelmInstallOptions `json:"installOptions,omitempty"`
```
wahabmk, line 231: I think we should have sveltos-agnostic fields in our API as much as possible. The existing sveltos fields are already marked deprecated.

api/v1beta1/multiclusterservice_types.go
kylewuolle: Done!

api/v1beta1/multiclusterservice_types.go
kylewuolle: I agree but I feel like the sveltos options were well designed and we'll end up using the same fields if we start adding more providers.

api/v1beta1/multiclusterservice_types.go
wahabmk, line 169: This is part of `HelmInstallOptions` so should be marked Deprecated.

api/v1beta1/multiclusterservice_types.go
wahabmk: I think functionally there is no problem with referencing sveltos fields directly. However, there is precedent where we opted to recreate 3rd party fields in our API rather than reusing them. E.g. in https://github.com/k0rdent/kcm/pull/1203#discussion_r2008789639 in which case I actually argued in favour of reusing the 3rd party fields at that time 😅 but we decided to go the route of recreating them to avoid dependency. It's a trade-off: low maintenance of importing & reusing 3rd party fields VS the low dependence of recreating them but then updating them as needed with new features.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -488,9 +489,28 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
...
+		})
+
+		/*
+			err := rgnClient.Get(ctx, key, profile)
+				// if IgnoreNotFound returns non-nil error, this means that
+				// an error occurred while trying to request kube-apiserver
+				if client.IgnoreNotFound(err) != nil {
+					return fmt.Errorf("failed to get Profile: %w", err)
+				}
+		*/
```
wahabmk: leftover commented code.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -488,9 +489,28 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
...
-		if err = rgnClient.Update(ctx, profile); err != nil {
-			return fmt.Errorf("failed to update ClusterProfile for ServiceSet %s: %w", serviceSet.Name, err)
+
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			err := rgnClient.Get(ctx, key, profile)
+			// if IgnoreNotFound returns non-nil error, this means that
+			// an error occurred while trying to request kube-apiserver
+			if client.IgnoreNotFound(err) != nil {
+				return fmt.Errorf("failed to get Profile: %w", err)
+			}
```
wahabmk:
```suggestion
if err := rgnClient.Get(ctx, key, profile);  err != nil {
  return fmt.Errorf("failed to get Profile: %w", err)
}
```
I am thinking we should not filter the Not Found error here, here's my reasoning.

We get profile from kube at the start of this function. So within this case of the switch statement we shouldn't encounter Profile not found (as it is handled by the case above it). However, now that we are getting profile from kube again with `err := rgnClient.Get(ctx, key, profile)`, there is a minuscule chance that the profile object becomes deleted between the 1st time profile was retrieved and this time.

If that happens then `client.IgnoreNotFound(err)` will return nil and the `profile` reference will still be referencing data for the now deleted object and it will be assigned spec and attempted to be updated to the kube server. 

I am not sure what impact that might have but it's probably safer to avoid this scenario, so that if we get the Not Found in a rare circumstance, the function simply returns with that error and the reconciler attempts again.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -488,9 +489,28 @@ func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, r
 	// are defaulted otherwise comparison will always return false.
 	case annotationsUpdated || !equality.Semantic.DeepEqual(profile.Spec, *spec):
 		profile.OwnerReferences = []metav1.OwnerReference{*ownerReference}
-		profile.Spec = *spec
-		if err = rgnClient.Update(ctx, profile); err != nil {
-			return fmt.Errorf("failed to update ClusterProfile for ServiceSet %s: %w", serviceSet.Name, err)
+
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
wahabmk: Can we add this logic to reduce conflicts to the `createOrUpdateProfile` function as well?

internal/controller/adapters/sveltos/serviceset_controller.go
wahabmk: Thinking about it more. Maybe we can do something like this where we don't need to call `rgnClient.Get(ctx, key, profile)` twice and reduce an extra call to kube API:

```go
func (*ServiceSetReconciler) createOrUpdateClusterProfile(ctx context.Context, rgnClient client.Client, serviceSet *kcmv1.ServiceSet, spec *addoncontrollerv1beta1.Spec) error {
	ownerReference := metav1.NewControllerRef(serviceSet, kcmv1.GroupVersion.WithKind(kcmv1.ServiceSetKind))
	profile := new(addoncontrollerv1beta1.ClusterProfile)
	key := client.ObjectKeyFromObject(serviceSet)

	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
		err := rgnClient.Get(ctx, key, profile)
		if client.IgnoreNotFound(err) != nil {
			return fmt.Errorf("failed to get Profile: %w", err)
		}

		annotationsUpdated := handlePauseAnnotations(&profile.ObjectMeta, serviceSet)

		switch {
		case err != nil:
			. . . . .
			if err = rgnClient.Create(ctx, profile); err != nil {
				return fmt.Errorf("failed to create ClusterProfile for ServiceSet %s: %w", serviceSet.Name, err)
			}
		case annotationsUpdated || !equality.Semantic.DeepEqual(profile.Spec, *spec):
			profile.OwnerReferences = []metav1.OwnerReference{*ownerReference}
			profile.Spec = *spec
			if err := rgnClient.Update(ctx, profile); err != nil {
				return fmt.Errorf("failed to update ClusterProfile for ServiceSet %s: %w", serviceSet.Name, err)
			}
		}
		return nil
	})
	return retryErr
}
```
For the cases where we create the Profile instead of updating it, returning error from `rgnClient.Create(ctx, profile)` should be safe within `retry.RetryOnConflict` as it will simply return non-conflict errors without retrying if I understand it correctly.

internal/controller/adapters/sveltos/serviceset_controller.go
wahabmk, line 419: LGTM but can we also add the `RetryOnConflict` block to this function as well? It is almost the same as `createOrUpdateClusterProfile` func.

wahabmk (approved):

Issue #1863, opened 2025-08-16
[bug] KCM templates should avoid deleting Flux CRDs as it triggers


aglarendil:
**Describe the bug**
KCM top release deletion leads to all the helm releases deletion along with all the child clusters. KCM release deletion can be triggered inadvertently by some bug in release controllers or by a user unknowingly trying to reinstall it crushing the production. IMO, this is a critical issue and is barely acceptable.

**To Reproduce**
1. deploy KCM and leave happily with some of your clusters
2. KCM helm chart deletion is triggered manually or due to a bug in Flux controller (e.g. fluxcd deletes a release when chart name is changed, which is a behaviour not similar to basic helm binary)
3. observe that all your clusters installed as helm releases are now deleted with all their data, hopes and dreams

**Expected behavior**
clusters should be deleted only when they are explicitly deleted by a user, not when there is some cascade of deletion

**Additional context**
Suggestion: all the CRDs that bear critical meaning should be protected somehow with guardrails or installed with specific helm annotations, or managed outside of the context of kcm install/upgrade/reinstall scenarios.
there are similar issues discussed within kof team

aglarendil:
my current issue is that CRDs were actually removed and the clusters were removed as well, I did not remove keep annotations from them, so what happens it that flux CRDs get cleaned up, not CRDs for k0rdent, and my clusters are helmreleases

```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.16.1
    meta.helm.sh/release-name: kcm
    meta.helm.sh/release-namespace: kcm-system
  creationTimestamp: "2025-09-01T12:50:39Z"
  generation: 1
  labels:
    app.kubernetes.io/component: helm-controller
    app.kubernetes.io/instance: kcm-system
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/part-of: flux
    app.kubernetes.io/version: 2.5.1
    helm.sh/chart: flux2-2.15.0
    helm.toolkit.fluxcd.io/name: kcm
    helm.toolkit.fluxcd.io/namespace: kcm-system
  name: helmreleases.helm.toolkit.fluxcd.io
  resourceVersion: "2100"
  uid: 3cf00632-eaa4-494e-b690-93544465d08a
```

tests for clusters non-deletion and guardrails are needed, I think.
we should unmanage CRDs of flux, so we do not delete them

[...]

PR #2337, opened 2026-01-16
fix: mark CRDs with helm-keep annotation where possible

Kshatrix:
Remaining CRDs that are not annotated (since the configuration is not exposed in the corresponding charts):
* all Sveltos CRDs
* rbac-manager CRD

Fixes: #1863

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,10 +69,12 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: MANAGEMENT_CRDS_DIR := $(PROVIDER_TEMPLATES_DIR)/kcm/templates/crds
 manifests: REGIONAL_CRDS_DIR := $(PROVIDER_TEMPLATES_DIR)/kcm-regional/templates/crds
-manifests: controller-gen ## Generate CustomResourceDefinition objects.
+manifests: controller-gen yq ## Generate CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) crd paths="./..." output:crd:artifacts:config=$(MANAGEMENT_CRDS_DIR)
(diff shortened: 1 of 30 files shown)


eromanova (dismissed):

Makefile
```diff
@@ -69,10 +69,12 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: MANAGEMENT_CRDS_DIR := $(PROVIDER_TEMPLATES_DIR)/kcm/templates/crds
 manifests: REGIONAL_CRDS_DIR := $(PROVIDER_TEMPLATES_DIR)/kcm-regional/templates/crds
-manifests: controller-gen ## Generate CustomResourceDefinition objects.
+manifests: controller-gen yq ## Generate CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) crd paths="./..." output:crd:artifacts:config=$(MANAGEMENT_CRDS_DIR)
+	find $(MANAGEMENT_CRDS_DIR) -maxdepth 1 -name "*.yaml" -exec $(YQ) eval -i '.metadata.annotations["helm.sh/resource-policy"] = "keep"' {} \;
```
zerospiel, line 74: do we want to call `yq` in a way, it does not change the current formatting?

zerospiel (commented):
I was under the impression that the idea is to keep flux-related CRDs, and that it could be achieved with the [provided cofnig option](https://github.com/fluxcd-community/helm-charts/blob/e07cfacf7dc57ff2f264a9a20add570861e02640/charts/flux2/values.yaml#L4-L6)

zerospiel (approved):

Issue #2291, opened 2025-12-22
[ci] [tests] add ability to skip mothership/ksm tests


zerospiel:
We have to have an ability to skip mothership and docker tests, and configure them directly through the currently existing mechanics (comment msg in the deployment's approve)

Original:
~~If no config has been provided in the approval comment, only [mothership tests](https://github.com/k0rdent/kcm/actions/runs/20430843344/job/58701390413#step:10:998) are running, which is unexpected, and at least `aws` cloud provider should be executed~~

Fixed in #2292

PR #2342, opened 2026-01-18
fix: run ksm tests via test config

wahabmk:
# Description
Fixes https://github.com/k0rdent/kcm/issues/2291.

Made mothership and functional tests run only when the config has `mothership` and `docker` entries respectively. Similar to [how it is done for aws tests](https://github.com/k0rdent/kcm/blob/f1af46d67aac7cf2f1447a72a1091e2e5151d59d/test/e2e/provider_aws_test.go#L145).

# Verification

I ran the e2e tests in CI and verified the following:
1. With empty or `aws: []` only the AWS test was run.
2. With `docker: []` only the functional tests were run.
3. With `mothership: []` only the mothership tests were run,

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,7 +209,7 @@ load-e2e-config: yq
 		echo "E2E_CONFIG_B64 is empty, the default configuration from test/e2e/config/config.yaml will be used"; \
 		exit 0; \
 	fi; \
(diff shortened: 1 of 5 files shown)


wahabmk:
> LGTM. Besides, is there an issue the PR is addressing?

I usually link the issue via the "Development" field on the right bar on github GUI but I've updated the description with it as well.

PR #2350, opened 2026-01-20
fix(cleanup): delete cld objects after servicesets

zerospiel:
This ensures that no new PVCs/PVs/LBs or other objects will be created after initial ones got deleted. If (cluster)profiles marked for deletion, it guarantees none helm releases rollouts, thus the leftovers could finally be collected for further deletion.

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -1225,6 +1225,26 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, mgmt 
 		return ctrl.Result{}, fmt.Errorf("failed to aggregate conditions from CAPI Cluster for ClusterDeployment %s: %w", client.ObjectKeyFromObject(cd), err)
 	}
 
+	if err := r.releaseProviderCluster(ctx, mgmt, scope); err != nil {
+		if r.IsDisabledValidationWH && errors.Is(err, errClusterTemplateNotFound) {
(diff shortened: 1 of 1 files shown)


internal/controller/clusterdeployment_controller.go
```diff
@@ -1225,6 +1225,26 @@ func (r *ClusterDeploymentReconciler) reconcileDelete(ctx context.Context, mgmt
 		return ctrl.Result{}, fmt.Errorf("failed to aggregate conditions from CAPI Cluster for ClusterDeployment %s: %w", client.ObjectKeyFromObject(cd), err)
 	}
 
+	if err := r.releaseProviderCluster(ctx, mgmt, scope); err != nil {
+		if r.IsDisabledValidationWH && errors.Is(err, errClusterTemplateNotFound) {
+			r.setCondition(cd, kcmv1.DeletingCondition, err)
+			l.Error(err, "failed to release provider cluster object due to absent ClusterTemplate, will not retrigger")
+			// there is not much to do, we cannot release the clusterdeployment without the clustertemplate
+			return ctrl.Result{}, nil
```
eromanova, line 1233: When the cluster template does not exist, we should still proceed with cleaning up service sets, PVCs, and related resources. I know this behavior has existed for a long time and is not introduced by your PR, but it might be worth addressing as part of this fix.

eromanova: Discussed in DM. It does not make much sense to cleanup other resources if we cannot release the infra cluster.

PR #2362, opened 2026-01-22
fix: helm values are not updated correctly

kylewuolle:
Simple fix for an issue where helm value updates (and other fields) are not reflected in the cluster profile / cluster summary correctly.

https://github.com/k0rdent/k0rdent-enterprise/issues/236


api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,6 +1,6 @@
 //go:build !ignore_autogenerated
 
-// Copyright 2025
+// Copyright 2026
```
a13x5, line 3: Could be omitted

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1,6 +1,6 @@
 //go:build !ignore_autogenerated
 
-// Copyright 2025
+// Copyright 2026
```
a13x5, line 3: Could be omitted

internal/serviceset/util_test.go
```diff
@@ -215,7 +215,7 @@ func Test_ServicesToDeploy(t *testing.T) {
 				{
 					Name:      "service2",
 					Namespace: metav1.NamespaceDefault,
-					Template:  "template2-1-0-0",
+					Template:  "template2-2-0-0",
```
a13x5, line 218: Is that necessary? Probably could be omitted

PR #2419, opened 2026-02-20
fix: resolve ServiceSet reconciliation issue (#2413)

### LINKED ISSUE #2413: [enh] Show number of services in summary of ServiceSet
# Description

Currently I have 2 services deployed via `mcs1` and 1 service deployed via `mcs2`. However, I can't tell that from looking at the kubectl summary of either the MCS:
```sh
➜  ~ kubectl get mcs
NAME   CLUSTERS   PROVIDER             SELF-MANAGEMENT   AGE
mcs1   1/1        ksm-projectsveltos   true              15m
mcs2   1/1        ksm-projectsveltos   true              15m
```

Or the ServiceSets:
```sh
➜  ~ kkcm get serviceset
NAME                  CLUSTER   MULTICLUSTERSERVER   PROVIDER             SELF-MANAGEMENT   AGE
management-3556efa5             mcs1                 ksm-projectsveltos   true              3m17s
management-fc79ea26             mcs2                 ksm-projectsveltos   true              2m39s
```

# Proposal
For the MCS I understand its not useful to convey service status because it can match many clusters. But currently we have only 1 ServiceSet for every unique combination of CD and MCS, so I think showing the status of services like so can be useful for convenience:
```sh
➜  ~ kkcm get serviceset
NAME                  CLUSTER   MULTICLUSTERSERVER   SERVICES  PROVIDER             SELF-MANAGEMENT   AGE
management-3556efa5             mcs1                 2/2       ksm-projectsveltos   true              3m17s
management-fc79ea26             mcs2                 1/1       ksm-projectsveltos   true              2m39s
```

sudheerpailu1017:

Adds a SERVICES column to kubectl get serviceset output showing the count of deployed services out of total in deployed/total format (e.g., 2/3). This gives users quick visibility into service deployment status without having to inspect each ServiceSet individually.

Fixes #2413

### DIFF
--- a/api/v1beta1/serviceset_types.go
+++ b/api/v1beta1/serviceset_types.go
@@ -174,6 +174,9 @@ type ServiceSetStatus struct {
 	// Cluster contains [k8s.io/api/core/v1.ObjectReference] to the cluster object.
 	Cluster *corev1.ObjectReference `json:"cluster,omitempty"`
 
+	// ServicesSummary shows the count of deployed services out of total (e.g., "2/2")
+	ServicesSummary string `json:"servicesSummary,omitempty"`
+
 	// +patchMergeKey=type
 	// +patchStrategy=merge
 	// +listType=map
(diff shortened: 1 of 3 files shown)


zerospiel (approved):

api/v1beta1/serviceset_types.go
```diff
@@ -174,6 +174,9 @@ type ServiceSetStatus struct {
 	// Cluster contains [k8s.io/api/core/v1.ObjectReference] to the cluster object.
 	Cluster *corev1.ObjectReference `json:"cluster,omitempty"`
 
+	// ServicesSummary shows the count of deployed services out of total (e.g., "2/2")
+	ServicesSummary string `json:"servicesSummary,omitempty"`
```
wahabmk, line 178: This can be achieved without introducing a new field in the API like it is [done](https://github.com/k0rdent/kcm/blob/a5d49276f4810ed7e088c5d7dc552e035acd19f6/api/v1beta1/clusterdeployment_types.go#L145) for ClusterDeployment, which I think is preferable.

wahabmk (changes requested):

PR #2423, opened 2026-02-20
feat(event)!: migrate to the new event recorder

### LINKED ISSUE #2420: Migrate to a new event recorder

zerospiel:
Migrates to the new EventRecorder iface, effectively substituting the current generalized wrappers.

Closes #2420

### DIFF
--- a/internal/controller/release_controller.go
+++ b/internal/controller/release_controller.go
@@ -175,7 +175,7 @@ func (r *ReleaseReconciler) validateProviderTemplates(ctx context.Context, relea
 	return nil
 }
 
-func (r *ReleaseReconciler) updateTemplatesValidCondition(release *kcmv1.Release, err error) (changed bool) {
+func (*ReleaseReconciler) updateTemplatesValidCondition(release *kcmv1.Release, err error) (changed bool) {
 	condition := metav1.Condition{
 		Type:               kcmv1.TemplatesValidCondition,
 		Status:             metav1.ConditionTrue,


internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -189,28 +189,28 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		Suspended: smp.Spec.Suspend,
 	}
 	if !smp.Status.Ready {
-		record.Eventf(serviceSet, serviceSet.Generation, kcmv1.StateManagementProviderNotReadyEvent,
+		record.Eventf(serviceSet, smp, kcmv1.StateManagementProviderNotReadyEvent, "reconciling",
```
eromanova: As said in the EventRecorder interface https://github.com/kubernetes/client-go/blob/v0.35.1/tools/internal/events/interfaces.go#L39:
```
	// 'action' explains what happened with regarding/what action did the ReportingController
	// (ReportingController is a type of a Controller reporting an Event, e.g. k8s.io/node-controller, k8s.io/kubelet.)
	// take in regarding's name; it should be in UpperCamelCase format (starting with a capital letter).
```

action should be in UpperCamelCase format. Can we change the `reconciling` action accordingly? Applicable everywhere.

internal/util/kube/secrets.go
```diff
@@ -124,7 +124,7 @@ func SetPredeclaredSecretsCondition(ctx context.Context, cl client.Client, base
 	missingMsg := fmt.Sprintf("Some of the predeclared Secrets (%v) are missing (%v) in the %s namespace", names, missingSecrets, namespace)
 
 	if changed = meta.SetStatusCondition(base.GetConditions(), getPredeclaredSecretsExistCondition(base.GetGeneration(), missingMsg)); changed {
-		eventFn(base, base.GetGeneration(), "MissingDeclaredSecrets", missingMsg)
+		record.Warnf(base, nil, "MissingDeclaredSecrets", "checking existence", missingMsg)
```
eromanova:
```suggestion
		record.Warnf(base, nil, "MissingDeclaredSecrets", "CheckingSecretsExistence", missingMsg)
```

internal/controller/adapters/sveltos/serviceset_controller.go
zerospiel: Thanks for catching, I was paying attention only to the `reason` and missed that the `action` should also be in the same case

internal/util/kube/secrets.go
zerospiel: Done here and elsewhere

internal/controller/management_controller.go
```diff
@@ -740,10 +740,12 @@ func (r *ManagementReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return managedController.Complete(r)
 }
 
+// TODO: FIXME: pass meaningful non-empty action
 func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message string, args ...any) {
-	record.Eventf(mgmt, mgmt.Generation, reason, message, args...)
+	record.Eventf(mgmt, nil, reason, "Reconcile", message, args...)
```
kylewuolle, line 745: Just wondering if we need this?

zerospiel: what do you mean by `this`? you mean the ~~`reason`~~ `action`? then yes, we must have it, it is a required field

PR #2425, opened 2026-02-20
fix: do not validate template / multi cluster service relationship if mcs is being deleted

kylewuolle:
This fix allows service templates to be deleted in the case where the multi cluster service that references it is being deleted.

https://github.com/k0rdent/k0rdent-enterprise/issues/191

### DIFF
--- a/internal/webhook/template_webhook.go
+++ b/internal/webhook/template_webhook.go
@@ -155,15 +155,15 @@ func (v *ServiceTemplateValidator) ValidateDelete(ctx context.Context, obj *kcmv
 			return nil, err
 		}
 
-		var mscNames []string
+		mcsNames := make([]string, 0, len(multiSvcClusters.Items))
 		for _, msc := range multiSvcClusters.Items {
 			if msc.GetDeletionTimestamp().IsZero() {
-				mscNames = append(mscNames, msc.Name)
+				mcsNames = append(mcsNames, msc.Name)
 			}
 		}
 
-		if len(mscNames) > 0 {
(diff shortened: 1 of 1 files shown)


internal/webhook/template_webhook.go
```diff
@@ -155,11 +155,14 @@ func (v *ServiceTemplateValidator) ValidateDelete(ctx context.Context, obj *kcmv
 			return nil, err
 		}
 
-		if len(multiSvcClusters.Items) > 0 {
-			mscNames := make([]string, len(multiSvcClusters.Items))
-			for i, msc := range multiSvcClusters.Items {
-				mscNames[i] = msc.Name
+		var mscNames []string
```
zerospiel: we might still want to preallocate the slice with `mcsNames := make([]string, 0, len(multiSvcClusters.Items))`

internal/webhook/template_webhook.go
```diff
@@ -155,11 +155,14 @@ func (v *ServiceTemplateValidator) ValidateDelete(ctx context.Context, obj *kcmv
...
+		for _, msc := range multiSvcClusters.Items {
+			if msc.GetDeletionTimestamp().IsZero() {
```
zerospiel, line 160: we might calculate this just once before the loop: `isDeleting := !msc.GetDeletionTimestamp().IsZero()`

internal/webhook/template_webhook.go
kylewuolle: Done!

internal/webhook/template_webhook.go
kylewuolle: Wish we could but we're looping through the mcs items here.

zerospiel: Ugh, you're right, I missed it, sorry! Thanks for pointing

PR #2446, opened 2026-02-25
fix: remote machines naming

### LINKED ISSUE #2445: [bug] Remote cluster machine names depend on list index, causing incorrect machine removal
**Describe the bug**
In the `remote-cluster` template, machine names depend on the index in the machines list. This causes incorrect behavior when a machine is removed from the middle of the list, as indices shift and the wrong machine is deleted.

**To Reproduce**

1. Create a ClusterDeployment using the remote-cluster template with multiple machines defined
2. Remove a machine from the ClusterDeployment configuration that is not the last entry in the machines list

**Expected behavior**
Only the specified machine is removed.

eromanova:
This PR fixes remote machine naming so it no longer depends on the list index.

Now it's possible to explicitly specify the name of the machine under `machines[*].name` or, if unspecified, the name will be generated automatically (`<clusterName>-<machineAddress>`).

Fixes #2445

### DIFF
--- a/templates/cluster/remote-cluster/templates/machines.yaml
+++ b/templates/cluster/remote-cluster/templates/machines.yaml
@@ -1,7 +1,7 @@
 {{- $global := .Values.global | default dict }}
 {{- range $machine := .Values.machines }}
-{{- $addressDash := replace "." "-" $machine.address }}
-{{- $defaultName := printf "%s-%s" (include "cluster.name" $) $addressDash -}}
+{{- $shaSum := sha256sum (printf "%s%s" $machine.address $machine.port) | trunc 6 }}
+{{- $defaultName := printf "%s-%s" (include "cluster.name" $) $shaSum -}}
 {{- $name := default $defaultName $machine.name -}}
 apiVersion: cluster.x-k8s.io/v1beta1
 kind: Machine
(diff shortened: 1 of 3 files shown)


templates/cluster/remote-cluster/templates/machines.yaml
```diff
@@ -1,25 +1,28 @@
 {{- $global := .Values.global | default dict }}
-{{- range $index, $machine := .Values.machines }}
+{{- range $machine := .Values.machines }}
+{{- $addressDash := replace "." "-" $machine.address }}
+{{- $defaultName := printf "%s-%s" (include "cluster.name" $) $addressDash -}}
+{{- $name := default $defaultName $machine.name -}}
```
zerospiel: can't we use something like `randBytes` function? or maybe sha sums shrinked down to X bytes to fit into the 256 symbols? in that way there won't be any strict reqs for the naming but just maybe 200-220 symbols

eromanova: I guess, using `randBytes` here is likely incorrect, because during upgrades it would generate a new name and trigger machine recreation. I’ve updated the code to derive the name from a SHA sum of the machine’s address and port instead.

PR #2458, opened 2026-02-27
test: add release controller integration tests

### LINKED ISSUE #2405: [tests] Cover release controller with integration tests
At the moment, the release controller code does not have any test coverage.

eromanova:

This PR adds integration tests for the release controller logic and covers most functionality.
Note: it does not verify the initial management object creation yet, since currently we create a single default management object before all tests start (it already exists at the moment the release controller tests run), and integration tests must be refactored to address this use case. 

Increases release controller test coverage 0% -> 66,1%

Fixes #2405

### DIFF
--- a/internal/controller/release_controller.go
+++ b/internal/controller/release_controller.go
@@ -374,7 +374,7 @@ func (r *ReleaseReconciler) reconcileKCMTemplates(ctx context.Context, releaseNa
 
 	opts := helm.ReconcileHelmReleaseOpts{
 		ChartRef: &helmcontrollerv2.CrossNamespaceSourceReference{
-			Kind:      helmChart.Kind,
+			Kind:      sourcev1.HelmChartKind,
 			Name:      helmChart.Name,
 			Namespace: helmChart.Namespace,
 		},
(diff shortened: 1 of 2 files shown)


internal/controller/release_controller.go
```diff
@@ -374,7 +374,7 @@ func (r *ReleaseReconciler) reconcileKCMTemplates(ctx context.Context, releaseNa
 
 	opts := helm.ReconcileHelmReleaseOpts{
 		ChartRef: &helmcontrollerv2.CrossNamespaceSourceReference{
-			Kind:      helmChart.Kind,
+			Kind:      sourcev1.HelmChartKind,
```
zerospiel, line 377: out of curiosity: can there be another `kind` set in the reference?

eromanova: In general, the `Kind` field of `helmcontrollerv2.CrossNamespaceSourceReference` may be one of HelmChart, OCIRepository, or ExternalArtifact (see https://fluxcd.io/flux/components/helm/helmreleases/#chart-reference).
In the release controller, when reconciling default provider templates, we always create and reference HelmChart objects exclusively, other reference types are not supported in this path.
This change was made because, in test environments, `helmChart.Kind` is not reliably populated. Since the reference kind is known and invariant in this context, there is no benefit in deriving it from the object. Using a constant is simpler.

zerospiel: I got the reason behind the change, but I was wondering whether it would have been possible to have something else. Okay, thanks for the clarification

Issue #2470, opened 2026-03-02
[bug] continuous events producing cause mem leak


BROngineer:
**Describe the bug**
When reconciling service sets we produce events related to underlying profiles disregard of actual state transition.
We need to guard the event production to only fire when the actual state changes.

PR #2471, opened 2026-03-02
fix: reduce number of emitted events for ServiceSet

BROngineer:
This PR reduces the number of events emitted during the ServiceSet objects reconciliation. At the moment we emit events (Normal and Warnings) on every reconciliation disregard of whether the actual state was changed.
This PR fixes this and adds the conditional events emitting: events will be emitted only when the event-related state was changed.

Fixes: #2470

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -190,7 +190,7 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		// we'll emit StateManagementProviderNotReadyEvent
 		// only in case the previous observed state was "ready".
 		if clone.Status.Provider.Ready {
-			record.Eventf(serviceSet, smp, kcmv1.StateManagementProviderNotReadyEvent, "Reconcile",
+			record.Eventf(serviceSet, smp, kcmv1.StateManagementProviderNotReadyEvent, kcmv1.ServiceSetReconcileEventAction,
 				"StateManagementProvider %s not ready, skipping ServiceSet %s reconciliation", smp.Name, serviceSet.Name)
 		}
 		l.Info("StateManagementProvider is not ready, skipping", "provider", serviceSet.Spec.Provider)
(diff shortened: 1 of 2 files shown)


internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -187,29 +187,70 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			record.Warnf(serviceSet, nil, kcmv1.ServiceSetEnsureProfileFailedEvent, "EnsureProfile",
+				"Failed to ensure Profile for ServiceSet %s: %v", serviceSet.Name, err)
+		}
+
+		// we'll emit failure-specific events in case failure reason was changed
+		if !conditionReasonChanged(conditionOldState, conditionNewState) {
+			return ctrl.Result{}, err
+		}
+
+		switch conditionNewState.Reason {
```
eromanova, line 226: `ensureProfile` function already has this logic:
1. Find the new status condition
2. It evaluates the error and constructs the appropriate reason and message

Given that, I think we could emit the events directly within `ensureProfile`, but only when the condition’s reason actually changes, instead of duplicating the reason checks in the switch/case. For example, we could pass the previous status condition (or just the previous reason) into `ensureProfile` and compare it with the newly computed one before emitting the event. Please let me know if I’m missing something and there’s a specific reason we need this.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -187,29 +187,70 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	}
 
 	// first we'll ensure the profile exists and up-to-date
 	if err = r.ensureProfile(ctx, rgnClient, serviceSet); err != nil {
-		record.Warnf(serviceSet, nil, kcmv1.ServiceSetEnsureProfileFailedEvent, "EnsureProfile",
-			"Failed to ensure Profile for ServiceSet %s: %v", serviceSet.Name, err)
+		conditionOldState := apimeta.FindStatusCondition(clone.Status.Conditions, kcmv1.ServiceSetProfileCondition)
+		conditionNewState := apimeta.FindStatusCondition(serviceSet.Status.Conditions, kcmv1.ServiceSetProfileCondition)
+		// we'll emit ServiceSetEnsureProfileFailedEvent warning
+		// only in case the previous observed state was ok.
```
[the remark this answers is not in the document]
BROngineer: At the lines 163-182 there is a defer function which will update the status of the object being reconciled even if there is an error during the reconciliation. So, we update the in-memoory state while reconciling an object -> if error occurs we exit the reconciliation -> if there are any changes in the object state we update the status -> thus next reconciliation won't emit the same warning.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: @eromanova I'd avoid passing the old state to ensureProfile function for the ONLY reason to emit the proper event. Since we already emitting warning in case of error in ensureProfile function, it makes sense to me to move all events logic there. We also already have all we need for this - different condition reasons for different failures, different error types for the same purpose, so we can handle all warnings emitting in one place.

eromanova: Ok, good. My main intention was just to keep this logic centralized, but I’m fine with it being handled outside of ensureProfile if that works better.

PR #2476, opened 2026-03-04
revert: k0s bump and all the related changes

zerospiel:
This reverts commit https://github.com/k0rdent/kcm/commit/46fab8fa1cdbd9880c405ab4cebd50273bad54d4.

This reverts commit https://github.com/k0rdent/kcm/commit/d10d38a76e505ac4c92a7e5806f56be8b0ad45bc.

test(e2e): fix openstack env var

chore(providers): adds capi v1beta2 anno where applicable without actual version bump

Bumps all provider and cluster templates due to incorrect helm push logic
and already existing artifacts.

### DIFF
--- a/templates/cluster/aws-eks/Chart.yaml
+++ b/templates/cluster/aws-eks/Chart.yaml
@@ -6,7 +6,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.4
+version: 1.0.6
 annotations:
   cluster.x-k8s.io/provider: infrastructure-aws
   cluster.x-k8s.io/infrastructure-aws: v1beta2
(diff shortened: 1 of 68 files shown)


templates/cluster/aws-eks/Chart.yaml
```diff
@@ -6,7 +6,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.4
+version: 1.0.6
```
eromanova, line 9:
```suggestion
version: 1.0.5
```
and `make generate-all`

eromanova (approved):

templates/cluster/aws-eks/Chart.yaml
zerospiel: It was on purpose due to bump-and-revert to ensure we won't miss the new (old) chart because of the intermediate being in the packages

Issue #2473, opened 2026-03-03
ServiceSet generation keeps raising indefinitely


Danil-Grigorev:
**Describe the bug**

K0rdent managed `ServiceSet` objects and the created `Profile` sveltos objects are getting constant generation bumps, even though the source `ClusterDeployment` service list or values does not change.

From what I observed, `ServiceSet` is created from `ClusterDeployment`, and `Profile` is then created/updated during `ServiceSet` reconciliation. The problem is that on repeated reconcile cycles, both resources get services added in a different order each time:

- `ServiceSet.spec.services` gets reordered multiple times
- as a result `Profile.spec.helmCharts` gets reordered

```yaml
apiVersion: k0rdent.mirantis.com/v1beta1
kind: ServiceSet
metadata:
  creationTimestamp: "2026-03-03T11:20:19Z"
  finalizers:
  - k0rdent.mirantis.com/service-set
  generation: 1772
  labels:
    ksm.k0rdent.mirantis.com/adapter: kcm-controller-manager
  name: self-adopted-cluster-mke
  namespace: k0rdent
  ownerReferences:
  - apiVersion: k0rdent.mirantis.com/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: ClusterDeployment
    name: self-adopted-cluster-mke
    uid: a644bf09-8af3-4be3-9844-a6c45d999724
  resourceVersion: "214412"
  uid: 4130ef2f-5a50-4ba4-bc95-02e1b7ddfddd
```

```yaml
apiVersion: config.projectsveltos.io/v1beta1
kind: Profile
metadata:
  creationTimestamp: "2026-03-03T11:20:20Z"
  finalizers:
  - profilefinalizer.projectsveltos.io
  generation: 1733
  labels:
    k0rdent.mirantis.com/managed: "true"
    projectsveltos.io/cluster-name: self-adopted-cluster-mke
    projectsveltos.io/cluster-type: Sveltos
    projectsveltos.io/profile-name: self-adopted-cluster-mke
  name: self-adopted-cluster-mke
  namespace: k0rdent
  ownerReferences:
  - apiVersion: k0rdent.mirantis.com/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: ServiceSet
    name: self-adopted-cluster-mke
    uid: 4130ef2f-5a50-4ba4-bc95-02e1b7ddfddd
  resourceVersion: "214410"
[...]

PR #2478, opened 2026-03-04
fix: continuous serviceset updates

BROngineer:
This PR mitigates a bug leading to continuous ServiceSet object updates even if there are no changes in ClusterDeployment/MultiClusterService have been done.

Two issues were discovered:
1. dependencies resolving returned the service list in non-deterministic order. Now it will return sorted list
2. while computing whether target ServiceSet should be updated, the ServiceSet-specific service representation could be defined with empty namespace, forwarded from ClusterDepployment/MultiClusterService-specific service representation. In the meantime, we always pass non-empty namespace to ServiceSet service representation. This made controller to always update ServiceSet object. Now the comparison is done with effective namespace, hence no discrepancy will be there.

Fixes: #2473

### DIFF
--- a/internal/serviceset/util.go
+++ b/internal/serviceset/util.go
@@ -384,9 +384,10 @@ func appendIfNotPresent(
 	s kcmv1.Service,
 	minimumUpgrade kcmv1.AvailableUpgrade,
 ) []kcmv1.ServiceWithValues {
+	serviceNamespace := effectiveNamespace(s.Namespace)
 	exists := slices.ContainsFunc(services, func(c kcmv1.ServiceWithValues) bool {
 		return c.Name == s.Name &&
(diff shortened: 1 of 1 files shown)


internal/serviceset/util.go
```diff
@@ -372,7 +386,7 @@ func appendIfNotPresent(
 ) []kcmv1.ServiceWithValues {
 	exists := slices.ContainsFunc(services, func(c kcmv1.ServiceWithValues) bool {
 		return c.Name == s.Name &&
-			c.Namespace == s.Namespace &&
+			c.Namespace == effectiveNamespace(s.Namespace) &&
```
zerospiel: nit: we can calculate the effectiveNamespace just once instead of calling it each iteration of the loop inside the `ContainsFunc`

Issue #2358, opened 2026-01-21
[bug] ServiceSet not reconciling when recovering service from Failed state


dmitriy-m2:
**Describe the bug**
When a service enters `Failed` state due to invalid values, and then the ClusterDeployment is patched to fix the template/values, the ServiceSet does not automatically reconcile. Manual deletion of the ServiceSet is required to trigger recovery.

**To Reproduce**

1. Deploy a ClusterDeployment with services that have dependencies (e.g., cert-manager and ingress-nginx, where ingress-nginx depends on cert-manager):
```shell
❯ kubectl get clusterdeployments openstack-upgrade-test -n kcm-system -o jsonpath='{.status.services}' | jq .
[
  {
    "lastStateTransitionTime": "2026-01-21T20:08:49Z",
    "name": "cert-manager",
    "namespace": "cert-manager",
    "state": "Deployed",
    "template": "cert-manager-1-17-1",
    "type": "Helm",
    "version": "v1.17.1"
  },
  {
    "lastStateTransitionTime": "2026-01-21T20:09:29Z",
    "name": "ingress-nginx",
    "namespace": "ingress-nginx",
    "state": "Deployed",
    "template": "ingress-nginx-4-12-0",
    "type": "Helm",
    "version": "4.12.0"
  }
]
```

2. Simulate a failure by patching ClusterDeployment with invalid values:
```shell
❯ kubectl patch clusterdeployments.k0rdent.mirantis.com openstack-upgrade-test -n kcm-system \
    --type='json' -p='[{"op": "replace", "path": "/spec/serviceSpec/services/0/template", "value": "cert-manager-1-17-2-broken"}, {"op": "replace", "path": "/spec/serviceSpec/services/0/values", "value": "replicaCount: \"invalid-not-a-number\""}]'
```

3. Verify services enter `Failed` state:
```shell
❯ kubectl get clusterdeployments openstack-upgrade-test -n kcm-system -o jsonpath='{.status.services}' | jq .
[
  {
    "failureMessage": "values don't meet the specifications of the schema(s) in the following chart(s):\ncert-manager:\n- at '/replicaCount': got string, want number\n",
    "lastStateTransitionTime": "2026-01-21T21:50:46Z",
    "name": "cert-manager",
    "namespace": "cert-manager",
    "state": "Failed",
    "template": "cert-manager-1-17-2-broken",
    "type": "Helm",
    "version": "v1.17.2"
  },
  {
[...]

PR #2482, opened 2026-03-05
chore: refactor serviceSet creation-or-update

BROngineer:
This PR contains refactoring of how ServiceSet objects are got created or updated. Previously the flow was the following:
1. Get serviceSet object and the required operation - Create/Update/None - basing on whether serviceSet exist and whether there is difference between desired state, coming from CLD/MCS, and state defined in ServiceSet spec.
2. Resolve versions and dependencies
3. Compute resulting services to be deployed
4. Create or update ServiceSet with computed service list

This brought an issue, that we decide whether to update service set **before** we know whether there is any difference between desired and actual states.

To solve this issue:
- versions and dependencies resolving is moved to GetServiceSetWithOperation from CLD and MCS controllers;
- The decision whether to update serviceSet is now being made **after** versions and dependencies are resolved;

The second issue fixed in this PR is that during computing whether service requires update the update came from the serviceSet.spec instead of service definition from CLD/MCS. This caused that in case of failure (due to invalid values passed to helm chart for instance) these values were never updated in ServiceSet.

Fixes: #2358

### DIFF
--- a/internal/serviceset/util.go
+++ b/internal/serviceset/util.go
@@ -515,36 +515,38 @@ func servicesToBeUpdated(
 
 	// we'll check whether deployed services could be upgraded to the desired version
 	for _, svc := range serviceSet.Spec.Services {
-		key := client.ObjectKey{Namespace: effectiveNamespace(svc.Namespace), Name: svc.Name}
+		effectiveServiceNs := effectiveNamespace(svc.Namespace)
+		key := client.ObjectKey{Namespace: effectiveServiceNs, Name: svc.Name}
 		desiredVersion := desiredServiceVersions[key]
 		// check upgrade availability
 		upgradeAvailable[key] = svc.Version != nil && desiredVersion < *svc.Version ||
 			desiredVersionInUpgradePaths(upgradePaths, svc, desiredVersion)
 		for _, serviceState := range serviceSet.Status.Services {
 			if serviceState.State == kcmv1.ServiceStateDeployed &&
-				serviceState.Namespace == svc.Namespace && serviceState.Name == svc.Name && serviceState.Version != nil {
+				serviceState.Namespace == effectiveServiceNs && serviceState.Name == svc.Name && serviceState.Version != nil {
 				deployedServiceVersions[key] = *serviceState.Version
 			}
 		}
 
-		if svc.Version != nil && *svc.Version != deployedServiceVersions[key] {
-			// Merge mutable fields from the desired service spec (e.g. updated values after a
-			// failed deploy) while preserving the in-flight version for upgrade-path tracking.
-			svcNamespace := effectiveNamespace(svc.Namespace)
-			for i := 0; i < len(desiredServices); {
-				ds := desiredServices[i]
-				if ds.Name == svc.Name && effectiveNamespace(ds.Namespace) == svcNamespace {
-					svc.Values = ds.Values
-					svc.ValuesFrom = ds.ValuesFrom
-					svc.HelmOptions = ds.HelmOptions
-					svc.HelmAction = ds.HelmAction
-					desiredServices = slices.Delete(desiredServices, i, i+1)
-				} else {
-					i++
-				}
-			}
-			services = append(services, svc)
+		if svc.Version == nil || *svc.Version == deployedServiceVersions[key] {
+			continue
 		}
+
+		// Merge mutable fields from the desired service spec (e.g. updated values after a
+		// failed deploy) while preserving the in-flight version for upgrade-path tracking.
+		for i := 0; i < len(desiredServices); {
+			ds := desiredServices[i]
+			if ds.Name == svc.Name && effectiveNamespace(ds.Namespace) == effectiveServiceNs {
+				svc.Values = ds.Values
+				svc.ValuesFrom = ds.ValuesFrom
+				svc.HelmOptions = ds.HelmOptions
+				svc.HelmAction = ds.HelmAction
+				desiredServices = slices.Delete(desiredServices, i, i+1)
+			} else {
+				i++
+			}
+		}
+		services = append(services, svc)
 	}
 	return services
 }


internal/serviceset/util_test.go
```diff
@@ -191,6 +191,131 @@ func Test_ServicesToDeploy(t *testing.T) {
...
+					Services: []kcmv1.ServiceWithValues{
+						{
+							Name:      "service1",
+							Namespace: metav1.NamespaceDefault,
+							Template:  "template1-1-0-0",
+							Version:   new("1.1.0.0"),
+							// Wrong values that caused the failure.
+							Values: "replicaCount: two\n",
+						},
+					},
```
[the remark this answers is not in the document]
zerospiel: Starting `go` `1.26.0`, this is not true. Please learn the new features.

internal/serviceset/util.go
```diff
@@ -525,6 +503,52 @@ func ServicesToDeploy(
...
+		key := client.ObjectKey{Namespace: effectiveNamespace(svc.Namespace), Name: svc.Name}
+		desiredVersion := desiredServiceVersions[key]
+		// check upgrade availability
+		upgradeAvailable[key] = svc.Version != nil && desiredVersion < *svc.Version ||
+			desiredVersionInUpgradePaths(upgradePaths, svc, desiredVersion)
+		for _, serviceState := range serviceSet.Status.Services {
+			if serviceState.State == kcmv1.ServiceStateDeployed &&
+				serviceState.Namespace == svc.Namespace && serviceState.Name == svc.Name && serviceState.Version != nil {
+				deployedServiceVersions[key] = *serviceState.Version
+			}
```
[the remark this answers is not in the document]
BROngineer: In serviceSet service namespace both in spec and status is mandatory field and won't be empty. Hence it's not necessary to normalize it to effective namespace.

internal/serviceset/util.go
```diff
@@ -427,34 +426,13 @@ func ServicesToDeploy(
...
-			}
-		}
-	}
+	services := serivesToBeUpdated(
+		serviceSet,
+		desiredServices,
+		deployedServiceVersions,
+		desiredServiceVersions,
+		upgradeAvailable,
+		upgradePaths)
```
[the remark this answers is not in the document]
BROngineer: This is intended. `slices.Delete` for `desiredServices` is needed for the `for` loop and was not expected to mutate initial slice.

internal/serviceset/util.go
```diff
@@ -525,6 +503,52 @@ func ServicesToDeploy(
...
+	deployedServiceVersions map[client.ObjectKey]string,
+	desiredServiceVersions map[client.ObjectKey]string,
+	upgradeAvailable map[client.ObjectKey]bool,
+	upgradePaths []kcmv1.ServiceUpgradePaths,
+) []kcmv1.ServiceWithValues {
+	services := make([]kcmv1.ServiceWithValues, 0)
+
+	// we'll check whether deployed services could be upgraded to the desired version
+	for _, svc := range serviceSet.Spec.Services {
+		key := client.ObjectKey{Namespace: effectiveNamespace(svc.Namespace), Name: svc.Name}
```
zerospiel: nit: it is a bit inconsistent, here we get the effective namespace, and several lines below it's already absolutely irrelevant, thus we take the raw namespace. for the code's reader it might be very hard to understand the reasoning behind, as well as it was for the copilot

internal/serviceset/util.go
```diff
@@ -525,6 +503,52 @@ func ServicesToDeploy(
...
+		upgradeAvailable[key] = svc.Version != nil && desiredVersion < *svc.Version ||
+			desiredVersionInUpgradePaths(upgradePaths, svc, desiredVersion)
+		for _, serviceState := range serviceSet.Status.Services {
+			if serviceState.State == kcmv1.ServiceStateDeployed &&
+				serviceState.Namespace == svc.Namespace && serviceState.Name == svc.Name && serviceState.Version != nil {
+				deployedServiceVersions[key] = *serviceState.Version
+			}
+		}
+
+		if svc.Version != nil && *svc.Version != deployedServiceVersions[key] {
```
zerospiel: nit: could be done conversely to dec

```go
if svc.Version == nil || *svc.Version == deployedServiceVersions[key] {
  continue
}
// the rest of the logic with the merge
```

internal/serviceset/util.go
```diff
@@ -525,6 +503,52 @@ func ServicesToDeploy(
...
+			// Merge mutable fields from the desired service spec (e.g. updated values after a
+			// failed deploy) while preserving the in-flight version for upgrade-path tracking.
+			svcNamespace := effectiveNamespace(svc.Namespace)
```
zerospiel: can't we reuse the already calculated namespace from the `key` or create a dedicated variable for this purpose, avoiding stacking unnecessary variables on the heap?

internal/serviceset/util.go
```diff
@@ -553,136 +577,101 @@ func desiredVersionInUpgradePaths(
...
+	providerSpec, err := StateManagementProviderConfigFromServiceSpec(serviceSpec)
+	if err != nil {
+		return nil, kcmv1.ServiceSetOperationNone, fmt.Errorf("failed to convert ServiceSpec to provider config: %w", err)
+	}
+	provider := new(kcmv1.StateManagementProvider)
+	if err := c.Get(ctx, client.ObjectKey{Name: providerSpec.Name}, provider); err != nil {
+		return nil, kcmv1.ServiceSetOperationNone, fmt.Errorf("failed to get StateManagementProvider %s: %w", providerSpec.Name, err)
+	}
+
+	// Get ServiceSet, create if it does not exist
```
zerospiel, line 619: something is off with the comment, i cannot locate the creation logic, could you advise on it? or is it the caller's responsibility to create a `serviceset` depending on the returned operation value?

internal/serviceset/util.go
```diff
@@ -553,136 +577,101 @@ func desiredVersionInUpgradePaths(
...
 	ctx context.Context,
 	c client.Client,
 	operationReq OperationRequisites,
 ) (*kcmv1.ServiceSet, kcmv1.ServiceSetOperation, error) {
 	l := ctrl.LoggerFrom(ctx)
+
+	// Determine desired services and service spec from MCS or CD.
+	var desiredServices []kcmv1.Service
+	var serviceSpec kcmv1.ServiceSpec
+	if operationReq.MCS != nil {
```
eromanova, line 601: Just curious: if both `operationReq.MCS` and `operationReq.CD` are provided, the `serviceSpec` from MCS takes precedence. Is this expected? Should the desired services be merged instead, or am I missing something?

internal/serviceset/util.go
zerospiel: It's the caller's responsibility. Sorry for not checking it in prior

internal/serviceset/util.go
BROngineer: Yes, it is expected. By design when MCS got reconciled, the ServiceSet for each matching cluster got produced. Such ServiceSets will have:
1. name computed from the corresponding CD name AND MCS name
2. CD AND MCS references set in the spec
3. services came from MCS

When services are defined in CD and CD got reconciled separate ServiceSet will be produced. It will have:
1. name computed from the owning CD
2. only CD reference in the spec
3. services came from the CD

internal/serviceset/util.go
BROngineer: fixed

internal/serviceset/util.go
BROngineer: fixed

internal/serviceset/util.go
BROngineer: fixed

internal/serviceset/util.go
```diff
@@ -553,136 +579,101 @@ func desiredVersionInUpgradePaths(
...
-			Namespace:   svcNamespace,
-			Template:    s.Template,
-			Values:      s.Values,
-			ValuesFrom:  s.ValuesFrom,
-			HelmOptions: s.HelmOptions,
-			Version:     &s.Version,
-			HelmAction:  s.HelmAction,
-		}
+	if err := ResolveServiceVersions(ctx, c, templateNamespace, filteredServices); err != nil {
+		return nil, kcmv1.ServiceSetOperationNone, fmt.Errorf("failed to resolve versions for filtered services: %w", err)
```
wahabmk, line 648: I know this isn't introduced in this PR but I don't see the need to create a [new slice of ptrs](https://github.com/k0rdent/kcm/blob/a5d49276f4810ed7e088c5d7dc552e035acd19f6/internal/serviceset/util.go#L65-L84) in the `ResolveServiceVersions` function.  I think we can simply remove this function.

We can edit the [`fillServiceVersions`](https://github.com/k0rdent/kcm/blob/a5d49276f4810ed7e088c5d7dc552e035acd19f6/internal/serviceset/util.go#L86-L106) function slightly to access each element via its index like so:
```go
func fillServiceVersions(ctx context.Context, c client.Client, namespace string, services []kcmv1.Service) error {
	for i := range services {
		if services[i].Version == "" && services[i].Template != "" {
			template := kcmv1.ServiceTemplate{}
			if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: services[i].Template}, &template); err != nil {
				return fmt.Errorf("failed to fetch template %s: %w", services[i].Template, err)
			}

			version := template.Spec.Version
			if version == "" && template.Spec.Helm != nil && template.Spec.Helm.ChartSpec != nil {
				version = template.Spec.Helm.ChartSpec.Version
			}
			services[i].Version = version

			if services[i].Version == "" {
				services[i].Version = services[i].Template
			}
		}
	}
	return nil
}
```

and then we can simply pass the slice header to it like so:
```suggestion
	if err := fillServiceVersions(ctx, c, templateNamespace, filteredServices); err != nil {
		return nil, kcmv1.ServiceSetOperationNone, fmt.Errorf("failed to resolve versions for filtered services: %w", err)
```

Now that each element of the slice is accessed by its index, the versions will get updated in the original slice. Example: https://goplay.tools/snippet/foWRy8gIMTh

internal/serviceset/util.go
```diff
@@ -553,136 +579,101 @@ func desiredVersionInUpgradePaths(
...
-func areProvidersEqual(p1, p2 *kcmv1.StateManagementProviderConfig) (bool, error) {
-	if p1 == nil && p2 == nil {
-		return true, nil
-	}
-	if p1 == nil || p2 == nil {
-		return false, nil
+	serviceSetServices := serviceSet.Spec.Services
+	if err := ResolveServiceVersions(ctx, c, templateNamespace, serviceSetServices); err != nil {
+		return nil, kcmv1.ServiceSetOperationNone, fmt.Errorf("failed to resolve versions for service set services: %w", err)
 	}
```
wahabmk, line 654: I don't understand the need for creating the `serviceSetServices` reference as it is not used anywhere else in the function. Is there a reason I'm missing as to why we can't just do:

```suggestion
	if err := ResolveServiceVersions(ctx, c, templateNamespace, serviceSet.Spec.Services); err != nil {
		return nil, kcmv1.ServiceSetOperationNone, fmt.Errorf("failed to resolve versions for service set services: %w", err)
	}
```

wahabmk: Infact, we can do away with the `ResolveServiceVersions` function entirely and change it to:
```suggestion
	if err := fillServiceWithValueVersions(ctx, c, templateNamespace, serviceSet.Spec.Services); err != nil {
		return nil, kcmv1.ServiceSetOperationNone, fmt.Errorf("failed to resolve versions for service set services: %w", err)
	}
```
Similar to as explained in [this](https://github.com/k0rdent/kcm/pull/2482/changes#r2911521068) comment.

internal/serviceset/util.go
```diff
@@ -525,6 +503,54 @@ func ServicesToDeploy(
...
+) []kcmv1.ServiceWithValues {
+	services := make([]kcmv1.ServiceWithValues, 0)
+
+	// we'll check whether deployed services could be upgraded to the desired version
+	for _, svc := range serviceSet.Spec.Services {
+		effectiveServiceNs := effectiveNamespace(svc.Namespace)
+		key := client.ObjectKey{Namespace: effectiveServiceNs, Name: svc.Name}
+		desiredVersion := desiredServiceVersions[key]
+		// check upgrade availability
+		upgradeAvailable[key] = svc.Version != nil && desiredVersion < *svc.Version ||
```
wahabmk, line 522: These are semver versions so I think we should use semver's Compare or LessThan functions, like:
```go
desired, _ := semver.NewVersion(desiredVersion)
actual, _ := semver.NewVersion(*svc.Version)
. . . . = . . . . desired.LessThan(actual) . . . .
```

wahabmk: Btw I don't quite understand this logic. Why does `desiredVersion < svc.Version (current version)` mean that an upgrade is available? Should this be the opposite?

wahabmk: > These are semver versions so I think we should use semver's Compare or LessThan functions

I just realized that the version can also have the name of the template (as done [here](https://github.com/k0rdent/kcm/blob/a5d49276f4810ed7e088c5d7dc552e035acd19f6/internal/serviceset/util.go#L122) & [here](https://github.com/k0rdent/kcm/blob/a5d49276f4810ed7e088c5d7dc552e035acd19f6/internal/serviceset/util.go#L101)). So version can contain either template name or a semver string, maybe that's why semver comparison isn't used.

Still I think there is potential for issues having to compare semver version strings using raw string comparison.

I am not quite familiar with how the sequential upgrade logic was implemented, but do we need template name in the version field?

wahabmk: Also the template name can be any user-defined name (not necessarily in `ingress-nginx-4-11-3` format).

PR #2509, opened 2026-03-11
fix: cleanup sveltos util code for service set upgrades

### LINKED ISSUE #2506: Refactoring request for serviceset utils
Please consider whether the following comments are relevant and should be resolved:
https://github.com/k0rdent/kcm/pull/2482#discussion_r2911521068
https://github.com/k0rdent/kcm/pull/2482#discussion_r2911541403
https://github.com/k0rdent/kcm/pull/2482#discussion_r2911543623
https://github.com/k0rdent/kcm/pull/2482#discussion_r2911605773

Otherwise, elaborate in comments, why they are irrelevant.
Thanks.

kylewuolle:
This PR cleans up some code in the sveltos util and also fixes a small issue where conflicts can occur because of stale updates to a service set object in the functional tests.

Fixes #2506

### DIFF
--- a/internal/serviceset/util.go
+++ b/internal/serviceset/util.go
@@ -21,8 +21,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"regexp"
 	"slices"
 
+	"github.com/Masterminds/semver/v3"
 	addoncontrollerv1beta1 "github.com/projectsveltos/addon-controller/api/v1beta1"
 	libsveltosv1beta1 "github.com/projectsveltos/libsveltos/api/v1beta1"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -482,6 +484,36 @@ func ServicesToDeploy(
 	return services
 }
 
+func normalizeVersion(v string) string {
+	versionRegex := regexp.MustCompile(`\d+\.\d+\.\d+|\d+-\d+-\d+`)
+	m := versionRegex.FindStringSubmatch(v)
+	if len(m) != 4 {
+		return ""
+	}
+	return m[1] + "." + m[2] + "." + m[3]
+}
+
+func isUpgradeAvailable(current, desired string) bool {
+	cur := normalizeVersion(current)
+	des := normalizeVersion(desired)
+
+	if cur == "" || des == "" {
+		return false
+	}
+
+	curVer, err := semver.NewVersion(cur)
+	if err != nil {
+		return false
+	}
+
+	desVer, err := semver.NewVersion(des)
+	if err != nil {
+		return false
+	}
+
+	return curVer.GreaterThan(desVer)
+}
+
 func servicesToBeUpdated(
 	serviceSet *kcmv1.ServiceSet,
 	desiredServices []kcmv1.Service,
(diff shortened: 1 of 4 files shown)


BROngineer (commented):
@kylewuolle I think it also makes sense to add few unit-tests for `fillServiceVersions` and `GetServiceSetWithOperation`

wahabmk:
Do we also need to consider the following concerns highlighted in https://github.com/k0rdent/kcm/pull/2482#discussion_r2911605773 in this PR:
1. Comparison of semver strings using raw string comparison. Since these are semver strings, raw string comparison may be inaccurate. E.g `1.0.0-alpha < 1.0.0` as per semver, but when comparing them as raw strings `1.0.0-alpha > 1.0.0`.
2. Version containing both semver string as well as template name if version is empty. So string comparison done on version when it contains the template name may also be inaccurate as the template name can be anything.
3. Is `desiredVersion < *svc.Version` correct in following? As it seems counter-intuitive on the surface to have upgrade available if desired version is less that current version:
https://github.com/k0rdent/kcm/blob/de6df1602f17e1affbf577eb549785c4e5902513/internal/serviceset/util.go#L522

We can resolve 1 & 2 if we can find a way to not use template name as version when version is empty, then we can use the semver package's functions to compare versions. Maybe we can use the version from within the ServiceTemplate ([`.spec.version`](https://github.com/k0rdent/kcm/blob/de6df1602f17e1affbf577eb549785c4e5902513/api/v1beta1/servicetemplate_types.go#L59)) instead of the template name when `.serviceSpec.services[].version` is empty, Would that work? The ([`.spec.version`](https://github.com/k0rdent/kcm/blob/de6df1602f17e1affbf577eb549785c4e5902513/api/v1beta1/servicetemplate_types.go#L59)) can also be empty though.

BROngineer (approved):
LGTM overall

internal/serviceset/util.go
```diff
@@ -503,6 +484,36 @@ func ServicesToDeploy(
 	return services
 }
 
+func normalizeVersion(v string) string {
+	versionRegex := regexp.MustCompile(`\d+\.\d+\.\d+|\d+-\d+-\d+`)
```
zerospiel, line 488: I believe there is absolutely no need to compile exactly the same expression each time (each reconciliation several calls). This is very expensive

zerospiel: Besides, I believe for such a normalization, less expensive logic could be used, like slicing and converting strings to numbers

internal/serviceset/util.go
```diff
@@ -503,6 +484,36 @@ func ServicesToDeploy(
...
+	m := versionRegex.FindStringSubmatch(v)
```
wahabmk, line 489: I ran this regex match function and it is returning empty for all of the following:
```
ingress-nginx-4-11-3
4.11.3
ingress-nginx-4-11-3-alpha1
4.11.3-alpha1
```
Is that the intended behavior?

zerospiel: That means that the tests do not cover the basic cases.

The probable desired behavior (as I guess) is to extract each number to a dedicated (un)named group, and then compare against it. As of now, the substring is exactly a single string, thus an empty return.

In any case, I hope that could've been rewritten without REs.

Good catch!

Issue #1889, opened 2025-08-26
[bug] Hosted templates failed to be deployed with custom registry and custom CA in case if subpath present in the registry url


eromanova:
**Describe the bug**
CCM and CSI charts can not be pulled during os hosted CP install.
```
status:
  error: 'can''t locate chart `oci://172.19.121.177:5000/k0rdent-enterprise/charts/openstack-cloud-controller-manager-2.31.1`:
    failed to do request: Head "https://172.19.121.177:5000/v2/k0rdent-enterprise/charts/openstack-cloud-controller-manager/manifests/2.31.1":
    tls: failed to verify certificate: x509: certificate signed by unknown authority'
  updated: 2025-08-26 14:38:24.188902124 +0000 UTC m=+726.801795267
  valuesHash: 0d994320878dd0a5f0a63e7eed9e5b6de9bee33e70f3c7d62bfe1759baaff0f7
```

**To Reproduce**
For example, on openstack:

1. Deploy k0rdent v1.2.0 with a custom registry and a single custom CA
2. Create standalone os deployment
3. Repeat step 1 for created cluster
4. Deploy hosted openstack cp

**Expected behavior**
Everything deployed fine

**Additional context**
Custom repo contains a subpath. E.g. `oci://172.19.121.177:5000/k0rdent-enterprise`

The issue is not reproduced when using the registry without a subpath, e.g.: `oci://172.19.121.177:5000`.

Root cause: 
The registry URL is incorrectly configured when the registry contains a subpath. As said in the k0s project: `# OCI registry URL must not include any path elements`, see:

https://github.com/k0sproject/k0s/pull/5901/files#diff-5a07081f272ac5d5dcc8fbd240e2eb0dbea08f2717c9031ec779ab7d7cc3ba0dR21

PR #2521, opened 2026-03-13
feat: support for registry with authentication when creating child clusters.

a13x5:

Adds all needed configuration to support registry with authentication when creating child clusters.

Major changes:
- Switch from k0s extension configuration to direct creation of the `Chart` object (now recommended way of using chart installation)
- Additional secret creation to reference credentials from inside cluster templates
- capo-orc installed to the same namespace as all capo components (kcm-system) to avoid additional logic for pull secret propagation to `orc-system` namespace
- https://github.com/k0rdent/kcm/issues/1889 (PR https://github.com/k0rdent/kcm/pull/1890) fix shouldn't be needed, but it should be re-tested

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,8 +141,10 @@ update-dev-confs: yq
 capo-orc-fetch: CAPO_DIR := $(PROVIDER_TEMPLATES_DIR)/cluster-api-provider-openstack
 capo-orc-fetch: CAPO_ORC_VERSION := 2.1.0
 capo-orc-fetch: CAPO_ORC_TEMPLATE := "$(CAPO_DIR)/templates/orc.yaml"
-capo-orc-fetch:
+capo-orc-fetch: yq
 	@curl -L --fail -s https://github.com/k-orc/openstack-resource-controller/releases/download/v$(CAPO_ORC_VERSION)/install.yaml | \
+	$(YQ) 'del(select(.kind == "Namespace"))' | \
+	$(YQ) '(select(.metadata.namespace) | .metadata.namespace) = "{{ .Release.Namespace }}"' | \
 	awk 'NR==1{print "{{- $$global := .Values.global | default dict }}"} \
 	{ \
 	  if ($$0 ~ /controller-gen.kubebuilder.io\/version/) { \
@@ -158,24 +160,24 @@ capo-orc-fetch:
 	    for(i=2;i<=length(arr);i++) { \
 	      image_path = image_path ((i>2)?"/":"") arr[i]; \
 	    } \
-	    print "        image: {{ default \"" registry "\" $$global.registry }}/" image_path; \
+	    print "          image: {{ default \"" registry "\" $$global.registry }}/" image_path; \
 	    next; \
 	  } \
 \
 	  print; \
 \
 	  if ($$0 ~ /^[ \t]*name: manager$$/) { \
-	    print "        {{- $$proxyEnv := include \"infrastructureProvider.proxyEnv\" . | fromYaml }}"; \
-	    print "        {{- if $$proxyEnv }}"; \
-	    print "        env:"; \
-	    print "        {{ toYaml $$proxyEnv.env | nindent 8 }}"; \
-	    print "        {{- end }}"; \
+	    print "          {{- $$proxyEnv := include \"infrastructureProvider.proxyEnv\" . | fromYaml }}"; \
+	    print "          {{- if $$proxyEnv }}"; \
+	    print "          env:"; \
+	    print "          {{ toYaml $$proxyEnv.env | nindent 8 }}"; \
+	    print "          {{- end }}"; \
 	  } \
 \
 	  if ($$0 ~ /serviceAccountName: orc-controller-manager/) { \
-	    print "      {{- if $$global.imagePullSecrets }}"; \
-	    print "      imagePullSecrets: {{ toYaml $$global.imagePullSecrets | nindent 8 }}"; \
-	    print "      {{- end }}"; \
+	    print "        {{- if $$global.imagePullSecrets }}"; \
+	    print "        imagePullSecrets: {{ toYaml $$global.imagePullSecrets | nindent 8 }}"; \
+	    print "        {{- end }}"; \
 	  } \
 	}' > $(CAPO_ORC_TEMPLATE)
 
--- a/internal/controller/management_controller.go
+++ b/internal/controller/management_controller.go
@@ -16,16 +16,20 @@ package controller
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"os"
 	"sort"
+	"strings"
 	"time"
 
+	"github.com/BurntSushi/toml"
 	helmcontrollerv2 "github.com/fluxcd/helm-controller/api/v2"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	addoncontrollerv1beta1 "github.com/projectsveltos/addon-controller/api/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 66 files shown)


internal/controller/management_controller.go
```diff
@@ -73,6 +78,8 @@ type ManagementReconciler struct {
 	sveltosDependentControllersStarted bool
 }
 
+const CldRegSecretName string = "cld-registry-credentials"
```
eromanova: Can we make this variable unexported? Looks like it's in use in this package only.
```suggestion
const cldRegSecretName string = "cld-registry-credentials"
```

internal/controller/management_controller.go
eromanova: nit: It would be good to add `r.RegistryCredentialSecretName` here as a parameter to set the correct condition if the registry credential secret is not found as well.

internal/controller/management_controller.go
```diff
@@ -189,6 +196,13 @@ func (r *ManagementReconciler) update(ctx context.Context, management *kcmv1.Man
 	}
 	management.Status.ObservedGeneration = management.Generation
 
+	if r.RegistryCredentialsSecretName != "" {
+		if err := r.createCldRegistryCredSecret(ctx); err != nil {
```
eromanova, line 200: I guess this should be added to the region controller as well to ensure the secret exists on the regional cluster before deploying providers.

templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
```diff
@@ -76,86 +88,116 @@ spec:
...
+    spec:
+      {{- if not $global.registry }}
+      repository:
+        url: https://kubernetes-sigs.github.io/aws-ebs-csi-driver
+      {{- end }}
+      {{- if $global.registryCredentialSecret }}
+      repository:
+        configFrom:
+          secretRef:
+            name: {{ $global.registryCredentialSecret }}
```
eromanova, line 160: Just curious: if the namespace is not specified, it defaults to the Chart's namespace: https://docs.k0sproject.io/stable/helm-charts/?h=chart, which is `kube-system` in this case.

But if I understand correctly, the `registryCredentialSecret` will be created in the cluster namespace, which may be different. Does it work properly?

internal/controller/management_controller.go
eromanova: Please, ignore. I was mistaken, this is not needed. The secret will be generated once in the management controller and then copied by the cluster deployment controller to the correct mothership cluster (to the region in case of regional deployments).

internal/controller/management_controller.go
```diff
@@ -672,6 +687,91 @@ func (r *ManagementReconciler) getRelease(ctx context.Context, mgmt *kcmv1.Manag
...
+	containerdAuth, err := makeContainerdAuth(r.GlobalRegistry, string(user), string(pass))
+	if err != nil {
+		return err
+	}
+
+	helmCredsSecretTpl := `apiVersion: v1
+kind: Secret
+metadata:
+  name: %s
+  namespace: kube-system
```
eromanova, line 748: Do we need `namespace: kube-system` here? Eventually, the secret will be created in the system namespace, and this line may be confusing.

templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
```diff
@@ -51,6 +51,119 @@ spec:
...
+          releaseName: aws-cloud-controller-manager
+          namespace: kube-system
+          values: |
+            nodeSelector:
+              node-role.kubernetes.io/control-plane: "true"
+            image:
+              {{- if $global.registry }}
+              repository: {{ $global.registry }}/provider-aws/cloud-controller-manager
+              {{- end }}
+              tag: v1.30.3
```
zerospiel, line 102: Don't we want to update the CCM version (along with the CSI version, which is now at 2.56.0)? We do have 1.35 k8s version still using the CCM for the 1.30

The same question applies to all of the other CCM/CSI versions of the other templates, if applicable.

Maybe this should not be part of this particular PR, but it might be worth doing it in this particular cadence

Makefile
```diff
@@ -158,24 +160,24 @@ capo-orc-fetch:
 	    for(i=2;i<=length(arr);i++) { \
 	      image_path = image_path ((i>2)?"/":"") arr[i]; \
 	    } \
-	    print "        image: {{ default \"" registry "\" $$global.registry }}/" image_path; \
+	    print "          image: {{ default \"" registry "\" $$global.registry }}/" image_path; \
```
zerospiel, line 163: not sure why extra 2 spaces have been added here

Makefile
```diff
@@ -141,8 +141,10 @@ update-dev-confs: yq
 capo-orc-fetch: CAPO_DIR := $(PROVIDER_TEMPLATES_DIR)/cluster-api-provider-openstack
 capo-orc-fetch: CAPO_ORC_VERSION := 2.1.0
 capo-orc-fetch: CAPO_ORC_TEMPLATE := "$(CAPO_DIR)/templates/orc.yaml"
-capo-orc-fetch:
+capo-orc-fetch: yq
 	@curl -L --fail -s https://github.com/k-orc/openstack-resource-controller/releases/download/v$(CAPO_ORC_VERSION)/install.yaml | \
+	$(YQ) 'del(select(.kind == "Namespace"))' | \
+	$(YQ) '(select(.metadata.namespace) | .metadata.namespace) = "{{ .Release.Namespace }}"' | \
```
zerospiel, line 147: NOTE: according changes to the docs should be added (regarding the proxy and the image pull secrets)

internal/controller/management_controller.go
```diff
@@ -16,16 +16,20 @@ package controller
...
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"os"
 	"sort"
+	"strings"
 	"time"
 
+	"github.com/BurntSushi/toml"
```
zerospiel, line 27: `go mod tidy`

internal/controller/management_controller.go
```diff
@@ -672,6 +687,91 @@ func (r *ManagementReconciler) getRelease(ctx context.Context, mgmt *kcmv1.Manag
 	return release, r.Client.Get(ctx, client.ObjectKey{Name: mgmt.Spec.Release}, release)
 }
 
+func makeContainerdAuth(registry, user, pass string) ([]byte, error) {
+	registryHost := strings.Split(registry, "/")[0]
```
zerospiel, line 697: potential runtime error: better use `strings.Cut`

and here is also the same problem `internal/util/pullsecret/pullsecret.go:67`, that I haven't noticed earlier

internal/controller/management_controller.go
```diff
@@ -189,6 +196,13 @@ func (r *ManagementReconciler) update(ctx context.Context, management *kcmv1.Man
 	}
 	management.Status.ObservedGeneration = management.Generation
 
+	if r.RegistryCredentialsSecretName != "" {
+		if err := r.createCldRegistryCredSecret(ctx); err != nil {
+			r.warnf(management, "RegistryCredentialSecretCreationFailed", "Failed to create registry credential secret: %v", err)
+			return ctrl.Result{}, err
```
zerospiel, line 203: please log the error; otherwise, it might be cumbersome to debug potential issues

internal/controller/management_controller.go
```diff
@@ -672,6 +687,91 @@ func (r *ManagementReconciler) getRelease(ctx context.Context, mgmt *kcmv1.Manag
...
+		},
+	}
+
+	return toml.Marshal(cfg)
+}
+
+func (r *ManagementReconciler) createCldRegistryCredSecret(ctx context.Context) error {
+	regSecret := new(corev1.Secret)
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: r.RegistryCredentialsSecretName, Namespace: r.SystemNamespace}, regSecret); err != nil {
+		return err
```
zerospiel: please wrap all of the errors with meaningful messages and extra info, such as what the secret name was

```go
// errors.New("failed to get username field from credential secret provided")
fmt.Errorf("failed to get 'username' field from credentials Secret %s/%s provided", r.SystemNamespace, r.RegistryCredentialsSecretName)
```
, or what operation failed

```go
	containerdAuth, err := makeContainerdAuth(r.GlobalRegistry, string(user), string(pass))
	if err != nil {
		// return err
        return fmt.Errorf("creating containerd auth config from %s registry for user %s: %w", r.GlobalRegistry, user, err)
	}
```

This ensures that in the event of an issue, we can easily determine the source of the problem from the logs

templates/cluster/aws-hosted-cp/templates/k0smotroncontrolplane.yaml
a13x5: I used the same name for both secrets: the one in the kcm-system namespace created by controller and the one that will be created in kube-system on the child cluster. This result better convergence and separate handling of 2 names, which pretty much the same thing

internal/controller/management_controller.go
a13x5: This secret will go to the child cluster. See https://github.com/k0rdent/kcm/pull/2521#discussion_r2932641355
Namespace field is mandatory when using k0s manifest deployer, otherwise manifest won't be applied

templates/cluster/aws-standalone-cp/templates/k0scontrolplane.yaml
a13x5: Agree, but not in this PR - it has enough changes.
Created a separate task: https://github.com/k0rdent/kcm/issues/2523

Makefile
a13x5: previous yq processing changes list indentation. Without these additional spaces yaml will be incorrect

PR #2539, opened 2026-03-16
feat: add a flag to remove k0rdent CRDs on management removal

### LINKED ISSUE #2374: [feat] Optional flag to cleanup k0rdent-related CRDs on Management removal
Currently, we persist all CRDs after the Management object is removed to prevent accidental cascading deletion of important infrastructure. This approach leaves us with manual cleanup. Cleanup could be automated using a separate flag (disabled by default).

kylewuolle:
This PR adds a flag which defaults to false that will cleanup k0rdernt CRDs when the management object is deleted.

Fixes #2374


cmd/main.go
```diff
@@ -148,6 +150,7 @@ func main() {
 	flag.DurationVar(&defaultHelmTimeout, "default-helm-timeout", 0, "Specifies the timeout duration for Helm install or upgrade operations. If unset, Flux’s default value will be used")
 	flag.IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 10, "Specifies the maximum number of concurrent reconciles that will be run for each controller.")
 	flag.BoolVar(&fluxEnabled, "flux-enabled", true, "The flag that indicates whether Flux integration is enabled")
+	flag.BoolVar(&cleanupCRDs, "cleanup-crds", false, "The flag that indicates whether installed CRDs should be cleaned up after removal")
```
Kshatrix: You have to add this flag to the kcm helm chart because, essentially, there’s currently no way to use it. However, I would suggest making this flag part of the management CRD spec instead.

internal/controller/management_controller.go
```diff
@@ -749,3 +758,38 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
 func (*ManagementReconciler) warnf(mgmt *kcmv1.Management, reason, message string, args ...any) {
 	record.Warnf(mgmt, nil, reason, "Reconcile", message, args...)
 }
+
+func (r *ManagementReconciler) removeCRDs(ctx context.Context) (bool, error) {
```
Kshatrix: The current implementation leaves quite a number of CRDs initially installed by k0rdent in place (e.g., velero CRDs and CRDs of all installed capi providers). Should those be taken care of as well?

Kshatrix (changes requested):

api/v1beta1/management_types.go
```diff
@@ -42,6 +42,11 @@ type ManagementSpec struct {
 	// Release references the Release object.
 	Release string `json:"release"`
 
+	// + optional
+
+	// CleanupCRD is a flag that indicates whether installed CRDs should be removed when a management is deleted
+	CleanupCRD *bool `json:"cleanupCRD,omitempty"`
```
zerospiel: why is the field pointer-typed? the current logic does not treat it as a tri-state, but only as a binary. thus, if there is absolutely no need for the pointer, please make it just `bool`

nit: probably plural could be better in terms of readability (`cleanupCRDs`), wdyt?

internal/controller/management_controller.go
```diff
@@ -791,5 +802,65 @@ func (r *ManagementReconciler) removeCRDs(ctx context.Context) (bool, error) {
...
+	}
+
+	for _, crd := range rel.Chart.CRDObjects() {
+		crds[crd.Name] = struct{}{}
+	}
+
+	for _, f := range rel.Chart.Files {
+		if strings.HasPrefix(f.Name, "crds/") {
+			var obj apiextv1.CustomResourceDefinition
+			_ = yaml.Unmarshal(f.Data, &obj)
```
zerospiel: do i undestand it correctly, that we perform a whole CRD object deserialization (which is normally huge) just to extract its name and along the way skipping the error checking? what if it is not a CRD? it is not a big deal in this case, but one redundant `delete` API call will be made

i'd suggest something more lightweight, maybe extracting the raw data could help, if it can avoid object's deserialization

internal/controller/management_controller.go
```diff
@@ -791,5 +802,65 @@ func (r *ManagementReconciler) removeCRDs(ctx context.Context) (bool, error) {
...
+		LabelSelector: fmt.Sprintf("name=%s,owner=helm", kcmv1.CoreKCMName),
+	})
+	if err != nil {
+		return crds, fmt.Errorf("failed to get helm release: %w", err)
+	}
+
+	var latestSecretName string
+	for _, s := range secrets.Items {
+		if latestSecretName == "" || s.Name > latestSecretName {
+			latestSecretName = s.Name
```
zerospiel: this is not the latest secret name, `v9` > `v10` for instance, so this is an incorrect assumption

there is `"helm.sh/helm/v3/pkg/storage".Init(<driver>).Last(<release name>)` for this purpose

internal/controller/management_controller.go
```diff
@@ -791,5 +802,65 @@ func (r *ManagementReconciler) removeCRDs(ctx context.Context) (bool, error) {
...
+		}
+	}
+
+	rel, err := drv.Get(latestSecretName)
```
zerospiel: if there are no secret objects, we will fail there. Is that expected? however, with another comment regarding finding the latest release, this should not be the case

internal/controller/management_controller.go
```diff
@@ -749,3 +769,98 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+		l.Info("Deleting CRD installed by Helm release", "crd", crdName)
+		crd := apiextv1.CustomResourceDefinition{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: r.SystemNamespace,
+				Name:      crdName,
+			},
+		}
+		err := r.Client.Delete(ctx, &crd)
+		if err != nil && !apierrors.IsNotFound(err) {
+			l.Error(err, "Failed deleting CRD", "crd", crd)
```
zerospiel:
```suggestion
			l.Error(err, "deleting CRD", "crd name", crdName)
```

nit-picking: we don't want logging the whole CRD object, do we? if so, the `crd` key can be more exact like `crd name`, and that is applicable elsewhere

internal/controller/management_controller.go
```diff
@@ -749,3 +769,98 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+			l.Error(err, "Failed deleting CRD", "crd", crd.Name)
+			return true, err
+		}
+	}
+
+	for crdName := range subCrds {
+		l.Info("Deleting CRD installed by Helm release", "crd", crdName)
+		crd := apiextv1.CustomResourceDefinition{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: r.SystemNamespace,
```
zerospiel:
```suggestion
```

internal/controller/management_controller.go
```diff
@@ -749,3 +769,98 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
 func (*ManagementReconciler) warnf(mgmt *kcmv1.Management, reason, message string, args ...any) {
 	record.Warnf(mgmt, nil, reason, "Reconcile", message, args...)
 }
+
+func (r *ManagementReconciler) removeCRDs(ctx context.Context, subCrds map[string]struct{}) (bool, error) {
+	l := ctrl.LoggerFrom(ctx)
+	l.Info("Ensuring all CRDs owned by KCM are removed")
+
+	var crdList apiextv1.CustomResourceDefinitionList
+	if err := r.Client.List(ctx, &crdList); err != nil {
```
zerospiel: let me clarify this: we list the whole CRDs and deserialize all of them in order to find the KCM release's CRD objects?

1. isn't that kind of a duplication of what has already been done in the `extractCRDs` method?
2. otherwise, we could potentially use partial objects, avoiding deserialization of such resources in whole, couldn't we?

internal/controller/management_controller.go
```diff
@@ -749,3 +769,98 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+			continue
+		}
+
+		if ann["meta.helm.sh/release-namespace"] != r.SystemNamespace {
+			continue
+		}
+
+		l.Info("Deleting CRD installed by Helm release", "crd", crd.Name)
+
+		err := r.Client.Delete(ctx, &crd)
```
zerospiel: if we still have to list the CRDs (see the previous comment), I believe we can freely remove the succesfully deleted CRD name from the `subCrds` set, wdyt?

zerospiel (changes requested):

internal/controller/management_controller.go
```diff
@@ -749,3 +769,98 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+	if err != nil {
+		return crds, fmt.Errorf("failed to get helm release: %w", err)
+	}
+
+	if rel.Chart == nil {
+		return crds, nil
+	}
+
+	for _, crd := range rel.Chart.CRDObjects() {
+		crds[crd.Name] = struct{}{}
```
zerospiel: if I understand this correctly, the `crd.Name` in this case still has the `crds/` prefix, and that is not a valid k8s object name

internal/controller/management_controller.go
```diff
@@ -749,3 +769,98 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+func (r *ManagementReconciler) extractCRDs(ctx context.Context) (map[string]struct{}, error) {
+	crds := make(map[string]struct{})
+	k8sClient, err := kubernetes.NewForConfig(r.Config)
+	if err != nil {
+		return crds, fmt.Errorf("failed to build client: %w", err)
+	}
+
+	drv := driver.NewSecrets(k8sClient.CoreV1().Secrets(r.SystemNamespace))
+	secrets, err := k8sClient.CoreV1().Secrets(r.SystemNamespace).List(ctx, metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("name=%s,owner=helm", kcmv1.CoreKCMName),
```
zerospiel: this does not fully address https://github.com/k0rdent/kcm/pull/2539#discussion_r2944453525 since it covers only the `kcm`-related CRDs however not the CRDs installed by CAPI, CAPI-related providers, sveltos and etc.

internal/controller/management_controller.go
```diff
@@ -749,3 +769,98 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+
+// extractCRDs pulls CRDs from the Helm chart (including subcharts)
+func (r *ManagementReconciler) extractCRDs(ctx context.Context) (map[string]struct{}, error) {
+	crds := make(map[string]struct{})
+	k8sClient, err := kubernetes.NewForConfig(r.Config)
+	if err != nil {
+		return crds, fmt.Errorf("failed to build client: %w", err)
+	}
+
+	drv := driver.NewSecrets(k8sClient.CoreV1().Secrets(r.SystemNamespace))
```
zerospiel: i cannot understand how this addresses the https://github.com/k0rdent/kcm/pull/2539#discussion_r2944453525

Our charts (for instance, for CAPA) do not have any CRDs; they hold only CAPI-operator-related objects, and the operator, in turn, downloads the corresponding manifest (eg from https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/latest/infrastructure-components.yaml) and installs it into the cluster. Could you please elaborate on this?

internal/controller/management_controller.go
zerospiel: Revoled in favor of [this](https://github.com/k0rdent/kcm/pull/2539#discussion_r2952513182), because I duplicate it accidentally

cmd/main.go
kylewuolle: Changed it to a flag in the management spec instead.

internal/controller/management_controller.go
kylewuolle: Turns out the only reliable way to delete the CRDs installed is to delete them by api group. I've updated the code to do that.

api/v1beta1/management_types.go
kylewuolle: Changed it to a non-pointer and made it plural

internal/controller/management_controller.go
kylewuolle: Looks like the only way to remove them all is to delete by api group so I've changed the code to do that instead.

internal/controller/management_controller.go
```diff
@@ -749,3 +758,41 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+		"config.projectsveltos.io",
+		"infrastructure.cluster.x-k8s.io",
+		"ipam.cluster.x-k8s.io",
+		"k0rdent.mirantis.com",
+		"kubernetesconfiguration.azure.com",
+		"lib.projectsveltos.io",
+		"network.azure.com",
+		"openstack.k-orc.cloud",
+		"resources.azure.com",
+		"source.toolkit.fluxcd.io",
```
eromanova: We have a scenario where k0rdent is deployed with Flux disabled, since flux can be already installed on the cluster beforehand. The current logic removes all Flux CRDs (even unmanaged) during k0rdent deinstallation, which is likely not the desired behavior. The same issue applies to other components such as velero, capi operator, etc.
I think we need to check labels on CRDs instead (for example, all kcm-managed management and regional components have `meta.helm.sh/release-name: kcm` label) or think about a better option for identifying k0rdent-related CRDs for dependent components.

internal/controller/management_controller.go
```diff
@@ -749,3 +758,41 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+		"velero.io",
+	}
+
+	var crdList apiextv1.CustomResourceDefinitionList
```
eromanova: Potential improvement: This cleanup logic should likely be extended to Regions as well. When a Region is deleted, all k0rdent-related CRDs within that Region should also be cleaned up.

internal/controller/management_controller.go
kylewuolle: The problem is a lot of CRDs don't have the kcm label set. We could restrict deletion to just CRDs with that label but it'd leave many others in place. 


Here's the list of CRDs without the kcm label currently:

```
       awsclustercontrolleridentities.infrastructure.cluster.x-k8s.io
       awsclusterroleidentities.infrastructure.cluster.x-k8s.io
       awsclusters.infrastructure.cluster.x-k8s.io
       awsclusterstaticidentities.infrastructure.cluster.x-k8s.io
       awsclustertemplates.infrastructure.cluster.x-k8s.io
       awsfargateprofiles.infrastructure.cluster.x-k8s.io
       awsmachinepools.infrastructure.cluster.x-k8s.io
       awsmachines.infrastructure.cluster.x-k8s.io
       awsmachinetemplates.infrastructure.cluster.x-k8s.io
       awsmanagedclusters.infrastructure.cluster.x-k8s.io
       awsmanagedclustertemplates.infrastructure.cluster.x-k8s.io
       awsmanagedcontrolplanes.controlplane.cluster.x-k8s.io
       awsmanagedcontrolplanetemplates.controlplane.cluster.x-k8s.io
       awsmanagedmachinepools.infrastructure.cluster.x-k8s.io
       azureasomanagedclusters.infrastructure.cluster.x-k8s.io
       azureasomanagedclustertemplates.infrastructure.cluster.x-k8s.io
       azureasomanagedcontrolplanes.infrastructure.cluster.x-k8s.io
       azureasomanagedcontrolplanetemplates.infrastructure.cluster.x-k8s.io
       azureasomanagedmachinepools.infrastructure.cluster.x-k8s.io
       azureasomanagedmachinepooltemplates.infrastructure.cluster.x-k8s.io
       azureclusteridentities.infrastructure.cluster.x-k8s.io
       azureclusters.infrastructure.cluster.x-k8s.io
       azureclustertemplates.infrastructure.cluster.x-k8s.io
       azuremachinepoolmachines.infrastructure.cluster.x-k8s.io
       azuremachinepools.infrastructure.cluster.x-k8s.io
       azuremachines.infrastructure.cluster.x-k8s.io
       azuremachinetemplates.infrastructure.cluster.x-k8s.io
       azuremanagedclusters.infrastructure.cluster.x-k8s.io
       azuremanagedclustertemplates.infrastructure.cluster.x-k8s.io
       azuremanagedcontrolplanes.infrastructure.cluster.x-k8s.io
       azuremanagedcontrolplanetemplates.infrastructure.cluster.x-k8s.io
       azuremanagedmachinepools.infrastructure.cluster.x-k8s.io
       azuremanagedmachinepooltemplates.infrastructure.cluster.x-k8s.io
       backuprepositories.velero.io
       backups.velero.io
       backupstoragelocations.velero.io
       bastionhosts.network.azure.com
       clusterclasses.cluster.x-k8s.io
       clusterresourcesetbindings.addons.cluster.x-k8s.io
       clusterresourcesets.addons.cluster.x-k8s.io
       clusters.cluster.x-k8s.io
       clusters.k0smotron.io
       datadownloads.velero.io
       datauploads.velero.io
       deletebackuprequests.velero.io
       devclusters.infrastructure.cluster.x-k8s.io
       devclustertemplates.infrastructure.cluster.x-k8s.io
       devmachines.infrastructure.cluster.x-k8s.io
       devmachinetemplates.infrastructure.cluster.x-k8s.io
       dockerclusters.infrastructure.cluster.x-k8s.io
       dockerclustertemplates.infrastructure.cluster.x-k8s.io
       dockermachinepools.infrastructure.cluster.x-k8s.io
       dockermachinepooltemplates.infrastructure.cluster.x-k8s.io
       dockermachines.infrastructure.cluster.x-k8s.io
       dockermachinetemplates.infrastructure.cluster.x-k8s.io
       downloadrequests.velero.io
       eksconfigs.bootstrap.cluster.x-k8s.io
       eksconfigtemplates.bootstrap.cluster.x-k8s.io
       extensionconfigs.runtime.cluster.x-k8s.io
       extensions.kubernetesconfiguration.azure.com
       fleetsmembers.containerservice.azure.com
       gcpclusters.infrastructure.cluster.x-k8s.io
       gcpclustertemplates.infrastructure.cluster.x-k8s.io
       gcpmachines.infrastructure.cluster.x-k8s.io
       gcpmachinetemplates.infrastructure.cluster.x-k8s.io
       gcpmanagedclusters.infrastructure.cluster.x-k8s.io
       gcpmanagedclustertemplates.infrastructure.cluster.x-k8s.io
       gcpmanagedcontrolplanes.infrastructure.cluster.x-k8s.io
       gcpmanagedcontrolplanetemplates.infrastructure.cluster.x-k8s.io
       gcpmanagedmachinepools.infrastructure.cluster.x-k8s.io
       gcpmanagedmachinepooltemplates.infrastructure.cluster.x-k8s.io
       gkeconfigs.bootstrap.cluster.x-k8s.io
       gkeconfigtemplates.bootstrap.cluster.x-k8s.io
       globalinclusterippools.ipam.cluster.x-k8s.io
       inclusterippools.ipam.cluster.x-k8s.io
       infobloxinstances.ipam.cluster.x-k8s.io
       infobloxippools.ipam.cluster.x-k8s.io
       ipaddressclaims.ipam.cluster.x-k8s.io
       ipaddresses.ipam.cluster.x-k8s.io
       jointokenrequests.k0smotron.io
       k0scontrollerconfigs.bootstrap.cluster.x-k8s.io
       k0scontrolplanes.controlplane.cluster.x-k8s.io
       k0scontrolplanetemplates.controlplane.cluster.x-k8s.io
       k0smotroncontrolplanes.controlplane.cluster.x-k8s.io
       k0smotroncontrolplanetemplates.controlplane.cluster.x-k8s.io
       k0sworkerconfigs.bootstrap.cluster.x-k8s.io
       k0sworkerconfigtemplates.bootstrap.cluster.x-k8s.io
       kubevirtclusters.infrastructure.cluster.x-k8s.io
       kubevirtclustertemplates.infrastructure.cluster.x-k8s.io
       kubevirtmachines.infrastructure.cluster.x-k8s.io
       kubevirtmachinetemplates.infrastructure.cluster.x-k8s.io
       machinedeployments.cluster.x-k8s.io
       machinedrainrules.cluster.x-k8s.io
       machinehealthchecks.cluster.x-k8s.io
       machinepools.cluster.x-k8s.io
       machines.cluster.x-k8s.io
       machinesets.cluster.x-k8s.io
       managedclusters.containerservice.azure.com
       managedclustersagentpools.containerservice.azure.com
       natgateways.network.azure.com
       openstackclusteridentities.infrastructure.cluster.x-k8s.io
       openstackclusters.infrastructure.cluster.x-k8s.io
       openstackclustertemplates.infrastructure.cluster.x-k8s.io
       openstackfloatingippools.infrastructure.cluster.x-k8s.io
       openstackmachines.infrastructure.cluster.x-k8s.io
       openstackmachinetemplates.infrastructure.cluster.x-k8s.io
       openstackservers.infrastructure.cluster.x-k8s.io
       podvolumebackups.velero.io
       podvolumerestores.velero.io
       pooledremotemachines.infrastructure.cluster.x-k8s.io
       privateendpoints.network.azure.com
       remoteclusters.infrastructure.cluster.x-k8s.io
       remoteclustertemplates.infrastructure.cluster.x-k8s.io
       remotemachines.infrastructure.cluster.x-k8s.io
       remotemachinetemplates.infrastructure.cluster.x-k8s.io
       resourcegroups.resources.azure.com
       restores.velero.io
       rosaclusters.infrastructure.cluster.x-k8s.io
       rosacontrolplanes.controlplane.cluster.x-k8s.io
       rosamachinepools.infrastructure.cluster.x-k8s.io
       rosanetworks.infrastructure.cluster.x-k8s.io
       rosaroleconfigs.infrastructure.cluster.x-k8s.io
       schedules.velero.io
       serverstatusrequests.velero.io
       virtualnetworks.network.azure.com
       virtualnetworkssubnets.network.azure.com
       volumesnapshotlocations.velero.io
       vsphereclusteridentities.infrastructure.cluster.x-k8s.io
       vsphereclusters.infrastructure.cluster.x-k8s.io
       vsphereclustertemplates.infrastructure.cluster.x-k8s.io
       vspheredeploymentzones.infrastructure.cluster.x-k8s.io
       vspherefailuredomains.infrastructure.cluster.x-k8s.io
       vspheremachines.infrastructure.cluster.x-k8s.io
       vspheremachinetemplates.infrastructure.cluster.x-k8s.io
       vspherevms.infrastructure.cluster.x-k8s.io
```

Kshatrix: Hardcoding CRDs is definitely not an option, as some out-of-tree (OOT) providers might not be cleaned up. 

Generally, CRDs fall into two categories: those installed directly by k0rdent (identifiable by the _helm.toolkit.fluxcd.io/name: release name_ label) and those installed by the CAPI operator (identifiable by labels such as _cluster.x-k8s.io/provider_). It may be beneficial to separate these groups using different cleanup flags.

There will be some CRDs that fall into neither of these groups. These should be identified, and the installation process for them should be reworked (as a separate task).

Kshatrix: And as I see, there are only two components that don’t fit into these buckets: velero and orc

cmd/main.go
Kshatrix: please remove the redundant flag

internal/controller/management_controller.go
```diff
@@ -931,3 +951,19 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
 func (*ManagementReconciler) warnf(mgmt *kcmv1.Management, reason, message string, args ...any) {
 	record.Warnf(mgmt, nil, reason, "Reconcile", message, args...)
 }
+
+func (r *ManagementReconciler) removeCRDsWithSelector(ctx context.Context, selector labels.Selector) (bool, error) {
+	crdList := &metav1.PartialObjectMetadataList{}
+	crdList.SetGroupVersionKind(apiextv1.SchemeGroupVersion.WithKind("CustomResourceDefinitionList"))
```
Kshatrix: please use kubeutil.EnsureDeleteAllOf instead to ensure all CRDs got removed

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -380,3 +380,8 @@ rules:
   resources:
   - inclusterippools
   verbs: {{ include "rbac.editorVerbs" . | nindent 4 }}
+- apiGroups:
+    - apiextensions.k8s.io
+  resources:
+    - customresourcedefinitions
+  verbs: {{ include "rbac.editorVerbs" . | nindent 4 }}
```
Kshatrix: please remove excessive permissions

templates/provider/kcm/templates/rbac/user-facing/management-editor.yaml
```diff
@@ -18,3 +18,8 @@ rules:
     verbs: {{ include "rbac.viewerVerbs" . | nindent 6 }}
       - create
       - delete
+  - apiGroups:
```
Kshatrix: I don't think this should be part of the management-editor role, let's just remove it

api/v1beta1/management_types.go
```diff
@@ -44,6 +46,16 @@ type ManagementSpec struct {
 
 	// ComponentsCommonSpec defines the desired state of management components.
 	ComponentsCommonSpec `json:",inline"`
+
```
Kshatrix, line 54: let's group it under _cleanup:_
something like:
type ManagementCleanup struct {
	// CRDs indicates whether installed CRDs should be removed when a Management is deleted.
	CRDs bool `json:"crds,omitempty"`
	// CAPIProviderCRDs indicates whether CAPI provider CRDs should be removed on Management deletion.
	CAPIProviderCRDs bool `json:"capiProviderCRDs,omitempty"`
}

also let's add a note that all CRDs installed by the capi operator will be removed (even those unrelated to k0rdent)

internal/controller/management_controller.go
```diff
@@ -583,6 +585,24 @@ func (r *ManagementReconciler) delete(ctx context.Context, management *kcmv1.Man
 		return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, err
 	}
 
+	if management.Spec.CleanupCRDs {
```
Kshatrix: please also add tests for the functionality

Kshatrix (changes requested):

internal/controller/management_controller.go
```diff
@@ -931,3 +951,29 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+const managementCRDName = "managements.k0rdent.mirantis.com"
+
+func (r *ManagementReconciler) removeCRDsWithSelector(ctx context.Context, selector labels.Selector) (bool, error) {
+	l := ctrl.LoggerFrom(ctx)
+	l.Info("Ensuring CRDs are removed")
+	gvk := apiextv1.SchemeGroupVersion.WithKind("CustomResourceDefinition")
+	listOpts := &client.ListOptions{
+		LabelSelector: selector,
+	}
+	if err := kubeutil.EnsureDeleteAllOf(ctx, r.Client, gvk, listOpts, managementCRDName); err != nil {
```
Kshatrix: What’s the purpose of excluding it here? The CRD won’t be deleted until the management object’s finalizer is removed anyway.

internal/controller/management_controller.go
```diff
@@ -583,6 +585,24 @@ func (r *ManagementReconciler) delete(ctx context.Context, management *kcmv1.Man
 		return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, err
 	}
 
+	if management.Spec.Cleanup != nil && management.Spec.Cleanup.CRDs {
+		sel := labels.SelectorFromSet(map[string]string{kcmv1.FluxHelmChartNameKey: kcmv1.CoreKCMName})
```
Kshatrix: the release name should not be hardcoded, please check https://github.com/k0rdent/kcm/commit/9a98e0c17a98cce34fcd330429520bb067439f80

Kshatrix (dismissed):

internal/controller/management_controller.go
kylewuolle: The problem is that we get a circular dependency here. We're waiting for the CRDs to be confirmed deleted but the management one can't be deleted since the finalizer is not removed on the management object yet. We could flip it to remove the finalizer and then remove the CRDs but it could lead to them not being deleted and then no further reconciliation.

api/v1beta1/management_types.go
```diff
@@ -32,10 +32,17 @@ const (
 
 	K0rdentManagementClusterLabelKey   = "k0rdent.mirantis.com/management-cluster"
 	K0rdentManagementClusterLabelValue = "true"
+
+	CAPIProviderLabelKey = "cluster.x-k8s.io/provider"
```
zerospiel: nit: this is a mere `"sigs.k8s.io/cluster-api/api/core/v1beta2".ProviderNameLabel`

internal/controller/management_controller.go
```diff
@@ -601,6 +604,25 @@ func (r *ManagementReconciler) delete(ctx context.Context, management *kcmv1.Man
 		return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, nil
 	}
 
+	kcmReleaseName := config.KCMHelmReleaseName()
+	if management.Spec.Cleanup != nil && management.Spec.Cleanup.CRDs {
+		sel := labels.SelectorFromSet(map[string]string{kcmv1.FluxHelmChartNameKey: kcmReleaseName})
+		requeue, err = r.removeCRDsWithSelector(ctx, sel)
+		if err != nil || requeue {
```
zerospiel: Please avoid using the anti-pattern requeue in combination with a non-nil error. It is either an error -> `return {}, err`, or requeue -> `return {Requeue}, nil`

internal/controller/management_controller.go
```diff
@@ -601,6 +604,25 @@ func (r *ManagementReconciler) delete(ctx context.Context, management *kcmv1.Man
...
+		}
+	}
+
+	if management.Spec.Cleanup != nil && management.Spec.Cleanup.CAPIProviderCRDs {
+		req, err := labels.NewRequirement(kcmv1.CAPIProviderLabelKey, selection.Exists, nil)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		requeue, err = r.removeCRDsWithSelector(ctx, labels.NewSelector().Add(*req))
+		if err != nil || requeue {
```
zerospiel: Ditto about requeue+error

internal/controller/management_controller.go
```diff
@@ -974,3 +996,29 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+
+func (r *ManagementReconciler) removeCRDsWithSelector(ctx context.Context, selector labels.Selector) (bool, error) {
+	l := ctrl.LoggerFrom(ctx)
+	l.Info("Ensuring CRDs are removed")
+	gvk := apiextv1.SchemeGroupVersion.WithKind("CustomResourceDefinition")
+	listOpts := &client.ListOptions{
+		LabelSelector: selector,
+	}
+	if requeue, err := kubeutil.EnsureDeleteAllOf(ctx, r.Client, gvk, listOpts, managementCRDName); err != nil {
+		l.Error(err, "Not all CRDs are removed")
```
zerospiel: the logging is inconsistent: here the error is being logged; however, the one below is not. Generally speaking, it's a rule of thumb to wrap an error, and the most outer caller will log it eventually (in this case, the `delete` method). So, before returning from the reconciliation loop, log the error (it is easier to debug rather than trying to chase an error from the internals of the controller-runtime)

internal/controller/management_controller.go
```diff
@@ -974,3 +996,29 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+		return requeue, err
+	}
+
+	mgmtCRD := &apiextv1.CustomResourceDefinition{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: managementCRDName}, mgmtCRD); err == nil {
+		if selector.Matches(labels.Set(mgmtCRD.Labels)) && mgmtCRD.DeletionTimestamp.IsZero() {
+			if err := r.Client.Delete(ctx, mgmtCRD); client.IgnoreNotFound(err) != nil {
+				return true, err
```
zerospiel: ditto about requeue+error

internal/controller/management_controller.go
```diff
@@ -974,3 +996,29 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
 func (*ManagementReconciler) warnf(mgmt *kcmv1.Management, reason, message string, args ...any) {
 	record.Warnf(mgmt, nil, reason, "Reconcile", message, args...)
 }
+
+// managementCRDName is the API server object name of the Management CRD.
+const managementCRDName = "managements.k0rdent.mirantis.com"
+
+func (r *ManagementReconciler) removeCRDsWithSelector(ctx context.Context, selector labels.Selector) (bool, error) {
```
zerospiel:
```suggestion
func (r *ManagementReconciler) removeCRDsWithSelector(ctx context.Context, selector labels.Selector) error {
```

I suppose we do not need the requeue flag here: if an error occurred, we do not need the requeue; if no error has occurred, we also do not need to requeue the event

internal/controller/management_controller.go
```diff
@@ -974,3 +996,29 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+	l := ctrl.LoggerFrom(ctx)
+	l.Info("Ensuring CRDs are removed")
+	gvk := apiextv1.SchemeGroupVersion.WithKind("CustomResourceDefinition")
+	listOpts := &client.ListOptions{
+		LabelSelector: selector,
+	}
+	if requeue, err := kubeutil.EnsureDeleteAllOf(ctx, r.Client, gvk, listOpts, managementCRDName); err != nil {
```
zerospiel: the return-argument `requeue` is not respected here: `err != nil` means `no requeue`, but if `err == nil`, the `requeue` is not being checked, meaning some of the objects might still sit in the cluster.

We have another similar function `kubeutil.DeleteAllExceptAndWait` which waits for the given time and fails if any objects are still there, however, I am not sure whether it'd work with cluster-scoped objects (please check if you consider switching to it)

internal/controller/management_controller.go
```diff
@@ -974,3 +996,29 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+		l.Error(err, "Not all CRDs are removed")
+		return requeue, err
+	}
+
+	mgmtCRD := &apiextv1.CustomResourceDefinition{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: managementCRDName}, mgmtCRD); err == nil {
+		if selector.Matches(labels.Set(mgmtCRD.Labels)) && mgmtCRD.DeletionTimestamp.IsZero() {
```
zerospiel:
```suggestion
		if mgmtCRD.DeletionTimestamp.IsZero() && selector.Matches(labels.Set(mgmtCRD.Labels)) {
```

nit: short-circuit evaluation

But there is another problem: if we delete `managements` CRD object, still having the object of the CRD, we may never delete the object, because the finalizer still holds the object. I'd say we have to release the finalizer first, and then try to delete the CRD, without any hopes that it succeeds

zerospiel (changes requested):

internal/controller/management_controller.go
```diff
@@ -601,6 +604,25 @@ func (r *ManagementReconciler) delete(ctx context.Context, management *kcmv1.Man
 		return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, nil
 	}
 
+	kcmReleaseName := config.KCMHelmReleaseName()
```
zerospiel: Question to you and @eromanova: do we need to extend the feature onto the `regions` as well, or not? Currently, the scope is focused on the `managements` resource.

zerospiel: Looks like a dedicated issue should be created

Kshatrix: let's handle regions as a separate task

internal/controller/management_controller.go
zerospiel: Okay, seems like marking the CRD with the deletion timestamp is a viable solution: mark the CRD with the deletion TS, then delete the object

internal/controller/management_controller.go
eromanova: I agree, we can handle regions separately.

api/v1beta1/management_types.go
```diff
@@ -36,6 +36,11 @@ const (
 
 // ManagementSpec defines the desired state of Management
 type ManagementSpec struct {
+	// +optional
+
+	// Cleanup configures CRD removal behaviour when the Management object is deleted.
+	Cleanup *ManagementCleanup `json:"cleanup,omitempty"`
```
zerospiel:
```suggestion
	Cleanup ManagementCleanup `json:"cleanup,omitempty"`
```

nit: both `bool`-typed fields are `false` by default and the pointer on the struct here is unnecessary. Up to the implementer

api/v1beta1/management_types.go
```diff
@@ -46,6 +51,21 @@ type ManagementSpec struct {
 	ComponentsCommonSpec `json:",inline"`
 }
 
+// ManagementCleanup controls which CRDs are removed when a Management object is deleted.
+type ManagementCleanup struct {
+	// +optional
+
+	// CRDs indicates whether k0rdent-owned CRDs should be removed when a Management is deleted.
+	CRDs bool `json:"crds,omitempty"`
```
zerospiel:
```suggestion
	K0rdentCRDs bool `json:"k0rdentCRDs,omitempty"`
```

nit: just a suggestion (or `KCMCRDs` or something else) to avoid ambiguity 

```yaml
cleanup:
  crds: true
  capiProviderCRDs: true
```
vs
```yaml
cleanup:
  k0rdentCRDs: true
  capiProviderCRDs: true
```

internal/controller/management_controller.go
```diff
@@ -1016,3 +997,30 @@ func (*ManagementReconciler) eventf(mgmt *kcmv1.Management, reason, message stri
...
+// removeCRDsWithSelector issues deletion of all CRDs matching the given selector
+// without waiting for the deletion to complete: CRD termination is asynchronous
+// and requires no further action from this controller. In particular, the
+// Management CRD itself finishes terminating only after the Management object's
+// finalizer is removed and the object ceases to exist.
+func (r *ManagementReconciler) removeCRDsWithSelector(ctx context.Context, selector labels.Selector) error {
+	l := ctrl.LoggerFrom(ctx)
+	l.Info("Removing CRDs", "selector", selector.String())
+
+	itemsList := new(metav1.PartialObjectMetadataList)
```
zerospiel, line 980: nit: `PartialObjectMetadataList` spins a cluster-wide watcher (**see the RBAC-related comment below**), which we could avoid spinning using the direct reader `mgr.GetAPIReader()` which is uncached.

This also might be hard to implement: we need to introduce a new field for the direct reader, initialize it, and use it

Up to the implementer

internal/controller/management_controller.go
```diff
@@ -605,6 +605,25 @@ func (r *ManagementReconciler) delete(ctx context.Context, management *kcmv1.Man
...
+	if management.Spec.Cleanup != nil && management.Spec.Cleanup.CAPIProviderCRDs {
+		req, err := labels.NewRequirement(clusterapiv1.ProviderNameLabel, selection.Exists, nil)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.removeCRDsWithSelector(ctx, labels.NewSelector().Add(*req)); err != nil {
+			l.Error(err, "removing CAPI provider CRDs")
+			return ctrl.Result{}, err
+		}
+	}
```
zerospiel:
```suggestion
	if cleanup := management.Spec.Cleanup; cleanup != nil {
		if cleanup.CRDs {
			sel := labels.SelectorFromSet(map[string]string{kcmv1.FluxHelmChartNameKey: config.KCMHelmReleaseName()})
			if err := r.removeCRDsWithSelector(ctx, sel); err != nil {
				l.Error(err, "removing k0rdent CRDs")
				return ctrl.Result{}, err
			}
		}

		if cleanup.CAPIProviderCRDs {
			req, err := labels.NewRequirement(clusterapiv1.ProviderNameLabel, selection.Exists, nil)
			if err != nil {
				return ctrl.Result{}, err
			}

			if err := r.removeCRDsWithSelector(ctx, labels.NewSelector().Add(*req)); err != nil {
				l.Error(err, "removing CAPI provider CRDs")
				return ctrl.Result{}, err
			}
		}
	}
```

nit: we could combine into a dedicated block

This also avoids calling `config.KCMHelmReleaseName()` each time even when the `cleanup` is not set. If we are to convert the pointer to a value, then `cleanup != nil` guard converts to two `if ...path.CRDs` and `if ...path.CAPIProviderCRDs`

Up to the implementer

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -381,3 +380,11 @@ rules:
...
   - inclusterippools
   verbs: {{ include "rbac.editorVerbs" . | nindent 4 }}
+- apiGroups:
+    - apiextensions.k8s.io
+  resources:
+    - customresourcedefinitions
+  verbs:
+    - list
+    - delete
+    - get
```
zerospiel:
```suggestion
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs:
  - get
  - list
  - watch
  - delete
```

nit (because the copilot highlighted it as well): we need to add the `watch` term because of the cluster-wide watcher being spun when the `list` is called (**see the comment** about the direct reader **above**)

The indentation and the "read-then-write" order are optional and up to the implementer

internal/controller/management_controller_test.go
```diff
@@ -495,6 +496,93 @@ var _ = Describe("Management Controller", func() {
...
+			Eventually(k8sClient.Get).WithArguments(ctx, client.ObjectKeyFromObject(crd), &apiextv1.CustomResourceDefinition{}).
+				WithTimeout(timeout).WithPolling(interval).Should(Succeed())
+
+			r := &ManagementReconciler{Client: k8sClient}
+			sel := labels.SelectorFromSet(map[string]string{kcmv1.FluxHelmChartNameKey: kcmv1.CoreKCMName})
+			Expect(r.removeCRDsWithSelector(ctx, sel)).To(Succeed())
+
+			Consistently(func() error {
+				return k8sClient.Get(ctx, client.ObjectKeyFromObject(crd), &apiextv1.CustomResourceDefinition{})
+			}).WithTimeout(2 * time.Second).WithPolling(interval).Should(Succeed())
```
zerospiel:
```suggestion
			}).WithTimeout(500 * time.Millisecond).WithPolling(interval).Should(Succeed())
```

nit: for "did not delete", 250 ms polling for `2s` is 8 polls that always run to completion. Consider `500ms` — fast tests matter and this negative check is cheap after the first successful poll

Up to the implementer

internal/controller/management_controller_test.go
```diff
@@ -495,6 +496,93 @@ var _ = Describe("Management Controller", func() {
 		})
 	})
 
+	Context("removeCRDsWithSelector", func() {
```
zerospiel, line 499: nit: I see a lack of the following checks:
- when the `cleanup` is not set (CRDs are preserved)
- skipping items marked for deletion (`deletionTimestamp` is set)
- multiple CRDs where one deletion fails

Maybe an over-check, up to the implementer

internal/controller/management_controller_test.go
```diff
@@ -495,6 +496,93 @@ var _ = Describe("Management Controller", func() {
...
+
+			r := &ManagementReconciler{Client: k8sClient}
+			sel := labels.SelectorFromSet(map[string]string{kcmv1.FluxHelmChartNameKey: kcmv1.CoreKCMName})
+			Expect(r.removeCRDsWithSelector(ctx, sel)).To(Succeed())
+
+			Consistently(func() error {
+				return k8sClient.Get(ctx, client.ObjectKeyFromObject(crd), &apiextv1.CustomResourceDefinition{})
+			}).WithTimeout(2 * time.Second).WithPolling(interval).Should(Succeed())
+
+			Expect(k8sClient.Delete(ctx, crd)).To(Succeed())
```
zerospiel, line 602: nit: could be deferred

```go
DeferCleanup(func() {
    _ = k8sClient.Delete(ctx, crd)
})
```

Up to the implementer

api/v1beta1/management_types.go
```diff
@@ -46,6 +51,21 @@ type ManagementSpec struct {
 	ComponentsCommonSpec `json:",inline"`
 }
 
+// ManagementCleanup controls which CRDs are removed when a Management object is deleted.
+type ManagementCleanup struct {
+	// +optional
+
+	// CRDs indicates whether k0rdent-owned CRDs should be removed when a Management is deleted.
```
zerospiel: nit: the user-facing docstring should say `(a)` deletion is fire-and-forget, `(b)` enabling `CRDs` will remove the `Management` CRD itself as part of the sweep

I'd say `(b)` may worth to mention it, `(a)` is very optional

zerospiel (commented):
In general LGTM, please fix the RBAC-related commit from Copilot.

From my side, I left a bunch of nit-picking comments, they ALL are optional and up to the implementer (except the RBAC-related one), feel free to either address them or simply resolve.

zerospiel:
Oh, and please rebase the branch one more time. Thanks!

zerospiel (approved):
LGTM, weird `go.sum` diff but can be done in one of the other PRs

PR #2563, opened 2026-03-23
fix: multiple fixes for ClusterDeployment conditions

### LINKED ISSUE #2481: [bug] ClusterDeployment Ready condition uses reason: Failed for transient in-progress states
## Describe the bug

When a ClusterDeployment is still converging (e.g., CNI initializing after node creation, workers waiting for control plane), the Ready condition is set to:

```yaml
type: Ready
status: "False"
reason: Failed
message: |
  * ControlPlaneMachinesReady:
    * Machine e2e-test-01-cp-cr54d:
      * NodeHealthy:
        * Node.Ready: container runtime network not ready: NetworkReady=false
          reason:NetworkPluginNotReady message:Network plugin returns error:
          cni plugin not initialized
  * WorkersAvailable:
    * MachineDeployment e2e-test-01-md: 0 available replicas, at least 1 required
  * WorkerMachinesReady:
    * Machine e2e-test-01-md-mlk2q-d6dzd:
      * NodeHealthy:
        * Node.Ready: container runtime network not ready
```

This cluster eventually becomes Ready without intervention — the CNI initializes within a couple of minutes. However, `reason: Failed` is indistinguishable from an actual terminal failure (e.g., invalid credentials, quota exceeded, misconfigured template).

## Root cause

The [`setCondition` helper](https://github.com/k0rdent/kcm/blob/main/internal/controller/clusterdeployment_controller.go#L1038) uses a binary mapping — any non-nil error produces `FailedReason`, any nil error produces `SucceededReason`:

```go
func (*ClusterDeploymentReconciler) setCondition(cd *kcmv1.ClusterDeployment, typ string, err error) (changed bool) {
	reason, cstatus, msg := kcmv1.SucceededReason, metav1.ConditionTrue, ""
	if err != nil {
		reason, cstatus, msg = kcmv1.FailedReason, metav1.ConditionFalse, err.Error()
	}
	return apimeta.SetStatusCondition(&cd.Status.Conditions, metav1.Condition{
		Type:               typ,
		Status:             cstatus,
		Reason:             reason,
		Message:            msg,
		ObservedGeneration: cd.Generation,
	})
}
```

There is no intermediate state. The condition aggregation logic (`aggregateCapiConditions`, `updateStatus`) rolls up sub-resource conditions into the top-level `Ready` condition using the same binary approach — if any CAPI sub-condition is not yet satisfied, the result is `Ready=False, reason=Failed`, regardless of whether the underlying state is transient (e.g., `WaitingForAvailableMachines`, CNI initializing) or terminal (e.g., invalid credentials, quota exceeded).

## Impact

Consumers of the ClusterDeployment status (automation, dashboards, workflow engines) cannot programmatically distinguish between:

- **"Still provisioning, will converge"** — CNI not ready yet, machines booting, control plane starting
- **"Actually failed, needs human intervention"** — bad credentials, resource limits, template not found

The only workaround is to keep polling and rely on a timeout to decide whether the failure is permanent — which delays error detection and makes progress reporting misleading.

This also affects multiple providers. For example, the [[KubeVirt troubleshooting docs](https://docs.k0rdent.io/latest/troubleshooting/known-issues-kubevirt/)](https://docs.k0rdent.io/latest/troubleshooting/known-issues-kubevirt/) show a ClusterDeployment with `reason: Failed` while the cluster is simply waiting for infrastructure to provision — demonstrating this is not provider-specific.

## Expected behavior

Use distinct `reason` values for in-progress vs terminal states:

| State | Suggested `reason` | When |
|---|---|---|
| Sub-resources converging | `Provisioning` or `Reconciling` | Machines booting, CNI initializing, control plane starting |
| Terminal error | `Failed` | Invalid credentials, quota exceeded, template not found |
| Successfully ready | `Provisioned` | All sub-conditions satisfied |

This aligns with [[Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) where `Reason` is intended to be a machine-readable CamelCase token that indicates the specific cause, enabling programmatic decision-making.

Note that the `Management` resource within KCM already uses more granular reasons (`AllComponentsHealthyReason`, `NotAllComponentsHealthyReason`, `ReleaseIsNotReadyReason`, `HasIncompatibleContractsReason`) — so there is precedent within the codebase for this pattern.

## Suggested fix

The fix involves two levels:

1. **`setCondition` helper** — either accept a `reason` parameter, or introduce typed/sentinel errors that the helper can inspect to choose between `Provisioning` and `Failed`.

2. **Condition aggregation** (`aggregateCapiConditions` / `updateStatus`) — when rolling up CAPI sub-conditions, inspect the underlying reasons. If sub-conditions indicate transient progress (e.g., `WaitingForAvailableMachines`, `WaitingForControlPlaneInitialization`, nodes not yet ready), produce `reason: Provisioning`. Reserve `reason: Failed` for conditions that indicate terminal errors.

## To Reproduce

1. Create a ClusterDeployment using any standalone template (e.g., `aws-standalone-cp-1-0-20`)
2. Observe the Ready condition immediately after creation — it shows `reason: Failed` while infrastructure is being provisioned
3. Wait 10-15 minutes — the cluster becomes Ready without intervention
4. The `reason: Failed` was misleading for the entire provisioning duration

## Environment

- k0rdent version: v1.2.3 (KCM)
- Template: `aws-standalone-cp-1-0-20`
- Provider: AWS (also reproducible with KubeVirt per docs)

eromanova:
This PR introduces several fixes and improvements to ClusterDeployment conditions:

1. When the CAPI Cluster is not yet ready, ClusterDeployment now reports the `Progressing` reason instead of `Failed` in the Ready condition.
2. If the CAPI Cluster remains unready for more than 30 minutes, ClusterDeployment reports `Failed` in the `Ready` condition, indicating that provider logs should be checked for more details.
3. Improved handling of deletion conditions, ensuring accurate reasons are reported for different deletion stages.
5. Added tests to verify that the `Ready` condition is set correctly.

Fixes #2481

### DIFF
--- a/api/v1beta1/clusterdeployment_types.go
+++ b/api/v1beta1/clusterdeployment_types.go
@@ -66,8 +66,9 @@ const (
 	// WaitingForClusterDeletionReason indicates the cluster is being deleted, and the controller is waiting for
 	// the cluster deletion to complete before deleting the [ClusterDeployment].
 	WaitingForClusterDeletionReason = "WaitingForClusterDeletion"
-	// WaitingForHelmReleaseDeletionReason indicates the controller is waiting for the [HelmRelease] deletion to
-	// complete before proceeding with the [ClusterDeployment] deletion.
+	// WaitingForHelmReleaseDeletionReason indicates the controller is waiting for
+	// the [github.com/fluxcd/helm-controller/api/v2.HelmRelease] deletion to complete before proceeding with
+	// the [ClusterDeployment] deletion.
 	WaitingForHelmReleaseDeletionReason = "WaitingForHelmReleaseDeletion"
 	// WaitingForServiceSetsDeletionReason indicates that the controller is waiting for the deletion of the [ServiceSet]
 	// objects to complete before proceeding with the [ClusterDeployment] deletion.


api/v1beta1/common.go
```diff
@@ -29,6 +29,12 @@ const (
 	// resource or an action has started.
 	ProgressingReason string = "Progressing"
 
+	// NotDeletingReason appears when an object is not being deleted (when DeletionTimestamp is not set).
+	NotDeletingReason string = "NotDeleting"
+
+	// DeletingReason indicates a condition or event for the deletion of the object.
+	DeletingReason string = "Deleting"
```
zerospiel: these reasons look redundant (at least the `NotDeletingReason`), do we actually need them? I haven't delved into the code, though

`NotDeletingReason` is always set to the `False` status (!deleting == deleting, thus `DeletingReason` is `True`), and that indicates a probable issue with it

api/v1beta1/clusterdeployment_types.go
```diff
@@ -59,6 +59,26 @@ const (
...
+	// deleted before proceeding with the [ClusterDeployment] deletion.
+	WaitingForCloudResourcesDeletionReason = "WaitingForCloudResourcesDeletion"
+	// WaitingForClusterDeletionReason indicates the cluster is being deleted, and the controller is waiting for
+	// the cluster deletion to complete before deleting the [ClusterDeployment].
+	WaitingForClusterDeletionReason = "WaitingForClusterDeletion"
+	// WaitingForHelmReleaseDeletionReason indicates the controller is waiting for the HelmRelease deletion to
+	// complete before proceeding with the [ClusterDeployment] deletion.
+	WaitingForHelmReleaseDeletionReason = "WaitingForHelmReleaseDeletion"
+	// WaitingForServiceSetsDeletionReason indicates that the controller is waiting for the deletion of the [ServiceSet]
+	// objects to complete before proceeding with the ClusterDeployment deletion.
```
zerospiel:
```suggestion
	// objects to complete before proceeding with the [ClusterDeployment] deletion.
```

nit-pick

internal/controller/clusterdeployment_controller.go
```diff
@@ -993,10 +1008,16 @@ func (*ClusterDeploymentReconciler) initClusterConditions(cd *kcmv1.ClusterDeplo
 		if typ == kcmv1.HelmReleaseReadyCondition && cd.Spec.DryRun {
 			continue
 		}
+		status := metav1.ConditionUnknown
+		reason := kcmv1.ProgressingReason
+		if typ == kcmv1.DeletingCondition {
+			status = metav1.ConditionFalse
+			reason = kcmv1.NotDeletingReason
+		}
```
zerospiel: during the init this condition will have `NotDeleting` and the `DeletingCondition == False`. Looks reasonable

zerospiel: Okay, it seems there is no issue with that

api/v1beta1/common.go
eromanova: Technically, it doesn't make much sense to initialize the Deleting condition when the object is not actually being deleted. The approach to set the `Deleting` condition with the `NotDeleting` reason is currently implemented for CAPI Clusters. I've double-checked the code, and it looks like we can get rid of it and set the `Deleting` condition only when the object is being deleted.

zerospiel: Okay, seems there are no issues with the reasons. But at this point, I am not sure. 
Maybe it might not be a very bad idea to reuse the conditions/reasons directly from CAPI (we already have the API package as a dependency). But I believe that is up to the implementer in this case

zerospiel: Oh, sorry, I was a bit slower. Yes, there is no problem, it is a bit confusing (to me), having the fact that there are other very detailed condition reasons. So as I've written above, it is up to you whether to leave them as is/reuse from CAPI/get rid of them

internal/controller/clusterdeployment_controller.go
eromanova: Here we'll have the following condition:
```
type: Deleting
status: False
reason: NotDeleting
```
Which is correct and indicates that the object is not being deleted (Deleting: False with the reason: `NotDeleting`), same as in CAPI Clusters. But I decided to address your previous comment since it looks like NotDeleting is not really needed.

api/v1beta1/common.go
eromanova: Thanks, Michael. You’re right that keeping the NotDeleting reason for the Deleting condition wouldn’t cause any issues. Conditions should reflect the actual state, and I don’t see much value in maintaining a Deleting condition with a NotDeleted reason at all times. When deletion is in progress, it makes sense to mirror the CAPI Deleting condition summary. When it’s not, the condition doesn’t add much value. For that reason, I decided to remove it.

api/v1beta1/clusterdeployment_types.go
```diff
@@ -66,11 +66,11 @@ const (
 	// WaitingForClusterDeletionReason indicates the cluster is being deleted, and the controller is waiting for
 	// the cluster deletion to complete before deleting the [ClusterDeployment].
 	WaitingForClusterDeletionReason = "WaitingForClusterDeletion"
-	// WaitingForHelmReleaseDeletionReason indicates the controller is waiting for the HelmRelease deletion to
+	// WaitingForHelmReleaseDeletionReason indicates the controller is waiting for the [HelmRelease] deletion to
```
zerospiel: nit-picking: if we want the direct link, then we have to ensure the correct link: `[github.com/fluxcd/helm-controller/api/v2.HelmRelease]`. Up to you whether to apply it or skip

internal/util/conditions/ready_condition.go
```diff
@@ -0,0 +1,109 @@
...
+			// If False and NotPaused, that's normal operation - no need to include in status
+			continue
+		}
+		switch cond.Status {
+		case metav1.ConditionTrue:
+			// Do nothing
+		case metav1.ConditionUnknown:
+			unknownMsgs = append(unknownMsgs, cond.Message)
+		case metav1.ConditionFalse:
+			errMsg, warningMsg := handleFalseConditionFunc(cond)
```
wahabmk, line 66: We are calling `handleFalseConditionFunc` in the case when condition is False. However, the func that is passed in internal/controller/multiclusterservice_controller.go:345 (in this PR) is `handleMultiClusterServiceFailedCondition`, which does not seem to handle False conditions. Is this correct?

wahabmk: I see, this is just a refactor of what already existed.

eromanova: Correct, I didn’t change this logic. Initially, the `multiClusterService` has a `Ready` condition with a `Failed` reason when some services or clusters are not ready. This behavior remains unchanged, but it can be refactored as part of a separate task if needed.

Issue #2567, opened 2026-03-25
[bug] MCS unable to deploy services if ServiceTemplate not in system ns


wahabmk:
# Steps to Reproduce
1. Create a new Namespace `dev`.
2. Create HelmRepository in `dev`:
```yaml
apiVersion: source.toolkit.fluxcd.io/v1
kind: HelmRepository
metadata:
  name: k0rdent-catalog
  namespace: dev
  labels:
    k0rdent.mirantis.com/managed: "true"
spec:
  insecure: true
  interval: 10m0s
  provider: generic
  type: oci
  url: oci://ghcr.io/k0rdent/catalog/charts
```
3. Create ServiceTemplate in `dev`:
```yaml
apiVersion: k0rdent.mirantis.com/v1beta1
kind: ServiceTemplate
metadata:
  annotations:
    helm.sh/resource-policy: keep
  name: ingress-nginx-4-13-0
  namespace: dev
spec:
  # k8sConstraint: "<v1.30"
  helm:
    chartSpec:
      chart: ingress-nginx
      sourceRef:
        kind: HelmRepository
        name: k0rdent-catalog
      version: 4.13.0
```
4. Verify that the ServiceTemplate has successfully reconciled:
```yaml
➜  ~ kubectl -n dev get servicetemplate
NAME                   VALID
ingress-nginx-4-13-0   true
```
5. Create a ClusterDeployment in `dev` namespace with `owner=dev-team` label.
6. Create the following MCS matching the CD and using the above ServiceTemplate:
```yaml
apiVersion: k0rdent.mirantis.com/v1beta1
kind: MultiClusterService
metadata:
  name: mcs-dev
spec:
  clusterSelector:
    matchLabels:
      owner: dev-team
  serviceSpec:
    provider:
      name: ksm-projectsveltos
      # selfManagement: true
      config:
        continueOnError: true
        priority: 100
    services:
      - template: ingress-nginx-4-13-0
        name: nginx
        namespace: nginx
```
7. ❌ The MCS will show the following error in its status:
```yaml
apiVersion: k0rdent.mirantis.com/v1beta1
kind: MultiClusterService
metadata:
  . . .
spec:
  . . . 
status:
  conditions:
  - lastTransitionTime: "2026-03-25T17:26:08Z"
    message: |-
      some services have invalid templates
[...]

PR #2582, opened 2026-03-31
fix: don't overwrite err in defer func in mcs controller

wahabmk:
Fixes part of https://github.com/k0rdent/kcm/issues/2567.

### DIFF
--- a/internal/controller/multiclusterservice_controller.go
+++ b/internal/controller/multiclusterservice_controller.go
@@ -72,17 +72,6 @@ func (r *MultiClusterServiceReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
-	clone := mcs.DeepCopy()
-	defer func() {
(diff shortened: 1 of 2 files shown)


internal/controller/multiclusterservice_controller.go
```diff
@@ -72,17 +72,6 @@ func (r *MultiClusterServiceReconciler) Reconcile(ctx context.Context, req ctrl.
...
-	clone := mcs.DeepCopy()
-	defer func() {
-		// we need to explicitly requeue MultiClusterService object,
-		// otherwise we'll miss if some ClusterDeployment will be updated
-		// with matching labels.
-		var requeue bool
-		if requeue, err = r.updateStatus(ctx, clone, mcs); requeue {
-			result = ctrl.Result{RequeueAfter: r.defaultRequeueTime}
-		}
-	}()
```
wahabmk, line 84: We don't need this defer func because we have the exact same defer function in reconcileUpdates func.

PR #2586, opened 2026-04-01
fix: support dynamic kcm release name

### LINKED ISSUE #2585: [bug] Helm installation fails when using non-default release name
Currently, the installation of the kcm helm chart only works when the release name is explicitly set to `kcm`, as shown in the documentation:
```
helm install kcm oci://ghcr.io/k0rdent/kcm/charts/kcm --version 1.8.0 -n kcm-system --create-namespace
```

When installing with a different release name, for example:
```
helm install k0rdent oci://ghcr.io/k0rdent/kcm/charts/kcm --version 1.8.0 -n kcm-system --create-namespace
```
the installation does not complete successfully and appears to get stuck.

**Expected Behavior**
* The chart should install and function correctly regardless of the helm release name
* No implicit dependency on a specific release name should exist

eromanova:
This PR fixes the controller code to support KCM installation with a custom name. It includes:
1. KCM helm chart now passes the `.Release.name` as an env var, reads this value in the controller code, and uses it as the `spec.release` field in the KCM HelmRelease object.
2. Minor code refactoring to address linter issues.
3. Override kcm helm release name in e2e tests.

Fixes #2585

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -16,6 +16,7 @@ package config
 
 import (
 	"os"
+	"sync"
 
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
 )
@@ -25,10 +26,10 @@ const kcmHelmReleaseNameEnvVar = "HELM_RELEASE_NAME"
 // KCMHelmReleaseName returns the name of the helm release with core KCM components. The name is expected
 // to be provided via the HELM_RELEASE_NAME environment variable.
 // If HELM_RELEASE_NAME is not set or is empty, it falls back to the default core KCM release name.
-func KCMHelmReleaseName() string {
+var KCMHelmReleaseName = sync.OnceValue(func() string {
 	releaseName, ok := os.LookupEnv(kcmHelmReleaseNameEnvVar)
 	if !ok || releaseName == "" {
 		return kcmv1.CoreKCMName
 	}
 	return releaseName
-}
+})
--- a/internal/controller/components/cleanup.go
+++ b/internal/controller/components/cleanup.go
@@ -27,6 +27,7 @@ import (
 
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
 	"github.com/K0rdent/kcm/internal/config"
+	"github.com/K0rdent/kcm/internal/helm"
 	"github.com/K0rdent/kcm/internal/record"
 	releaseutil "github.com/K0rdent/kcm/internal/util/release"
 )
@@ -77,14 +78,14 @@ func Cleanup(
 		}
 
 		componentName := hr.Name // providers(components) names map 1-1 to the helmreleases names
-
-		if componentName == releaseutil.Name(cluster.HelmReleasePrefix(), kcmv1.CoreCAPIName) ||
-			componentName == releaseutil.Name(cluster.HelmReleasePrefix(), kcmChartName) ||
+		hrPrefix := cluster.HelmReleasePrefix()
+		if componentName == helm.ReleaseName(hrPrefix, kcmv1.CoreCAPIName) ||
+			componentName == helm.ReleaseName(hrPrefix, kcmChartName) ||
 			slices.ContainsFunc(releasesList.Items, func(r metav1.PartialObjectMetadata) bool {
 				return componentName == releaseutil.TemplatesChartFromReleaseName(r.Name)
 			}) ||
 			slices.ContainsFunc(cluster.Components().Providers, func(newComp kcmv1.Provider) bool {
-				return componentName == releaseutil.Name(cluster.HelmReleasePrefix(), newComp.Name)
+				return componentName == helm.ReleaseName(hrPrefix, newComp.Name)
 			}) {
 			continue
 		}
--- a/internal/controller/components/reconciler.go
+++ b/internal/controller/components/reconciler.go
@@ -46,7 +46,6 @@ import (
 	"github.com/K0rdent/kcm/internal/record"
 	kubeutil "github.com/K0rdent/kcm/internal/util/kube"
 	pullsecretutil "github.com/K0rdent/kcm/internal/util/pullsecret"
-	releaseutil "github.com/K0rdent/kcm/internal/util/release"
 )
 
 type ReconcileComponentsOpts struct {
@@ -172,6 +171,8 @@ func Reconcile(
 		return false, fmt.Errorf("copy regional proxy secret: %w", err)
 	}
 
+	hrPrefix := cluster.HelmReleasePrefix()
+
 	for _, component := range components {
 		l.V(1).Info("reconciling components", "component", component)
 		var notReadyDeps []string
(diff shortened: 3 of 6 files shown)


internal/controller/components/reconciler.go
```diff
@@ -83,14 +83,20 @@ type clusterInterface interface {
...
 	kcmv1.Component
 
-	name            string
+	// name is the name of the component (core component of provider).
+	name string
+	// helmReleaseName is the name of the HelmRelease object created for this component.
 	helmReleaseName string
+	// releaseName is the name of the Helm release to be passed to spec.releaseName of HelmRelease object
+	// created for this component. Defaults to `name` if not set.
+	releaseName     string
```
zerospiel, line 100: I just wonder, what is the difference between the `"components.component".helmReleaseName` and `"components.component".releaseName`? If they are the same, I think we can have just one entity, e.g. `releaseName` and drop the rest: `helmReleaseName` and `name`, as they cause confusion (to me)

internal/controller/components/reconciler.go
```diff
@@ -392,6 +368,51 @@ func getWrappedComponents(ctx context.Context, cluster clusterInterface, release
 	return components, nil
 }
 
+func getHelmReleaseReconcileOpts(
```
zerospiel, line 373: If the linter bothers, setting `nolint` is okay. I will revise linters settings to determine sub-optimal ones

internal/controller/components/reconciler.go
eromanova: These two fields are closely related but serve different purposes:
`helmReleaseName` is the name of the Flux HelmRelease object (ex. `kcm`, `capi`, etc.).
`releaseName` is the Helm release name used during installation (ex. what is passed to helm install <releaseName>). This value is stored in `spec.helmRelease` of the Flux HelmRelease.
For most components, these values are the same. But kcm is an exception. Its Helm release name must remain consistent with the original installation name, while I wanted to keep the HelmRelease name equal to `kcm`, since other parts still depend on it (in the Management and Release specs).

Because of this, I intentionally kept both fields to distinguish between the Flux HelmRelease resource name (helmReleaseName), and the actual Helm release name used at install time (releaseName), especially to preserve the original kcm release identity regardless of how the installation was named.

internal/controller/components/reconciler.go
eromanova: Yeah, but I thought moving this part into a separate function wouldn’t be a bad choice, as it would make the code more readable

internal/controller/components/reconciler.go
zerospiel: Correct. I've checked that the `helmReleaseName` is used both as the `ReleaseName` of the `flux.HelmRelease` object to be created (if no `ReleaseName` is not set in our HelmOptions) and as the name of the object. So merging these fields into a single one will change:

| option | `helm install` name | flux obj name | flux obj `spec.ReleaseName` |
| --- | --- | --- | --- |
| two fields | `k0rdent` | `kcm` | `k0rdent` |
| two fields | `kcm` | `kcm` | `kcm` |
| single field | `k0rdent` | `k0rdent` | `k0rdent` |
| single field | `kcm` | `kcm` | `kcm` |

Please correct me if I missed something. So I am not sure whether we actually need to restrain the given (customized) name.

Regarding the `name` field – it is used a bit more widely and changing it might take up to two releases to support backward compatibility and the migration process.

In any case, the only issue that I see with these 3 fields is confusion for the dev team and more complex code maintenance, using extra cognitive resources to properly use one of these.

To sum up, feel free to resolve and go forward, LGTM otherwise

eromanova: I understand these struct parameters can be a bit confusing. I’ve done a small refactoring (as a result, removed the `helmReleaseName` parameter), which should make things clearer.

internal/controller/components/cleanup.go
```diff
@@ -74,12 +78,14 @@ func Cleanup(
 
 		componentName := hr.Name // providers(components) names map 1-1 to the helmreleases names
 
-		if componentName == cluster.HelmReleaseName(kcmv1.CoreCAPIName) ||
-			componentName == cluster.HelmReleaseName(cluster.KCMHelmChartName()) ||
+		if componentName == releaseutil.Name(cluster.HelmReleasePrefix(), kcmv1.CoreCAPIName) ||
```
zerospiel: I think we can evaluate `cluster.HelmReleasePrefix` just once

internal/util/release/release.go
```diff
@@ -44,3 +44,13 @@ func ReleaseNameFromVersion(version string) (string, error) {
 func TemplatesChartFromReleaseName(releaseName string) string {
 	return releaseName + "-tpl"
 }
+
+// Name constructs a release name by combining an optional prefix with a name. If prefix is empty, it returns
+// the name unchanged. This is used for HelmRelease naming conventions where regional resources
+// are prefixed with the region name.
+func Name(prefix, name string) string {
```
zerospiel: nit: would it be better to rename to `HelmReleaseName`? Because the current code is exclusively related to `kcmv1.Release` object rather than something else

internal/config/config.go
```diff
@@ -0,0 +1,34 @@
...
+
+	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+)
+
+const kcmHelmReleaseNameEnvVar = "HELM_RELEASE_NAME"
+
+// KCMHelmReleaseName returns the name of the helm release with core KCM components. The name is expected
+// to be provided via the HELM_RELEASE_NAME environment variable.
+// If HELM_RELEASE_NAME is not set or is empty, it falls back to the default core KCM release name.
+func KCMHelmReleaseName() string {
```
zerospiel: nit: could be evaluated just once via the `sync.OnceValue`

internal/controller/components/reconciler.go
```diff
@@ -229,53 +237,24 @@ func Reconcile(
...
-			hrReconcileOpts.Labels = opts.Labels
-		}
-
-		if template.Spec.Helm.ChartSpec != nil {
-			hrReconcileOpts.ReconcileInterval = &template.Spec.Helm.ChartSpec.Interval.Duration
-		}
-
-		_, operation, err := helm.ReconcileHelmRelease(ctx, mgmtClient, component.helmReleaseName, opts.Namespace, hrReconcileOpts)
+		hrReconcileOpts := getHelmReleaseReconcileOpts(cluster, component, template, opts)
+		hrPrefix := cluster.HelmReleasePrefix()
```
zerospiel: nit: ditto, can be evaluated once

internal/controller/components/reconciler.go
```diff
@@ -426,34 +449,34 @@ func updateComponentsStatus(
 // checkProviderStatus checks the status of a provider associated with a given
 // ProviderTemplate name. Since there's no way to determine resource Kind from
 // the given template iterate over all possible provider types.
-func checkProviderStatus(ctx context.Context, mgmtClient, rgnlClient client.Client, component component, systemNamespace string) error {
-	helmReleaseName := component.helmReleaseName
+func checkProviderStatus(ctx context.Context, cluster clusterInterface, mgmtClient, rgnlClient client.Client, component component, systemNamespace string) error {
```
zerospiel: nit: can we pass here either the prefix or the helm release name directly?

PR #2609, opened 2026-04-09
test: improve e2e test performance

### LINKED ISSUE #474: Support failing tests in `validate*` methods when certain errors appear for longer than a desired threshold
Currently we have a few validations like this one:

```
	if cluster != nil {
		phase, _, _ := unstructured.NestedString(cluster.Object, "status", "phase")
		if phase != "Deleting" {
			return fmt.Errorf("cluster: %q exists, but is not in 'Deleting' phase", clusterName)
		}
```

Where we are trying to validate a deletion and are waiting on a certain phase.  In instances like this it doesn't make sense to spend 90m waiting for the cluster to enter `'Deleting'` phase, instead we should probably only wait 60s here.  It'd be nice to support a threshold where we call `Fail` after hitting a state that should exist for some time but not for a long time.

kylewuolle:
Improves the performance of e2e tests
Fixes #474

### DIFF
--- a/hack/test-e2e.bash
+++ b/hack/test-e2e.bash
@@ -23,9 +23,17 @@ if [[ -n "${GINKGO_LABEL_FILTER:-}" ]]; then
   ginkgo_label_flag="-ginkgo.label-filter=${GINKGO_LABEL_FILTER}"
 fi
 
+ginkgo_focus_flag=""
+if [[ -n "${GINKGO_FOCUS:-}" ]]; then
+  ginkgo_focus_flag="-ginkgo.focus=${GINKGO_FOCUS}"
+fi
+
 go_test_args=(./test/e2e/ -v -ginkgo.v -ginkgo.timeout=6h -timeout=6h)
 if [[ -n "${ginkgo_label_flag}" ]]; then
   go_test_args+=("${ginkgo_label_flag}")
+fi
+if [[ -n "${ginkgo_focus_flag}" ]]; then
+  go_test_args+=("${ginkgo_focus_flag}")
 fi
 
 if [[ "${GITHUB_ACTIONS:-}" == "true" ]]; then
--- a/test/e2e/clusterdeployment/providervalidator.go
+++ b/test/e2e/clusterdeployment/providervalidator.go
@@ -229,8 +229,7 @@ func (p *ProviderValidator) Validate(ctx context.Context, kc *kubeclient.KubeCli
 		if err := validator(ctx, kc, p.clusterName); err != nil {
 			_, _ = fmt.Fprintf(GinkgoWriter, "[%s/%s] validation error: %v\n", p.templateType, name, err)
 
-			var thresholdErr *ThresholdError
-			if errors.As(err, &thresholdErr) {
+			if thresholdErr, ok := errors.AsType[*ThresholdError](err); ok {
 				if firstSeen, ok := p.thresholdFirstSeen[name]; !ok {
 					p.thresholdFirstSeen[name] = time.Now()
 				} else if time.Since(firstSeen) > thresholdErr.Threshold {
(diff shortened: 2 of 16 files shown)


hack/test-e2e.bash
```diff
@@ -23,10 +23,18 @@ if [[ -n "${GINKGO_LABEL_FILTER:-}" ]]; then
   ginkgo_label_flag="-ginkgo.label-filter=${GINKGO_LABEL_FILTER}"
 fi
 
+ginkgo_focus_flag=""
```
zerospiel, line 26: I probably missed this, but why was the new envvar added?

test/e2e/functional_test.go
```diff
@@ -54,14 +56,16 @@ const (
 	openWebuiVersion     = "8.10.0"
 	nginxServiceName     = "managed-ingress-nginx"
 	validatorTimeout     = 30 * time.Minute
-	validatorPoll        = 10 * time.Second
+	validatorPoll        = 5 * time.Second
```
zerospiel: nit-picking: the const (along with many others above) is set for the package but is being used exclusively in this file, do this (and the other) const need to be package-wide? For instance, in the other tests, you've introduced local constants

test/e2e/provider_vsphere_test.go
```diff
@@ -87,6 +87,8 @@ var _ = Context("vSphere Templates", Label("provider:onprem", "provider:vsphere"
 		}
 	})
 
+	const pollInterval = 5 * time.Second
```
zerospiel: Tests-wide question: why has the polling period been decreased? Does it speed up the test? If deployment is taking, e.g., 15 Mins on average, it means that instead of ~90 log msgs, ~180 log msgs will be produced

test/e2e/clusterdeployment/providervalidator.go
```diff
@@ -220,8 +228,21 @@ func (p *ProviderValidator) Validate(ctx context.Context, kc *kubeclient.KubeCli
 
 		if err := validator(ctx, kc, p.clusterName); err != nil {
 			_, _ = fmt.Fprintf(GinkgoWriter, "[%s/%s] validation error: %v\n", p.templateType, name, err)
+
+			var thresholdErr *ThresholdError
+			if errors.As(err, &thresholdErr) {
```
zerospiel:
```suggestion
			if thresholdErr, ok := errors.AsType[*ThresholdError](err); ok {
```

test/e2e/functional_test.go
```diff
@@ -151,6 +155,10 @@ var _ = Describe("Functional e2e tests", Label("provider:cloud", "provider:docke
 		templates.CreateTemplateChain(context.Background(), kc.CrClient, kubeutil.DefaultSystemNamespace, templateChainName, kcmv1.TemplateChainSpec{
 			SupportedTemplates: supportedTemplates,
 		})
+
+		By("Creating shared cluster for tests that don't require custom service dependency specs")
```
zerospiel: General question: What do these changes have to do with addressing the original issue with the deleting phase?

zerospiel: And one more general question/request: if the major changes come to the `functional` tests, I'd say it makes sense to test these changes along with others, wdyt?

zerospiel:
Forgot to mention the deadcode:
```
test/e2e/functional_test.go:766:6: unreachable func: waitForSveltosResourcesDeleted
test/e2e/clusterdeployment/clusterdeployment.go:102:6: unreachable func: FilterAllProviders
test/e2e/clusterdeployment/clusterdeployment.go:112:6: unreachable func: GetProviderLabel
```

test/e2e/provider_vsphere_test.go
kylewuolle: Yes this will speed up tests because validations can succeed or fail faster

test/e2e/functional_test.go
kylewuolle: Done!

BROngineer:
@kylewuolle please either address Copilot's review comments or explain why they are irrelevant.

Issue #2406, opened 2026-02-16
[tests] Improve KCM controller test coverage


eromanova:
Some parts of the KCM controller currently lack sufficient test coverage. This issue tracks adding or expanding unit/integration tests to improve reliability.

PR #2618, opened 2026-04-12
test: improve ClusterDeployment integration tests

eromanova:

This PR improves the `ClusterDeployment` controller integration tests.

Changes:
1. Init the Event recorder in the controller testing suite setup
2. Improve ClusterDeployment controller integration tests (`internal/controller/clusterdeployment_controller_test.go`)

All changes are in test infrastructure and test code. No production logic is modified.

Related issue #2406

### DIFF
--- a/internal/controller/clusterdeployment_controller_test.go
+++ b/internal/controller/clusterdeployment_controller_test.go
@@ -16,28 +16,104 @@ package controller
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"strings"
 	"time"
 
 	helmcontrollerv2 "github.com/fluxcd/helm-controller/api/v2"
 	fluxmeta "github.com/fluxcd/pkg/apis/meta"
+	fluxconditions "github.com/fluxcd/pkg/runtime/conditions"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chartutil"
 	corev1 "k8s.io/api/core/v1"
-	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apimachinery/pkg/api/meta"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	apiserverv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	clusterapiv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	. "sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+	ctrl "sigs.k8s.io/controller-runtime"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/yaml"
 
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
-	kubeutil "github.com/K0rdent/kcm/internal/util/kube"
 )
+
+const (
+	kubernetesVersion = "v1.35.1"
+	testFinalizer     = "kcm.cluster.k0sproject.io/test-finalizer"
+
+	// Global certificate secret names
+	k0sURLCertSecretName   = "test-k0s-url-cert"
+	registryCertSecretName = "test-registry-cert"
+
+	// Region configuration
+	regionName               = "test-region-cld"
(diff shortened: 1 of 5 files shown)


internal/controller/clusterdeployment_controller_test.go
```diff
@@ -67,537 +144,1269 @@ func (*fakeHelmActor) EnsureReleaseWithValues(_ context.Context, _ *action.Confi
...
+		}
+		Eventually(func(g Gomega) {
+			g.Expect(apierrors.IsNotFound(k8sClient.Get(ctx, crclient.ObjectKeyFromObject(cld), &ss))).To(BeTrue())
+		}).Should(Succeed())
+	})
+
+	if tc.dryRun {
+		By("DryRun mode: deletion should complete immediately", func() {
+			result, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: cldName})
+			Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
eromanova: This concern does not apply here. With the current implementation, the ServiceSet object is being created for both dryRun and non-dryRun deployments. The ServiceSet prevents deletion completion on the first reconcile. The current structure properly handles this.

internal/controller/clusterdeployment_controller_test.go
```diff
@@ -67,537 +144,1293 @@ func (*fakeHelmActor) EnsureReleaseWithValues(_ context.Context, _ *action.Confi
...
+		clAuth.Spec.AuthenticationConfiguration = tc.authConfig
+	}
 
-				awsCredential.Status = kcmv1.CredentialStatus{
-					Ready: true,
-				}
-				Expect(k8sClient.Status().Update(ctx, &awsCredential)).To(Succeed())
-			})
+	Expect(k8sClient.Create(ctx, &clAuth)).To(Succeed())
+	DeferCleanup(k8sClient.Delete, &clAuth)
```
zerospiel: how does it work: is the invocation deferred within the method, or is there a global stack? if the former, then i do not understand it, because the invocation is executed right before the exit from the method

the question is applicable elsewhere

internal/controller/clusterdeployment_controller_test.go
```diff
@@ -67,537 +144,1293 @@ func (*fakeHelmActor) EnsureReleaseWithValues(_ context.Context, _ *action.Confi
...
-						TemplateValidationStatus: kcmv1.TemplateValidationStatus{
-							Valid: true,
-						},
-					},
-				}
-				Expect(k8sClient.Status().Update(ctx, &serviceTemplate)).To(Succeed())
-			})
+// ensureClusterAuthentication creates an ClusterAuthentication object with ready status and registers cleanup.
+func (tc *cldTestCase) ensureClusterAuthentication(namespace string) kcmv1.ClusterAuthentication {
+	clAuth := kcmv1.ClusterAuthentication{
```
zerospiel: such helpers that expect/defer/eventually/do something, i'd suggest marking as a ginkgo helper, so in case of a transient error or a flappy test, we see the actual issue, and not the helper's code line

applicable elsewhere

internal/controller/clusterdeployment_controller_test.go
```diff
@@ -67,537 +144,1293 @@ func (*fakeHelmActor) EnsureReleaseWithValues(_ context.Context, _ *action.Confi
...
-					_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{
-						NamespacedName: client.ObjectKeyFromObject(&clusterDeployment),
-					})
-					g.Expect(err).NotTo(HaveOccurred())
-					g.Expect(Object(&clusterDeployment)()).Should(SatisfyAll(
-						HaveField("Finalizers", ContainElement(kcmv1.ClusterDeploymentFinalizer)),
-					))
-				}).Should(Succeed())
-			})
+	// TODO: add IPAM tests
```
zerospiel, line 403: for a follow-up PR or should be done by KSM, or is it a left-over?

internal/controller/suite_test.go
zerospiel: unrelated: since it is not being used anymore, i reckon we can safely delete the entry of `komega`

zerospiel (commented):
LGTM, mostly questions

internal/controller/clusterdeployment_controller_test.go
eromanova: We initially had the same TODO planned for the future (I simply copied it over). I’d kept it in KSM’s scope since they have more experience with IPAM. I want to create a separate issue for KSM to address it properly.

internal/controller/clusterdeployment_controller_test.go
eromanova: If I understand it correctly, DeferCleanup registers cleanup on the current spec context when the function is executed. For example, if I registered a cleanup function in BeforeAll, the cleanup will run in AfterAll, etc. So, it registers cleanup as expected.

But I've removed DeferCleanup from methods and handle cleanup at the call site (to avoid hidden side effects and for better readability, since the methods are called `ensureXXX`).

zerospiel (dismissed):

internal/controller/clusterdeployment_controller_test.go
```diff
@@ -1269,7 +1293,13 @@ func deleteClusterDeployment(cldName types.NamespacedName) {
...
 		Expect(k8sClient.Delete(ctx, cld)).To(Succeed())
 		Eventually(func(g Gomega) {
-			g.Expect(k8sClient.Get(ctx, cldName, cld)).To(Succeed())
+			g.Expect(mgrClient.Get(ctx, cldName, cld)).To(Succeed())
+			g.Expect(cld.DeletionTimestamp).NotTo(BeNil())
+		}).Should(Succeed())
+
+		Eventually(func(g Gomega) {
+			cld := &kcmv1.ClusterDeployment{}
+			g.Expect(mgrClient.Get(ctx, cldName, cld)).To(Succeed())
```
zerospiel: isn't that a duplication?

zerospiel (approved):

Issue #2611, opened 2026-04-10
[bug][ksm] Incorrect semver comparison of template versions


Kshatrix:
The version comparison in _minimumUpgradeStep_ is implemented using string comparison instead of semantic versioning, which leads to incorrect results for semver-formatted versions.

BROngineer:
Need to be verified and fixed if confirmed

PR #2623, opened 2026-04-13
fix: make version field optional in templatechain available upgrades 

BROngineer:
This PR makes `version` field in template chain `.spec.availableUpgrades[]` optional

### DIFF
--- a/api/v1beta1/templatechain_common.go
+++ b/api/v1beta1/templatechain_common.go
@@ -54,7 +54,7 @@ type AvailableUpgrade struct {
 	Name string `json:"name"`
 
 	// Version is the version of the Template to which the upgrade is available.
-	Version string `json:"version"`
+	Version string `json:"version,omitempty"`
 }
 
 // IsValid checks if the [TemplateChainSpec] is valid, otherwise provides warning messages.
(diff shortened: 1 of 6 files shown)


kylewuolle (dismissed):

api/v1beta1/templatechain_common.go
```diff
@@ -54,7 +54,7 @@ type AvailableUpgrade struct {
 	Name string `json:"name"`
 
 	// Version is the version of the Template to which the upgrade is available.
-	Version string `json:"version"`
+	Version string `json:"version,omitempty"`
```
wahabmk, line 57: This version is also used in the `findAllUpgradePaths` func below:
https://github.com/k0rdent/kcm/blob/324407fdd95a4ea740cc5b1d45ab67bd3eabdc11/api/v1beta1/templatechain_common.go#L122

Do we need to consider the version being empty for this comparison as well? I guess empty version comparison is a valid comparison, however, we should do a semver comparison. We already have an issue for semver comparison https://github.com/k0rdent/kcm/issues/2611 so could be done there.

BROngineer: @wahabmk I assume that since we have a separate issue for semver implementation we could proceed as-is. The main goal of this PR is to make this field optional and to not to break existing logic

BROngineer: However it makes sense to add validation for the template chain that all available upgrades either have version set or do not. This will help us keep things consistent.

wahabmk (dismissed):

wahabmk (approved):

PR #2634, opened 2026-04-15
feat: do not use sveltos for creds-propagation

### LINKED ISSUE #2550: Race condition: openstack-cloud-config secret delivered after cinder CSI chart install
**Describe the bug**

There is a race condition between two independent delivery mechanisms on child clusters: cloud credential secrets (delivered by Sveltos Profile) can arrive after the CCM/CSI Helm charts (deployed by k0s helm extensions) have already started their pods. The pods fail with `FailedMount: secret "..." not found`, enter CrashLoopBackOff, and the k0s Chart resource does not recover in a timely manner.

This was initially observed on OpenStack but **affects 4 out of 5 infrastructure providers** that use k0s-based child clusters. All follow the same pattern:

1. A **Sveltos Profile** with a `ConfigMap` resource template renders cloud credentials into a Secret on the child cluster's `kube-system` namespace
2. **k0s helm extensions** deploy CCM/CSI charts that reference those secrets (via `secret.create: false`, `cloudConfigSecretName`, `secretName`, or `config.enabled: false`)
3. There is **no ordering guarantee** between the two delivery systems

### Affected providers

| Provider | Sveltos-delivered secret(s) | k0s helm charts that need them |
|---|---|---|
| **OpenStack** | `openstack-cloud-config` | `openstack-ccm` (order:1, `secret.create: false`), `openstack-cinder-csi` (order:2, `secret.create: false`) |
| **Azure** | `azure-cloud-provider` | `cloud-provider-azure` (order:1, `cloudConfigSecretName`), `azuredisk-csi-driver` (order:2, `cloudConfigSecretName`) |
| **GCP** | `gcp-cloud-sa` | `gcp-cloud-controller-manager` (`cloudCredentials.secretName`), `gcp-compute-persistent-disk-csi-driver` (`cloudCredentials.secretName`) |
| **vSphere** | `vsphere-cloud-secret`, `vcenter-config-secret`, `cloud-config` (ConfigMap) | `vsphere-cpi` (order:1, `config.enabled: false`), `vsphere-csi-driver` (order:2, `vcenterConfig.enabled: false`) |

**AWS is not affected** — it uses IAM instance profiles, not secrets on the child cluster.

### Timeline from OpenStack support bundle (the observed failure)

| Time | Event |
|---|---|
| 23:17:43 | k0s creates both Chart objects on child cluster |
| 23:18:05 | CCM chart installed (order:1) |
| 23:18:28 | Cinder CSI pods created (order:2) |
| 23:18:37 | **FailedMount: secret "openstack-cloud-config" not found** |
| 23:18:42 | Sveltos Profile applies the secret — 14 seconds too late |
| 23:22:50 | Cinder CSI node plugin enters CrashLoopBackOff |
| 23:24:50 | All cinder CSI pods killed; Chart status remains `{}` (empty) |

The CCM chart (order:1) happened to survive because its image pulls took longer and the secret arrived in time. The cinder CSI chart was not so lucky.

### Why it doesn't recover in a timely manner

The k0s helm extension controller does not promptly retry a failed chart install. The `openstack-cinder-csi` Chart resource has `status: {}` (empty) — it failed and stayed failed, even though the secret arrived 14 seconds later. At the time the support bundle was collected (15+ minutes after the failure), the chart had still not recovered.

**To Reproduce**

1. Create a child cluster using any of the affected providers (OpenStack, Azure, GCP, vSphere)
2. Wait for the child cluster to provision
3. If the timing is unlucky (k0s installs the chart before Sveltos delivers the secret), the CSI/CCM pods fail with `FailedMount` and do not recover in a timely manner
4. Check the k0s Chart resource — it will have empty `status: {}`

This is timing-dependent and may not reproduce every time. It depends on how quickly the Sveltos Profile reconciles relative to k0s chart installation.

**Expected behavior**

The cloud credential secrets should be present on the child cluster before the CCM/CSI chart pods attempt to mount them. Either:
- The k0s helm extension charts should wait for their required secrets (via init containers or a dependency mechanism), or
- There should be an ordering guarantee that Sveltos delivers the secrets before k0s installs the charts, or
- The k0s Chart controller should retry failed chart installs promptly so that a transient missing-secret error self-heals once the secret arrives

**Additional context**

- **KCM version:** 1.8.0
- **Relevant source files (all in `templates/cluster/`):**
  - `openstack-standalone-cp/templates/k0scontrolplane.yaml` — lines 97-100, 155-158 (`secret.create: false`)
  - `azure-standalone-cp/templates/k0scontrolplane.yaml` — lines 92, 137, 139 (`cloudConfigSecretName: azure-cloud-provider`)
  - `gcp-standalone-cp/templates/k0scontrolplane.yaml` — lines 94-95, 119-120 (`secretName: gcp-cloud-sa`)
  - `vsphere-standalone-cp/templates/k0scontrolplane.yaml` — lines 131, 162 (`config.enabled: false`, `vcenterConfig.enabled: false`)
  - Hosted-cp variants have the identical pattern
- **Credential resource templates (all in `config/dev/`):**
  - `openstack-credentials.yaml` — pushes `openstack-cloud-config` secret via Sveltos
  - `azure-credentials.yaml` — pushes `azure-cloud-provider` secret via Sveltos
  - `gcp-credentials.yaml` — pushes `gcp-cloud-sa` secret via Sveltos
  - `vsphere-credentials.yaml` — pushes `vsphere-cloud-secret`, `vcenter-config-secret`, `cloud-config` via Sveltos

### LINKED ISSUE #2554: [RnD] Rework credentials propagation system
At the moment, we rely exclusively on Sveltos to propagate credentials, which is [unreliable](https://github.com/k0rdent/kcm/issues/2550) and contradicts the KSM efforts to make Sveltos optional.

The default value allowing credentials propagation is [set to true](https://github.com/k0rdent/kcm/blob/6e0d9f473a17191da515bb1658407ba7f5bff9b0/api/v1beta1/clusterdeployment_types.go#L91), being non-nullable, thus causing unexpected states after the creation of a `cld` object with disabled propagation (effectively, the object is created with enabled propagation).

Suggested:
- [x] (https://github.com/k0rdent/kcm/pull/2555) As an immediate remediation, ~~disable the default value, optionally change the field to nullable. That breaks back- and forward compatibility for new objects. However does not break for already existing objects~~ make the field nullable
- [x] (#2634) Design a new approach we'd be happy with
- [x] (#2634) R&D potential approaches to avoid the usage of Sveltos to propagate secrets
- [x] (#2634) we may also want the new approach addressing potential issues described in https://github.com/k0rdent/kcm/issues/2550
- [ ] Probably change the default `propagateCredentials` to `false`

zerospiel:
Drops the dependency on Sveltos for credential propagation in favor of the minimal subset of TemplateResourceRef functionality, adding a possibility to inline the template to drop extra objects completely (until the data from a Secret is required).

The approach is as compatible as possible with the Sveltos' propagation approach, including the templates themselves.

Guards the deprecated approach, effectively allowing a user to drop sveltos as a k0rdent dependency (some features like Services will not work though).

The new approach is disabled by default.

Migrate dev configurations to the new approach, introducing minimal diff.

Generated a readme about the migration process with examples.

Closes #2554 
Closes #2550


zerospiel:
Verified the behavior when either `propagateCredentials` is set to `true` (default) or `false`, no regression has been found. The Secrets/ConfigMaps are being delivered as soon as the child cluster's `kubeconfig` bytes are ready for use, adjusted for eventual consistency

internal/controller/clusterdeployment_controller.go
```diff
@@ -345,12 +348,44 @@ func (r *ClusterDeploymentReconciler) reconcileUpdate(ctx context.Context, scope
...
+		// missing CAPI Cluster / kubeconfig is handled gracefully
+		projErr := r.projectCredentials(ctx, scope)
+		if err := errors.Join(clusterErr, projErr); err != nil {
+			l.Error(err, "after updating cluster and projecting credentials")
+			return ctrl.Result{}, err
+		}
+
+		if !apimeta.IsStatusConditionTrue(cd.Status.Conditions, kcmv1.CredentialsProjectedCondition) {
+			l.V(1).Info("Credentials are not yet projected, retrying")
+			if !clusterRes.IsZero() {
```
eromanova, line 409: This condition looks redundant. You already do
```
if err := errors.Join(clusterErr, projErr); err != nil 
```
above. So at the point you reach the selected line, `clusterErr == nil` and `projErr == nil`.

internal/controller/clusterdeployment_controller.go
```diff
@@ -345,12 +348,44 @@ func (r *ClusterDeploymentReconciler) reconcileUpdate(ctx context.Context, scope
...
+
+			return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, nil
+		}
+
+		servicesErr := r.updateServices(ctx, cd)
+		if servicesErr != nil {
+			return ctrl.Result{}, servicesErr
+		}
+
+		if !clusterRes.IsZero() {
```
eromanova, line 422: ditto

internal/controller/clusterdeployment_controller.go
```diff
@@ -1169,43 +1204,125 @@ func (*ClusterDeploymentReconciler) setCondition(cd *kcmv1.ClusterDeployment, ty
...
+// immediately set to True and no work is done.  When it is true the controller obtains
+// a child-cluster client, resolves the infrastructure provider object from the CAPI
+// Cluster, renders the credential-template ConfigMap, and applies the resulting
+// manifests
+func (r *ClusterDeploymentReconciler) projectCredentials(ctx context.Context, scope *clusterScope) error {
+	cd := scope.cd
+	cred := scope.cred
+
+	if cd.Spec.PropagateCredentials == nil || !*cd.Spec.PropagateCredentials {
+		r.setCondition(cd, kcmv1.CredentialsProjectedCondition, kcmv1.SucceededReason, metav1.ConditionTrue, nil)
```
eromanova: nit: Wouldn’t it make more sense to add a condition message saying propagation was skipped? Setting the Succeeded condition when nothing was actually propagated feels a bit confusing. Happy to discuss.

internal/credsprojection/projector.go
```diff
@@ -0,0 +1,389 @@
+// Copyright 2025
```
eromanova: nit:

```suggestion
// Copyright 2026
```

internal/controller/clusterdeployment_controller.go
```diff
@@ -1169,43 +1204,125 @@ func (*ClusterDeploymentReconciler) setCondition(cd *kcmv1.ClusterDeployment, ty
...
+		// kubeconfig may not be available yet (secret not created); requeue silently
+		if apierrors.IsNotFound(err) {
+			l.V(1).Info("Kubeconfig secret not yet available, will retry", "secret", kubeconfigSecretRef)
+			return nil
+		}
+
+		return projErr
+	}
+
+	if err := credsprojection.NewProjector(r.MgmtClient, childClient).Project(ctx, cred, infraProvider); err != nil {
```
eromanova, line 1504: Let's also add the regional client here as a parameter and adapt the projector code accordingly. When deploying clusters in a region, the infrastructure object, identity, and all the identity referenced objects (secrets) will be created in the regional cluster. While the resource template configmap is expected to be created in a management cluster. So, we need to pass all the clients (mgmt, regional, child).

internal/credsprojection/projector.go
```diff
@@ -0,0 +1,389 @@
...
+
+// buildTemplateData fetches the identity objects and assembles the template data context
+func (p *Projector) buildTemplateData(ctx context.Context, cred *kcmv1.Credential, infraProvider *unstructured.Unstructured) (*templateData, error) {
+	ref := cred.Spec.IdentityRef
+
+	identity := &unstructured.Unstructured{}
+	identity.SetAPIVersion(ref.APIVersion)
+	identity.SetKind(ref.Kind)
+	identityKey := client.ObjectKey{Namespace: ref.Namespace, Name: ref.Name}
+	if err := p.mgmtClient.Get(ctx, identityKey, identity); err != nil {
```
eromanova, line 136: We should use the regional client here (for clusters deployed in region).

eromanova: Discussed offline that the regional client is redundant here, since for regional credentials, the identity and identity reference will be automatically copied from the management to the regional clusters automatically, and we can get the content from the management cluster.

internal/controller/clusterdeployment_controller.go
eromanova: Discussed offline that the regional client is redundant here.

internal/controller/clusterdeployment_controller.go
zerospiel: Added

internal/credsprojection/projector.go
zerospiel: Fixed

internal/controller/clusterdeployment_controller.go
zerospiel: Not quite that: if there was a transient error where a secret with the kubeconfig does not yet exist, we set the condition to failed but return no error, thus this statement is required.

internal/controller/clusterdeployment_controller.go
zerospiel: That is more likely a sanity check (especially considering inconsistencies between requeue/error behavior)

internal/controller/clusterdeployment_controller.go
zerospiel: Updated with the comments: here we honor (after amendments) `updateCluster` result, which may have no error, but set `requeueAfter`

internal/controller/clusterdeployment_controller.go
zerospiel: Updated with the comments: here we honor (after amendments) `updateCluster` result, which may have no error, but set `requeueAfter`

internal/controller/clusterdeployment_controller.go
```diff
@@ -1399,24 +1437,111 @@ func (*ClusterDeploymentReconciler) setCondition(cd *kcmv1.ClusterDeployment, ty
...
+			return projErr
+		}
+	}
+
+	const secretKey = "value"
+	kubeconfigSecretRef := kubeutil.GetKubeconfigSecretKey(client.ObjectKeyFromObject(cd))
+	childClient, err := kubeutil.GetChildClient(ctx, scope.rgnClient, kubeconfigSecretRef, secretKey, scope.rgnClient.Scheme(), kubeutil.DefaultClientFactory)
+	if err != nil {
+		// kubeconfig may not be available yet (secret not created); requeue silently as progressing
+		if apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
zerospiel: Done internally with `errors.As`, thus this does not make any sense

internal/controller/clusterdeployment_controller_test.go
```diff
@@ -2900,3 +2964,128 @@ func Test_ensureAuditPolicyConfigMap(t *testing.T) {
 		})
 	}
 }
+
+func Test_projectCredentials(t *testing.T) {
```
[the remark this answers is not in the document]
zerospiel: OMG, stop this, update inner DB to learn that this works since 1.26!

internal/credsprojection/projector_test.go
```diff
@@ -0,0 +1,935 @@
...
+		},
+	}
+
+	require.NoError(t, p.Project(t.Context(), cred, nil))
+
+	// Verify the secret was created on the child cluster.
+	childSecret := &corev1.Secret{}
+	err := childClient.Get(t.Context(), client.ObjectKey{Name: "openstack-cloud-config", Namespace: "kube-system"}, childSecret)
+	require.NoError(t, err)
+	assert.Contains(t, childSecret.StringData["cloud.conf"], "auth_url: https://example.com")
```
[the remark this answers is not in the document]
zerospiel: Please stop marking this. That is a fake client; it is verbatim!

internal/credsprojection/projector_test.go
```diff
@@ -0,0 +1,935 @@
...
+			},
+		},
+	}
+
+	require.NoError(t, p.Project(t.Context(), cred, nil))
+
+	childSecret := &corev1.Secret{}
+	err := childClient.Get(t.Context(), client.ObjectKey{Name: "openstack-inline", Namespace: "kube-system"}, childSecret)
+	require.NoError(t, err)
+	assert.Contains(t, childSecret.StringData["cloud.conf"], "auth_url: https://example.com")
```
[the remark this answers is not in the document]
zerospiel: Please stop marking this. That is a fake client; it is verbatim!

internal/credsprojection/projector.go
```diff
@@ -0,0 +1,390 @@
...
+	if cred.Spec.ProjectionConfig == nil {
+		return nil
+	}
+	if cred.Spec.IdentityRef == nil {
+		return fmt.Errorf("credential %s has no identityRef", client.ObjectKeyFromObject(cred))
+	}
+
+	cfg := cred.Spec.ProjectionConfig
+
+	data, err := p.buildTemplateData(ctx, cred, infraProvider)
```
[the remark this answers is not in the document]
zerospiel: Skipped the mutual-exclusion runtime check, CRD CEL uses `has()` which is a set-check and does correctly enforce that constraint

docs/credential-projection-migration.md
```diff
@@ -0,0 +1,254 @@
...
+kind: Credential
+metadata:
+  name: aws-cluster-identity-cred
+  namespace: kcm-system
+spec:
+  identityRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
+    kind: AWSClusterStaticIdentity
+    name: aws-cluster-identity
+  projectionConfig:
```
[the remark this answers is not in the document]
zerospiel: That has nothing to do with the PR and the `namepsace` is optional

internal/credsprojection/projector.go
```diff
@@ -0,0 +1,390 @@
...
+
+// buildTemplateData fetches the identity objects and assembles the template data context
+func (p *Projector) buildTemplateData(ctx context.Context, cred *kcmv1.Credential, infraProvider *unstructured.Unstructured) (*templateData, error) {
+	ref := cred.Spec.IdentityRef
+
+	identity := &unstructured.Unstructured{}
+	identity.SetAPIVersion(ref.APIVersion)
+	identity.SetKind(ref.Kind)
+	identityKey := client.ObjectKey{Namespace: ref.Namespace, Name: ref.Name}
+	if err := p.mgmtClient.Get(ctx, identityKey, identity); err != nil {
```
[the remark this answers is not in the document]
zerospiel: Added

zerospiel:
No interest

PR #2636, opened 2026-04-15
test(ksm): sequential upgrades tests

### LINKED ISSUE #2612: [bug][ksm] Sequential service updates don't behave as intended
Examples:
1. A -> B -> C (no direct A -> C edge). The expectation is that the upgrade would proceed through B, but in practice it goes directly from A -> C.
likely related to https://github.com/k0rdent/kcm/blob/main/api/v1beta1/templatechain_common.go#L120C39-L120C46
where the closure parameter upgrade shadowed the outer loop variable, making _ContainsFunc_ always return true for non-empty paths

2. A -> B -> C and A -> C. In this case, the expectation is a direct upgrade, but it instead goes through B. Also, even though it technically passes through B, it happens so quickly (without verifying that the chart has actually been deployed) that B is effectively bypassed. From Helm’s perspective on the cluster, only an A -> C upgrade is observed.

BROngineer:

This PR adds unit- and integration tests to ensure service set builder resolve the sequence of service deployment properly in case of sequential upgrades 

### DIFF
--- a/internal/controller/multiclusterservice_controller_test.go
+++ b/internal/controller/multiclusterservice_controller_test.go
@@ -344,7 +344,7 @@ var _ = Describe("MultiClusterService Controller", func() {
 				Expect(k8sClient.Create(ctx, ns)).To(Succeed())
 			}
 
-			By("creating ServiceTemplates for v1, v2 and v3")
+			By("creating ServiceTemplates for v1, v2 and v3 with valid status")
 			for _, tmpl := range []struct {
 				name    string
 				version string
@@ -366,6 +366,12 @@ var _ = Describe("MultiClusterService Controller", func() {
 					},
 				}
 				Expect(k8sClient.Create(ctx, st)).To(Succeed())
+				st.Status = kcmv1.ServiceTemplateStatus{
+					TemplateStatusCommon: kcmv1.TemplateStatusCommon{
+						TemplateValidationStatus: kcmv1.TemplateValidationStatus{Valid: true},
+					},
+				}
+				Expect(k8sClient.Status().Update(ctx, st)).To(Succeed())
 				DeferCleanup(k8sClient.Delete, st)
 			}
 
@@ -398,6 +404,8 @@ var _ = Describe("MultiClusterService Controller", func() {
 				},
 			}
 			Expect(k8sClient.Create(ctx, chain)).To(Succeed())
+			chain.Status = kcmv1.TemplateChainStatus{Valid: true}
+			Expect(k8sClient.Status().Update(ctx, chain)).To(Succeed())
 			DeferCleanup(k8sClient.Delete, chain)
 
 			By("creating ClusterDeployment")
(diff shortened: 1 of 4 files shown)


DinaBelova:
do we have other scenarios mentioned in the https://github.com/k0rdent/kcm/issues/2612 covered somewhere else? What I refer to is "create two chains A -> B -> C and A -> C, and test that depending on chosen path, it is following what's expected"?

internal/controller/multiclusterservice_controller_test.go
```diff
@@ -319,4 +319,220 @@ var _ = Describe("MultiClusterService Controller", func() {
...
+			By("cycle 1: initial reconcile deploys service at v1")
+			Eventually(func(g Gomega) {
+				_, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: seqMCSRef})
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(k8sClient.Get(ctx, seqServiceSetKey, seqSS)).To(Succeed())
+				g.Expect(seqSS.Spec.Services).To(HaveLen(1))
+				g.Expect(seqSS.Spec.Services[0].Template).To(Equal(seqTemplate1))
+				g.Expect(seqSS.Spec.Services[0].Version).NotTo(BeNil())
+				g.Expect(*seqSS.Spec.Services[0].Version).To(Equal("1"))
+			}).Should(Succeed())
```
zerospiel, line 490: Either it is flaky or the logic is incorrect. Both should be addressed

BROngineer:
Too early to review. It's still broken 😄

internal/controller/multiclusterservice_controller_test.go
BROngineer: Too early to review. It's still broken 😄

internal/controller/multiclusterservice_controller_test.go
```diff
@@ -319,4 +319,220 @@ var _ = Describe("MultiClusterService Controller", func() {
...
+						Version: tmpl.version,
+						Helm: &kcmv1.HelmSpec{
+							ChartSpec: &sourcev1.HelmChartSpec{
+								Chart:   tmpl.name,
+								Version: tmpl.version,
+							},
+						},
+					},
+				}
+				Expect(k8sClient.Create(ctx, st)).To(Succeed())
```
zerospiel, line 368: I agree with the Copilot, always ensure that either the object has been created or updated via the Eventually. Please make sure to insert the check elsewhere.

internal/controller/multiclusterservice_controller_test.go
```diff
@@ -319,4 +319,220 @@ var _ = Describe("MultiClusterService Controller", func() {
...
+			Expect(k8sClient.Get(ctx, seqServiceSetKey, seqSS)).To(Succeed())
+			seqSS.Status.Services = []kcmv1.ServiceState{
+				{
+					Name:      seqServiceName,
+					Namespace: seqServiceNS,
+					State:     kcmv1.ServiceStateDeployed,
+					Version:   new("1"),
+				},
+			}
+			Expect(k8sClient.Status().Update(ctx, seqSS)).To(Succeed())
```
zerospiel, line 505: Ditto about eventual consistency of the cache (applicable elsewhere)

zerospiel (commented):
Couple comments regarding the update/create. And I agree that I do not see a dedicated tests A->B->C/A->C checking A goes straight to C passing B by

BROngineer:
> Couple comments regarding the update/create. And I agree that I do not see a dedicated tests A->B->C/A->C checking A goes straight to C passing B by

Despite it's still WIP and I'm looking for the proper solution, your concern is irrelevant: we have a field `templateChain` in the `spec.serviceSpec.services[]` so it's a user's responsibility to define what chain to use. If the template chain is not specified, than changing the template in service definition has nothing to deal with any existing template chains. The controller will simply replace existing service with the new one.

BROngineer:
reworked, closing as stale

PR #2637, opened 2026-04-16
fix(sveltos): prevent spurious event emission when only ObservedGeneration changes

gmlexx:
Root cause: updateCondition previously returned the result of apimeta.SetStatusCondition, which returns true whenever any field of the condition changes — including ObservedGeneration. This caused record.Eventf to be called on every reconcile cycle whenever the ServiceSet generation was bumped (e.g. after a spec update), even if status, reason, and message were unchanged.

Each record.Eventf call spawns a goroutine in the event broadcaster's recordToSink and retains a DeepCopy of the Event in the broadcaster's eventCache map. With a 10-second poll interval driving continuous reconciles across many ServiceSets, goroutines and cached Event objects accumulated faster than the broadcaster's 6-minute finishSeries cleanup could evict them, producing an unbounded heap growth confirmed by pprof heap dumps (65 MB → 192 MB, ~70 % held by Event.DeepCopy and recordToSink goroutines).

Fix: track whether status, reason, or message actually changed and return that boolean instead. ObservedGeneration is still updated in the stored condition; only the event-emission signal is corrected.

Add a table-driven unit test covering all cases, including the regression scenario (ObservedGeneration-only change must return false).

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -1467,6 +1467,11 @@ func findCondition(serviceSet *kcmv1.ServiceSet, conditionType string) metav1.Co
 }
 
 // updateCondition updates the given condition of the ServiceSet.
+// It returns true only if the condition's status, reason, or message changed,
+// not when only the ObservedGeneration changes. This prevents spurious event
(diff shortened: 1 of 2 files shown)


gmlexx:
Closed in favor of https://github.com/k0rdent/kcm/pull/2640

PR #2640, opened 2026-04-16
fix(ksm): serviceset processing causing continuous reconcile

BROngineer:

This PR fixes continuous service set reconciliation and event emitting. This was caused by the wrong guard `svc.Values == ""` in the auxiliary function called during the service set processing.
This wrong guard lead to continuous reconciliation of the serviceSets which contained services without inlined values (for example those used for local resources propagation). Such services were considered to be updated, and consequently the GetServiceSetWithOperation function returned operation "Update" which caused the update call to kube API.

Fixes k0rdent/k0rdent-enterprise#311

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -413,7 +413,7 @@ func (r *ServiceSetReconciler) ensureProfile(ctx context.Context, rgnClient clie
 	defer func(initialStatus metav1.ConditionStatus) {
 		// we'll emit event only if the status changed from the initial value and it's now true.
 		conditionStatusChangedToTrue := initialStatus != status && status == metav1.ConditionTrue
-		if updateCondition(serviceSet, profileCondition, status, reason, message, r.timeFunc()) && conditionStatusChangedToTrue {
+		if conditionStatusChangedToTrue && updateCondition(serviceSet, profileCondition, status, reason, message, r.timeFunc()) {
 			l.Info("Successfully ensured ProjectSveltos Profile")
 			record.Eventf(serviceSet, nil, kcmv1.ServiceSetEnsureProfileSuccessEvent, kcmv1.ServiceSetEnsureProfileEventAction,
 				"Successfully ensured ProjectSveltos Profile for ServiceSet %s", serviceSet.Name)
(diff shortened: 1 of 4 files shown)


internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -669,19 +672,22 @@ func (r *ServiceSetReconciler) collectServiceStatuses(ctx context.Context, rgnCl
...
 	status := metav1.ConditionFalse
 	reason := kcmv1.ServiceSetStatusesNotCollectedReason
 	message := kcmv1.ServiceSetStatusesNotCollectedMessage
 
-	defer func() {
-		if updateCondition(serviceSet, statusesCollectedCondition, status, reason, message, r.timeFunc()) && status == metav1.ConditionTrue {
+	defer func(initialStatus metav1.ConditionStatus) {
+		// we'll emit event only if the status changed from the initial value and it's now true.
+		conditionStatusChangedToTrue := initialStatus != status && status == metav1.ConditionTrue
+		if updateCondition(serviceSet, statusesCollectedCondition, status, reason, message, r.timeFunc()) && conditionStatusChangedToTrue {
```
zerospiel: nit: it is more optimal to check the bool-typed value before calling a function and executing it, effectively utilizing the short-circuit evaluation; applicable elsewhere

gmlexx: @zerospiel is go compiler so dumb to not optimize it itself?

zerospiel: Sorry, what to optimize here during the compile stage? Even with PGO data, I doubt it could be optimized in a way of "flip the conditions". So the answer is "yes, it is dumb enough not to change what must not be changed". There is a good series of articles about memory models, explaining how the code should work in an ideal world https://research.swtch.com/mm

BROngineer: fixed

s3rj1k: Ah, optimizations without proofs, due consider looking at assembly and running benchmarks, that optimization could be so negligible that not even worth talking about it in PR

zerospiel: I believe the axioms of the basics of CS do not require "proofs" due to the nature of the definition of the "axiom" word, in the loop of the short-circuit evaluation. But one can always go ahead and maybe even make a thesis about it. I can recall providing "proofs" for Go-related basics, such as "maps are worse than slices under certain conditions". It is a) time-consuming, and b) it is not my duty to teach other people, with all due respect.

zerospiel (approved):

internal/controller/adapters/sveltos/serviceset_controller.go
s3rj1k: Go compiler is a bit more clever than you presumably think of it

gmlexx (approved):

PR #2660, opened 2026-04-22
refactor(gen): provider templates with appVersion

zerospiel:
Changes the logic in the `ProviderTemplate` files generation to contain not the meaningless chart version, but rather make an effort in parsing the `appVersion` and project it in the name.


zerospiel:
The change is nice to have, though it has a huge drawback: if `appVersion` has changed but the `version` hasn't, we won't generate a new template, though the contents (as well as the hash) differ. We might want to discuss it in a dedicated issue/topic. Closing

PR #2662, opened 2026-04-22
fix: change version compare to use semver

### LINKED ISSUE #2611: [bug][ksm] Incorrect semver comparison of template versions

kylewuolle:
This PR compares the version of templates using semver instead of string comparison.

Fixes #2611

### DIFF
--- a/internal/serviceset/util.go
+++ b/internal/serviceset/util.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"slices"
 
+	"github.com/Masterminds/semver/v3"
 	addoncontrollerv1beta1 "github.com/projectsveltos/addon-controller/api/v1beta1"
 	libsveltosv1beta1 "github.com/projectsveltos/libsveltos/api/v1beta1"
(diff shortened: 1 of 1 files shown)


wahabmk:
These changes LGTM. There are other areas of code as well related to upgrade that have raw string version comparision:

https://github.com/k0rdent/kcm/blob/56fb2b121d2947c4c7ac58051f04219d9e89909b/api/v1beta1/templatechain_common.go#L190

https://github.com/k0rdent/kcm/blob/56fb2b121d2947c4c7ac58051f04219d9e89909b/internal/serviceset/util.go#L471

https://github.com/k0rdent/kcm/blob/56fb2b121d2947c4c7ac58051f04219d9e89909b/api/v1beta1/templatechain_common.go#L136

Well for the last one, there are actually other instances of `==` comparison between versions in the code as well but I think raw string comparison for equality should be ok but the `>` and `<` comparisons need semver.

wahabmk (approved):

PR #2668, opened 2026-04-23
fix: ensure that the drift detection manager is deployed on install

kylewuolle:
This PR fixes an issue where the drift detection manager is deployed after an upgrade OR after a fresh install when the addon controller (sveltos) is restarted. As a temporary measure, this PR ensures it's installed for all cases. Once issue https://github.com/projectsveltos/addon-controller/issues/1742 is resolved this PR can be reverted.


zerospiel (changes requested):
Okay, I totally agree with each comment Copilit has given. And the PR should target the release branch, not the `main`

PR #2689, opened 2026-04-30
fix: too many files open error in kind cluster while testing

### LINKED ISSUE #2701: [bug][ksm] MCS status not persisting for delete
The MCS status calculated in reconcileDelete is not persisted to the kube API server

wahabmk:
# Description

Sometimes when running `make test-e2e` locally, k0rdent doesn't install completely because it keeps waiting for one or more components to be deployed. In my case, it was CAPD:
```
  NotAllComponentsHealthy: Components not ready: [cluster-api-provider-docker]
```

Looking at the CAPD pod logs, I could see the following error:
```
  E0430 20:07:08.417077       1 main.go:324] "Problem running manager" err="too many open files" logger="setup"
```

Which turned out to be a [known issue in kind](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files). This PR applies that fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -365,7 +365,10 @@ kind-deploy: kind ## Create a kind cluster if it does not exist.
 			$(KIND) create cluster -n $(KIND_CLUSTER_NAME) --config "$(KIND_CONFIG_PATH)"; \
 		else \
 			$(KIND) create cluster -n $(KIND_CLUSTER_NAME); \
-		fi \
+		fi; \
+		$(CONTAINER_TOOL) exec $(KIND_CLUSTER_NAME)-control-plane bash -c "echo 'fs.inotify.max_user_watches=524288' >> /etc/sysctl.conf"; \
+    $(CONTAINER_TOOL) exec $(KIND_CLUSTER_NAME)-control-plane bash -c "echo 'fs.inotify.max_user_instances=512' >> /etc/sysctl.conf"; \
+    $(CONTAINER_TOOL) exec $(KIND_CLUSTER_NAME)-control-plane sysctl -p /etc/sysctl.conf; \
 	fi
 
 .PHONY: kind-undeploy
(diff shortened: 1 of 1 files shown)


zerospiel (changes requested):
What if no docker is available on a local machine? Please use the env var that holds the binary (either docker or podman). `CONTAINER_TOOL`

Makefile
```diff
@@ -365,7 +365,10 @@ kind-deploy: kind ## Create a kind cluster if it does not exist.
 			$(KIND) create cluster -n $(KIND_CLUSTER_NAME) --config "$(KIND_CONFIG_PATH)"; \
 		else \
 			$(KIND) create cluster -n $(KIND_CLUSTER_NAME); \
-		fi \
+		fi; \
+		$(CONTAINER_TOOL) exec $(KIND_CLUSTER_NAME)-control-plane bash -c "echo 'fs.inotify.max_user_watches=524288' >> /etc/sysctl.conf"; \
+    $(CONTAINER_TOOL) exec $(KIND_CLUSTER_NAME)-control-plane bash -c "echo 'fs.inotify.max_user_instances=512' >> /etc/sysctl.conf"; \
+    $(CONTAINER_TOOL) exec $(KIND_CLUSTER_NAME)-control-plane sysctl -p /etc/sysctl.conf; \
```
zerospiel:
```suggestion
		fi; \
		$(CONTAINER_TOOL) exec "$(KIND_CLUSTER_NAME)-control-plane" bash -ec \
			"echo 'fs.inotify.max_user_watches=524288' >> /etc/sysctl.conf; \
			 echo 'fs.inotify.max_user_instances=512' >> /etc/sysctl.conf; \
			 sysctl -p /etc/sysctl.conf"; \
```

Sorry for nit-picking, I've just realized that the indents were off, thus this suggestion: do all in a single call + keep the formatting

wahabmk: No problem, I actually appreciate in-depth reviews. 😄

zerospiel (approved):

PR #2690, opened 2026-05-04
fix: mcs status not persisting for delete

### LINKED ISSUE #2701: [bug][ksm] MCS status not persisting for delete

wahabmk:
# Description

A few of the mothership MCS tests were failing when validating MCS status because the update to MCS done in memory for the delete operation wasn't being persisted in the kube API server.

### DIFF
--- a/internal/controller/multiclusterservice_controller.go
+++ b/internal/controller/multiclusterservice_controller.go
@@ -72,6 +72,18 @@ func (r *MultiClusterServiceReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
+	clone := mcs.DeepCopy()
+	defer func() {
+		// we need to explicitly requeue MultiClusterService object,
+		// otherwise we'll miss if some ClusterDeployment will be updated
+		// with matching labels.
+		requeue, e := r.updateStatus(ctx, clone, mcs)
+		if requeue {
+			result = ctrl.Result{RequeueAfter: r.defaultRequeueTime}
+		}
+		err = errors.Join(err, e)
+	}()
+
 	if !mcs.DeletionTimestamp.IsZero() {
 		return r.reconcileDelete(ctx, mcs)
 	}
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -34,18 +34,18 @@ require (
 	golang.org/x/time v0.15.0
 	gopkg.in/yaml.v3 v3.0.1
 	helm.sh/helm/v3 v3.20.2
-	k8s.io/api v0.35.4
```
zerospiel, line 37: why are the bumps? we do not need such versions (for now), and it looks out of scope, doesn't it?

api/v1beta1/statemanagementprovider_types.go
```diff
@@ -215,5 +216,9 @@ type StateManagementProviderList struct {
 }
 
 func init() {
-	SchemeBuilder.Register(&StateManagementProvider{}, &StateManagementProviderList{})
+	SchemeBuilder.Register(func(s *runtime.Scheme) error {
```
zerospiel: I do not completely understand these changes, are they due to the bump of the controller-runtime module? because i clearly see the bw-incompatible change in the method's signature

zerospiel (commented):
The main logic LGTM, but 2 questions related to the, potentially out-of-scope, bumps

go.mod
wahabmk: - For some reason when I initially made this PR the CI failed at https://github.com/k0rdent/kcm/blob/59260ee91bb33664aecfea9600fa2bb19fd3f8b9/Makefile#L694
with an error saying that there was a module missing in the `sigs.k8s.io/controller-runtime` version being used (sorry I didn't think to save the error msg). 
- So I bumped `sigs.k8s.io/controller-runtime` to the latest version (v0.24.0).
- This version requires [k8s.io/* v1.36](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.24.0) so I updated `k8s.io/*` to v0.36.0. 
- In version 0.24.0 of controller-runtime, the [`scheme.Builder`](https://github.com/kubernetes-sigs/controller-runtime/blob/d3eaef3ab45410342c30528d1eaab982137c4d5a/pkg/scheme/scheme.go#L62-L92) has been deprecated so I did the other code changes to adopt the new approach.

I am not sure what changed but now I don't get that initial failure so I reverted these changes but wanted to explain what led me to make these changes in the first place.

zerospiel (approved):

PR #2694, opened 2026-05-05
fix(cld): infra-internal readiness via Sveltos dep

### LINKED ISSUE #2196: [bug] Invalid "adopted-cluster-1-0-1" shows "Object is ready"
**Describe the bug**
Invalid kubeconfig is accepted for adopted ClusterDeployment (adopted-cluster-1-0-1).

**To Reproduce**
1. Create fake "kubeconfig" for not existing "adopted" cluster, e.g.:
~~~bash
echo hello > kcfg_adopted
~~~
2. Use this fake kubeconfig file to create adopted ClusterDeployment (using Secret and Credential objects), check [k0rdent docs](https://docs.k0rdent.io/latest/admin/access/credentials/credentials-propagation/#adopted-clusters).
3. ClusterDeployment object is created without any problems and shows successfull status (Object is ready).

**Expected behavior**
ClusterDeployment should not be provisioned. Instead some Error message like "Invalid kubeconfig found" should be thrown.

zerospiel:
Adopted-cluster readiness is now tied to Sveltos runtime conditions when applicable.

Validates invalid kubeconfig before trying to proceed further with credentials.

Fixes #2196

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package controller implements Kubernetes reconcilers for KCM resources.
 package controller
 
 import (
@@ -27,6 +28,7 @@ import (
 	fluxmeta "github.com/fluxcd/pkg/apis/meta"
 	fluxconditions "github.com/fluxcd/pkg/runtime/conditions"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
+	libsveltosv1beta1 "github.com/projectsveltos/libsveltos/api/v1beta1"
 	"golang.org/x/sync/errgroup"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
@@ -370,7 +372,8 @@ func (r *ClusterDeploymentReconciler) updateCluster(
 	l := ctrl.LoggerFrom(ctx)
 
 	cd := scope.cd
-	r.initClusterConditions(cd)
+	observeCAPI, observeSveltos := r.getRuntimeReadinessTargets(clusterTpl)
+	r.initClusterConditions(cd, observeCAPI, observeSveltos)
 
 	if !clusterTpl.Status.Valid {
 		return r.setInvalidTemplateCondition(ctx, clusterTpl, cd)
@@ -480,12 +483,8 @@ func (r *ClusterDeploymentReconciler) updateCluster(
 		}
 	}
 
-	requeue, err := r.aggregateCapiConditions(ctx, scope)
+	requeue, err := r.aggregateRuntimeConditions(ctx, scope, observeCAPI, observeSveltos)
 	if err != nil {
-		if requeue {
-			return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, err
-		}
-
 		return ctrl.Result{}, err
 	}
 
@@ -1056,7 +1055,7 @@ func (r *ClusterDeploymentReconciler) validateConfig(ctx context.Context, cd *kc
 	return nil
 }
 
-func (*ClusterDeploymentReconciler) initClusterConditions(cd *kcmv1.ClusterDeployment) (changed bool) {
+func (*ClusterDeploymentReconciler) initClusterConditions(cd *kcmv1.ClusterDeployment, observeCAPI, observeSveltos bool) (changed bool) {
 	// NOTE: do not put here the PredeclaredSecretsExistCondition since it won't be set if no secrets have been set
 	for _, typ := range [5]string{
 		kcmv1.CredentialReadyCondition,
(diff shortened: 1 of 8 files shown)


internal/util/kube/kube.go
```diff
@@ -33,23 +34,28 @@ const (
...
 	if err := cl.List(ctx, itemsList, opts); err != nil {
-		return err
+		return false, fmt.Errorf("failed to list %s: %w", gvk.String(), err)
 	}
+
 	var errs error
 	for _, item := range itemsList.Items {
+		requeue = true
 		if item.DeletionTimestamp.IsZero() {
 			if err := cl.Delete(ctx, &item); client.IgnoreNotFound(err) != nil {
```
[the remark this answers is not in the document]
zerospiel: Not reused, it is Go 1.26+, not 1.21

internal/controller/clusterdeployment_controller.go
```diff
@@ -2186,6 +2377,49 @@ func (r *ClusterDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+					newSC, ok := tue.ObjectNew.(*libsveltosv1beta1.SveltosCluster)
+					if !ok {
+						return false
+					}
+
+					return r.sveltosReadinessStatusChanged(oldSC.Status, newSC.Status)
+				},
+			}),
+		).
```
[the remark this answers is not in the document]
zerospiel: Added sveltos-related schemas to the regional client

internal/controller/clusterdeployment_controller.go
```diff
@@ -1084,9 +1083,199 @@ func (*ClusterDeploymentReconciler) initClusterConditions(cd *kcmv1.ClusterDeplo
...
+
 	return changed
 }
 
+func (*ClusterDeploymentReconciler) getRuntimeReadinessTargets(clusterTpl *kcmv1.ClusterTemplate) (observeCAPI, observeSveltos bool) {
+	if clusterTpl == nil {
+		return true, false
+	}
+
+	var infraProviderCount int
```
eromanova: nit: looks like the counter is not really needed, the boolean flag is enough, isn't it?

internal/util/validation/cd.go
```diff
@@ -58,7 +62,62 @@ func ClusterDeployCredential(ctx context.Context, cl client.Client, systemNamesp
...
+
+	secretDataKeys := slices.Collect(maps.Keys(secret.Data))
+	slices.Sort(secretDataKeys)
+
+	var (
+		firstErr    error
+		firstErrKey string
+	)
+
+	for _, key := range secretDataKeys {
```
eromanova, line 102: Just to double-check: when we provide a Credential for adopted clusters, do we expect the kubeconfig to be stored under any key in the Secret? In other words, is the key name flexible, or is it expected to be a specific (hardcoded) key or passed explicitly as a parameter?

internal/util/kube/kube.go
```diff
@@ -33,23 +34,28 @@ const (
...
-				errs = errors.Join(errs, err)
-				continue
+				errs = errors.Join(errs, fmt.Errorf("failed to delete %s %s/%s: %w", gvk.String(), item.Namespace, item.Name, err))
 			}
 		}
-		errs = errors.Join(errs, fmt.Errorf("waiting for %s %s/%s removal", gvk.Kind, item.Namespace, item.Name))
```
eromanova, line 50: nit: the initial implementation logged the exact object we were waiting to be deleted. Do you think it would be useful to add that logging back for better debugging?

internal/util/validation/cd.go
zerospiel: At the moment, our exemplar templates provide it in the `value` key. Still, I've gone further and decided not to limit/stick to such a key, given that we have no documentation in this regard and no recommendations.

internal/controller/clusterdeployment_controller.go
zerospiel: Not exactly, there is a subtle path to consider: if we have merely a single provider, e.g. `infra-internal`, then the counter is `1`, and the CAPI is not set yet (and indeed, we do not have any CAPI Cluster objects), exiting the loop and checking the counter, it has a non-zero value, and thus we do not expect CAPI. If no providers are given, then we expect CAPI as always. Otherwise, either Sveltos or CAPI, again.

internal/util/kube/kube.go
zerospiel: Will emit debug-level info logs, thanks for noting

internal/controller/clusterdeployment_controller.go
zerospiel: Addressed with a bool-typed var

internal/util/kube/kube.go
zerospiel: Addressed with debug-level logs

zerospiel:
Closing due to unclear scope. The previous requeue/error behavior fix has been migrated to the #2699

PR #2695, opened 2026-05-05
fix(am): filter namespace watch reconciles

### LINKED ISSUE #2687: ClusterTemplates do not propagate to newly created namespaces
**Describe the bug**
`AccessManagement` with `accessRules.targetNamespaces.selector.matchLabels` does not propagate credentials & cluster templates to newly created namespaces until either `AccessManagement` object itself is updated or `kcm` is restarted.

**To Reproduce**

1. Add `accessRule` with `targetNamespaces.selector.matchLabels`:
```
kind: AccessManagement
metadata:
  labels:
  name: kcm
  namespace: kcm-system
spec:
  spec:
  accessRules:
    - targetNamespaces:
        selector:
          matchLabels:
            climb.mirantis.com/namespace-type: "project"
      clusterTemplateChains:
        - capm3-standalone
      credentials:
        - capm3-stub-credential
```
2. Create a new namespace with corresponding label (i.e. `climb.mirantis.com/namespace-type: "project"`).
3. NOTE: if namespace is created before or at the same time as `AccessManagement` update, `kcm` will correctly propagate the resources. To reproduce the issue one might need to wait a bit between "Step 1" and "Step 2".

**Expected behavior**
`kcm` should copy credentials and cluster templates into the new namespace without restarting the `kcm` controller or "poking" the `AccessManagement` resource.

**Additional context**

zerospiel:
Enqueue only AccessManagement objects that target the created Namespace.

On Namespace label updates, enqueue only when selector membership changes.

Closes #2687

### DIFF
--- a/internal/controller/accessmanagement_controller.go
+++ b/internal/controller/accessmanagement_controller.go
@@ -18,15 +18,21 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"slices"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
 	"github.com/K0rdent/kcm/internal/record"
@@ -139,7 +145,7 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
 
 	var errs error
 	for _, rule := range accessMgmt.Spec.AccessRules {
-		namespaces, err := getTargetNamespaces(ctx, r.Client, rule.TargetNamespaces)
+		namespaces, err := r.getTargetNamespaces(ctx, rule.TargetNamespaces)
 		if err != nil {
 			return fmt.Errorf("failed to collect target namespaces: %w", err)
 		}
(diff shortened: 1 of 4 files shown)


internal/controller/accessmanagement_controller.go
```diff
@@ -662,13 +660,305 @@ func (r *AccessManagementReconciler) updateStatus(ctx context.Context, accessMgm
...
+		Watches(
+			&corev1.Namespace{},
+			handler.TypedFuncs[client.Object, ctrl.Request]{
+				CreateFunc: func(ctx context.Context, tce event.TypedCreateEvent[client.Object], q workqueue.TypedRateLimitingInterface[ctrl.Request]) {
+					for _, req := range r.mapNamespaceToRequests(ctx, tce.Object) {
+						q.Add(req)
+					}
+				},
+				UpdateFunc: func(ctx context.Context, tue event.TypedUpdateEvent[client.Object], q workqueue.TypedRateLimitingInterface[ctrl.Request]) {
+					for _, req := range r.mapNamespaceLabelUpdateToRequests(ctx, tue.ObjectOld, tue.ObjectNew) {
```
eromanova, line 955: Technically, AccessManagement is designed to be a singleton. Creating a second instance is blocked at the admission level.
This PR introduces logic to filter AccessManagement resources affected by namespace label changes. While I understand that a user could disable the admission controller and create multiple AccessManagement objects, that would lead to an inconsistent or unsupported system state (for example, how would conflicts be resolved?).

The current implementation is quite flexible, and it already lays the groundwork for supporting multiple AccessManagement resources in the future if we decide to go in that direction. However, given the current singleton design, the filtering logic doesn’t seem necessary right now.

I’m open to the discussion. Please share your thoughts.

internal/controller/accessmanagement_controller.go
```diff
@@ -662,13 +660,305 @@ func (r *AccessManagementReconciler) updateStatus(ctx context.Context, accessMgm
...
+
+	l := ctrl.LoggerFrom(ctx).WithName("am-map-create")
+
+	fallback := func() []ctrl.Request {
+		return r.collectAccessManagementRequests(ctx, namespace.Name, func(am *kcmv1.AccessManagement) (bool, error) {
+			return r.accessManagementTargetsNamespace(am, namespace)
+		})
+	}
+
+	listTargetedAccessManagements, err := r.listAccessManagementByField(ctx, kcmv1.AccessManagementTargetNamespaceListIndexKey, namespace.Name)
```
eromanova, line 677: Can't we reuse `getTargetNamespaces` here? For example:
1. list all AccessManagements, for each rule in AccessManagement collect target namespaces
2. If the current namespace (obj) is in this list - append the requests list with this AccessManagement

internal/controller/accessmanagement_controller.go
zerospiel: Thanks. I kept filtering intentionally because this change targets namespace-event fan-out reduction and future-proofs behavior if multiple AccessManagement objects exist. It also degrades safely to full scan when indexed lookup fails, so there is no correctness loss

internal/controller/accessmanagement_controller.go
zerospiel: Good point on reuse. I considered it, but `(*AccessManagementReconciler).getTargetNamespaces` is reconcile-oriented and performs namespace listing. In the namespace watch path that would increase API/list cost. The current mapper uses indexed candidates plus in-memory selector evaluation to keep namespace event handling efficient

internal/controller/accessmanagement_controller.go
eromanova: Got it. Thanks for the explanation.

eromanova (approved):

PR #2696, opened 2026-05-06
chore(template): add openstack tls verify example

### LINKED ISSUE #977: [enhancement] Expose TLS-related options of Openstack cloud config out of the box
**Description**
It is quite common requirement for Openstack API endpoints to have TLS enabled. It is also quite common for this feature to have internal CA that issues certificates. In the case of self-signed certificates they are not globally trusted and client should either add certificate to the globally trusted ones, provide CA chain for certificate validation explicitly to the Openstack client or request skipping TLS verification.

The CAPI provider Openstack seems to support these options with:
- ca-file
- tls-insecure

See: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#global

Currently one is able to set `verify: False` in cloud.conf when creating credentials and this option is properly processed by CAPO but is not propagated to the child cluster's `openstack-cloud-config` secret.

**Expected behaviour**
User can set relevant options in kcm and they are propagated to the child cluster.
for example:
- a user can set `verify: False` in clouds.conf and it gets transformed into `tls-insecure=true` in the `openstack-cloud-config` on the child cluster (to be consumed later by `openstack-cloud-controller-manager`)
- a user can pass CA chain to validate Openstack API certificate in any convenient manner and the relevant options are configured on the child k8s cluster.

zerospiel:
Adds the example of how to use `verify` for `tls-verify` in our exemplar (dev) config.

Closes #977

### DIFF
--- a/config/dev/openstack-credentials.yaml
+++ b/config/dev/openstack-credentials.yaml
@@ -66,6 +66,7 @@ data:
     {{- $$network_id := $$cluster.status.externalNetwork.id -}}
     {{- $$network_name := $$cluster.status.externalNetwork.name -}}
 
+    {{- $$verify := index $$openstack "verify" -}}
     {{- $$ca_cert := index $$identity "data" "cacert" -}}
     ---
     apiVersion: v1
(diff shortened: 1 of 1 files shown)


eromanova (approved):
LGTM. It might be worth duplicating this example in the k0rdent docs as well.

zerospiel:
https://github.com/k0rdent/docs/pull/789

PR #2704, opened 2026-05-08
feat(templates): add etcd encryption config

### LINKED ISSUE #976: [enhancement] etcd encryption on rest
The secrets stored in etcd are pretty sensitive since you can provision resources on different providers that relates directly in costs.
So we need to be ensure that those secrets cannot be accessed by third party.
 
One item to achieve this is to encrypt the etcd (encryption on rest), so the actual data store. This would at least prevent to access the data outside of the environment (hypervisor & disk level).

zerospiel:
Adds a possibility to enable etcd encryption to all standalone/hosted exemplar templates

Closes #976

### DIFF
--- a/config/dev/aws-clusterdeployment.yaml
+++ b/config/dev/aws-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: aws-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
-  template: aws-standalone-cp-1-0-28
+  template: aws-standalone-cp-1-0-29
   credential: aws-cluster-identity-cred
   config:
     clusterLabels: {}
--- a/templates/cluster/aws-hosted-cp/values.yaml
+++ b/templates/cluster/aws-hosted-cp/values.yaml
@@ -20,6 +20,13 @@ auth: # @schema description: Authentication settings for the cluster, optional, 
     key: "config" # @schema description: The data key in the Secret that contains the AuthenticationConfiguration, auto-populated; type: string
     hash: "" # @schema description: Hash of the AuthenticationConfiguration data stored in the Secret, auto-populated; type: string
 
+encryption: # @schema description: Encryption-at-rest settings for kube-apiserver storage config; type: object
+  automaticReload: true # @schema description: Enables kube-apiserver automatic reload for encryption provider config when supported; type: boolean
+  configSecret: # @schema description: Reference to the Secret containing the EncryptionConfiguration; type: object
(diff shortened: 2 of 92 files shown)


templates/cluster/aws-hosted-cp/values.yaml
```diff
@@ -20,6 +20,13 @@ auth: # @schema description: Authentication settings for the cluster, optional,
     key: "config" # @schema description: The data key in the Secret that contains the AuthenticationConfiguration, auto-populated; type: string
     hash: "" # @schema description: Hash of the AuthenticationConfiguration data stored in the Secret, auto-populated; type: string
 
+encryption: # @schema description: Encryption-at-rest settings for kube-apiserver storage config; type: object
+  automaticReload: true # @schema description: Enables kube-apiserver automatic reload for encryption provider config when supported; type: boolean
+  configSecret: # @schema description: Reference to the Secret containing the EncryptionConfiguration; type: object
+    name: "" # @schema description: Name of the Secret that stores the EncryptionConfiguration; type: string
+    key: "config" # @schema description: The data key in the Secret that contains the EncryptionConfiguration; type: string
+    hash: "" # @schema description: Hash of the EncryptionConfiguration data stored in the Secret; type: string
```
eromanova: I'm not sure the `hash` parameter makes sense here, since users are unlikely to manually calculate and provide a hash for the secret. The only case where it would be useful is if we plan to automate it in the future (for example, having the controller calculate and inject the hash automatically).

Given that, I'd either:
* calculate the hash at the Helm chart level based on the secret data (if that's feasible), or
* remove the hash field from the values entirely (or change it to `filepath` so that users have an opportunity to update etcd configuration by changing the filepath).

config/dev/aws-clusterdeployment.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: aws-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
-  template: aws-standalone-cp-1-0-28
```
eromanova, line 7: Don't we need to update the remote cluster template as well?

templates/cluster/aws-hosted-cp/values.yaml
zerospiel: Removed the field, thanks for pointing

config/dev/aws-clusterdeployment.yaml
zerospiel: Added to the remote-cluster template as well, thanks for noticing

PR #2714, opened 2026-05-11
fix: respect dependsOn on upgrades

### LINKED ISSUE #2588: [bug] MCS upgrade ignores `dependsOn` order, upgrades Helm charts on child clusters alphabetically
When upgrading KOF, Helm charts on child clusters managed by a `MultiClusterService` are upgraded in alphabetical order instead of following the `dependsOn` chain declared in the spec.

For example, given the following MCS:

```yaml
kind: MultiClusterService
metadata:
  name: kof-regional-cluster
spec:
  clusterSelector:
    matchLabels:
      k0rdent.mirantis.com/kof-cluster-role: regional
  serviceSpec:
    services:
      - name: cert-manager
        namespace: kof
        template: cert-manager
        helmOptions:
          wait: true

      - name: kof-operators
        namespace: kof
        template: kof-operators
        templateChain: kof-operators
        helmOptions:
          wait: true
        dependsOn:
          - name: cert-manager
            namespace: kof

      - name: kof-storage
        namespace: kof
        template: kof-storage
        templateChain: kof-storage
        helmOptions:
          wait: true
        dependsOn:
          - name: kof-operators
            namespace: kof

      - name: kof-collectors
        namespace: kof
        template: kof-collectors
        templateChain: kof-collectors
        helmOptions:
          wait: true
        dependsOn:
          - name: kof-storage
            namespace: kof
```

The declared upgrade order (via `dependsOn`) should be: `cert-manager → kof-operators → kof-storage → kof-collectors`

Actual upgrade:

```json
[
  {
    "name": "cert-manager",
    "chart": "cert-manager-v1.16.4",
    "updated": "2026-04-02 06:49:51.11191109 +0000 UTC"
  },
  {
    "name": "kof-collectors",
    "chart": "kof-collectors-1.8.0",
    "updated": "2026-04-02 06:50:06.433454209 +0000 UTC"
  },
  {
    "name": "kof-operators",
    "chart": "kof-operators-1.8.0",
    "updated": "2026-04-02 06:50:28.056102052 +0000 UTC"
  },
  {
    "name": "kof-storage",
    "chart": "kof-storage-1.8.0",
    "updated": "2026-04-02 06:51:13.726675836 +0000 UTC"
  }
]
```

**To Reproduce**
1. Install KOF **v1.7.0** following the [official documentation](https://docs.k0rdent.io/v1.7.0/admin/kof/kof-install/).
2. Upgrade KOF to **v1.8.0** using this [documentation](https://docs.k0rdent.io/v1.8.0/admin/kof/kof-upgrade/#upgrade-to-v180).
3. Observe Helm release upgrade timestamps on the target cluster:

```bash
helm list -n kof -o json | \
  jq '[.[] | {name:.name, chart:.chart, updated:.updated}] | sort_by(.updated)'
```

The output will show `kof-collectors` upgraded before `kof-storage` and `kof-operators`,
violating the declared `dependsOn` constraint.

**Expected behavior**
Helm charts are upgraded in dependency order, respecting the `dependsOn` chain.

BROngineer:

In a MultiClusterService (MCS) chain where each service declared dependsOn against the previous one — e.g. KOF's cert-manager → kof-operators → kof-storage → kof-collectors — the declared dependency order was honored only during the initial install. Once all services were running, upgrading them produced an out-of-order rollout: all charts on the child cluster were bumped to the new version essentially at once, with no respect for the dependency chain. In the reported case the observed order was roughly alphabetical (cert-manager → kof-collectors → kof-operators → kof-storage) instead of the declared chain.

The dependency gate that controls which services may be promoted into the runtime ServiceSet only checked whether each upstream service was in a Deployed state. It did not consider which version was deployed. On a fresh install this worked, because each upstream had to first reach Deployed before its dependents became eligible. On an upgrade, however, every service was already Deployed (at the previous version), so the gate immediately let every service through in a single reconcile, and the controller wrote the new versions for all of them at once. The downstream state-management adapter then applied them in its own order, breaking the chain.

The dependency gate was tightened from a single state check into a three-part check: a service satisfies its dependents only when (1) it is in the Deployed state, (2) what's actually running on the cluster matches what the controller currently promised it would run, and (3) what the controller currently promised matches the user's desired version. Whenever any of those drift — because the user just bumped a version, an upstream is mid-upgrade, or an upstream is failing — dependents stay locked at their stored versions until the upstream is fully caught up. This restores one-step-per-reconcile promotion along the declared chain on upgrades, while preserving the existing first-install behavior and the existing protection against mutating dependents whose upstream is in a failed state.

Fixes #2588

### DIFF
--- a/internal/serviceset/util.go
+++ b/internal/serviceset/util.go
@@ -233,6 +233,13 @@ func ServicesUpgradePaths(
 // NOTE: Every service referenced in a DependsOn field must also appear in
 // the desired services list; referencing an absent service is an error.
 //
+// NOTE: desiredServices is expected to have its Versions already resolved by
+// the caller (e.g. via ResolveServiceVersions) so the version-aware gate can
+// compare against the resolved form that previous reconciles persisted in
+// ServiceSet.Spec. Passing unresolved services is tolerated — comparison
+// falls back to Template — but mixing resolved Spec versions with unresolved
+// desired versions will lock dependents.
+//
 // NOTE: This function works under the assumption that there will
 // always be just 1 ServiceSet for every unique combination of CD & MCS.
 //
@@ -328,7 +335,16 @@ func FilterServiceDependencies(
 
 	// Build version maps from the existing ServiceSet(s) so we can compare each
 	// service's currently-promised spec step against what the cluster actually
-	// runs and against the user's final desired version.
+	// runs and against the user's final desired version. Spec and status entries
+	// are normalised the same way so the comparison is symmetric: prefer Version,
+	// fall back to Template when Version is nil/empty.
+	//
+	// NOTE: When more than one ServiceSet matches (cd-only call with multiple
+	// MCS targeting the same cluster, or analogous), values overwrite on key
+	// collision. Current reconciler wiring guarantees at most one relevant
+	// ServiceSet per call (the MCS reconciler scopes by both cluster and MCS;
+	// the CD reconciler operates on the CD's own spec/ServiceSet only), so the
+	// overwrite is not observable in practice. Revisit if that wiring changes.
 	specVersion := make(map[client.ObjectKey]string)
 	statusVersion := make(map[client.ObjectKey]string)
 	statusState := make(map[client.ObjectKey]string)
@@ -344,24 +360,25 @@ func FilterServiceDependencies(
 			specVersion[ServiceKey(svc.Namespace, svc.Name)] = v
 		}
 		for _, svc := range sset.Status.Services {
-			k := ServiceKey(svc.Namespace, svc.Name)
+			v := ""
 			if svc.Version != nil {
-				statusVersion[k] = *svc.Version
+				v = *svc.Version
 			}
+			if v == "" {
+				v = svc.Template
+			}
+			k := ServiceKey(svc.Namespace, svc.Name)
+			statusVersion[k] = v
 			statusState[k] = svc.State
 		}
 	}
 
-	// Resolve versions on a copy of desiredServices so the gate below compares
-	// against the same resolved form that previous reconciles persisted in
-	// ServiceSet.Spec. Working on a copy avoids mutating the caller's slice.
-	resolvedDesired := make([]kcmv1.Service, len(desiredServices))
-	copy(resolvedDesired, desiredServices)
-	if err := ResolveServiceVersions(ctx, c, namespace, resolvedDesired); err != nil {
-		return nil, fmt.Errorf("failed to resolve desired service versions: %w", err)
-	}
-	desiredVersion := make(map[client.ObjectKey]string, len(resolvedDesired))
-	for _, svc := range resolvedDesired {
+	// desiredServices is expected to have Versions resolved by the caller
+	// (see ResolveServicesToApply). For each service, prefer Version; fall back
+	// to Template — same normalisation used above for spec/status — so the
+	// gate comparisons are like-for-like.
+	desiredVersion := make(map[client.ObjectKey]string, len(desiredServices))
+	for _, svc := range desiredServices {
 		v := svc.Version
 		if v == "" {
 			v = svc.Template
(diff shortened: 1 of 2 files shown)


internal/serviceset/util.go
```diff
@@ -319,24 +326,69 @@ func FilterServiceDependencies(
...
+			statusState[k] = svc.State
+		}
+	}
+
+	// Resolve versions on a copy of desiredServices so the gate below compares
+	// against the same resolved form that previous reconciles persisted in
+	// ServiceSet.Spec. Working on a copy avoids mutating the caller's slice.
+	resolvedDesired := make([]kcmv1.Service, len(desiredServices))
+	copy(resolvedDesired, desiredServices)
+	if err := ResolveServiceVersions(ctx, c, namespace, resolvedDesired); err != nil {
```
wahabmk: Would it be better to call `ResolveDependencies` before `FilterServiceDependencies` in the `ResolveServicesToApply` below (switch the order) so that we don't have to resolve dependencies twice?

https://github.com/k0rdent/kcm/blob/bd61f54121b9adceb0514549c6ded1dc85877c11/internal/serviceset/util.go#L620-L627

internal/serviceset/util.go
```diff
@@ -319,24 +326,69 @@ func FilterServiceDependencies(
...
 
-	// Map of services (with their states) from the status of fetched ServiceSets.
-	servicesFromStatus := make(map[client.ObjectKey]string)
+	// Build version maps from the existing ServiceSet(s) so we can compare each
+	// service's currently-promised spec step against what the cluster actually
+	// runs and against the user's final desired version.
+	specVersion := make(map[client.ObjectKey]string)
+	statusVersion := make(map[client.ObjectKey]string)
+	statusState := make(map[client.ObjectKey]string)
 	for _, sset := range serviceSets.Items {
```
wahabmk, line 351: I think there is a potential problem in the logic for this for loop. To illustrate consider the example below.

Let's say we have the following MCS with equal priority with both deploying on the same cluster.

- `mcs0` which deploys `nginx-4-13-0`.
```yaml
  serviceSpec:
    services:
        name: nginx
        namespace: nginx
```
- `mcs1` which deploys `nginx-4-14-3` with `B` depending on it.
```yaml
  serviceSpec:
    services:
      - template: ingress-nginx-4-14-3
        name: nginx
        namespace: nginx
      - template: service-B
         name: B
         namespace: B
         dependsOn:
           - namespace: nginx
              name: nginx
```

Since there is a conflict, either the nginx service from `mcs0` or `mcs1` will get deployed. Both have equal priority so whichever is processed before will deploy its service.

Let's say that `mcs1` gets to deploy nginx:
```yaml
kind: ServiceSet
metadata:
  . . .
spec:
  services:
  - name: nginx
    namespace: nginx
    template: ingress-nginx-4-14-3
    version: 4.14.3
status:
  services:
  - lastStateTransitionTime: "2026-05-12T13:20:03Z"
    name: nginx
    namespace: nginx
    state: Deployed
    template: ingress-nginx-4-14-3
    type: Helm
    version: 4.14.3
```
Whereas `mcs0` gets a conflict error:
```yaml
kind: ServiceSet
metadata:
  . . .
spec:
  services:
  - name: nginx
    namespace: nginx
    template: ingress-nginx-4-13-0
    version: 4.13.0
status:
  services:
  - failureMessage: ClusterSummary management-6ec16280-sveltos-mgmt managing it
    lastStateTransitionTime: "2026-05-12T13:20:27Z"
    name: nginx
    namespace: nginx
    state: Failed
    template: ingress-nginx-4-13-0
    type: Helm
    version: 4.13.0
```

In this case when the ClusterDeploymentReconciler runs this function will be called with `cd != nil` and `mcs == nil` so both the ServiceSets will be fetched. So if the order of the fetched ServiceSets in `serviceSets.Items` is `[<sset.for.mcs1>,<sset.for.mcs0>]` then it could happen that the version and status is taken from `mcs0` which has failed, so:
```
specVersion["nginx/nginx"] = "4.13.0"
statusVersion["nginx/nginx"] = "4.13.0"
statusState["nginx/nginx"] = "Failed"
```

So even though version 4.14.3 of the nginx service was successfully deployed, the service B which depends on nginx will remain blocked.

wahabmk: Hmm I think in the case of ClusterDeploymentReconciler when `cd != nil` and `mcs == nil`, in the above scenario the `desiredServices` will be empty (because the services are deployed via MCS not CD) so the ServiceSet belonging to the CD will remain empty.

And then once the MultiClusterServiceReconciler is triggered for mcs1 with `cd != nil` and `mcs != nil` then only 1 ServiceSet will be fetched so `serviceSet.Items` will always be `[<sset.for.mcs1]`.

So perhaps this is fine for now with how we have wired the reconcilers, however, I think the more future-proof approach would still be to have the map as:
```go
statusState := make(map[client.ObjectKey][]string)
```
so that when more than 1 ServiceSets are fetched (as in example above) then the map has a list of the states for the service like:
```
statusState["nginx/nginx"] = ["Failed", "Deployed"]
```
Then we can see if any of the states in this list is Deployed or not to know if the service has actually been deployed by any of the fetched ServiceSets.

BROngineer: Dependencies between MCSs are out of scope this PR and related issue.

kylewuolle (approved):

internal/serviceset/util.go
wahabmk: The example above is not an example of dependencies between MCSs (which we use with [`.spec.dependsOn`](https://github.com/k0rdent/kcm/blob/24aa88a7eabe378cfe837b1cebf40d3139da2b0f/api/v1beta1/multiclusterservice_types.go#L322). It is an example of dependencies between services within 1 MCS but there is also another MCS matching the same cluster (but there is no explicit dependency between mcs0 and mcs1).

wahabmk (approved):

PR #2715, opened 2026-05-11
feat: allow to keep deployed services when changing MultiClusterService selector

### LINKED ISSUE #2682: [ksm] Option for MCS to keep services when clusterSelector stopped matching
* Please see the description of https://github.com/k0rdent/kof/issues/867
  and the plan in [this comment](https://github.com/k0rdent/kof/issues/867#issuecomment-4075103331).
* When such MCS with `.spec.clusterSelector.matchLabels.kof-version: {{ $.Chart.Version }}` is upgraded,
  it stops matching the `ClusterDeployment` having old `.metadata.labels.kof-version`,
  which [currently leads](https://github.com/k0rdent/kcm/blob/v1.9.0/internal/controller/multiclusterservice_controller.go#L158-L159) to deletion of `ServiceSet`, Sveltos `Profile`,
  and uninstallation of services (Helm charts) from the cluster, which may result in data loss.
* Adding [helm.sh/resource-policy: keep](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource) to annotations of such MCS does not help,
  because as soon as the old version of MCS stops matching the cluster,
  it tries to delete helm charts from the cluster,
  while the new version of this MCS tries to install helm charts to the cluster,
  with an unpredictable result due to the race condition on the ownership change.
* So we will not use `helm.sh/resource-policy: keep`,
  and there will be only one version of such MCS at a time,
  to keep things simple and reliable.
* Please add a new option for MCS to keep services when `clusterSelector` stopped matching,
  e.g. `.spec.keepServices: true` with `false` as the default value for backward compatibility.
* Please also test that once `ClusterDeployment.metadata.labels` are updated to match such MCS again,
  the MCS will apply minimal required upgrades to the helm charts on the cluster,
  avoiding full uninstall and re-install.

BROngineer:

When a MultiClusterService (MCS) is used to roll out services across many clusters via clusterSelector, changing the selector (e.g., bumping a version label from kof-version: v1.0.0 to v2.0.0) causes the controller to delete the underlying ServiceSet objects on clusters that no longer match. That deletion cascades into uninstalling the deployed services from those clusters — there is no way to perform a gradual, per-cluster rollout without losing the existing deployment.

Introduced an opt-in boolean field KeepServicesOnSelectorMismatch on the MCS spec. When set to true, the controller preserves ServiceSets on clusters whose labels no longer match the selector — including the case where the selector is cleared entirely. This enables granular, opt-in upgrade flows: operators can relabel clusters one at a time to migrate them to a new MCS version, while clusters still carrying the old labels continue running their existing services untouched.

The field defaults to false, so the existing behavior is preserved for all current users.

Fixes #2682

### DIFF
--- a/api/v1beta1/multiclusterservice_types.go
+++ b/api/v1beta1/multiclusterservice_types.go
@@ -317,10 +317,22 @@ type ServiceSpec struct {
 type MultiClusterServiceSpec struct {
 	// ClusterSelector identifies target clusters to manage services on.
 	ClusterSelector metav1.LabelSelector `json:"clusterSelector,omitempty"`
+
 	// DependsOn is a list of other MultiClusterServices this one depends on.
 	DependsOn []string `json:"dependsOn,omitempty"`
+
 	// ServiceSpec is spec related to deployment of services.
 	ServiceSpec ServiceSpec `json:"serviceSpec,omitempty"`
+
+	// +kubebuilder:default:=false
+
+	// KeepServicesOnSelectorMismatch indicates whether ServiceSets owned by
+	// this MultiClusterService should be preserved on clusters whose labels
+	// no longer match ClusterSelector, including the case where
+	// ClusterSelector is cleared. When true, services already deployed on
+	// such clusters keep running, enabling per-cluster opt-in rollouts driven
+	// by selector changes.
+	KeepServicesOnSelectorMismatch bool `json:"keepServicesOnSelectorMismatch,omitempty"`
 }
 
 // ServiceStatus contains details for the state of services.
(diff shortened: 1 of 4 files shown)


kylewuolle (dismissed):

wahabmk:
Wouldn't this create a problem of dangling ServiceSet with ownerReference set to an MCS which might not exist anymore?

Sveltos has a [StopMatchingBehavior](https://github.com/projectsveltos/addon-controller/blob/1f15e55ba8dea9955de141cc1dbb767ba38c25e2/api/v1beta1/spec.go#L814) field which leaves the services deployed on the target cluster. 

Would this apporach be feasible where if `KeepServicesOnSelectorMismatch` is set then we may set `StopMatchingBehavior="LeavePolicies"` before deleting the ServiceSet?

BROngineer:
> Wouldn't this create a problem of dangling ServiceSet with ownerReference set to an MCS which might not exist anymore?
> 
> Sveltos has a [StopMatchingBehavior](https://github.com/projectsveltos/addon-controller/blob/1f15e55ba8dea9955de141cc1dbb767ba38c25e2/api/v1beta1/spec.go#L814) field which leaves the services deployed on the target cluster.
> 
> Would this apporach be feasible where if `KeepServicesOnSelectorMismatch` is set then we may set `StopMatchingBehavior="LeavePolicies"` before deleting the ServiceSet?

@wahabmk the MCS object will still be there. No issues. Leaving services on the cluster **with** deleting the ServiceSets + Profiles may then lead to issues when the cluster will match selector again. Hence, I want to keep all deployment-related objects and just stop processing them.

wahabmk (approved):

PR #2719, opened 2026-05-11
feat(template): Allow to disable ingress deployment

uwej711:

With the addition of the ingress for openstack hosted cp template the deploy property is always set to true. In case you want to deploy the ingress yourself or use a TLSRoute with a Gateway you should be able to disable the deployment of the ingress.

This change exposes ingress.deploy in the values.yaml, so that you can set it to false if you have to.

### DIFF
--- a/templates/cluster/openstack-hosted-cp/Chart.yaml
+++ b/templates/cluster/openstack-hosted-cp/Chart.yaml
@@ -7,7 +7,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.26
+version: 1.0.27
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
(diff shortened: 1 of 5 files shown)


zerospiel:
@uwej711 hey, thanks for the contribution! Could you please kindly:
1. follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) spec, amending the current commit msg
2. run `make generate-all` (possibly with the `BASE_COMMIT=upstream/main` to ensure everything is up-to-date)
3. extend the same logic to the standalone template, since they both have limited ingress support

@eromanova please advise whether it makes sense along with extending the logic to the second template

uwej711:
@zerospiel ran make generate-all as requested (after rebasing to main). As far as I can see, openstack-hosted-cp is the only template having ingress at the moment. It could make sense for other clouds where the controlplane ist hosted but I have no way to test any changes.

zerospiel:
Thanks for the amendments! You're right, I've overlooked the templates, so there are no other templates to add to. 

Regarding the tests: we have our own internal test suites that target the master branch. We look forward to introducing a new set of GH Actions that will be available for all contributors.

At the moment, there is no request to support other providers in our default cluster templates (though, everyone can make their own templates!), so we do not expect extending the functionality to other cloud providers

zerospiel (approved):

eromanova (approved):

PR #2722, opened 2026-05-12
fix(templates): proper merge user and defaults vals

### LINKED ISSUE #2709: [bug] Incorrect values merging when setting provider certain provider configuration
**Describe the bug**

When overriding provider's deployment configuration via `Management` object the merging is incorrect.

For example when setting container env in the management like so:

```yaml
   - config:
        infrastructure:
          deployment:
            containers:
            - env:
              - name: SSH_KNOWN_HOSTS
                value: ""
              name: manager
      name: cluster-api-provider-k0sproject-k0smotron
```

The whole containers array will be replaced by the literal provided in the `Management,` which causes a problem when custom registry is used and containers override already present (i.e. the `controller.globalRegistry` parameter is set).

**To Reproduce**

1. Install k0rdent with `controller.globalRegistry` parameter
2. Override any provider deployment configuration in the `Management`

**Expected behavior**

Array merge happens with user supplied values having a higher precedence. _**The priority of global image values must be discussed separately**_ 

**Additional context**

The behavior was introduced by https://github.com/k0rdent/kcm/pull/2279 which uses `merge` helm function which does not merge arrays or any compound structures and can only merge simple dicts.

As an option the `deepCopy` function may be considered or even more controlled values merge in the controller instead of using  inferior helm templating functions.

zerospiel:
Use a deployment merge helper in provider charts that merges containers by name instead of replacing container arrays.

This preserves default manager fields (including global registry image URLs) while keeping user overrides and sidecars.

Also add Helm unittest automation with AWS and k0smotron providers regression suite.

Fixes #2709

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,6 +215,7 @@ $(IMAGES_PACKAGE_DIR): | $(LOCALBIN)
 	mkdir -p $(IMAGES_PACKAGE_DIR)
 
 TEMPLATE_FOLDERS = $(patsubst $(TEMPLATES_DIR)/%,%,$(wildcard $(TEMPLATES_DIR)/*))
+TEST_CHARTS = $(sort $(patsubst %/tests/,%,$(dir $(wildcard $(TEMPLATES_DIR)/*/*/tests/*_test.yaml))))
 
 .PHONY: helm-package
 helm-package: $(CHARTS_PACKAGE_DIR) $(EXTENSION_CHARTS_PACKAGE_DIR) helm ## Package all template charts.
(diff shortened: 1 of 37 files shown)


templates/provider/cluster-api-provider-aws/templates/_helpers.tpl
```diff
@@ -100,11 +100,104 @@ Build the default deployment settings
...
+{{- $mergedContainers = append $mergedContainers $mergedContainer -}}
+{{- else -}}
+{{- $mergedContainers = append $mergedContainers $defaultContainer -}}
+{{- end -}}
+{{- end -}}
+
+{{- range $userContainer := $userContainers -}}
+{{- $name := $userContainer.name | default "" -}}
+{{- if or (eq $name "") (not (hasKey $defaultNames $name)) -}}
+{{- $mergedContainers = append $mergedContainers $userContainer -}}
```
eromanova, line 183: Suggestion: The PR specifically addresses `containers` but along with containers, the CAPI operator Provider object's deployment settings also have `tolerations` and `imagePullSecrets` arrays. The user-defined `imagePullSecrets` will override the default ones.
`imagePullSecrets` entries are identified by `.name` (a `LocalObjectReference`), so they could be deduplicated and concatenated similarly to containers. Or simplier: just concatenated since duplicate pull secrets are harmless:
```
{{- $defaultSecrets := get $default “imagePullSecrets” | default list -}}
{{- $userSecrets := get $user “imagePullSecrets” | default list -}}
{{- if or $defaultSecrets $userSecrets -}}
{{- $_ := set $merged “imagePullSecrets” (concat $userSecrets $defaultSecrets | uniq) -}}
{{- end -}}
```
The same pattern could be applied to `tolerations` for consistency, even if defaults are currently empty.

PS: my comment is not related to this fix or the original issue, since this behaviour existed previously, but can also be addressed in this or in a separate PR.

eromanova (commented):
LGTM overall. I only have one suggestion that can be addressed as well (in templates and test files).

templates/provider/cluster-api-provider-aws/templates/_helpers.tpl
zerospiel: Thanmks for the suggestion, will address it promptly

zerospiel: [Done](https://github.com/k0rdent/kcm/compare/c669db32ab5b85d937ab8e3d0305777aac1294c3..d6ea9765cad94e4cb8a0514b710842b8755d5673)

Makefile
```diff
@@ -227,9 +228,23 @@ package-chart-%: lint-chart-%
...
+		echo "No Helm unittest suites found."; \
+		exit 0; \
+	fi
+	@status=0; \
+	for chart in $(TEST_CHARTS); do \
+		echo "Running helm unittest in $$chart"; \
+		$(HELM) dependency update "$$chart" >/dev/null || status=$$?; \
+		$(HELM) unittest "$$chart" -f 'tests/*_test.yaml' -q || status=$$?; \
+	done; \
+	exit $$status
```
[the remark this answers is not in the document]
zerospiel: Already considered this, pretty minor so I tend to skip it for now (the tests are already guarded by having test files)

eromanova (approved):

PR #2738, opened 2026-05-14
fix: default createnamespace to false rather than true when absent

### LINKED ISSUE #2644: `createNamespace: false` is silently overridden to true in MCS
Setting `helmOptions.installOptions.createNamespace: false` in a `MultiClusterService` service spec has no effect, the value is always persisted as `true`.

**To Reproduce**

1. Create a test MCS: 
```yaml
kubectl apply -f - <<'EOF'
kind: MultiClusterService
metadata:
  name: test-namespace
spec:
  clusterSelector:
    matchLabels:
      test: "true"
  serviceSpec:
    services:
      - name: test
        namespace: test-namespace
        template: cert-manager-v1-19-3 #change to your template
        helmOptions:
          installOptions:
            createNamespace: false
EOF
```

2. Run this command to verify the MCS:
```bash
kubectl get mcs test-namespace -o yaml | grep "services:" -A 10

# OUTPUT: 
# 
# services:
# - helmOptions:
#     installOptions:
#       createNamespace: true   <-- True???
#       disableHooks: false
#       replace: true
#       takeOwnership: false
#   name: test
#   namespace: test-namespace
#   template: cert-manager-v1-19-3
```

**Expected behavior**
`helmOptions.installOptions.createNamespace` should be `false` if it is defined as false in MCS, and the namespace should not be created on child clusters.

kylewuolle:
This PR makes createNamespace optional and defaults to false if not specified. The upstream struct in sveltos addoncontoller has a default value of true. While this could be changed upstream it looks like it would be a breaking change to so on the sveltos side so instead the struct is being copied into our codebase to resolve this issue.

Fixes #2644


api/v1beta1/multiclusterservice_types.go
```diff
@@ -228,7 +228,33 @@ type ServiceHelmOptions struct {
...
+	// +optional
+
+	// If set to true, prevents hooks from running during install, overriding DisableHooks in HelmOptions.
+	DisableHooks bool `json:"disableHooks,omitempty"`
+
+	// +kubebuilder:default:=false
+	// +optional
+
+	// If set, install will ignore the check for helm annotations and take ownership of existing resources.
+	TakeOwnership bool `json:"takeOwnership,omitempty"`
```
zerospiel:
```suggestion
	Replace *bool `json:"replace,omitempty"`

	// +kubebuilder:default:=false
	// +optional

	// If set to true, prevents hooks from running during install, overriding DisableHooks in HelmOptions.
	DisableHooks *bool `json:"disableHooks,omitempty"`

	// +kubebuilder:default:=false
	// +optional

	// If set, install will ignore the check for helm annotations and take ownership of existing resources.
	TakeOwnership *bool `json:"takeOwnership,omitempty"`
```

I understand that this is a mere copy from the upstream but:
- having this is now a de facto standard
- this aligns with the change required by the initial PR initiative/issue
- we want to avoid the situation we'd had with `clusterdeployments` where we also had default `true` having a simple `bool` without the tri-state, which led to situations such as "we set `false` in the spec, but it rendered with `true`" due to the schema

kylewuolle: True that's probably a good idea. I've updated the PR to make the rest the same.

api/v1beta1/multiclusterservice_types.go
```diff
@@ -228,7 +228,30 @@ type ServiceHelmOptions struct {
...
 	// InstallOptions are options specific to helm install
-	InstallOptions *addoncontrollerv1beta1.HelmInstallOptions `json:"installOptions,omitempty"`
+	InstallOptions *HelmInstallOptions `json:"installOptions,omitempty"`
+}
+
+// HelmInstallOptions are options specific to helm install.
+type HelmInstallOptions struct {
+	// +optional
+
+	// Create the release namespace if not present. Defaults to false.
```
Kshatrix, line 238: You mention that it’s going to be a breaking change if it’s changed in sveltos, which is true - but isn’t the same concern applicable to the KSM API as well?

kylewuolle: That's true it is breaking if a user specifies helmInstallOptions on new MCS but existing ones will keep the old behavior.

wahabmk, line 238: I think it is more intuitive to keep `CreateNamespace` as `true` as Sveltos does it.

The reason is that when we deploy a service as the following by providing it's name and namespace, the `nginx` namespace gets created. In this example the `helmOptions.installOptions.createNamespace` field is missing but the namespace still gets created so that implies that the missing createNamespace field is treated as true implicitly. 
```yaml
. . .
    services:
        name: nginx
        namespace: nginx 
. . .
```

If the user instead wants to not create the namespace, they'll have to explicitly specify not to create the namespace like so:
```yaml
. . .
    services:
        name: nginx
        namespace: nginx
        helmOptions:
          installOptions:
            createNamespace: false
. . . 
```

wahabmk: I just ran the changes in this PR with the following spec:
```yaml
. . .
    services:
        name: nginx
        namespace: nginx
        helmOptions:
          installOptions:
            createNamespace: false
. . .
```
❌ My expectation was that the `nginx` namespace won't get created but it did:
```sh
➜  ~ k -n nginx get pod
NAME                                              READY   STATUS    RESTARTS   AGE
nginx-ingress-nginx-controller-667966cf44-dv22w   1/1     Running   0          39s
```
The Sveltos ClusterSummary converted createNamespace to true:
```yaml
apiVersion: config.projectsveltos.io/v1beta1
kind: ClusterSummary
metadata:
  . . .
spec:
  clusterName: mgmt
  clusterNamespace: mgmt
  clusterProfileSpec:
    clusterRefs:
    - apiVersion: lib.projectsveltos.io/v1beta1
      kind: SveltosCluster
      name: mgmt
      namespace: mgmt
    clusterSelector: {}
    continueOnConflict: true
    continueOnError: true
    helmCharts:
    - chartName: ingress-nginx
      chartVersion: 4.13.0
      helmChartAction: Install
      options:
        atomic: false
        dependencyUpdate: false
        disableHooks: false
        disableOpenAPIValidation: false
        enableClientCache: false
        installOptions:
          createNamespace: true
          disableHooks: false
          replace: true
          takeOwnership: false
        runTests: false
        skipCRDs: false
        skipSchemaValidation: false
        uninstallOptions:
          disableHooks: false
        upgradeOptions:
          cleanupOnFail: false
          disableHooks: false
          force: false
          maxHistory: 2
          recreate: false
          resetThenReuseValues: false
          resetValues: false
          reuseValues: false
          subNotes: false
          takeOwnership: false
          upgradeCRDs: false
        wait: false
        waitForJobs: false
      registryCredentialsConfig:
        plainHTTP: true
      releaseName: nginx
      releaseNamespace: nginx
      repositoryName: ingress-nginx
      repositoryURL: oci://ghcr.io/k0rdent/catalog/charts
    reloader: false
    stopMatchingBehavior: WithdrawPolicies
    syncMode: Continuous
    tier: 2147483547
  clusterType: Sveltos
status:
  dependencies: no dependencies
  featureSummaries:
  - featureID: Helm
    hash: ra22futrnCE/YH0yZ75Kw75atIP/5uhMdJqVPqKvWMs=
    lastAppliedTime: "2026-05-22T14:40:47Z"
    resourceSummaryDeployed: false
    status: Provisioned
  helmReleaseSummaries:
  - releaseName: nginx
    releaseNamespace: nginx
    status: Managing
    valuesHash: yj0WO6sFU4GCciYUBWjzvvfqrBh869doeOC2Pp5EI1Y=
```

zerospiel: This is exactly the problem with the `default: true` having a plain `bool` instead of the tri-state `*bool` in the upstream, and the `false` set to the upstream's struct is being rejected by the schema (`false` is the deafult value, thus it is being omitted), and this leads to the "okay thre is nothing set, thus i set the default `true` now"

So my opinion is that it must be fixed on the upstream side (`bool` -> `*bool`)

kylewuolle:
Closing this PR as the fix needs to be made in sveltos. Once the sveltos PR is merged a follow on PR is needed to upgrade KCM references to sveltos.

https://github.com/projectsveltos/addon-controller/pull/1812

PR #2741, opened 2026-05-15
fix(templates): add openstack floatingippoolref

### LINKED ISSUE #2740: OpenStack ClusterTemplates: expose floatingIPPoolRef on worker (and control plane) machines
The OpenStack ClusterTemplates (`openstack-standalone-cp`, `openstack-hosted-cp`) currently do not expose CAPO’s `floatingIPPoolRef` field on `OpenStackMachineTemplate.spec.template.spec`.

Without this field, users cannot attach a floating IP from an `OpenStackFloatingIPPool` to each provisioned VM through k0rdent. Even if the key is set in `ClusterDeployment.spec.config`, it is silently ignored because the Helm template never references it.

CAPO already supports this field in both `v1beta1` and `v1beta2` (`openstackmachine_types.go`, lines 161–165). The missing piece is the Helm chart plumbing.

---

Add `floatingIPPoolRef` support to the OpenStack ClusterTemplates by:

* Adding `floatingIPPoolRef` under:

  * `worker:`
  * optionally `controlPlane:`
    in `values.yaml`
* Rendering the field in the worker `OpenStackMachineTemplate`

This would allow users to associate floating IPs from an `OpenStackFloatingIPPool` with provisioned VMs through standard ClusterDeployment configuration.

---

Currently, the only workaround is to manually modify the generated templates or maintain custom chart patches/forks to inject `floatingIPPoolRef`.

Setting the field directly in `ClusterDeployment.spec.config` is not sufficient because the value is dropped during Helm template rendering.

---

**Additional context**
Users would continue managing the `OpenStackFloatingIPPool` CR independently.

Relevant CAPO support already exists in:

* `v1beta1`
* `v1beta2`

The request is only to expose the existing CAPO functionality through the ClusterTemplate Helm charts.

bnallapeta:

Closes #2740

### DIFF
--- a/templates/cluster/openstack-hosted-cp/templates/openstackmachinetemplate.yaml
+++ b/templates/cluster/openstack-hosted-cp/templates/openstackmachinetemplate.yaml
@@ -39,3 +39,7 @@ spec:
       {{- if $sshKey }}
       sshKeyName: {{ $sshKey }}
       {{- end }}
+      {{- with .Values.floatingIPPoolRef }}
+      floatingIPPoolRef:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
--- a/config/dev/openstack-clusterdeployment.yaml
+++ b/config/dev/openstack-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: openstack-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
-  template: openstack-standalone-cp-1-0-30
+  template: openstack-standalone-cp-1-0-31
   credential: openstack-cluster-identity-cred
   config:
     clusterLabels: {}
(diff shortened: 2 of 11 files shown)


templates/cluster/openstack-hosted-cp/templates/openstackmachinetemplate.yaml
```diff
@@ -39,3 +39,7 @@ spec:
       {{- if $sshKey }}
       sshKeyName: {{ $sshKey }}
       {{- end }}
+      {{- if .Values.worker.floatingIPPoolRef }}
```
zerospiel: this is no `values.worker` rather than `values.k0smotron`

zerospiel: Fixed

PR #2748, opened 2026-05-18
fix: ensure kcm helm chart is updated while dev-upgrade

### LINKED ISSUE #2747: make dev-upgrade does not apply updated kcm helm chart changes
**Describe the bug**
`make dev-upgrade` does not apply updates from the KCM Helm chart. Instead, it only updates the kcm controller image.

As a result, any changes introduced in the helm chart (for example new values, manifests or CRDs) are not applied when running `make dev-upgrade`.

**To Reproduce**

1. Run `git checkout v1.9.0-release`
2. `make dev-apply`
3. `git checkout main`
4. `make dev-upgrade`

**Expected behavior**
`make dev-upgrade` should simulate a real upgrade scenario.

**Additional context**

eromanova:
1. Ensure kcm helm chart is updated during `make dev-upgrade`.
2. Use kcm label selector to ensure the KCM controller Deployment is correctly rollout-restarted even when a custom Helm release name is used.

Fixes #2747

### DIFF
--- a/hack/dev-upgrade.bash
+++ b/hack/dev-upgrade.bash
@@ -24,6 +24,7 @@ set -euo pipefail
 : "${VALUES_FILE:?VALUES_FILE must be set}"
 : "${NAMESPACE:?NAMESPACE must be set}"
 : "${READINESS_TIMEOUT:?READINESS_TIMEOUT must be set}"
+: "${KCM_DEPLOYMENT_NAME:?KCM_DEPLOYMENT_NAME must be set}"
 
 echo "Applying new Release object: kcm-${FQDN_VERSION}"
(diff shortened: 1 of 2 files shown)


hack/dev-upgrade.bash
```diff
@@ -40,7 +41,7 @@ trap 'rm -f "${tmp}"' EXIT
 echo "Patching Management object to use Release: kcm-${FQDN_VERSION}"
 "${KUBECTL}" patch management kcm --type=merge -p '{"spec":{"release":"kcm-'"${FQDN_VERSION}"'"}}'
 
-"${KUBECTL}" rollout restart -n "${NAMESPACE}" deployment/kcm-controller-manager
+"${KUBECTL}" rollout restart -n "${NAMESPACE}" deployment/"${KCM_DEPLOYMENT_NAME}"
```
Kshatrix: can we use a label selector instead?

Kshatrix (changes requested):

Kshatrix (approved):

PR #2760, opened 2026-05-21
feat: add `<deployed>/<total>` services printcolumn to ServiceSet

### LINKED ISSUE #2413: [enh] Show number of services in summary of ServiceSet

wahabmk:
# Description

Implements https://github.com/k0rdent/kcm/issues/2413.

The management ServiceSet deploying 2 services looks like as follows with this change:
```sh
NAME                  CLUSTER   MULTICLUSTERSERVER   SERVICES   PROVIDER             SELF-MANAGEMENT   AGE
management-6ec16280             mcs0                 0/2        ksm-projectsveltos   true              0s
````
And eventually:
```sh
NAME                  CLUSTER   MULTICLUSTERSERVER   SERVICES   PROVIDER             SELF-MANAGEMENT   AGE
management-6ec16280             mcs0                 2/2        ksm-projectsveltos   true              29s
```

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -712,10 +712,30 @@ func (r *ServiceSetReconciler) collectServiceStatuses(ctx context.Context, rgnCl
 		return fmt.Errorf("failed to collect service statuses: %w", err)
 	}
 
+	r.updateServicesInReadyStateCondition(serviceSet)
+
 	status = metav1.ConditionTrue
 	reason = kcmv1.ServiceSetStatusesCollectedReason
 	message = kcmv1.ServiceSetStatusesCollectedMessage
 	return nil
+}
+
+// updateServicesInReadyStateCondition updates the "ServicesInReadyState" condition.
+func (r *ServiceSetReconciler) updateServicesInReadyStateCondition(serviceSet *kcmv1.ServiceSet) {
+	deployedCount := 0
+	for _, svc := range serviceSet.Status.Services {
+		if svc.State == kcmv1.ServiceStateDeployed {
+			deployedCount++
+		}
+	}
+	totalCount := len(serviceSet.Spec.Services)
+	svcReadyCond := findCondition(serviceSet, kcmv1.ServicesInReadyStateCondition)
+	svcReadyStatus := metav1.ConditionFalse
+	if totalCount > 0 && deployedCount == totalCount {
+		svcReadyStatus = metav1.ConditionTrue
+	}
+	updateCondition(serviceSet, svcReadyCond, svcReadyStatus, kcmv1.ServicesInReadyStateCondition,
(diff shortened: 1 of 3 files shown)


kylewuolle (approved):

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -712,12 +712,32 @@ func (r *ServiceSetReconciler) collectServiceStatuses(ctx context.Context, rgnCl
...
+func (r *ServiceSetReconciler) updateServicesInReadyStateCondition(serviceSet *kcmv1.ServiceSet) {
+	deployedCount := 0
+	for _, svc := range serviceSet.Status.Services {
+		if svc.State == kcmv1.ServiceStateDeployed {
+			deployedCount++
+		}
+	}
+	totalCount := len(serviceSet.Spec.Services)
+	svcReadyCond := findCondition(serviceSet, kcmv1.ServicesInReadyStateCondition)
+	svcReadyStatus := metav1.ConditionFalse
```
eromanova, line 733: @wahabmk When a ServiceSet is created without any services (for example, when a ClusterDeployment is deployed without services), the `svcReadyStatus` remains false (totalCount == 0). As a result, this condition is set to False, which might be confusing. It might be better to set it to True in this case, since the desired and actual states match. What do you think?

wahabmk: Oh sorry I merged this before this comment but yes I agree empty service set should be ready. I'll create a follow up!

PR #2764, opened 2026-05-22
fix: make ServicesInReady condition true when 0 services

wahabmk:
Make the ServicesInReady condition True when the ServiceSet has 0 services defined in spec. Follow-up to https://github.com/k0rdent/kcm/pull/2760.

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -730,9 +730,9 @@ func (r *ServiceSetReconciler) updateServicesInReadyStateCondition(serviceSet *k
 	}
 	totalCount := len(serviceSet.Spec.Services)
 	svcReadyCond := findCondition(serviceSet, kcmv1.ServicesInReadyStateCondition)
-	svcReadyStatus := metav1.ConditionFalse
-	if totalCount > 0 && deployedCount == totalCount {
-		svcReadyStatus = metav1.ConditionTrue
+	svcReadyStatus := metav1.ConditionTrue
+	if deployedCount != totalCount {
+		svcReadyStatus = metav1.ConditionFalse
 	}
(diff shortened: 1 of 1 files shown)


internal/controller/adapters/sveltos/serviceset_controller.go
eromanova, line 715: If `collectServiceStatuses` fails prior to the ready services condition update, the condition will be stale or missing. Maybe it'll be better to always call `updateServicesInReadyStateCondition` (in defer) regardless of whether status collection succeeded or failed. What do you think?

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -730,9 +730,9 @@ func (r *ServiceSetReconciler) updateServicesInReadyStateCondition(serviceSet *k
 	}
 	totalCount := len(serviceSet.Spec.Services)
 	svcReadyCond := findCondition(serviceSet, kcmv1.ServicesInReadyStateCondition)
-	svcReadyStatus := metav1.ConditionFalse
-	if totalCount > 0 && deployedCount == totalCount {
```
eromanova, line 734: Just curious about the status reporting behavior when a service is removed from the ServiceSet spec but still remains in the ServiceSet status (when totalCount < deployedCount)?
In that case, the condition would be reported as `False` with something like a `6/5` message. Is that scenario actually possible, and if so, is this the desired behavior?

internal/controller/adapters/sveltos/serviceset_controller.go
wahabmk: The `updateServicesInReadyStateCondition` depends on the status of each service in the serviceSet which is collected by the `collectServiceStatuses` func. So if `collectServiceStatuses` fails prior to reaching line 715 then the status of the services likely are not up to date anyway so I think it is okay to not call `updateServicesInReadyStateCondition` in this iteration of reconcile. As the func will return error so reconcile will be called again.

eromanova (approved):

PR #2785, opened 2026-06-02
feat(templates): allow 0 workers in standalone-cp charts

### LINKED ISSUE #1981: [enhancement] Allow for 0 workers in standalone templates
Currently with standalone-cp templates the smallest cluster possible to make is 2 node and smallest HA cluster would involve 5 nodes 3 controllers + 2 maybe 3 workers.

standalone templates already do `--enable-worker` on controller nodes and only require `--no-taints` in `cpArgs` to allow scheduling workloads on them, however MachineDeployment does not accept value of 0 for amount of replicas.

I propose to wrap worker's MachineTemplate, MachineDeployment, and k0sWorkerConfigTemplate with:
```tpl
{{- if gt (int .Values.workersNumber) 0 }}
# Whole Manifest 
{{- end }} 
```
> NOTE: the int casting is required because kcm passes that value to helm as string for whatever reason

This would allow for creating usable single node clusters or HA 3 node controller+worker clusters

Additional enhancement might be to expose `noTaints` as helm value in controlPlane section but it's not strictly necessairy

zerospiel:
Wrap the worker `MachineDeployment`, `K0sWorkerConfigTemplate` and provider-specific worker `MachineTemplate` in each `*-standalone-cp` chart with a `{{ if gt (int .Values.workersNumber) 0 }}` guard so that setting `workersNumber=0` skips the dedicated worker `MachineDeployment` (which rejects `replicas=0`) and schedules workloads on the control-plane nodes, which already run with `--enable-worker`

Closes #1981

### DIFF
--- a/config/dev/aws-clusterdeployment.yaml
+++ b/config/dev/aws-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: aws-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
-  template: aws-standalone-cp-1-0-32
+  template: aws-standalone-cp-1-0-33
   credential: aws-cluster-identity-cred
   config:
     clusterLabels: {}
--- a/config/dev/azure-clusterdeployment.yaml
+++ b/config/dev/azure-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: azure-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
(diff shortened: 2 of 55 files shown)


eromanova (commented):
LGTM on the cluster templates. One thing I’d like to confirm. Do we need to make any corresponding changes to the ClusterDeployment controller’s CAPI condition aggregation logic?

zerospiel:
> Do we need to make any corresponding changes to the ClusterDeployment controller’s CAPI condition aggregation logic?

No need

eromanova (approved):

Issue #2788, opened 2026-06-03
[bug] 1.10.0-rc1 spec.auditPolicy change on single-CP ClusterDeployment leaves it permanently NotReady


bak0wski:
**Describe the bug**
On a single-control-plane `ClusterDeployment`, changing `spec.auditPolicy` triggers a rolling control-plane recreation. The first change (set) completes; a subsequent change (clear) leaves the previous control-plane Machine stuck in `Deleting` indefinitely and the cluster never returns to `Ready`.

**To Reproduce**

1. Create a ClusterDeployment on AWS with one control-plane machine.
2. Wait for it to reach Ready.
3. Create a ClusterAuditPolicy in the same namespace.
4. Set `spec.auditPolicy` on the ClusterDeployment to the policy name. Wait for the rolling recreation to complete and the cluster to return to Ready.
5. Clear `spec.auditPolicy` on the ClusterDeployment.

**Expected behavior**
The cluster returns to `Ready` after a successful rolling recreation, regardless of how many times `spec.auditPolicy` is changed.

**Additional context**
KCM/OSS 1.10.0-rc1, AWS, single control-plane ClusterDeployment.

- [x] This is NOT a request to change or create a new provider or cluster template

eromanova:
@bak0wski Could you please share the logs and the current status of the relevant resources from your environment (for example, the ClusterDeployment, K0sControlPlane, AWSMachines, etc.)? A support bundle would be especially helpful. I wanted to clarify the following:
> the previous control-plane Machine stuck in Deleting indefinitely and the cluster never returns to Ready

Does the previous control-plane machine still exist in the cloud provider, indicating that the deletion itself is stuck? Or was the machine successfully removed and the issue is that the ClusterDeployment (or related Cluster API resources) continues to report an incorrect status and never transitions back to Ready?

---

If the issue is limited to incorrect status reporting in the ClusterDeployment, that would fall under the KCM side and we can investigate it there.
If there is a Machine object stuck in the Deleting state, we'll likely need to involve the k0smotron/k0s team for further investigation.

bak0wski:
Hi @eromanova,  
I re-ran the test on a fresh cld.  Some logs, yamls and the support bundle are attached and I didn't delete the cluster this time.
[audit-repro.tgz](https://github.com/user-attachments/files/28579440/audit-repro.tgz)

eromanova:
[...]

PR #2792, opened 2026-06-04
fix: always requeue cld controller on hr update

eromanova:
When the HelmRelease was just created or updated, we need to requeue to ensure subsequent reconciles catch the CAPI cluster state changes that will inevitably follow from the helm chart update.
Without it, the controller can check CAPI Cluster readiness before the HelmRelease update takes effect, causing the ClusterDeployment to remain Ready temporarily while the CAPI Cluster transitions to NotReady.

Was found while trying to reproduce #2788 (not a fix for #2788)


eromanova:
Closing this, as the existing HelmRelease watchers already address this issue

PR #2800, opened 2026-06-06
fix: limit ClusterDeployment name to 22 characters for cloud compatibility

AndyCpX:
This PR adds a validation check to the ClusterDeployment admission webhook to limit the name length to 22 characters. This prevents provisioning failures in cloud providers like AWS and Azure, where the cluster name is used as a prefix for resources (e.g., Load Balancers) that have strict naming limits.


zerospiel:
@AndyCpX hey, thanks for the PR

Could you clarify where these requirements come from? Names of the cloud objects are not necessarily tied to a `cld` name, considering custom templating. Do you have any examples of the issues you stumbled upon?

PR #2816, opened 2026-06-12
feat(cld): periodic CAPI status poller

### LINKED ISSUE #2796: [enh] Periodically check CAPI Cluster status to reflect up-to-date info in ClusterDeployment

Currently, the ClusterDeployment controller monitors the CAPI Cluster status only during the provisioning phase. Once the ClusterDeployment reaches the Ready state, the controller stops reconciling based on the CAPI Cluster status.

This creates a gap during subsequent cluster updates. For example, when a user modifies the ClusterDeployment configuration and triggers a cluster update (such as a control plane machine rollout), the ongoing update is sometimes not reflected in the ClusterDeployment status. Although the controller continues to watch HelmRelease resources, CAPI Cluster updates may occur later. As a result, the controller can consider the ClusterDeployment ready and stop reconciliation before the underlying CAPI Cluster status reflects the update in progress.

This makes it difficult for users to understand the actual state of the cluster during upgrades or rollouts.

Ensure that the ClusterDeployment status reflects the current state of the underlying CAPI Cluster, not only during initial provisioning but also during subsequent cluster updates.

Checking the CAPI Cluster status during update operations.

zerospiel:
Add a periodic CAPI Cluster status check so the ClusterDeployment CAPIClusterSummary condition no longer waits for an unrelated watch event to be refreshed. The poller fetches each CAPI Cluster from the correct cluster (management or regional) and only enqueues a ClusterDeployment when the computed summary condition would actually change, keeping reconciles cheap.

The same shared, generic runnable backs the existing periodic ManagementBackup scheduler and the Sveltos ClusterSummary watcher, so all three pollers now share one cancellation, logging, and channel back-pressure path.

Sveltos regional clients are now cached per tick instead of being fetched per ClusterSummary, which removes the hot-path kubeconfig Secret read on management clusters with many ServiceSets.

The poll interval is exposed as a new flag and Helm value (set to 0 to disable).

Closes #2796


internal/controller/clusterdeployment_capi_poller.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, nil
+	}
+
+	// region name -> regional client, reused within a single tick
+	rgnClients := make(map[string]client.Client)
+
+	enqueue := make([]*kcmv1.ClusterDeployment, 0, len(cds.Items))
+	for i := range cds.Items {
+		cd := &cds.Items[i]
+		if !cd.DeletionTimestamp.IsZero() {
```
eromanova: It also makes sense to skip ClusterDeployments with `dryRun: true`, as no infrastructure is created for them.

internal/controller/clusterdeployment_capi_poller.go
```diff
@@ -0,0 +1,141 @@
...
+		client.Limit(1),
+		client.InNamespace(cd.Namespace),
+	); err != nil {
+		return false, fmt.Errorf("failed to list CAPI Clusters for %s: %w", client.ObjectKeyFromObject(cd), err)
+	}
+
+	curr := apimeta.FindStatusCondition(cd.Status.Conditions, kcmv1.CAPIClusterSummaryCondition)
+
+	if len(clusters.Items) == 0 {
+		// no CAPI Cluster; only enqueue if previously deleting so the controller
```
eromanova: If a user accidentally deletes the CAPI Cluster object, it may be removed before the next poll cycle (ex. faster than the default CAPI cluster check interval), while the corresponding ClusterDeployment remains (not deleted). In that case, this function returns `false, nil`, and the user never knows that the underlying CAPI cluster no longer exists.

internal/controller/clusterdeployment_capi_poller.go
zerospiel: Added, thanks

internal/controller/clusterdeployment_capi_poller.go
zerospiel: Handled via a new dedicated condition

eromanova (approved):

PR #2834, opened 2026-06-19
feat(templates): expose storage.etcd.autoDeletePVCs in hosted-cp templates

XDDDaniel:
The template changes lets us opt in to deleting etcd PVCs on cluster teardown, which avoids name reuse issues caused by leftover PVCs.

Added storage.etcd.autoDeletePVCs to hosted-cp templates (default false) and wires it through to K0smotronControlPlane.spec.etcd.

Includeed generated updates (values.schema.json, hosted-cp chart version bumps and regenerated kcm-templates provider template files).

### DIFF
--- a/config/dev/docker-clusterdeployment.yaml
+++ b/config/dev/docker-clusterdeployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: docker-${CLUSTER_NAME_SUFFIX}
   namespace: ${NAMESPACE}
 spec:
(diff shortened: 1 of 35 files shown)


Kshatrix (approved):

eromanova (approved):

eromanova (commented):
LGTM. We can adapt the `remote-cluster` template in a similar manner, but it can be addressed in a separate change.

PR #2842, opened 2026-06-22
fix: svc key collisions in FilteredDependencies func

### LINKED ISSUE #2827: [bug][ksm] Issues with service dependency filtering due to key collisions
# Environment

- Create following MCS with `owner: dev-team` matching label with 200 priority containing nginx service which fails:
```yaml
kind: MultiClusterService
metadata:
  name: mcs1
spec:
  clusterSelector:
    matchLabels:
      owner: dev-team
  serviceSpec:
    provider:
      name: ksm-projectsveltos
      config:
        continueOnError: true
        stopOnConflict: false
        priority: 200
    services:
        name: nginx
        namespace: nginx
        values: |
          ingress-nginx:
            global:
              image:
                registry: this.does.not.exist.com
        helmOptions:
          timeout: 1s
```
- Create following CD cluster with `owner: dev-team` label (so that the MCS matches it) with 300 priority containing nginx <- external-secrets services:
```yaml
kind: ClusterDeployment
metadata:
  name: wali-dev-1
  namespace: kcm-system
  labels:
    owner: dev-team
spec:
  template: aws-standalone-cp-1-0-35
  config:
    clusterIdentity:
      name: aws-cluster-identity
      namespace: kcm-system
    controlPlane:
      amiID: ami-0fe18bc3cfa53a248
      instanceProfile: control-plane.cluster-api-provider-aws.sigs.k8s.io
      instanceType: t3.small
    controlPlaneNumber: 1
    publicIP: true
    region: us-east-2
    worker:
      amiID: ami-0fe18bc3cfa53a248
      instanceProfile: nodes.cluster-api-provider-aws.sigs.k8s.io
      instanceType: t3.small
    workersNumber: 1
  serviceSpec:
    provider:
      name: ksm-projectsveltos
      config:
        priority: 300
        continueOnError: true
        stopOnConflict: false
    services:
      - template: ingress-nginx-4-14-3
        name: nginx
        namespace: nginx
      - template: external-secrets-2-0-0
        name: external-secrets
        namespace: external-secrets
        dependsOn:
          - name: nginx
            namespace: nginx
```

# The Problem

* Two ServiceSets are created: `wali-dev-1` for the CD, and `wali-dev-1-3556efa5` for the MCS matching the CD.
* The `wali-dev-1-3556efa5` shows a conflict error for nginx@v4.13.0 as expected because it has a lower priority.
* The `wali-dev-1` ServiceSet showed only the nginx service in it's spec and the external-secrets service was not added to its spec for a long time (relatively) even though it showed Deployed status for nginx@v4.14.3.
* Ideally, the external-secrets service (which depends on nginx) should be added to the spec of `wali-dev-1` ServiceSet fairly quickly (~5 sec) after nginx becomes Deployed but I observed that it took longer (1+ min).

# The Reason

* After instrumenting the code, the reason I found was in this for loop:
https://github.com/k0rdent/kcm/blob/73df02e4d17badc298deabbaa5ab5d1eaf609a2e/internal/serviceset/util.go#L348-L374
* In this case with 1 CD and 1 MCS when the CD is being reconciled (`cd == "wali-dev-1" && mcs == ""`), the `serviceSet.Items` has 2 entries. Since both nginx@v4.13.0 and the nginx@v4.14.3 service have the same key (`nginx/nginx`), due to key collisions one is overwritten by the other in the `specVersion`, `statusVersion` & `statusState` maps.
* So when the order of the fetched `serviceSet.Items` is `[<wali-dev-1-3556efa5>,<wali-dev-1>]` then these maps contain the information for only nginx@v4.13.0, which has Failed status. This is the reason why the external-secrets service isn't added to the CD's ServiceSet's spec because  -> it's dependsOn count is never [decreased](https://github.com/k0rdent/kcm/blob/73df02e4d17badc298deabbaa5ab5d1eaf609a2e/internal/serviceset/util.go#L413-L417) because -> the nginx service that it depends on is never marked as [deployed](https://github.com/k0rdent/kcm/blob/73df02e4d17badc298deabbaa5ab5d1eaf609a2e/internal/serviceset/util.go#L398-L409) even though it has actually been deployed.
```
# INSTRUMENTED OUTPUT
>>>>> [cd=wali-dev-1, mcs=] len(serviceSets) = 2
---------------------------- IF COLLISION WAS HANDLED -----------------------------
>>>>> [cd=wali-dev-1, mcs=] _specVersion    = map[nginx/nginx:[4.14.3 4.13.0]]        
>>>>> [cd=wali-dev-1, mcs=] _statusVersion  = map[nginx/nginx:[4.14.3 4.13.0]]
>>>>> [cd=wali-dev-1, mcs=] _statusState    = map[nginx/nginx:[Deployed Failed]]
------------------------ ACTUAL WITHOUT COLLISION HANDLING ------------------------
>>>>> [cd=wali-dev-1, mcs=] specVersion     = map[nginx/nginx:4.13.0]    
>>>>> [cd=wali-dev-1, mcs=] statusVersion   = map[nginx/nginx:4.13.0]     
>>>>> [cd=wali-dev-1, mcs=] statusState     = map[nginx/nginx:Failed]       
-----------------------------------------------------------------------------------
>>>>> [cd=wali-dev-1, mcs=] desiredVersion    = map[external-secrets/external-secrets:2.0.0 nginx/nginx:4.14.3]
>>>>> [cd=wali-dev-1, mcs=] deployedServices  = map[]
>>>>> [cd=wali-dev-1, mcs=] dependsOnCount    = map[external-secrets/external-secrets:1 nginx/nginx:0]
>>>>> [cd=wali-dev-1, mcs=] filteredServices  = [(nginx/nginx,4.14.3)]
```
* The reason why the external-secrets service was eventually added to the ServiceSet's spec was because in one of the reconcile cycles (after some time), the order of the `serviceSet.Items` fetched from the Kube API was `[<wali-dev-1>,<<wali-dev-1-3556efa5>>]`. So now the maps contained the information for nginx@v4.14.3 only, which has Deployed status.
```
# INSTRUMENTED OUTPUT AFTER SOME TIME
>>>>> [cd=wali-dev-1, mcs=] len(serviceSets) = 2
---------------------------- IF COLLISION WAS HANDLED -----------------------------
>>>>> [cd=wali-dev-1, mcs=] _specVersion    = map[nginx/nginx:[4.13.0 4.14.3]]
>>>>> [cd=wali-dev-1, mcs=] _statusVersion  = map[nginx/nginx:[4.13.0 4.14.3]]
>>>>> [cd=wali-dev-1, mcs=] _statusState    = map[nginx/nginx:[Failed Deployed]]
------------------------ ACTUAL WITHOUT COLLISION HANDLING ------------------------
>>>>> [cd=wali-dev-1, mcs=] specVersion     = map[nginx/nginx:4.14.3]
>>>>> [cd=wali-dev-1, mcs=] statusVersion   = map[nginx/nginx:4.14.3]
>>>>> [cd=wali-dev-1, mcs=] statusState     = map[nginx/nginx:Deployed]
-----------------------------------------------------------------------------------
>>>>> [cd=wali-dev-1, mcs=] desiredVersion    = map[external-secrets/external-secrets:2.0.0 nginx/nginx:4.14.3]
>>>>> [cd=wali-dev-1, mcs=] deployedServices  = map[nginx/nginx:{}]
>>>>> [cd=wali-dev-1, mcs=] dependsOnCount    = map[external-secrets/external-secrets:0 nginx/nginx:0]
>>>>> [cd=wali-dev-1, mcs=] filteredServices  = [(external-secrets/external-secrets,2.0.0)(nginx/nginx,4.14.3)]
```
* In this example, there is a 50% chance that the ServiceSets will be fetched in the correct order but that will decrease significantly if there are more MCS deploying nginx (with failures) matching the same cluster, thus significantly increasing the time to deploy any subsequent dependents.

wahabmk:
# Description

Kindly see the detailed description of the problem in https://github.com/k0rdent/kcm/issues/2827. This PR fixes the problem by introducing the `fetchServiceSet` function which returns only 1 ServiceSet for each combination of CD & MCS. 

Previously, for `cd != nil && mcs == nil`, even though we were fetching all ServiceSets where cluster was set but implicitly we were only considering 1 ServiceSet because of not accounting for svc key collisions.

So this PR makes this assumption explicit via `fetchServiceSet` by:
* returning just 1 ServiceSet for any combination of cd && mcs.
* filtering out other ServiceSets and only returning 1 ServiceSet for `cd != nil && mcs == nil`. The other ServiceSets with cd set (that is  `cd != nil && mcs != nil`) will be processed by the MCS controller in other reconcile cycles.
* returning an error if there are more than 1 ServiceSets because there shouldn't be more than 1 ServiceSet for any combination of cd && mcs with the way KSM runs currently.

### DIFF
--- a/internal/serviceset/util_test.go
+++ b/internal/serviceset/util_test.go
@@ -647,56 +647,6 @@ func Test_FilterServiceDependencies(t *testing.T) {
 			expected: []testService{a, b},
 		},
 		{
-			testName:        "service A deployed & B provisioning in different servicesets when B->A",
-			desiredServices: []testService{a, b.dependsOn(a)},
-			objects: []client.Object{
-				&kcmv1.ServiceSet{
-					ObjectMeta: metav1.ObjectMeta{Namespace: cd.GetNamespace(), Name: cd.GetName()},
-					Spec:       kcmv1.ServiceSetSpec{Cluster: cd.GetName()},
-					Status: kcmv1.ServiceSetStatus{
-						Services: []kcmv1.ServiceState{
-							{Namespace: a.Namespace, Name: a.Name, State: kcmv1.ServiceStateDeployed},
-						},
-					},
-				},
-				&kcmv1.ServiceSet{
-					ObjectMeta: metav1.ObjectMeta{Namespace: cd.GetNamespace(), Name: cd.GetName() + "-7sc4gx"},
-					Spec:       kcmv1.ServiceSetSpec{Cluster: cd.GetName()},
-					Status: kcmv1.ServiceSetStatus{
-						Services: []kcmv1.ServiceState{
-							{Namespace: b.Namespace, Name: b.Name, State: kcmv1.ServiceStateProvisioning},
-						},
-					},
-				},
-			},
-			expected: []testService{a, b},
-		},
-		{
-			testName:        "service A & B deployed in different servicesets when B->A",
-			desiredServices: []testService{a, b.dependsOn(a)},
-			objects: []client.Object{
-				&kcmv1.ServiceSet{
-					ObjectMeta: metav1.ObjectMeta{Namespace: cd.GetNamespace(), Name: cd.GetName()},
-					Spec:       kcmv1.ServiceSetSpec{Cluster: cd.GetName()},
-					Status: kcmv1.ServiceSetStatus{
-						Services: []kcmv1.ServiceState{
-							{Namespace: a.Namespace, Name: a.Name, State: kcmv1.ServiceStateDeployed},
-						},
-					},
-				},
-				&kcmv1.ServiceSet{
-					ObjectMeta: metav1.ObjectMeta{Namespace: cd.GetNamespace(), Name: cd.GetName() + "-7sc4gx"},
-					Spec:       kcmv1.ServiceSetSpec{Cluster: cd.GetName()},
-					Status: kcmv1.ServiceSetStatus{
-						Services: []kcmv1.ServiceState{
-							{Namespace: b.Namespace, Name: b.Name, State: kcmv1.ServiceStateDeployed},
-						},
-					},
-				},
-			},
-			expected: []testService{a, b},
-		},
-		{
 			// A is failing. B (depends on A) is deployed at its current version.
 			// C (depends on A) was never deployed.
 			// Because A is not Deployed, B and C both have unsatisfied dependencies and
@@ -1023,28 +973,18 @@ func Test_FilterServiceDependencies_Operation(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{Namespace: cd.GetNamespace(), Name: cd.GetName()},
 				Spec:       kcmv1.ServiceSetSpec{Cluster: cd.GetName()},
 			}
-			ssetMCS := &kcmv1.ServiceSet{
-				ObjectMeta: metav1.ObjectMeta{Namespace: cd.GetNamespace(), Name: cd.GetName() + "gswge"},
-				Spec:       kcmv1.ServiceSetSpec{Cluster: cd.GetName()},
-			}
 
 			for itr := range tc.expectedServices {
-				// Divide expected services between 2 ServiceSets targeting the same cluster,
-				// where one serviceset belongs to ClusterDeployment and the other to MultiClusterService.
-				for j, svc := range filtered {
+				for _, svc := range filtered {
 					sstate := kcmv1.ServiceState{
 						Namespace: svc.Namespace, Name: svc.Name, State: kcmv1.ServiceStateDeployed,
 					}
-					if j%2 == 0 {
-						ssetCD.Status.Services = append(ssetCD.Status.Services, sstate)
-					} else {
-						ssetMCS.Status.Services = append(ssetMCS.Status.Services, sstate)
-					}
+					ssetCD.Status.Services = append(ssetCD.Status.Services, sstate)
 				}
 
 				client := fake.NewClientBuilder().
 					WithScheme(scheme).
-					WithObjects(ssetCD, ssetMCS).
+					WithObjects(ssetCD).
 					WithIndex(&kcmv1.ServiceSet{}, kcmv1.ServiceSetClusterIndexKey, kcmv1.ExtractServiceSetCluster).
 					Build()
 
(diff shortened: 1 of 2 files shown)


internal/serviceset/util_test.go
```diff
@@ -646,56 +646,6 @@ func Test_FilterServiceDependencies(t *testing.T) {
 			},
 			expected: []testService{a, b},
 		},
-		{
```
wahabmk, line 649: The following 2 removed tests tested across 2 ServiceSets. Now that the `FilterServiceDependencies` func only operates on 1 ServiceSet, these tests are no longer relevant.

kylewuolle (dismissed):

zerospiel (commented):
Generic question: i've read through the original issue, and why can't we use a stable sorting here? However, I may completely miss something, thus I'd appreciate an explanation on this one

highly likely that even with sorting, we have to do the same process searching for a suitable `servicesets` object, meaning the sort is a useless waste of CPU cycles

wahabmk:
@zerospiel I think you are correct that this could be resolved with sort.

Lets say we have a CD which deploys `nginx <- myapp` and the ClusterDeployment Reconciler calls this function and it fetches 3 ServiceSets from kube:
1. ServiceSet for CD with nginx/nginx@4.15.0 with Deployed status.
2. ServiceSet for MCS1 (matching CD) with nginx/nginx@4.14.0 with Failed status.
3. ServiceSet for MCS2 (matching CD) with nginx/nginx@v4.13.0 with Failed status.

So it has to decide whether `myapp` has been ungated by checking if its parent `nginx` has been Deployed or not.

Now since we are reconciling for the CD, we need to consider the status for services in the ServiceSet belonging to the CD (because the other ServiceSets will be handled by the MCS Controller when it calls this function). We can resolve the key collisions via the following approaches:

### Approach 1

Leave the [nested for loop](https://github.com/k0rdent/kcm/blob/048046e58545ff128195077f2f849601bca3eb4a/internal/serviceset/util.go#L348-L374) as is but sort the ServiceSets beforehand such that the ServiceSet we want is at the end of the list.

Pros: 
- Minimal code change

Cons: 
- We will have to call two sorts:
     - One for the `cd != nil && mcs == nil` case where the ServiceSet with `.spec.Cluster != "" && .spc.MultiClusterService == ""` is at the end of the list.
     - The other for `cd == nil && mcs != nil` case where the ServiceSet with `.spec.Cluster == "" && .spec.MultiClusterService != ""` is at the end of the list. Because in this case we will get multiple ServiceSets as well but we only want the ServiceSet belonging to the self-management MCS.
- Even if it can be done in a single sort, the worst case time cost will be `O(nlogn)`.

### Approach 2
Make it so that the we filter the unwanted ServiceSets so that we don't have to loop through multiple ServiceSets and will only have 1 ServiceSet at a time.

Pros: 
- Makes the intention of this function to operate on 1 ServiceSet at a time explicit so its easier to understand. This function already had the 1 ServiceSet assumption implicitly (as evident by it not handling key collisions as well as [this](https://github.com/k0rdent/kcm/blob/048046e58545ff128195077f2f849601bca3eb4a/internal/serviceset/util.go#L243-L244) comment) but with this change this assumption is made explicit via the `fetchServiceSet` helper.
- Always `O(n)` time cost.

Cons: 
- More code change.

So Approach 2 seems better IMO.

zerospiel (approved):

PR #2877, opened 2026-07-02
fix: rollback k0s version in kubevirt templates to v1.35.4+k0s.0

eromanova:
Reason for rollback: k0s `v1.36.1+k0s.0` deploys calico `v3.32.0` which has an issue https://github.com/projectcalico/calico/issues/12972 (the issue is fixed in `v3.32.1` and calico is not yet updated in recent k0s).

Related task: #2828


zerospiel (approved):

eromanova:
The k0s version has to be updated on a kubevirt workload cluster, thus, this fix isn't applicable.

PR #2891, opened 2026-07-06
feat(ksm): track deployment state

### LINKED ISSUE #2612: [bug][ksm] Sequential service updates don't behave as intended

### LINKED ISSUE #2844: [bug] dependencies between services take no effect
Due to the bug in Helm SDK when it does not respect `--wait` + `--timeout` flags and report deployment is done almost immediately, the dependencies management on KSM side does not work as intended.

BROngineer:
This PR fixes the deployed services' state tracking. Instead if relying on the release/service name and namespace and/or version along with the state reported by the sveltos controllers, the serviceSet controller will compute the hash of the applied workload which is a sum of

- release hash (obtained from the ClusterConfiguration project, computed from release name, namespace, app version, chart version and the source)
- values hash (reported by sveltos)
- patches hash (reported by sveltos)

This allows accurate tracking of what is currently deployed and whether there are any changes that must be applied or rolled back.

Aside from that this PR introduces health checks for deployed workloads: when the sveltos reports that the services are deployed, serviceSet controller verifies that the deployed resources (deployments, statefulsets, daemonsets, pods, PVs and PVCs) are in proper state. If checks pass, then the serviceSet state will be changed to `Deployed` otherwise it won't change.

#2612 
#2844

### DIFF
--- a/api/v1beta1/serviceset_types.go
+++ b/api/v1beta1/serviceset_types.go
@@ -228,13 +228,18 @@ type ProviderState struct {
 
 // ServiceState is the state of a Service
 type ServiceState struct {
+	// LastStateTransitionTime is the time the State was last transitioned
+	LastStateTransitionTime *metav1.Time `json:"lastStateTransitionTime"`
+
+	// +optional
+
+	// Version is the version of the Service
+	Version *string `json:"version,omitempty"`
+
 	// +kubebuilder:validation:Enum=Helm;Kustomize;Resource
 
 	// Type is the type of the deployment method for the Service
 	Type ServiceType `json:"type"`
-
-	// LastStateTransitionTime is the time the State was last transitioned
-	LastStateTransitionTime *metav1.Time `json:"lastStateTransitionTime"`
 
 	// Name is the name of the Service
 	Name string `json:"name"`
@@ -245,17 +250,19 @@ type ServiceState struct {
 	// Template is the name of the ServiceTemplate used to deploy the Service
 	Template string `json:"template"`
 
-	// +optional
-
-	// Version is the version of the Service
-	Version *string `json:"version,omitempty"`
-
 	// State is the state of the Service
 	// +kubebuilder:validation:Enum=Deployed;Provisioning;Failed;Pending;Deleting;Deleted
 	State string `json:"state"`
 
 	// FailureMessage is the reason why the Service failed to deploy
 	FailureMessage string `json:"failureMessage,omitempty"`
+
+	// LastDeployedHash is the verifier's fingerprint of the sveltos-side
+	// configuration this service was most recently confirmed Deployed at.
+	// Used to recognise that a Provisioning state reported by sveltos is
+	// transient noise from an unrelated apply (rather than a real change to
+	// this service) and to promote back to Deployed safely.
+	LastDeployedHash string `json:"lastDeployedHash,omitempty"`
 
 	// +patchMergeKey=type
 	// +patchStrategy=merge
@@ -264,13 +271,6 @@ type ServiceState struct {
 
 	// Conditions is a list of conditions for the Service
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
-
-	// LastDeployedHash is the verifier's fingerprint of the sveltos-side
-	// configuration this service was most recently confirmed Deployed at.
-	// Used to recognise that a Provisioning state reported by sveltos is
-	// transient noise from an unrelated apply (rather than a real change to
-	// this service) and to promote back to Deployed safely.
-	LastDeployedHash string `json:"lastDeployedHash,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/api/v1beta1/serviceset_types.go
+++ b/api/v1beta1/serviceset_types.go
@@ -257,9 +257,9 @@ type ServiceState struct {
 	// FailureMessage is the reason why the Service failed to deploy
 	FailureMessage string `json:"failureMessage,omitempty"`
 
-	// LastDeployedHash is the verifier's fingerprint of the sveltos-side
+	// LastDeployedHash is the verifier's fingerprint of the provider-side
 	// configuration this service was most recently confirmed Deployed at.
-	// Used to recognise that a Provisioning state reported by sveltos is
+	// Used to recognise that a progressing state reported by the provider is
 	// transient noise from an unrelated apply (rather than a real change to
 	// this service) and to promote back to Deployed safely.
 	LastDeployedHash string `json:"lastDeployedHash,omitempty"`
--- a/api/v1beta1/serviceset_types.go
+++ b/api/v1beta1/serviceset_types.go
@@ -264,6 +264,13 @@ type ServiceState struct {
 
 	// Conditions is a list of conditions for the Service
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// LastDeployedHash is the verifier's fingerprint of the sveltos-side
+	// configuration this service was most recently confirmed Deployed at.
+	// Used to recognise that a Provisioning state reported by sveltos is
+	// transient noise from an unrelated apply (rather than a real change to
+	// this service) and to promote back to Deployed safely.
+	LastDeployedHash string `json:"lastDeployedHash,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/internal/controller/adapters/sveltos/enqueue.go
+++ b/internal/controller/adapters/sveltos/enqueue.go
@@ -17,6 +17,8 @@ package sveltos
 import (
 	"context"
 	"fmt"
+	"sync"
+	"time"
 
 	addoncontrollerv1beta1 "github.com/projectsveltos/addon-controller/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -26,15 +28,129 @@ import (
 	pollerutil "github.com/K0rdent/kcm/internal/util/poller"
 )
 
-// enqueueClusterSummary returns a [pollerutil.EnqueueFunc] that walks every
-// [kcmv1.ServiceSet] and emits each one whose regional ClusterSummary is
-// reachable. The Reconcile loop is responsible for collapsing sveltos's
-// (source-of-truth) view into the ServiceSet status — comparing the two
-// here would be circular, because the verifier deliberately diverges
-// ServiceSet.Status.Services from servicesStateFromSummary when it knows
-// reality on cluster disagrees with sveltos's premature claim. Letting the
-// reconcile run on every tick is cheap (status update is gated on a
-// DeepEqual of the whole status) and avoids that loop.
+const (
+	// enqueueBaseBackoff is the initial polling interval for a ServiceSet
+	// that is in flight (Status.Deployed == false). Also the reset target
+	// on any observed ClusterSummary ResourceVersion change.
+	enqueueBaseBackoff = 10 * time.Second
+
+	// enqueueMaxBackoff caps the in-flight back-off. Beyond this, we stop
+	// slowing further — a genuinely stuck install still gets polled every
+	// enqueueMaxBackoff so the verifier's diagnostic conditions stay fresh.
+	enqueueMaxBackoff = 5 * time.Minute
+)
+
+// enqueueState tracks per-ServiceSet polling metadata across ticks. Its
+// dual purpose: (1) detect ClusterSummary-side changes cheaply via
+// ResourceVersion comparison, and (2) implement exponential back-off on
+// ServiceSets that are still in flight, without hammering the workload
+// cluster on every tick.
+//
+// Field ordering: time.Time first (its embedded *Location has an internal
+// pointer offset that fieldalignment sees), then string, then int64.
+type enqueueState struct {
+	nextEligibleTime  time.Time
+	lastSeenSummaryRV string
+	currentBackoff    time.Duration
+}
+
+// enqueueStates holds live enqueueState values keyed by ServiceSet
+// (namespace, name). Entries are created on first observation and pruned
+// once a ServiceSet disappears from the poller's List.
+var enqueueStates = struct {
+	entries map[client.ObjectKey]*enqueueState
+	sync.Mutex
+}{entries: make(map[client.ObjectKey]*enqueueState)}
+
+// evaluate is the per-tick state-machine decision for a single ServiceSet.
+// It updates the receiver in place and returns whether the ServiceSet
+// should be enqueued this tick.
+//
+// Extracted from enqueueClusterSummary so unit tests can exercise the
+// state transitions directly with an injectable clock, without spinning
+// up envtest. `deployed` is a state-machine axis (the "we are done"
+// arm), not a control-coupling flag.
+//
+//nolint:revive // deployed is an input axis of the state machine, not a caller-driven switch
+func (s *enqueueState) evaluate(now time.Time, summaryRV string, deployed bool) bool {
+	if s.lastSeenSummaryRV != summaryRV {
+		// A real sveltos-side change since we last looked. Enqueue,
+		// reset back-off, remember the new RV.
+		s.lastSeenSummaryRV = summaryRV
+		s.currentBackoff = enqueueBaseBackoff
+		s.nextEligibleTime = now.Add(s.currentBackoff)
+		return true
+	}
+	if deployed {
+		// Quiescent — sveltos unchanged AND every service confirmed
+		// Deployed by the verifier. No health controller exists in this
+		// codebase, so ongoing pod-death detection is out of scope; we
+		// unquiesce automatically when sveltos next bumps CS RV.
+		return false
+	}
+	if now.Before(s.nextEligibleTime) {
+		// In-flight but still within the back-off window from the last
+		// fruitless enqueue.
+		return false
+	}
+	// In-flight and the back-off window has elapsed — enqueue and
+	// double the back-off up to the cap.
+	s.currentBackoff *= 2
+	if s.currentBackoff > enqueueMaxBackoff {
+		s.currentBackoff = enqueueMaxBackoff
+	}
+	s.nextEligibleTime = now.Add(s.currentBackoff)
+	return true
+}
+
+// loadOrCreateEnqueueState returns the state for key, creating a fresh
+// one on first sight. Fresh state has an empty lastSeenSummaryRV so the
+// first observed CS RV triggers the "change detected" arm and enqueues
+// once — the correct behavior on process restart or newly created
+// ServiceSet.
+func loadOrCreateEnqueueState(key client.ObjectKey) *enqueueState {
+	enqueueStates.Lock()
+	defer enqueueStates.Unlock()
+	s, ok := enqueueStates.entries[key]
+	if !ok {
+		s = &enqueueState{}
+		enqueueStates.entries[key] = s
+	}
+	return s
+}
+
+// pruneEnqueueStates drops state entries for ServiceSets that were not
+// observed this tick — deleted, garbage-collected, or moved out of the
+// poller's List scope. Runs under the state map's lock; O(N) in current
+// map size.
+func pruneEnqueueStates(seen map[client.ObjectKey]struct{}) {
+	enqueueStates.Lock()
+	defer enqueueStates.Unlock()
+	for key := range enqueueStates.entries {
+		if _, ok := seen[key]; !ok {
+			delete(enqueueStates.entries, key)
+		}
+	}
+}
+
+// enqueueClusterSummary returns a [pollerutil.EnqueueFunc] that emits a
+// ServiceSet only when it needs work, rather than every tick. Three
+// gates decide, in order:
+//
+//  1. ClusterSummary.ResourceVersion advanced since last enqueue — a
+//     real sveltos-side change (apply, upgrade, patch). Enqueue and
+//     reset back-off to enqueueBaseBackoff.
+//  2. ServiceSet.Status.Deployed == true AND CS RV unchanged — the
+//     verifier confirmed every service is Deployed AND sveltos has not
+//     moved since then. Skip. There is no ongoing health controller in
+//     this codebase — the verifier's role in this PR is apply-time
+//     correctness, not continuous pod-death detection.
+//  3. Otherwise (in flight) — enqueue with exponentially-increasing
+//     back-off from enqueueBaseBackoff up to enqueueMaxBackoff. Any
+//     CS RV change resets to the base.
+//
+// Per-ServiceSet state is retained across ticks in enqueueStates;
+// entries are pruned when a ServiceSet disappears from the poller's List.
 //
 // Regional clients are cached per tick.
 func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.EnqueueFunc[*kcmv1.ServiceSet] {
@@ -50,18 +166,23 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 
 		// cluster object key -> regional client, reused within a single tick
 		rgnClients := make(map[client.ObjectKey]client.Client)
+		seenThisTick := make(map[client.ObjectKey]struct{}, len(serviceSetList.Items))
+		now := time.Now()
 
 		out := make([]*kcmv1.ServiceSet, 0, len(serviceSetList.Items))
 		for i := range serviceSetList.Items {
 			serviceSet := &serviceSetList.Items[i]
+			key := client.ObjectKeyFromObject(serviceSet)
+
 			if !serviceSet.GetDeletionTimestamp().IsZero() {
-				logger.V(1).Info("ServiceSet is being deleted, skipping polling", "service_set", client.ObjectKeyFromObject(serviceSet))
+				logger.V(1).Info("ServiceSet is being deleted, skipping polling", "service_set", key)
 				continue
 			}
+			seenThisTick[key] = struct{}{}
 
 			rgnClient, err := resolveRegionalClient(ctx, cl, serviceSet, systemNamespace, rgnClients)
 			if err != nil {
-				logger.V(1).Error(err, "failed to get regional client", "service_set", client.ObjectKeyFromObject(serviceSet))
+				logger.V(1).Error(err, "failed to get regional client", "service_set", key)
 				continue
 			}
 
@@ -71,8 +192,6 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 			} else {
 				profile = new(addoncontrollerv1beta1.Profile)
 			}
-
-			key := client.ObjectKeyFromObject(serviceSet)
 			if err := rgnClient.Get(ctx, key, profile); err != nil {
 				logger.V(1).Error(err, "failed to get Profile or ClusterProfile", "object_name", key)
 				continue
@@ -84,9 +203,18 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 				continue
 			}
 
-			logger.V(1).Info("Scheduling reconcile", "service_set", key, "cluster_summary", client.ObjectKeyFromObject(summary))
-			out = append(out, serviceSet)
+			state := loadOrCreateEnqueueState(key)
+			if state.evaluate(now, summary.ResourceVersion, serviceSet.Status.Deployed) {
+				logger.V(1).Info("Scheduling reconcile",
+					"service_set", key,
+					"cluster_summary", client.ObjectKeyFromObject(summary),
+					"next_backoff", state.currentBackoff,
+				)
+				out = append(out, serviceSet)
+			}
 		}
+
+		pruneEnqueueStates(seenThisTick)
 
 		return out, nil
 	}
--- a/internal/controller/adapters/sveltos/enqueue.go
+++ b/internal/controller/adapters/sveltos/enqueue.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 
 	addoncontrollerv1beta1 "github.com/projectsveltos/addon-controller/api/v1beta1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -28,8 +27,16 @@ import (
 )
 
 // enqueueClusterSummary returns a [pollerutil.EnqueueFunc] that walks every
-// [kcmv1.ServiceSet] and emits the ones whose observed Status.Services drift
-// from the regional ClusterSummary. Regional clients are cached per tick.
+// [kcmv1.ServiceSet] and emits each one whose regional ClusterSummary is
+// reachable. The Reconcile loop is responsible for collapsing sveltos's
+// (source-of-truth) view into the ServiceSet status — comparing the two
+// here would be circular, because the verifier deliberately diverges
+// ServiceSet.Status.Services from servicesStateFromSummary when it knows
+// reality on cluster disagrees with sveltos's premature claim. Letting the
+// reconcile run on every tick is cheap (status update is gated on a
+// DeepEqual of the whole status) and avoids that loop.
+//
+// Regional clients are cached per tick.
 func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.EnqueueFunc[*kcmv1.ServiceSet] {
 	return func(ctx context.Context) ([]*kcmv1.ServiceSet, error) {
 		logger := ctrl.LoggerFrom(ctx)
@@ -77,12 +84,8 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 				continue
 			}
 
-			logger.V(1).Info("Fetched ClusterSummary", "cluster_summary", client.ObjectKeyFromObject(summary))
-			serviceStatesFromSummary := servicesStateFromSummary(logger, summary, serviceSet)
-			if !equality.Semantic.DeepEqual(serviceSet.Status.Services, serviceStatesFromSummary) {
-				logger.V(1).Info("ClusterSummary status does not match observed ServiceSet status, scheduling reconcile", "service_set", key)
-				out = append(out, serviceSet)
-			}
+			logger.V(1).Info("Scheduling reconcile", "service_set", key, "cluster_summary", client.ObjectKeyFromObject(summary))
+			out = append(out, serviceSet)
 		}
 
 		return out, nil
--- a/internal/controller/adapters/sveltos/enqueue_test.go
+++ b/internal/controller/adapters/sveltos/enqueue_test.go
@@ -1,0 +1,195 @@
+// Copyright 2026
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sveltos
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TestEnqueueState_FirstSightEnqueuesAndSeedsRV asserts that a fresh
+// enqueueState (empty lastSeenSummaryRV) treats any observed CS RV as a
+// change, enqueues, and remembers the RV. This is the correct behavior
+// on process restart or newly-created ServiceSet — we have not verified
+// anything yet, so an initial verify is warranted.
+func TestEnqueueState_FirstSightEnqueuesAndSeedsRV(t *testing.T) {
+	now := time.Date(2026, 7, 20, 12, 0, 0, 0, time.UTC)
+	s := &enqueueState{}
+
+	got := s.evaluate(now, "rv-1", false)
+
+	assert.True(t, got, "first sight of any RV must enqueue")
+	assert.Equal(t, "rv-1", s.lastSeenSummaryRV, "RV must be recorded")
+	assert.Equal(t, enqueueBaseBackoff, s.currentBackoff, "back-off starts at base")
+	assert.Equal(t, now.Add(enqueueBaseBackoff), s.nextEligibleTime,
+		"next-eligible must be now+base")
+}
+
+// TestEnqueueState_CSChangedResetsBackoff asserts that any RV change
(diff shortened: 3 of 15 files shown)


api/v1beta1/serviceset_types.go
```diff
@@ -228,14 +228,19 @@ type ProviderState struct {
 
 // ServiceState is the state of a Service
 type ServiceState struct {
+	// LastStateTransitionTime is the time the State was last transitioned
+	LastStateTransitionTime *metav1.Time `json:"lastStateTransitionTime"`
+
+	// +optional
+
+	// Version is the version of the Service
+	Version *string `json:"version,omitempty"`
```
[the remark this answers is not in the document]
BROngineer: Wrong.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -878,8 +1072,10 @@ func getHelmCharts(ctx context.Context, c client.Client, serviceSet *kcmv1.Servi
 				Name:                    svc.Name,
 				Namespace:               svc.Namespace,
 				Template:                svc.Template,
-				Version:                 svc.Version,
-				State:                   kcmv1.ServiceStateProvisioning,
+				// Version is intentionally NOT seeded for Helm services —
+				// the verifier owns Status.Version for this type and stamps
```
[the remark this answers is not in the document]
BROngineer: wrong. starting from go1.26 it's a valid syntax

api/v1beta1/serviceset_types.go
```diff
@@ -245,18 +250,20 @@ type ServiceState struct {
...
 	State string `json:"state"`
 
 	// FailureMessage is the reason why the Service failed to deploy
 	FailureMessage string `json:"failureMessage,omitempty"`
 
+	// LastDeployedHash is the verifier's fingerprint of the sveltos-side
+	// configuration this service was most recently confirmed Deployed at.
+	// Used to recognise that a Provisioning state reported by sveltos is
+	// transient noise from an unrelated apply (rather than a real change to
+	// this service) and to promote back to Deployed safely.
```
wahabmk: We should probably not use Sveltos-specific language here as ServiceSet may be used with other state-management providers.

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+// "no fingerprint information available" so a sentinel error would only
+// force pointless errors.Is checks at every call site.
+func findOwnedClusterConfiguration(
+	ctx context.Context,
+	rgnClient client.Client,
+	namespace string,
+	profileUID types.UID,
+) (*addoncontrollerv1beta1.ClusterConfiguration, error) {
+	list := new(addoncontrollerv1beta1.ClusterConfigurationList)
+	if err := rgnClient.List(ctx, list, client.InNamespace(namespace)); err != nil {
```
wahabmk, line 1175: Perf Improvement. Might be better to create an index to query for ClusterConfigurations matching owner reference in case there are many clusters.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -250,9 +250,203 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	}
+	applyRuleLoadErrorCondition(serviceSet, loadErrs)
+
+	if len(rules) == 0 {
+		// No rules — keep sveltos-reported state.
+		return nil
+	}
+
+	// Fingerprint inputs from sveltos's view.
+	summary, cc, profileKind, profileName, err := fetchHelmArtifactsForVerifier(ctx, r.Client, serviceSet)
```
wahabmk: 🔴 Primary concern: inconsistent client selection across clusters

collectServiceStatuses reads the sveltos Profile/ClusterProfile/ClusterSummary via `rgnClient` (the regional cluster where sveltos runs — confirmed at `serviceset_controller.go:708-807`). But the new verifier reads the same sveltos objects via r.Client:

In the self-managed case `rgnClient == r.Client`, so everything works — and that is the only case the tests exercise (the integration test passes cl as target and sets `reconciler.SystemNamespace`, so `r.Client == rgnClient == target`). But in a regional deployment (`cred.Spec.Region != ""`), r.Client is the management cluster and the Profile/ClusterSummary live on the regional cluster. `fetchHelmArtifactsForVerifier` will hit IsNotFound, return nil artifacts → serviceHashes is empty → currentHash == "" for every service → the promotion arm (Case A) and the hash-stamp arm (Case C) can never fire. The verifier silently degrades to downgrade-only, and Status.Version/LastDeployedHash are never advanced.

- **Recommendation:** use `rgnClient` (the value passed into verifyServiceStates) for `fetchHelmArtifactsForVerifier`, matching `collectServiceStatuses`. If the split to `r.Client` is deliberate, please document why the two code paths read the same objects from different clusters.

wahabmk: **Related, please confirm:** `verifyHelmServiceOnCluster` lists the actual Deployment/Pod/… via target (= `rgnClient`). Helm releases deployed by sveltos land on the child/workload cluster, whereas `rgnClient` is the regional cluster where sveltos and the ClusterSummary objects live (it's the client used to read ClusterSummary, so it cannot also be the child cluster in a CAPI topology). If that's right, in non-self-managed setups the verifier is listing workloads on the wrong cluster and will report NoResourcesOnCluster → downgrade. The `getRegionalClient` doc comment ("the cluster where serviceSet's workload runs") suggests the intended semantics, but that conflicts with its use for reading ClusterSummary. This is the single most important thing to nail down, and it is entirely untested — every test collapses all three clusters into one fake client.

wahabmk (commented):
I have a suggestion for potential performance improvement and 1 critical concern from AI.

internal/controller/adapters/sveltos/verify.go
wahabmk: Since this doesn't affect correctness, can be done in a follow-up

kylewuolle (dismissed):

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+
+// discoveryLabels are the well-known label keys chart authors use to tag
+// resources with the release/instance identity. Sveltos does not stamp a
+// per-service label of its own, so we query each of these in turn (value =
+// service name) and union the results. A resource matching any one of them
+// is considered to belong to the service.
+//
+// Order is irrelevant — the discovery does N list calls per rule's GVK and
+// merges them, deduplicating by (namespace, name).
+var discoveryLabels = []string{
```
eromanova, line 110: I’m afraid these discovery labels are too broad. Labels such as `app`, `release`, and `app.kubernetes.io/name` are commonly shared across helm releases, so querying `app=<releaseName>` can match unrelated resources. As a result, unrelated resources could be included in the health check.
Am I correct that sveltos doesn't apply a unique label or annotation to the resources it manages that we could use instead?

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+	}
+	return prg, nil
+}
+
+// ruleCacheEntry stores the compiled rules and load errors for a single
+// ConfigMap, keyed by ResourceVersion so a CM edit invalidates the entry.
+// lastAccess is refreshed on every hit and consulted by sweepRuleParseCache
+// to evict entries whose owning CM has been deleted (or whose ServiceSet
+// no longer reconciles) after ruleParseCacheTTL.
+type ruleCacheEntry struct {
```
eromanova, line 290: Currently, CEL expressions are recompiled every reconcile. `rulesFromDocs` is called each time, which calls `compileExpr` for every rule. The `docsFromConfigMap` cache stores the raw parsed docs but not the compiled cel.Programs. Can the cache entry store compiled rules (cel.Programs), not just parsed yaml?

eromanova (commented):
A few comments from my side as well (I also share Wahab’s feedback).

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -250,9 +250,203 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
+	// Verify on-cluster state for services that sveltos reports as
+	// Deployed. The helm SDK can return success before pods are actually
+	// ready, so we cross-check by listing labeled resources on the target
+	// cluster and evaluating their state against the embedded rules.
+	// Verification is best-effort: a failure here is logged but does not
+	// fail the reconcile — keep the sveltos-reported state for that round.
+	if verifyErr := r.verifyServiceStates(ctx, rgnClient, serviceSet); verifyErr != nil {
```
zerospiel: **P0 - Verification uses the wrong Kubernetes clients.** 

This passes the regional management client as the workload client. Although regional Profiles, ClusterSummaries, and ClusterConfigurations live in the regional cluster. Non-self-managed services will generally find no workload resources, receive `Provisioning`, and never record hashes or versions.  

**Fix:** Keep three explicit clients: central management for ServiceSets/rule ConfigMaps, regional management for Sveltos artifacts and kubeconfig Secrets, and a workload client constructed from the matched cluster kubeconfig. Add a test using distinct fake clients for each API plane.

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+	verdicts, err := collectVerdicts(ctx, rgnClient, rules, serviceName, serviceNamespace)
+	if err != nil {
+		return "", nil, err
+	}
+
+	// LastTransitionTime is a required field per the ServiceSet CRD's
+	// metav1.Condition schema; status.Update fails validation if any
+	// condition omits it. Using a single now() value per call keeps the
+	// timestamp consistent across all conditions emitted from this round.
+	now := metav1.Now()
```
zerospiel: **P0 - Persistent unhealthy states can create a tight status-write reconcile loop.** 

This generates a new `LastTransitionTime` every invocation, while `applyVerifyConditions` replaces existing conditions directly. Every unhealthy reconciliation therefore changes status; the unfiltered watch `serviceset_controller.go:538` immediately enqueues another reconciliation.  

**Fix:** Merge conditions with `meta.SetStatusCondition`, preserving transition time when status is unchanged, and ignore ordinary status-only updates in the primary watch while still allowing deletion/finalizer transitions.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -250,9 +250,203 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		// Stamp Status.Version and Status.LastDeployedHash on a hash
+		// transition while Deployed. Under the switch above, `newState ==
+		// Deployed` with `currentHash != LastDeployedHash` is only
+		// reachable via the "sveltos=Deployed AND verifier=Deployed" path
+		// — the promotion arm requires `currentHash == LastDeployedHash`,
+		// which makes the hash-!= guard here false. So this stamp fires
+		// only when BOTH sveltos and the verifier confirm Deployed AND
+		// the fingerprint has actually advanced: the strongest agreement
+		// we can express with the signals sveltos exposes.
+		if newState == kcmv1.ServiceStateDeployed && currentHash != "" && currentHash != s.LastDeployedHash {
```
zerospiel, line 455: **P1 - Missing fingerprint artifacts can permanently block sequential deployment.**

An absent ClusterConfiguration or summary produces an empty hash. Sveltos can still leave the service `Deployed` with a stale or nil `status.version`. Dependency filtering then never sees the verified version.  

**Fix:** Treat missing fingerprint inputs as pending for Helm services, or continue polling until the hash/version is recorded. Watch or include ClusterConfiguration state in the poller's change signal.

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+//     These are the cluster-global defaults shipped with kcm.
+//  2. ConfigMaps in serviceSet.Namespace labelled health-rule-target=global.
+//     Tenant-authored defaults, scoped to a single namespace.
+//  3. ConfigMaps in serviceSet.Namespace labelled
+//     health-rule-target=<ownerName>, where ownerName is the name of the
+//     MultiClusterService or ClusterDeployment that owns the ServiceSet.
+//     Rules that speak to a specific workload. Users know the MCS/CD name
+//     at authoring time (the ServiceSet is auto-created, its name is not
+//     stable — the owner's is).
+func rulesFromConfigMaps(
```
zerospiel, line 416: **P1 - Health-rule ConfigMap changes do not enqueue deployed ServiceSets.**

Rules are loaded during in this function, but controller setup has no ConfigMap watch. Once `status.deployed=true`, edited, added, deleted, or repaired rules remain stale indefinitely.  


**Fix:** Watch labeled ConfigMaps and map system-global, namespace-global, and owner-targeted changes to affected ServiceSets.

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+
+		for _, labelKey := range discoveryLabels {
+			list := &unstructured.UnstructuredList{}
+			list.SetGroupVersionKind(*listGVK)
+
+			opts := []client.ListOption{client.MatchingLabels{labelKey: serviceName}}
+			if scope == scopeNamespaced {
+				opts = append(opts, client.InNamespace(serviceNamespace))
+			}
+			if err := rgnClient.List(ctx, list, opts...); err != nil {
```
zerospiel: **P1 - “No resources” conflates absence with inability to verify.** 

Includes valid charts containing only unmonitored kinds such as Jobs or CRDs. Unserved rule GVKs are skipped, and an actual List `NoMatch`, potentially producing the same false downgrade.  

**Fix:** Return a tri-state result: healthy, unhealthy, or not evaluated. Preserve the Sveltos state when no applicable rule was successfully evaluated; only report missing resources when authoritative release inventory says resources should exist.

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+
+		// Dedup objects discovered via multiple labels.
+		seen := make(map[client.ObjectKey]struct{})
+
+		for _, labelKey := range discoveryLabels {
+			list := &unstructured.UnstructuredList{}
+			list.SetGroupVersionKind(*listGVK)
+
+			opts := []client.ListOption{client.MatchingLabels{labelKey: serviceName}}
+			if scope == scopeNamespaced {
```
zerospiel: **P2 - ConfigMap-provided scope is trusted instead of Kubernetes discovery.** 

Validation only checks that scope is one of two strings, while collection. A Deployment rule accidentally marked `Cluster` lists across every namespace and can expose or evaluate unrelated releases.  

**Fix:** Derive scope from `RESTMapping.Scope` and reject rules whose declared scope disagrees.

internal/controller/adapters/sveltos/verify.go
zerospiel, line 110: **P2 - Resource discovery is both non-authoritative and API-heavy.**

The verifier treats `app` and `app.kubernetes.io/name` as release identities, although those normally identify an application/chart and can match another Helm release. Six defaults produce 24 target API calls per Helm service per round.  

**Fix:** Prefer Sveltos/Helm inventory. Otherwise restrict discovery to actual instance labels such as `app.kubernetes.io/instance` and `release`, then batch/cache lookups by target, namespace, and GVK across all services in the ServiceSet.

templates/provider/kcm/templates/rbac/controller/serviceset-verifier-role.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: ClusterRole
+metadata:
+  name: {{ include "kcm.fullname" . }}-serviceset-verifier-role
+  labels:
+    k0rdent.mirantis.com/aggregate-to-manager: "true"
+  {{- include "kcm.labels" . | nindent 4 }}
+# Read-only access used by the sveltos adapter's on-cluster service state
+# verifier (internal/controller/adapters/sveltos/verify.go) to list resources
+# sveltos deployed via helm and evaluate their health against embedded rules.
+rules:
```
zerospiel, line 11: **P2 - Arbitrary custom rule GVKs are accepted but not authorized for self-management.**

Rule parsing accepts any group/kind, while this grants only the default workload kinds. A custom Job, CronJob, or CRD rule receives `Forbidden` when the manager client is the target client.  

**Fix:** Either validate rules against an explicit supported allowlist or expose configurable extra read-only RBAC for custom GVKs. Avoid wildcard permissions.

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+	cond := metav1.Condition{
+		Type:               serviceSetHealthRulesCondition,
+		ObservedGeneration: serviceSet.Generation,
+	}
+	if len(loadErrs) == 0 {
+		cond.Status = metav1.ConditionTrue
+		cond.Reason = "AllRulesValid"
+		cond.Message = "All health rules loaded successfully"
+	} else {
+		msgs := make([]string, 0, len(loadErrs))
```
zerospiel: **P2 - Rule diagnostics can exceed Kubernetes Condition limits.**

Rule-load errors and CEL message output are also unbounded. The CRD limits condition messages to 32 KiB, so malformed or intentionally large rules can make every status update fail.


**Fix:** Cap the number of reported errors and truncate final condition messages safely below 32 KiB.

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1083 @@
...
+		panic(fmt.Errorf("create CEL env: %w", err))
+	}
+	return env
+}()
+
+// rulesFromDocs compiles a slice of source-tagged rule documents into a
+// ruleSet. Individual compile / validation errors are collected and
+// returned in []ruleLoadError so a single bad rule does not break the rest
+// of the set.
+func rulesFromDocs(docs []sourcedRuleDoc) (ruleSet, []ruleLoadError) {
```
zerospiel: **P3 - The rule cache avoids YAML parsing but still recompiles every CEL program per ServiceSet reconciliation.** 

This recompiles both expressions for every rule after each cached document load, creating avoidable CPU and allocator pressure.  

**Fix:** Cache immutable compiled rules by ConfigMap resource version and merge cached rule sets. Also create the CEL activation once per object rather than once per rule.

internal/controller/adapters/sveltos/enqueue.go
```diff
@@ -77,14 +203,19 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 				continue
 			}
 
-			logger.V(1).Info("Fetched ClusterSummary", "cluster_summary", client.ObjectKeyFromObject(summary))
-			serviceStatesFromSummary := servicesStateFromSummary(logger, summary, serviceSet)
-			if !equality.Semantic.DeepEqual(serviceSet.Status.Services, serviceStatesFromSummary) {
-				logger.V(1).Info("ClusterSummary status does not match observed ServiceSet status, scheduling reconcile", "service_set", key)
+			state := loadOrCreateEnqueueState(key)
```
zerospiel, line 206: **P3 - Poller backoff does not reduce regional API traffic.**

Every 10-second tick still fetches the Profile and ClusterSummary before the poller, including deployed and heavily backed-off ServiceSets.

**Fix:** Reuse long-lived regional clients/informers, or apply separate quiescent and failure retry schedules to remote reads.

internal/controller/adapters/sveltos/enqueue_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TestEnqueueState_FirstSightEnqueuesAndSeedsRV asserts that a fresh
+// enqueueState (empty lastSeenSummaryRV) treats any observed CS RV as a
+// change, enqueues, and remembers the RV. This is the correct behavior
+// on process restart or newly-created ServiceSet — we have not verified
+// anything yet, so an initial verify is warranted.
+func TestEnqueueState_FirstSightEnqueuesAndSeedsRV(t *testing.T) {
```
zerospiel, line 31: Generic: implementing time-related unit tests better with `testing/synctest` to ensure durable blocks and avoid flakiness of any kind.

templates/provider/kcm/templates/health-rules/default-rules.yaml
```diff
@@ -0,0 +1,101 @@
...
+  name: {{ include "kcm.fullname" . }}-default-health-rules
+  namespace: {{ .Release.Namespace }}
+  labels:
+    k0rdent.mirantis.com/health-rule-target: global
+  {{- include "kcm.labels" . | nindent 4 }}
+# Default workload-health CEL rules loaded by the sveltos adapter's
+# service-state verifier. These define the cluster-global tier — applied to
+# every ServiceSet on the management cluster. Tenants layer additional,
+# stricter rules in their own namespaces (same label, namespace-global tier)
+# or scoped to a single ServiceSet (label value = ServiceSet name).
```
zerospiel: **P3 - Shipped rule documentation names the wrong targeting key.**

This says targeted ConfigMaps use the ServiceSet name, while the implementation `serviceSetOwnerName` (`verify.go:414`) differs.

**Fix:** Align the manifest comments with the owner-name behavior or change the selector contract.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: Nice catch! Thanks. Totally messed up with all these clients.

templates/provider/kcm/templates/health-rules/default-rules.yaml
BROngineer: Thanks. Artifact from prior implementation.

internal/controller/adapters/sveltos/verify.go
BROngineer: Correct, it does not. However it provides an option to add the patches that we could potentially use to set labels. 

But this is a tricky part actually: sveltos does not inventory the resources being deployed using helm chart which mean we cannot reliably distinct one service from another to add the proper labels. Which means in turn that we either need to label all resources with common label and it will confuse the verifier or we need to pre-render helm charts, inventory resources and add a bunch of patches targeting inventoried resources. On large helm charts (like KOF) it will be really mess in the serviceSet and profile spec.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: > An absent ClusterConfiguration or summary produces an empty hash. Sveltos can still leave the service Deployed

Sveltos "Deployed" status comes from summary. And the ClusterConfiguration got created before the ClusterSummary. Therefore if we get any status from ClusterSummary, then the ClusterConfiguration exists.

internal/controller/adapters/sveltos/verify.go
BROngineer: Sveltos does not expose deployed resources inventory, hence we can't distinguish non-matching-kind from the missing (not-landed) object. We'll rely on sveltos-reported status. It will be on user's side to add health rules that matches resources being deployed.

internal/controller/adapters/sveltos/verify.go
BROngineer: Sveltos does not provide inventory for helm-based deployments. And there is no way to set some common label for the resources being deployed.

internal/controller/adapters/sveltos/verify.go
BROngineer: The verifier is a one-shot cross-check on sveltos's "Deployed" claim (working around helm SDK's "success too early" bug), not a continuous health monitor. Once a service is confirmed Deployed, re-verifying it against later rule changes would either be  misleading (rules tightened → service flips Provisioning based on a policy change, not a state change).

New rules DO take effect for:
- Services still in flight (not yet Deployed) — they're polled and reload rules each round.
- Newly created ServiceSets — first verification uses current rules.
- Any ServiceSet whose sveltos-side state changes (ClusterSummary RV advances) — the poller re-enqueues and rules reload.

The "Deployed and unchanged" case correctly preserves the historical acceptance decision.

internal/controller/adapters/sveltos/verify.go
BROngineer: Unmonitored kinds will be treated as healthy unless new rules appear

templates/provider/kcm/templates/rbac/controller/serviceset-verifier-role.yaml
BROngineer: > Rule parsing accepts any group/kind, while this grants only the default workload kinds. A custom Job, CronJob, or CRD rule receives Forbidden when the manager client is the target client.

If user need to create custom rules for custom (those not coming out-of-the-box) kinds, then it should take care of granting access to these kinds.

Added doc comment to the default health rules file.

internal/controller/adapters/sveltos/enqueue.go
BROngineer: The poller's exponential backoff was designed to reduce reconciler pressure — avoiding needless verifier re-runs on quiescent ServiceSets — not regional API load. That's why fetches happen before evaluate() decides whether to enqueue: we have to Get the ClusterSummary to know its ResourceVersion, and RV comparison is the primary "should we enqueue?" signal.

Regional API traffic is bounded by design on two axes:

- Lifecycle: it exists only while a ServiceSet is being verified (Status.Deployed == false). Once verification succeeds and the ServiceSet is marked Deployed, evaluate() early-returns without enqueuing and the poller stops driving regional Gets for that ServiceSet. The verifier is not a continuous health controller.
- Deployment scheme: for self-management and flat schemes, rgnClient == r.Client (the manager's cache-backed client), so Profile/ClusterSummary reads hit the local informer cache, not a live API. The distinct-regional-cluster case (case 3) is the only one where these reads go over the wire — and even there, bounded by the lifecycle axis above.

Steady-state load — the case that would motivate an informer refactor — is effectively zero. Active deployment windows on regional-scheme ServiceSets are the only paying traffic, which is a small fraction of production time on a small fraction of setups.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -1617,6 +1819,64 @@ func resolveRegionalClient(
...
+	}
+
+	cred := new(kcmv1.Credential)
+	credKey := client.ObjectKey{Namespace: cd.Namespace, Name: cd.Spec.Credential}
+	if err := cl.Get(ctx, credKey, cred); err != nil {
+		return nil, fmt.Errorf("failed to get Credential %s: %w", credKey, err)
+	}
+
+	if cred.Spec.Region == "" {
+		return rgnClient, nil
```
eromanova: This does not look correct. The function is intended to return the child client, but if the credential does not reference a region, it returns `rgnClient`. When the credential does not reference a region, the child client should be taken using the management client (for example, 
```
	child, err := kubeutil.GetChildClient(ctx, cl, secretRef, "value", scheme, kubeutil.DefaultClientFactory)
```
where `cl` is the management client.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -1617,6 +1819,64 @@ func resolveRegionalClient(
...
+	}
+
+	scheme, err := schemeutil.GetRegionalScheme()
+	if err != nil {
+		return nil, fmt.Errorf("build child cluster scheme: %w", err)
+	}
+
+	secretRef := kubeutil.GetKubeconfigSecretKey(clusterKey)
+	child, err := kubeutil.GetChildClient(ctx, rgnClient, secretRef, "value", scheme, kubeutil.DefaultClientFactory)
```
eromanova: The `kubeutil.GetChildClient` function retrieves the child client from the CAPI-generated kubeconfig Secret (`<clusterName>-kubeconfig`). This Secret may not exist for adopted clusters or for clusters with templates that do not include a CAPI Cluster resource, which is possible. As a result, the ServiceSet controller will always fail to reconcile such clusters because the kubeconfig Secret cannot be found.

Potential suggestion: Handle the `NotFound` error gracefully and skip this check for ClusterDeployments that do not contain a CAPI Cluster.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: > This does not look correct. The function is intended to return the child client, but if the credential does not reference a region, it returns `rgnClient`. When the credential does not reference a region, the child client should be taken using the management client (for example, 
> ```
> 	child, err := kubeutil.GetChildClient(ctx, cl, secretRef, "value", scheme, kubeutil.DefaultClientFactory)
> ```
> where `cl` is the management client.

But un case of region = nil, the "regional" cluster is a child in relation to mothership, or?

zerospiel: > But un case of region = nil, the "regional" cluster is a child in relation to mothership, or?

Not quite, it is a client from the management cluster in this case

BROngineer: Aaaaahhhh... you're right.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: Thanks for the tip!

internal/controller/adapters/sveltos/serviceset_controller.go
eromanova: In case the credential does not have a region specified, there may be no regional clusters at all, and all the ClusterDeployments that referenced this Credential (without a region) should take its kubeconfig secret from the management cluster.
So, every ClusterDeployment has its parent (the cluster where all the CAPI objects, including the kubeconfig secret are located):
* `parent == management cluster` when no region is specified in Credential
* `parent == regional cluster` when the region is specified in Credential
And we should use the `parent` client to get the kubeconfig secret. Hopefully I didn't make things even more confusing :)

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -1617,6 +1819,45 @@ func resolveRegionalClient(
...
+func getChildClient(ctx context.Context, cl, rgnClient client.Client, serviceSet *kcmv1.ServiceSet) (client.Client, error) {
+	return resolveChildClient(ctx, cl, rgnClient, serviceSet)
+}
+
+// resolveChildClient returns the kubernetes client for the cluster where
+// serviceSet's workloads actually run. Three schemes are handled:
+//
+//   - self-management: no CAPI child exists; return mgmt (cl).
+//   - flat: mgmt hosts CAPI, which provisions the workload cluster
+//     directly. rgnClient (resolved by getRegionalClient) is already
```
eromanova: Please update the comment accordingly. It currently says:
```
rgnClient (resolved by getRegionalClient) is already the workload cluster — no further hop; return rgnClient ...
```
this is incorrect.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -250,9 +250,208 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	// with the spec value at the moment it confirms a deploy. The Helm
+	// path makes Status.Version mean "verified on cluster" (not "what spec
+	// says") — Kustomize / Resource still mirror spec eagerly in state.go.
+	specVersions := make(map[client.ObjectKey]*string, len(serviceSet.Spec.Services))
+	for i := range serviceSet.Spec.Services {
+		svc := &serviceSet.Spec.Services[i]
+		specVersions[client.ObjectKey{Namespace: svc.Namespace, Name: svc.Name}] = svc.Version
+	}
+
+	childClient, err := getChildClient(ctx, r.Client, rgnClient, serviceSet)
```
eromanova, line 369: I don’t see the change I suggested in https://github.com/k0rdent/kcm/pull/2891#discussion_r3665597090.

If the child client cannot be created because the secret doesn’t exist yet, `getChildClient` will always return an error. Would it make sense to handle the secret `NotFound` case gracefully instead of treating it as a failure?

BROngineer:
The behaviour verified on clusters:
- self-management
- child aws standalone
- adopted kind

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1358 @@
...
+//     early bug — monitored Kinds present but unhealthy — is still
+//     caught by the Provisioning branch above.
+//
+// Returned per-Kind conditions carry no LastTransitionTime; the downstream
+// applyVerifyConditions delegates to meta.SetStatusCondition which
+// stamps timestamps correctly (preserving on same-Status renewals,
+// refreshing only on real transitions).
+func verifyHelmServiceOnCluster(
+	ctx context.Context,
+	rgnClient client.Client,
```
eromanova:
```suggestion
	childClient client.Client,
```

internal/controller/adapters/sveltos/verify.go
```diff
@@ -0,0 +1,1358 @@
...
+// tuples accumulate into an aggregate V(1) log line at the end of the
+// call — one summarised line per verifier round, not one per skipped GK.
+//
+// "Scope" disambiguates a namespaced vs. cluster-scoped list. When multiple
+// rules for the same (group, kind) disagree on scope (which shouldn't
+// happen in practice — kubernetes scope is intrinsic to the kind), the
+// first rule's scope is used.
+func collectVerdicts(
+	ctx context.Context,
+	rgnClient client.Client,
```
eromanova:
```suggestion
	childClient client.Client,
```

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -1633,3 +1902,43 @@ func clusterReference(serviceSet *kcmv1.ServiceSet) *corev1.ObjectReference {
...
+	if err := cl.Get(ctx, secretRef, secret); err != nil {
+		return secretRef, "", fmt.Errorf("failed to get secret %s: %w", secretRef, err)
+	}
+	// since iteration over map keys is non-deterministic in Go, we can't rely on
+	// Sveltos implementation to "get the first key in the Secret" (sveltoscluster_type.go, L229-232)
+	// instead, we'll reject secrets where there are more than 1 key defined.
+	if len(secret.Data) == 0 {
+		return secretRef, secretKey, fmt.Errorf("secret %s has no data", secretRef)
+	}
+	if len(secret.Data) > 1 {
```
eromanova, line 1937: If SveltosCluster does not have `spec.kubeconfigKeyName` defined, it takes the first key in the secret. Maybe it's better to introduce the same logic, otherwise returning an error in case there are >1 items in a secret?

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -1633,3 +1902,43 @@ func clusterReference(serviceSet *kcmv1.ServiceSet) *corev1.ObjectReference {
...
+		return secretRef, secretKey, fmt.Errorf("secret %s has more than one key", secretRef)
+	}
+	for k := range secret.Data {
+		secretKey = k
+	}
+	return secretRef, secretKey, nil
```
eromanova, line 1943: Can this block be simplified, for example:
```
keys := maps.Keys(secret.Data)
secretKey = keys[0]
return secretRef, secretKey, nil
```
?

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: The problem here is that sveltos simply iterates over the keys in secret data. The order is not guaranteed, so we literally can end up in state when sveltos does not get valid key and k0rdent does or vice versa. I'd suggest either expose `kubeconfigKeyName` in adopted cluster template (but we discussed it privately and you told that there are some tricky part related to our credential API) or explicitely write in docs that there must be one key with kubeconfig in the secret.

BROngineer: The problem here is that sveltos simply iterates over secret data keys, so the order is not guaranteed. Hence we can end up in the state when sveltos got the valid key and the k0rdent did not or vice versa. I'd either expose `kubeconfigKeyName` in the adopted cluster config (but as we discussed with you privately this may be not compatible with our current credential API) or explicitly write in docs that there MUST be one key with kubeconfig in the secret.

> otherwise returning an error in case there are >1 items in a secret?

it's already there:
```go
	// since iteration over map keys is non-deterministic in Go, we can't rely on
	// Sveltos implementation to "get the first key in the Secret" (sveltoscluster_type.go, L229-232)
	// instead, we'll reject secrets where there are more than 1 key defined.
	if len(secret.Data) == 0 {
		return secretRef, secretKey, fmt.Errorf("secret %s has no data", secretRef)
	}
	if len(secret.Data) > 1 {
		return secretRef, secretKey, fmt.Errorf("secret %s has more than one key", secretRef)
	}
	for k := range secret.Data {
		secretKey = k
	}
	return secretRef, secretKey, nil
```

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer:
```
❯ make lint test
go fmt ./...
go vet ./...
go fix ./...
internal/controller/adapters/sveltos/serviceset_controller.go:1941:10: golang.org/x/exp/maps.Keys() can be replaced by slices.AppendSeq(make([]T, 0, len(data)), maps.Keys(data)) (exptostd)
        keys := maps.Keys(secret.Data)
                ^
1 issues:
* exptostd: 1
make: *** [lint] Error 1
```

😄 can't say that this call will be easier to read than simple for loop. What do you think about leaving it as is?

internal/controller/adapters/sveltos/serviceset_controller.go
eromanova: Got it. Discussed in private. Sveltos takes the first random key in a secret.

internal/controller/adapters/sveltos/serviceset_controller.go
zerospiel:
```go
	keys := slices.Collect(maps.Keys(secret.Data))
	secretKey = keys[0]
	return secretRef, secretKey, nil
```

And please do not use the experiment package; it is outdated for several releases in this matter. STD's `"maps"` works well enough.

```bash
$ make lint
go fmt ./...
go vet ./...
go fix ./...
0 issues.
```

But in terms of optimization, I'd say that looping over a sequence (which is also allocated) and allocating a new slice might be more expensive in comparison to a simple loop over a map.

UPD. Sequence == iterator in the context of the message

eromanova (commented):
LGTM, but some else from the ksm team must approve

kylewuolle (approved):

wahabmk (approved):

PR #2901, opened 2026-07-08
refactor: drop PI lookup on CD delete path

### LINKED ISSUE #2888: Avoid ProviderInterface lookup in ClusterDeploymentReconciler.releaseProviderCluster
## Problem

`releaseProviderCluster` (internal/controller/clusterdeployment_controller.go) resolves the infra `Cluster` object's GVK on every delete-reconcile via `providerinterface.FindProviderInterfaceForInfra`:

ClusterTemplate.Status.Providers -> infra provider name -> List ProviderInterface (label-matched) -> candidate ClusterGVKs -> List each candidate GVK (label-matched) to find the actual object.

This is more indirection than needed and has real downsides:

- Several extra live List calls per deletion reconcile.
- `FindProviderInterfaceForInfra` swallows *any* List error and returns `nil`, so a transient API error against the regional cluster silently short-circuits `releaseProviderCluster` into a no-op success — the blocking finalizer never gets released, with no error surfaced.
- It guesses among multiple candidate Kinds (e.g. `AWSCluster` vs `AWSManagedCluster`) via List+label-match instead of knowing the answer directly.

## Proposed fix

The CAPI `Cluster` object for the `ClusterDeployment` is already fetched elsewhere (`getPartialCapiCluster`) and carries `Spec.InfrastructureRef` (`Kind`, `APIGroup`, `Name`) — an exact, unambiguous reference to the one infra `Cluster` resource CAPI actually created for this deployment.

- Fetch the CAPI `Cluster` as a typed object (`clusterapiv1.Cluster`, already registered in `GetRegionalScheme()`) instead of `PartialObjectMetadata`, so `Spec.InfrastructureRef` is populated.
- `Get` the infra cluster object directly via `InfrastructureRef.APIGroup` + `.Kind` + `.Name`, instead of `List`-by-label over guessed candidate GVKs.
- This removes the need for `getInfraProvidersNames`, `providerinterface.FindProviderInterfaceForInfra`, `getProviderGVKs`, and the multi-candidate loop in `getProviderCluster` for this code path.

Note: `Spec.InfrastructureRef` is a contract-versioned reference (CAPI v1beta2) with no explicit `APIVersion`, so the version would need to be resolved via the client's `RESTMapper` (`RESTMapping(GroupKind)` with no version), the same way CAPI's own controllers handle contract-versioned refs.

Out of scope: the other `FindProviderInterfaceForInfra` caller in `internal/util/validation/cd.go` (used to resolve supported `ClusterIdentityKinds` for Credential validation) has no alternative data source and should remain as-is.

### LINKED ISSUE #2894: ProviderInterface discovery fails for objects not installed via their provider's Helm chart
## Problem

`FindProviderInterfaceForInfra` finds the `ProviderInterface` for an infra provider by listing objects matching the `helm.toolkit.fluxcd.io/name` label - a label Flux stamps automatically only when the object was applied via that provider's own HelmRelease. Any `ProviderInterface` installed another way has no such label, so the lookup silently returns nil even though a valid `ProviderInterface` exists.

## Proposed solution

Match on the standard CAPI label `cluster.x-k8s.io/provider` instead (already used elsewhere for the same purpose):
​```yaml
metadata:
  labels:
    cluster.x-k8s.io/provider: <name>
​```
- Update `FindProviderInterfaceForInfra` to `client.MatchingLabels{clusterapiv1.ProviderNameLabel: infra}`.
- Add this label to every in-tree `templates/provider/*/templates/interface.yaml`.
- Document the label as the required contract for any custom/out-of-tree `ProviderInterface`.
- Keep the previous discovery method as a fallback

zerospiel:
Resolve the infra Cluster via CAPI Cluster.spec.infrastructureRef using external.GetObjectFromContractVersionedRef instead of listing by guessed GVKs.

Switch FindProviderInterfaceForInfra to the standard CAPI cluster.x-k8s.io/provider label with the flux helm-chart-name label kept as a fallback, and stamp the CAPI label on all in-tree ProviderInterface templates

Closes #2888 
Closes #2894

### DIFF
--- a/internal/controller/clusterdeployment_controller.go
+++ b/internal/controller/clusterdeployment_controller.go
@@ -38,11 +38,11 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/json"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
 	clusterapiv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/cluster-api/controllers/external"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	crcache "sigs.k8s.io/controller-runtime/pkg/cache"
@@ -58,7 +58,6 @@ import (
 	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
 	"github.com/K0rdent/kcm/internal/helm"
 	"github.com/K0rdent/kcm/internal/metrics"
-	"github.com/K0rdent/kcm/internal/providerinterface"
 	"github.com/K0rdent/kcm/internal/record"
 	"github.com/K0rdent/kcm/internal/serviceset"
 	authutil "github.com/K0rdent/kcm/internal/util/auth"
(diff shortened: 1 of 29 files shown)


zerospiel:
TODO: docs for the CAPI Provider label on PI objects

eromanova (commented):
LGTM overall. Please take a look at copilot’s comments on the tests and error handling, and I think we’re good to go.

eromanova (approved):

PR #2905, opened 2026-07-08
fix: do not add owner ref to profile when operating on regional clusters

### LINKED ISSUE #2820: [ksm] [bug] Regional sveltos Profiles carry an invalid cross-cluster ownerReference to the management ServiceSet
**Describe the bug**
When using regional clusters, the `ServiceSetReconciler` creates the ProjectSveltos `Profile` on the regional cluster but sets its controller `ownerReference` to the `ServiceSet`, which exists only on the management cluster. Kubernetes ownerReferences are cluster-local, so the regional Profile ends up with a permanently dangling controller-owner reference.

Because the `kcm-regional` chart installs only the `ProviderInterface` CRD (not `ServiceSet`), the regional clusters garbage collector cannot resolve the owners type and logs

```
unable to get REST mapping for k0rdent.mirantis.com/v1beta1, Kind=ServiceSet
```

whenever the GC reprocesses the affected objects.

The ownerReference is also unnecessary in the regional case: Deletion is already handled explicitly by the `ServiceSet` finalizer, and the Profile is located by name key + the `k0rdent.mirantis.com/managed` label rather than by ownerRef.

**To Reproduce**

1. Create a management cluster and a separate regional cluster as well as a child cluster on that regional cluster.
2. Deploy at least one service on the child cluster.
3. Confirm that the the owner of the Profile (the ServiceSet) does not exist on the regional cluster.
4. Observe the regional clusters kube-controller-manager logging `unable to get REST mapping for ... Kind=ServiceSet` repeatedly.

**Expected behavior**
Profiles created on a remote regional cluster should not carry a cross-cluster `ownerReference` to a management-cluster `ServiceSet`.

**Additional context**
- Relevant code: `internal/controller/adapters/sveltos/serviceset_controller.go`, `createOrUpdateProfile` builds the owner via `metav1.NewControllerRef(serviceSet, ...)` and sets it on the Profile (on both the create and update paths) written through `rgnClient`. `getRegionalClient` returns the management client for self-management ServiceSets (empty `.spec.cluster`) or when `cred.Spec.Region == ""`, and a remote regional client otherwise.

**Proposed fix**

Do not set the controller `ownerReference` when the Profile is created on a remote regional cluster, it is an invalid cross-cluster reference and unnecessary, since Profile deletion is already handled by the `ServiceSet` finalizer. Keep the ownerReference for the co-located non-regional case, where it is a valid same-cluster owner.

Happy to open a PR for this if you agree with the approach.

kylewuolle:
This change skips adding the owner ref to the sveltos profile when operating against a regional cluster.

Fixes #2820

### DIFF
--- a/internal/controller/adapters/sveltos/enqueue.go
+++ b/internal/controller/adapters/sveltos/enqueue.go
@@ -42,7 +42,7 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 		logger.V(1).Info("Listing ServiceSet objects", "service_set_count", len(serviceSetList.Items))
 
 		// cluster object key -> regional client, reused within a single tick
-		rgnClients := make(map[client.ObjectKey]client.Client)
+		rgnClients := make(map[client.ObjectKey]regionalClientEntry)
 
 		out := make([]*kcmv1.ServiceSet, 0, len(serviceSetList.Items))
 		for i := range serviceSetList.Items {
@@ -52,7 +52,7 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 				continue
 			}
 
-			rgnClient, err := resolveRegionalClient(ctx, cl, serviceSet, systemNamespace, rgnClients)
+			rgnClient, _, err := resolveRegionalClient(ctx, cl, serviceSet, systemNamespace, rgnClients)
 			if err != nil {
 				logger.V(1).Error(err, "failed to get regional client", "service_set", client.ObjectKeyFromObject(serviceSet))
 				continue
--- a/internal/controller/adapters/sveltos/enqueue.go
+++ b/internal/controller/adapters/sveltos/enqueue.go
@@ -42,7 +42,7 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 		logger.V(1).Info("Listing ServiceSet objects", "service_set_count", len(serviceSetList.Items))
 
 		// cluster object key -> regional client, reused within a single tick
-		rgnClients := make(map[client.ObjectKey]regionalClientEntry)
+		rgnClients := make(map[client.ObjectKey]client.Client)
 
 		out := make([]*kcmv1.ServiceSet, 0, len(serviceSetList.Items))
 		for i := range serviceSetList.Items {
@@ -52,7 +52,7 @@ func enqueueClusterSummary(cl client.Client, systemNamespace string) pollerutil.
 				continue
 			}
 
-			rgnClient, _, err := resolveRegionalClient(ctx, cl, serviceSet, systemNamespace, rgnClients)
+			rgnClient, err := resolveRegionalClient(ctx, cl, serviceSet, systemNamespace, rgnClients)
 			if err != nil {
 				logger.V(1).Error(err, "failed to get regional client", "service_set", client.ObjectKeyFromObject(serviceSet))
 				continue
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -128,7 +128,7 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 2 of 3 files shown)


zerospiel:
It looks like the base branch contains a 3-4 week-old stale state, please rebase it

zerospiel (commented):
I do not quite understand why we need to extract the whole string from the credentials and pass it within multiple levels, simply to check whether it is empty or not, when there is a simple option:

```go
if cl == rgnClient {
  // we are in mgmt
} else {
 // we are in a region
}
```

Could you please let me know if you considered such an option?

zerospiel (dismissed):

kylewuolle:
> I do not quite understand why we need to extract the whole string from the credentials and pass it within multiple levels, simply to check whether it is empty or not, when there is a simple option:
> 
> ```go
> if cl == rgnClient {
>   // we are in mgmt
> } else {
>  // we are in a region
> }
> ```
> 
> Could you please let me know if you considered such an option?

The problem is the regional client is set to be the management client when no regional cluster is set.

zerospiel:
> The problem is the regional client is set to be the management client when no regional cluster is set.

@kylewuolle correct, this is exactly what I proposed: simply compare the pointers, if they are equal, then we are in a mgmt cluster, thus there is no region, and thus you add an owner reference, otherwise we are in a regional cluster, so we do not need to set the reference. Does it make sense now? So exactly the code I've provided earlier:

```go
if cl == rgnClient {
  // SET OWNER REFERENCE
} else {
 // DO NOTHING
}
```

kylewuolle:
> > The problem is the regional client is set to be the management client when no regional cluster is set.
> 
> @kylewuolle correct, this is exactly what I proposed: simply compare the pointers, if they are equal, then we are in a mgmt cluster, thus there is no region, and thus you add an owner reference, otherwise we are in a regional cluster, so we do not need to set the reference. Does it make sense now? So exactly the code I've provided earlier:
> 
> ```go
> if cl == rgnClient {
>   // SET OWNER REFERENCE
> } else {
>  // DO NOTHING
> }
> ```

Fair enough but I think this could embed some implementation details in other places in the code and if we later change the regional client creation to do something else we could introduce a regression quite easily without realizing it. By passing the region name around it's clear we're distinguishing between regional and management "modes" and a change to how the client is created or whatever won't affect that.

zerospiel:
> Fair enough but I think this could embed some implementation details in other places in the code and if we later change the regional client creation to do something else we could introduce a regression quite easily without realizing it. By passing the region name around it's clear we're distinguishing between regional and management "modes" and a change to how the client is created or whatever won't affect that.

We already [use](https://github.com/k0rdent/kcm/blob/2d896f8e2d35b22c2fc7add8c751b0392250f7d2/internal/controller/components/reconciler.go#L710-L712) the same approach in the codebase. And I do not quite understand the "change the regional client creation to do something else": the pointer is the pointer, and they are literally the same. There is no wrapping or any other modification through any DI, so in conclusion, I can't properly understand your idea regarding the regression. And if we even change any of the corresponding code, that comparison will break immediately.

To sum up, I do not see any cons to this approach at the moment, considering how simple it is and provides the same answer to the bool-typed question "are we in a region or not?"

kylewuolle:
> > Fair enough but I think this could embed some implementation details in other places in the code and if we later change the regional client creation to do something else we could introduce a regression quite easily without realizing it. By passing the region name around it's clear we're distinguishing between regional and management "modes" and a change to how the client is created or whatever won't affect that.
> 
> We already [use](https://github.com/k0rdent/kcm/blob/2d896f8e2d35b22c2fc7add8c751b0392250f7d2/internal/controller/components/reconciler.go#L710-L712) the same approach in the codebase. And I do not quite understand the "change the regional client creation to do something else": the pointer is the pointer, and they are literally the same. There is no wrapping or any other modification through any DI, so in conclusion, I can't properly understand your idea regarding the regression. And if we even change any of the corresponding code, that comparison will break immediately.
> 
> To sum up, I do not see any cons to this approach at the moment, considering how simple it is and provides the same answer to the bool-typed question "are we in a region or not?"

Ok I've updated the code to do this compare instead.

zerospiel (commented):
Thanks for the changes! LGTM apart from the possible stale references after the upgrade, it'd be nice to address it as well (see https://github.com/k0rdent/kcm/pull/2905/changes#r3601973492)

zerospiel (approved):

PR #2954, opened 2026-07-23
fix: properly handle k0smotron service annotations for vsphere hosted

### LINKED ISSUE #2953: [v1.11.0-rc1] Vsphere hosted cluster issue: deprecated service annotations
**Describe the bug**

Since migrating to k0smotron v2.0.0+ (v1beta2 support), K0smotronControlPlane `spec.service.annotations` has been deprecated. The vSphere hosted cluster template still relies on this field, but should instead use `spec.patches` to apply Service annotations.

**Steps to reproduce**

1. Deploy a vSphere standalone cluster with k0rdent/kcm v1.11.0-rc1.
2. Deploy a regional cluster on top of it.
3. Deploy a vSphere hosted cluster following the official guide: https://docs.k0rdent.io/latest/admin/hosted-control-plane/hcp-vmware/

**Expected behavior**

The vSphere hosted cluster is deployed successfully.

eromanova:

The k0smotron v1beta2 API deprecated the `spec.service.annotations` field in K0smotronControlPlane. Previously, the `vsphere-hosted-cp` Helm chart allowed users to pass service annotations directly (e.g., `kube-vip.io/loadbalancerIPs`) via `k0smotron.service.annotations`, which were rendered inline under `spec.service`. After the migration to v1beta2, this field is no longer supported. Annotations must now be applied through `spec.patches` using a patch targeting the Service resource.

To maintain backward compatibility, the chart was updated to:
* Keep `spec.service` for non-deprecated fields (type, apiPort, konnectivityPort).
* Support user-provided arbitrary patches via a new `k0smotron.patches` field.

Fixes #2953

### DIFF
--- a/templates/cluster/vsphere-hosted-cp/values.yaml
+++ b/templates/cluster/vsphere-hosted-cp/values.yaml
@@ -65,11 +65,10 @@ network: ""
 k0smotron: # @schema description: K0smotron parameters; type: object
   controlPlaneFlags: [] # @schema description: ControlPlaneFlags allows to configure additional flags for k0s control plane and to override existing ones. The default flags are kept unless they are overridden explicitly. Flags with arguments must be specified as a single string, e.g. --some-flag=argument; type: array; item: string; uniqueItems: true
   patches: [] # @schema description: Additional patches to apply to k0smotron generated resources (StatefulSet, Service, ConfigMap, etc.). Patches are applied after generation and before apply. Target resources are matched by Kind and app.kubernetes.io/component label. Each patch should follow the k0smotron v1beta2 patch format; type: array; item: object
-  service: # @schema description: The API service configuration; type: object
+  service: # @schema description: The API service configuration; type: object; additionalProperties: false
     type: LoadBalancer # @schema description: An ingress methods for a service; type: string; enum: ClusterIP, NodePort, LoadBalancer; default: LoadBalancer
     apiPort: 6443 # @schema description: The kubernetes API port. If empty k0smotron will pick it automatically; type: number; minimum: 1; maximum: 65535
     konnectivityPort: 8132 # @schema description: The konnectivity port. If empty k0smotron will pick it automatically; type: number; minimum: 1; maximum: 65535
-    annotations: {} # @schema description: Annotations to add to the service (converted to spec.patches for v1beta2 compatibility); type: object; additionalProperties: true
 
 # K0s parameters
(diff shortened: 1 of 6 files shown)


templates/cluster/vsphere-hosted-cp/values.yaml
```diff
@@ -64,10 +64,12 @@ network: ""
...
+  patches: [] # @schema description: Additional patches to apply to k0smotron generated resources (StatefulSet, Service, ConfigMap, etc.). Patches are applied after generation and before apply. Target resources are matched by Kind and app.kubernetes.io/component label. Each patch should follow the k0smotron v1beta2 patch format; type: array; item: object
   service: # @schema description: The API service configuration; type: object
     type: LoadBalancer # @schema description: An ingress methods for a service; type: string; enum: ClusterIP, NodePort, LoadBalancer; default: LoadBalancer
     apiPort: 6443 # @schema description: The kubernetes API port. If empty k0smotron will pick it automatically; type: number; minimum: 1; maximum: 65535
     konnectivityPort: 8132 # @schema description: The konnectivity port. If empty k0smotron will pick it automatically; type: number; minimum: 1; maximum: 65535
+    annotations: {} # @schema description: Annotations to add to the service (converted to spec.patches for v1beta2 compatibility); type: object; additionalProperties: true
```
zerospiel: If they are not supported in `v1beta2` k0smotron APIs, what's the reason to introduce them then? The exemplar templates provide only the latest version where the field has no support; thus, this seems like a redundancy. I think it makes sense to fix/amend the tests/docs (whichever is applicable) in this case

zerospiel: And maybe tighten the `service` block, forbidding additional fields if applicable

eromanova: My initial concern was about `k0smotron.service.annotations`, which is much simpler and more intuitive than defining annotations via:
```
   patches:
     - target:
        kind: Service
        component: control-plane
      patch:
        type: merge
        content: |
          metadata:
            annotations:
              <key>: <value>
```
Even though k0smotron has moved to patches, I think the previous approach was more user-friendly. That said, I’m fine with using patches and updating our documentation to align with the current k0smotron style.

Regarding the `service` block, the explicit field-by-field version is better here, thanks!

zerospiel (approved):

PR #2975, opened 2026-07-29
fix: show failure if invalid kubeconfig is used for adopted cluster

### LINKED ISSUE #2196: [bug] Invalid "adopted-cluster-1-0-1" shows "Object is ready"

wahabmk:
Fixes https://github.com/k0rdent/kcm/issues/2196.

### DIFF
--- a/api/v1beta1/clusterdeployment_types.go
+++ b/api/v1beta1/clusterdeployment_types.go
@@ -94,6 +94,17 @@ const (
 	// CAPIClusterMissingReason indicates the underlying CAPI Cluster object is unexpectedly
 	// absent while the [ClusterDeployment] is still alive (e.g., it was deleted out-of-band).
 	CAPIClusterMissingReason = "CAPIClusterMissing"
+	// SveltosClusterMissingReason indicates the underlying SveltosCluster object is unexpectedly
+	// absent while the [ClusterDeployment] is still alive (e.g., it was deleted out-of-band).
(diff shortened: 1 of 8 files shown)


zerospiel (commented):
It resembles #2694 but now it does not have the corresponding watcher, that way the dependency is avoided and that is good. What is missing, though, is that in that way we might either be stuck forever or produce a stale status: there is no watcher for the `sveltosclusters` thus if it flips or changes in any way, there won't be any event for the cluster deployment controller, and the status may be stale at that point.

Please check out the comment https://github.com/k0rdent/kcm/issues/2196#issuecomment-4396707422 to get more details on my thoughts

IMO, we do not have to validate the given kubeconfig whatsoever.

Issue #2565, opened 2026-03-23
[test] Write tests for ClusterSummary status to ServiceSet status conversion


wahabmk:
Write tests (whether unit or e2e) targeting [servicesStateFromSummary](https://github.com/k0rdent/kcm/blob/fe7ffcfd8d6ef2eaf5c2baa8a37400b7bdc4018b/internal/controller/adapters/sveltos/serviceset_controller.go#L1511) (or any other function responsible for this conversion) which covers some scenarios when converting ClusterSummary status to ServiceSet status.

Especially implement the following flow via mocking unit tests:
1. Create ServiceSet with A.
2. ClusterSummary shows A is deployed.
3. Validate A is deployed in ServiceSet.
4. ClusterSummary shows A is Failed now (probably need to mock) with failure message.
5. Validate A in ServiceSet also shows failed with failure message.
6. ClusterSummary shows A is Pending now with the same failure message from failure previously.
7. Validate A in ServiceSet is also pending with the failure message.
8. ClusterSummary shows A has conflict now.
9. Validate A in ServiceSet shows conflict.
10. ClusterSummary now shows A is deployed again.
11. Validate ServiceSet also shows A as deployed.

The following flow in e2e tests is also important IMO:

**Changing status of a service:**
1. Create MCS on mothership cluster (can be a CD as well but MCS on mothership is cheap and quick) with service A.
2. Verify ServiceSet eventually shows A is deployed.
3. Induce forced failure in A. E.g with following:
```yaml
      - template: ingress-nginx-4-13-0
        name: nginx
        namespace: nginx
        helmOptions:
          timeout: 1s
        values: |
          ingress-nginx:
            global:
              image:
                registry: this.does.not.exist.com
      - template: postgres-operator-1-15-1
```
4. Verify the ClusterSummary shows error.
5. Verify the ServiceSet also shows error with the same failure message as reported in the ClusterSummary.
6. Now remove the invalid config from MCS for A so it can become deployed again.
7. Verify ServiceSet eventually shows A is deployed again.

**Status of service with conflict:**
1. Create MCS1 on mothership cluster with priority 100 with service A1.
2. Create MCS2 on mothership cluster with priority 200 with service A2.
_(Here A1 and A2 refer to the same service A but A2 just means a newer version of A than A1)_
3. Verify that ClusterSummary for MCS2 shows A deployed.
4. Verify that the ServiceSet for MCS2 also reflects that A is deployed.
[...]

PR #2996, opened 2026-08-07
test: add e2e tests for conversion of clustersummary status to serviceset status

kylewuolle:
This PR adds some end to end tests for the conversion of ClusterSummary status to ServiceSet status

#2565

### DIFF
--- a/internal/controller/adapters/sveltos/enqueue.go
+++ b/internal/controller/adapters/sveltos/enqueue.go
@@ -49,9 +49,11 @@ const (
 // Field ordering: time.Time first (its embedded *Location has an internal
 // pointer offset that fieldalignment sees), then string, then int64.
 type enqueueState struct {
-	nextEligibleTime  time.Time
-	lastSeenSummaryRV string
-	currentBackoff    time.Duration
+	nextEligibleTime   time.Time
+	deployedSince      time.Time
+	lastSeenSummaryRV  string
+	lastSeenGeneration int64
+	currentBackoff     time.Duration
 }
 
 // enqueueStates holds live enqueueState values keyed by ServiceSet
@@ -72,22 +74,53 @@ var enqueueStates = struct {
 // arm), not a control-coupling flag.
 //
 //nolint:revive // deployed is an input axis of the state machine, not a caller-driven switch
-func (s *enqueueState) evaluate(now time.Time, summaryRV string, deployed bool) bool {
-	if s.lastSeenSummaryRV != summaryRV {
-		// A real sveltos-side change since we last looked. Enqueue,
-		// reset back-off, remember the new RV.
+func (s *enqueueState) evaluate(now time.Time, summaryRV string, generation int64, deployed bool) bool {
+	if s.lastSeenSummaryRV != summaryRV || s.lastSeenGeneration != generation {
+		// A real sveltos-side change, or a newer desired Spec (e.g. a
+		// version bump) since we last looked. Enqueue, reset back-off,
+		// remember the new RV/generation.
+		//
+		// The generation half of this check matters even when quiescent:
+		// Status.Deployed==true only reflects the Spec that was current
+		// the last time this ServiceSet was reconciled. If Spec has since
+		// advanced (a new upgrade requested) while sveltos's ClusterSummary
+		// hasn't moved yet, CS RV alone would look unchanged and the old
+		// quiescence gate below would wrongly skip forever — nothing else
+		// resets Status.Deployed to false when Spec changes, so this is
+		// the only signal that un-quiesces a ServiceSet whose target moved
+		// again right after finishing a previous one.
 		s.lastSeenSummaryRV = summaryRV
+		s.lastSeenGeneration = generation
 		s.currentBackoff = enqueueBaseBackoff
 		s.nextEligibleTime = now.Add(s.currentBackoff)
+		s.deployedSince = time.Time{}
 		return true
 	}
 	if deployed {
-		// Quiescent — sveltos unchanged AND every service confirmed
-		// Deployed by the verifier. No health controller exists in this
-		// codebase, so ongoing pod-death detection is out of scope; we
-		// unquiesce automatically when sveltos next bumps CS RV.
+		// The verifier's per-service hash comes from a ClusterConfiguration
+		// read that is cached for up to clusterConfigCacheTTL (see
+		// verify.go). A reconcile can land inside that window right as
+		// sveltos finishes an apply: it reads the still-cached PREVIOUS
+		// chart info, computes an unchanged hash, and never re-stamps
+		// Status.Version — even though sveltos is already done and
+		// ClusterSummary/generation have stopped changing, so gate 1 above
+		// will never fire again to correct it. Trusting `deployed`
+		// immediately would freeze that stale version permanently.
+		// Instead, require this ServiceSet to keep reporting deployed
+		// across a full cache-TTL window before treating it as genuinely
+		// quiescent — one extra reconcile past the window reads the
+		// now-expired cache fresh and corrects any stale stamp.
+		if s.deployedSince.IsZero() {
+			s.deployedSince = now
+		}
+		if now.Sub(s.deployedSince) < clusterConfigCacheTTL {
+			s.currentBackoff = enqueueBaseBackoff
+			s.nextEligibleTime = now.Add(s.currentBackoff)
+			return true
+		}
 		return false
 	}
+	s.deployedSince = time.Time{}
 	if now.Before(s.nextEligibleTime) {
 		// In-flight but still within the back-off window from the last
 		// fruitless enqueue.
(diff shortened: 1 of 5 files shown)


test/e2e/serviceset/serviceset.go
```diff
@@ -92,3 +93,40 @@ func ValidateServiceSet(ctx context.Context, cl client.Client, serviceSetKey cli
 		return errs
 	}).WithTimeout(5 * time.Minute).WithPolling(3 * time.Second).Should(Succeed())
 }
+
+func WaitForServiceState(ctx context.Context, cl client.Client, serviceSetKey, svcKey client.ObjectKey, expectedState, failureMessageContains string) kcmv1.ServiceState {
```
wahabmk: This function is not used anywhere but in tests so so I think it will be more appropriate to have it under `test/e2e` package

test/e2e/functional_test.go
```diff
@@ -476,6 +494,119 @@ var _ = Describe("Functional e2e tests", Label("provider:cloud", "provider:docke
...
+			servicesete2e.WaitForServiceState(ctx, kc.CrClient, lowPriorityServiceSetKey, svcKey, kcmv1.ServiceStateDeployed, "")
+
+			highPriorityMCS := multiclusterservice.BuildMultiClusterService(sd, multiClusterServiceTemplate, headlampChartName, multiClusterServiceMatchLabel, highPriorityMCSName)
+			highPriorityMCS.Spec.ServiceSpec.Provider.Config = &apiextv1.JSON{Raw: []byte(highPriorityProviderConfig)}
+			multiclusterservice.CreateMultiClusterService(ctx, kc.CrClient, highPriorityMCS)
+
+			highPriorityServiceSetKey := serviceset.ObjectKey(kubeutil.DefaultSystemNamespace, sd, highPriorityMCS)
+
+			By("Verifying the higher priority MultiClusterService takes over and deploys the service")
+			waitForHelmReleaseSummaryStatus(ctx, kc, sd, highPriorityServiceSetKey, headlampChartName, multiClusterServiceTemplate, addoncontrollerv1beta1.HelmChartStatusManaging)
```
wahabmk: The tests seem ok to me. However, the `waitForHelmReleaseSummaryStatus` function fetches the status from Sveltos ClusterSummary and I don't think we should rely on the Sveltos ClusterSummary object for our tests. I think we should fetch the status from the ServiceSet object, so that our tests actually test if we're correctly capturing the status in the ServiceSet.

wahabmk (commented):
I left some high priority findings from AI review. Kindly verify if they are valid and need fixing.

internal/controller/adapters/sveltos/enqueue.go
```diff
@@ -72,22 +74,53 @@ var enqueueStates = struct {
...
+		// Status.Deployed==true only reflects the Spec that was current
+		// the last time this ServiceSet was reconciled. If Spec has since
+		// advanced (a new upgrade requested) while sveltos's ClusterSummary
+		// hasn't moved yet, CS RV alone would look unchanged and the old
+		// quiescence gate below would wrongly skip forever — nothing else
+		// resets Status.Deployed to false when Spec changes, so this is
+		// the only signal that un-quiesces a ServiceSet whose target moved
+		// again right after finishing a previous one.
 		s.lastSeenSummaryRV = summaryRV
+		s.lastSeenGeneration = generation
```
wahabmk, line 93: evaluate commits lastSeenGeneration before the caller knows the event was delivered, but Runner.tick silently drops events when eventC is full.

Line 99 is the head of the `if deployed {` settle-window block — the flapping issue. The fix is two deletions, not an addition.

**Root cause**

Two lines conspire:

- 117–118 reset `currentBackoff/nextEligibleTime` on every `true` tick, wiping the escalation the `false` ticks built up.
- 123 (`s.deployedSince = time.Time{}` on the in-flight arm) restarts the settle window on every `false` tick, so a flapper is permanently "inside the window" and the arm returns `true` on every `true` tick regardless of backoff.

Fixing only 117–118 isn't enough — the window itself bypasses the backoff, so it would still enqueue every tick. Both have to go.

**Patch**

```go
      if deployed {
              // ... existing comment block, plus:
              //
              // The window is keyed to the RV/generation, not to an
              // uninterrupted run of Deployed. The stale-stamp hazard only
              // arises from an *apply*, and an apply always moves CS RV or
              // generation — which re-arms this window via gate 1 above. A
              // verifier-driven flap (an unhealthy workload demoted back to
              // Provisioning) moves neither, so it must not restart the
              // window nor reset the back-off, or a permanently-flapping
              // ServiceSet is polled at the base interval forever.
              if s.deployedSince.IsZero() {
                      s.deployedSince = now
              }
              if now.Sub(s.deployedSince) < clusterConfigCacheTTL {
-                     s.currentBackoff = enqueueBaseBackoff
-                     s.nextEligibleTime = now.Add(s.currentBackoff)
                      return true
              }
              return false
      }
-     s.deployedSince = time.Time{}
      if now.Before(s.nextEligibleTime) {
```

`deployedSince` is still cleared in gate 1 (line 96) and starts zero from `loadOrCreateEnqueueState`, so removing line 123 doesn't strand it.

**Resulting behaviour**

- Happy path unchanged. Gate 1 fires on the apply → zeroes `deployedSince`, backoff to base. First `true` tick stamps `deployedSince`; ticks at +0/+10/+20s enqueue; +30s returns false. Same three enqueues the PR intends.
- Flapper now escalates. `deployedSince` holds, the window closes 30s after it first reports deployed, and from then on only the `false` ticks drive enqueues — through the doubling schedule up to the 5-minute cap.
- Not touching the backoff in the `true` arm restores exactly what `main` does today (the pre-PR `true` case was a bare `return false`), so this is not a new regression surface.

**Tests**

Both existing settle-window tests still pass — `DeployedKeepsPollingDuringSettleWindow` only asserts the return value and `deployedSince`; `QuiescentSkipsAfterSettleWindow` already asserts backoff must not change. No test asserts the line-123 reset. Add a regression test:

```go
// TestEnqueueState_FlappingDeployedStillEscalatesBackoff asserts that a
// ServiceSet whose Status.Deployed oscillates — an unhealthy workload the
// verifier keeps demoting, with CS RV and generation unchanged — does not
// get its back-off reset by the Deployed ticks. Without this, the
// escalation to enqueueMaxBackoff never happens and the ServiceSet is
// reconciled at the base interval forever, hitting the uncached regional
// client every round.
func TestEnqueueState_FlappingDeployedStillEscalatesBackoff(t *testing.T) {
      now := time.Date(2026, 7, 20, 12, 0, 0, 0, time.UTC)
      s := &enqueueState{
              lastSeenSummaryRV:  "rv-1",
              lastSeenGeneration: 3,
              currentBackoff:     enqueueBaseBackoff,
              nextEligibleTime:   now,
      }

      // enqueueBaseBackoff is the poller tick spacing by construction.
      deployed := true
      for range 100 {
              s.evaluate(now, "rv-1", 3, deployed)
              now = now.Add(enqueueBaseBackoff)
              deployed = !deployed
      }

      assert.Equal(t, enqueueMaxBackoff, s.currentBackoff,
              "flapping Deployed must not reset the back-off schedule")
}
```

One thing I checked while working this out: `pendingStamp` (`serviceset_controller.go:274`) only covers `currentHash == ""`, not the stale-cached-hash case — so the settle window really is the sole guard there, which is why the fix keeps the window intact rather than rate-limiting it.
```

File: internal/controller/adapters/sveltos/enqueue.go [hunk 4]
```diff
@@ -72,22 +74,53 @@ var enqueueStates = struct {
...
+		// will never fire again to correct it. Trusting `deployed`
+		// immediately would freeze that stale version permanently.
+		// Instead, require this ServiceSet to keep reporting deployed
+		// across a full cache-TTL window before treating it as genuinely
+		// quiescent — one extra reconcile past the window reads the
+		// now-expired cache fresh and corrects any stale stamp.
+		if s.deployedSince.IsZero() {
+			s.deployedSince = now
+		}
+		if now.Sub(s.deployedSince) < clusterConfigCacheTTL {
```
wahabmk, line 116: The settle window `deployedSince + clusterConfigCacheTTL` with a strict `<` relies on the unstated invariant deployedSince >= cacheWriteTime + 2×pollInterval, which only holds for the current constant values.

internal/controller/adapters/sveltos/enqueue.go
```diff
@@ -72,22 +74,53 @@ var enqueueStates = struct {
...
+			s.currentBackoff = enqueueBaseBackoff
```
wahabmk: The new settle-window arm resets currentBackoff to enqueueBaseBackoff on every tick where Deployed is true, so a flapping ServiceSet never escalates to enqueueMaxBackoff.

"Flapping" here means that `Status.Deployed` oscillates between `true` and `false` across poller ticks, rather than settling.

**How it happens.** `Status.Deployed` is an all-or-nothing aggregate — true only when every service is in state `Deployed` (`serviceset_controller.go:1052` and `:469`). One service bouncing is enough to flip the whole ServiceSet.

- The verifier (`serviceset_controller.go:440`) transitions per-service state from **live cluster reads**. An unhealthy workload — a pod crash-looping, a chart whose resources never go ready — reads healthy on one pass and unhealthy on the next. Crucially this path changes `Status.Deployed` without the ClusterSummary moving, so gate 1 (`RV/generation changed`) doesn't fire.
- Sveltos itself. `state.go:352` documents the observed pattern verbatim: `nginx is (Deployed->Provisioning->Failed->Provisioning)`.

**Why it defeats the backoff.** The `false` arm zeroes `deployedSince`, so the next `true` tick sets `deployedSince = now`, lands inside the `< clusterConfigCacheTTL` settle window, and resets `currentBackoff` to 10s. Traced out with the real constants (10s base, 30s TTL, 5min cap):

┌───────┬──────────┬────────────────────────────────────────────────────────────────────────────┐
│ tick  │ deployed │                                   result                                   │
├───────┼──────────┼────────────────────────────────────────────────────────────────────────────┤
│ N     │ true     │ `deployedSince=N`, 0 < 30s → backoff reset to 10s, enqueue                   │
├───────┼──────────┼────────────────────────────────────────────────────────────────────────────┤
│ N+10s │ false    │ past `nextEligibleTime` → enqueue, backoff → 20s                             │
├───────┼──────────┼────────────────────────────────────────────────────────────────────────────┤
│ N+20s │ true     │ `deployedSince` was zeroed, so 0 < 30s again → backoff reset to 10s, enqueue │
├───────┼──────────┼────────────────────────────────────────────────────────────────────────────┤
│ N+30s │ false    │ → enqueue, backoff → 20s                                                   │
└───────┴──────────┴────────────────────────────────────────────────────────────────────────────┘

It enqueues on every tick and the backoff just oscillates 10↔20s. Before this PR the `true` case was a bare `return false` that left the backoff untouched, so consecutive `false` ticks compounded normally: 10→20→40→80→…→5min.

Why it costs. Each enqueue drives a reconcile whose verifier issues LIST/GET calls against the regional cluster on an uncached client. A ServiceSet that never converges now hits that path every 10s indefinitely — and a permanently-unhealthy workload is precisely the case the 5-minute cap exists to throttle.

**The fix is narrow:** don't touch `currentBackoff`/`nextEligibleTime` in the settle-window arm, or only reset them on a genuine `false` → `true` edge rather than on every `true` tick.

test/e2e/functional_test.go
```diff
@@ -667,59 +797,89 @@ func waitForServiceSetVersions(
...
-				continue
+				return fmt.Errorf("not all expected versions observed: %+v", expectedVersions)
 			}
-
-			var svcSet kcmv1.ServiceSet
-			err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &svcSet)
-			Expect(err).NotTo(HaveOccurred(), "failed to convert unstructured to ServiceSet")
+			lastResourceVersion = obj.GetResourceVersion()
 
 			if event.Type != watch.Modified {
```
wahabmk: Only watch.Modified events update expectedVersions; the re-established watch uses an empty ResourceVersion, so replayed state arrives as Added and is discarded.

wahabmk (dismissed):

josef-hak (dismissed):
My [ksm tests](https://github.com/josef-hak/k0rdent-ksm-tests/actions/runs/34392191301/job/102603045613) passed!

internal/controller/adapters/sveltos/enqueue.go
```diff
@@ -72,20 +74,48 @@ var enqueueStates = struct {
...
-	if s.lastSeenSummaryRV != summaryRV {
-		// A real sveltos-side change since we last looked. Enqueue,
-		// reset back-off, remember the new RV.
+func (s *enqueueState) evaluate(now time.Time, summaryRV string, generation int64, deployed bool) bool {
+	if s.lastSeenSummaryRV != summaryRV || s.lastSeenGeneration != generation {
+		// A real sveltos-side change, or a newer desired Spec (e.g. a
+		// version bump) since we last looked. Enqueue, reset back-off,
+		// remember the new RV/generation.
+		//
+		// The generation half of this check matters even when quiescent:
```
josef-hak, line 83: This update adds longer and longer comments (it's AI, right?). I would avoid it. Nobody wants to read long comments. (IMHO it's also good to add ~/.claude/CLAUDE.md instruction "keep short comments in code" or so.

PR #3008, opened 2026-08-12
fix: share one RESTMapper per cluster instead of rebuilding per client

### LINKED ISSUE #3007: [ksm][bug] Regional and child clients rebuild their discovery RESTMapper on every poll tick
**Describe the bug**
Both client factories in `internal/util/kube/client.go` build their clients without an explicit `Mapper`, so controller-runtime gives each client its own dynamic discovery RESTMapper, and every fresh mapper discovers the target apiservers API surface on its first `RESTMapping`. Since these clients are built per cluster per poll tick and per reconcile, then discarded, KCM re-discovers remote apiservers continuously, resulting in load that scales with cluster count and never decays, even when nothing is changing.

PR #2816 already improved this path: the per-tick `rgnClients` map removed the repeated kubeconfig Secret read against the management cluster. The cost I am describing sits on the other side of the connection, in the client's own RESTMapper: since the map is scoped to a single tick, each clusters API surface is still discovered afresh on the next one, six times a minute per cluster at the pollers 10s interval.

Two places this shows up:

- **The sveltos ServiceSet poller** (regional deployments). It reads a Profile and a ClusterSummary for every ServiceSet on every tick. The enqueue back-off gates whether a ServiceSet is  reconciled rather than whether it is polled, so the discovery traffic does not decay once  the system settles.
- **`resolveChildClient`** (`serviceset_controller.go:1875`) builds a client against the workload cluster per reconcile, which the per-tick map does not cover. Worth flagging because it is not regional-specific: the verifier returns early only when no health rules are loaded, and KCM ships `templates/provider/kcm/templates/health-rules/default-rules.yaml` unconditionally, so flat deployments are affected too.

This looks like the same root cause as #2318 / #2439, which fixed `gvrFromResourceReference` by building one mapper at startup and reusing it. That fix was scoped to a single cluster, so it needed no keying; the regional and child client paths reach many clusters, which is presumably why they were handled separately.

**To Reproduce**

1. Setup a Management cluster plus a separate regional cluster, with child clusters whose `Credential` has `.spec.region` set.
2. Deploy services so a number of `ServiceSets` exist. The effect grows with the number of distinct `ClusterDeployment`s in the region.
3. On the regional apiserver, measure API discovery, e.g. `apiserver_response_sizes` for `verb="GET"` with empty `resource`/`group`. It sits at a sustained rate proportional to cluster count, bursting every 10s, and does not decay despite the system being idle.
5. Confirm the source by scaling the `kcm-controller-manager` to 0 and the discovery disappears; scale back and  it returns.

For the child-client path the same measurement on a workload cluster shows a discovery burst per ServiceSet reconcile, that one reproduces in a flat setup with no regional cluster involved.

Observed on a real regional deployment: ~31 ServiceSets ≈ 5–6 discovery req/s; ~90 ServiceSets ≈ ~30 req/s, with discovery dominating the apiserver's response bytes. On current `main` the per-tick map reduces that by roughly the ServiceSets-per-cluster ratio, but it stays linear in cluster count and never reaches zero.

**Expected behavior**
API discovery against a remote cluster should happen once per cluster and be reused across objects, poll ticks and reconciles, refreshing only when the clusters API surface or its credentials actually change. In steady state it should be near-zero and independent of ServiceSet and cluster count.

**Additional context**
Measured with a probe against a synthetic apiserver serving a 40-group discovery document, replaying 6 poll ticks × 30 clusters:

| | object GETs | discovery requests | discovery bytes |
|---|---|---|---|
| today | 180 | 540 | 1253.3 KB |
| one mapper per cluster | 180 | 3 | 7.0 KB |

The suggested fix is basically the remedy in #2439 but one level lower: Cache the RESTMapper per target cluster and pass it via `client.Options{Mapper: ...}`, so clients stay cheap to build and no call site changes.

srnbckr:
`DefaultClientFactory` and `DefaultClientFactoryWithRestConfig` build every regional and child cluster client without an explicit `Mapper`, so controller-runtime constructs a fresh dynamic discovery RESTMapper for each one. A fresh mapper performs a full API discovery against the target apiserver on its first `RESTMapping`, e.g. two requests, one of which carries the aggregated resource list for every group.

Because those clients are built per cluster per poll tick and per reconcile, then discarded, KCM re-discovers remote apiservers continuously. The sveltos ServiceSet poller runs every 10s and touches every ServiceSet on every tick, so discovery load against a regional apiserver scales with cluster count and never decays, even when the system is completely idle. `resolveChildClient` does the same against workload clusters on the reconcile path.

This PR caches the RESTMapper per target apiserver and passes it via `client.Options{Mapper: ...}`.

Measured while working on this, with a throwaway probe against a synthetic apiserver serving a 40-group discovery document, replaying 6 poll ticks × 30 clusters (60s at the poller's interval):

| | object GETs | discovery requests | discovery bytes |
|---|---|---|---|
| before | 180 | 540 | 1253.3 KB |
| after | 180 | 3 | 7.0 KB |

This is the same remedy as #2439, which fixed `gvrFromResourceReference` in the `StateManagementProvider` controller by building one `DeferredDiscoveryRESTMapper` at startup and reusing it.

This PR is also complementary to #2816, which already improved this path: its per-tick `rgnClients` map removed the repeated kubeconfig Secret read against the management cluster. This PR leaves that map in place and addresses a separate cost on the other side of the connection, the clients own RESTMapper, which is rebuilt whenever a client is, and so is unaffected by caching at tick scope.

Fixes #3007

### DIFF
--- a/internal/util/kube/restmapper_cache.go
+++ b/internal/util/kube/restmapper_cache.go
@@ -1,0 +1,88 @@
+// Copyright 2026
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kube
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+// restMapperCache holds one RESTMapper per remote cluster. A client.New that
+// leaves Options.Mapper unset gets a dynamic mapper of its own, and each fresh
+// mapper discovers the target apiserver's API surface on its first RESTMapping
+// — two requests, one of which carries every group on a server supporting
+// aggregated discovery.
+//
+// Keyed by apiserver URL: the API surface is a property of the cluster, not of
+// the credentials used to reach it, so this also bounds the map by cluster count
+// rather than growing on every credential rotation. The fingerprint is of the
+// kubeconfig the mapper was built from, because a mapper owns a discovery client with
+// those credentials baked in, so a rotation has to produce a new one.
+type restMapperCache struct {
+	entries map[string]restMapperEntry
+	mu      sync.RWMutex
+}
+
+type restMapperEntry struct {
+	mapper      meta.RESTMapper
+	fingerprint string
+}
+
+var sharedRESTMapperCache = &restMapperCache{entries: make(map[string]restMapperEntry)}
+
+func (c *restMapperCache) get(cfg *rest.Config, kubeconfig []byte) (meta.RESTMapper, error) {
+	sum := sha256.Sum256(kubeconfig)
+	fingerprint := hex.EncodeToString(sum[:])
+
+	c.mu.RLock()
+	entry, ok := c.entries[cfg.Host]
+	c.mu.RUnlock()
+	if ok && entry.fingerprint == fingerprint {
+		return entry.mapper, nil
+	}
+
+	httpClient, err := rest.HTTPClientFor(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create HTTP client: %w", err)
+	}
+	mapper, err := apiutil.NewDynamicRESTMapper(cfg, httpClient)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create REST mapper: %w", err)
+	}
+
+	// Re-check under the write lock and prefer whatever a racing caller stored:
+	// callers must converge on one mapper per cluster, or several would each run
+	// their own discovery.
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if cur, ok := c.entries[cfg.Host]; ok && cur.fingerprint == fingerprint {
+		return cur.mapper, nil
+	}
+	c.entries[cfg.Host] = restMapperEntry{mapper: mapper, fingerprint: fingerprint}
+
+	return mapper, nil
+}
+
+func (c *restMapperCache) len() int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return len(c.entries)
+}
--- a/internal/util/kube/restmapper_cache.go
+++ b/internal/util/kube/restmapper_cache.go
@@ -48,8 +48,9 @@ type restMapperCache struct {
 	// mu, so a test may swap it mid-flight without racing a concurrent get.
 	raceHook func()
 
-	ttl           time.Duration
-	sweepInterval time.Duration
+	ttl             time.Duration
+	refreshInterval time.Duration
+	sweepInterval   time.Duration
 
 	// generation increments on every store, so a build that started earlier can
 	// tell it is about to overwrite state a later caller installed.
@@ -59,25 +60,36 @@ type restMapperCache struct {
 	mu        sync.RWMutex
 }
 
-func newRESTMapperCache(ttl, sweepInterval time.Duration) *restMapperCache {
-	return newRESTMapperCacheWithClock(ttl, sweepInterval, time.Now)
+func newRESTMapperCache(ttl, refreshInterval, sweepInterval time.Duration) *restMapperCache {
+	return newRESTMapperCacheWithClock(ttl, refreshInterval, sweepInterval, time.Now)
 }
 
-func newRESTMapperCacheWithClock(ttl, sweepInterval time.Duration, nowFunc func() time.Time) *restMapperCache {
+func newRESTMapperCacheWithClock(ttl, refreshInterval, sweepInterval time.Duration, nowFunc func() time.Time) *restMapperCache {
 	return &restMapperCache{
-		entries:       make(map[string]*restMapperEntry),
-		nowFunc:       nowFunc,
-		ttl:           ttl,
-		sweepInterval: sweepInterval,
+		entries:         make(map[string]*restMapperEntry),
+		nowFunc:         nowFunc,
+		ttl:             ttl,
+		refreshInterval: refreshInterval,
+		sweepInterval:   sweepInterval,
 	}
 }
 
 type restMapperEntry struct {
-	mapper               meta.RESTMapper
+	mapper meta.RESTMapper
+	// createdAt is set once at store time and never refreshed by hits, unlike
+	// lastUsed, so an entry's absolute age keeps growing while it is in use.
+	createdAt            time.Time
 	fingerprint          string
 	canonicalFingerprint string
 	generation           uint64
 	lastUsed             atomic.Int64
+}
+
+// aged reports whether the entry has passed the absolute rebuild deadline. An
+// aged entry is treated as a miss even if recently used; lastUsed drives idle
+// eviction only.
+func (e *restMapperEntry) aged(now time.Time, refreshInterval time.Duration) bool {
+	return now.Sub(e.createdAt) >= refreshInterval
 }
 
 const (
@@ -85,13 +97,20 @@ const (
 	// clusters do not retain a discovery cache for the life of the process.
 	restMapperTTL = time.Hour
 
+	// restMapperRefreshInterval bounds a mapper's absolute age. The dynamic
+	// mapper re-discovers only on a NoMatch and serves a mapping it already
+	// knows from memory forever, so a cluster looked up more often than the TTL
+	// would otherwise keep a removed API version or a changed CRD scope alive
+	// until the process restarts. An aged entry is rebuilt on its next lookup.
+	restMapperRefreshInterval = 30 * time.Minute
+
 	// restMapperSweepInterval is how often eviction runs. It is driven by elapsed
 	// time rather than by cache misses: deleting a cluster produces no miss, so a
 	// miss-driven sweep would never reclaim a fleet that only shrinks.
 	restMapperSweepInterval = 10 * time.Minute
 )
 
-var sharedRESTMapperCache = newRESTMapperCache(restMapperTTL, restMapperSweepInterval)
+var sharedRESTMapperCache = newRESTMapperCache(restMapperTTL, restMapperRefreshInterval, restMapperSweepInterval)
 
 func normalizeHost(host string) string { return strings.TrimRight(host, "/") }
 
@@ -103,7 +122,7 @@ func (c *restMapperCache) get(cfg *rest.Config, kubeconfig []byte) (meta.RESTMap
 
 	c.mu.RLock()
 	entry, ok := c.entries[host]
-	if ok && entry.fingerprint == fingerprint {
+	if ok && entry.fingerprint == fingerprint && !entry.aged(now, c.refreshInterval) {
 		entry.lastUsed.Store(now.UnixNano())
 		c.mu.RUnlock()
 		return entry.mapper, nil
@@ -125,7 +144,7 @@ func (c *restMapperCache) get(cfg *rest.Config, kubeconfig []byte) (meta.RESTMap
 		entry, ok = c.entries[host]
 		if ok {
 			observedGeneration = entry.generation
-			if entry.canonicalFingerprint == canonicalFingerprint {
+			if entry.canonicalFingerprint == canonicalFingerprint && !entry.aged(now, c.refreshInterval) {
 				entry.lastUsed.Store(now.UnixNano())
 				c.mu.RUnlock()
 				return entry.mapper, nil
@@ -157,7 +176,7 @@ func (c *restMapperCache) get(cfg *rest.Config, kubeconfig []byte) (meta.RESTMap
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	cur, ok := c.entries[host]
-	if ok && (cur.fingerprint == fingerprint || cur.canonicalFingerprint == canonicalFingerprint) {
+	if ok && !cur.aged(now, c.refreshInterval) && (cur.fingerprint == fingerprint || cur.canonicalFingerprint == canonicalFingerprint) {
 		cur.lastUsed.Store(now.UnixNano())
 		return cur.mapper, nil
 	}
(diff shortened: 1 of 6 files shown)


internal/util/kube/restmapper_cache.go
```diff
@@ -0,0 +1,243 @@
...
+	host := normalizeHost(cfg.Host)
+	fingerprint := fingerprint(kubeconfig)
+
+	c.mu.RLock()
+	entry, ok := c.entries[host]
+	if ok && entry.fingerprint == fingerprint {
+		entry.lastUsed.Store(now.UnixNano())
+		c.mu.RUnlock()
+		return entry.mapper, nil
+	}
```
zerospiel: Active clusters can retain stale REST mappings forever

Here and `internal/util/kube/restmapper_cache.go:191-197`

A cache hit refreshes `lastUsed`, and eviction considers only `lastUsed`. A mapper for any cluster accessed more often than once per hour therefore never expires. controller-runtime's dynamic mapper refreshes after a `NoMatch`, but it does not revalidate a mapping that is already known. Removing an API version, changing a CRD's scope, or replacing an API while retaining its GroupKind can leave the old mapping usable in memory indefinitely. Clients then keep selecting the stale resource path or scope until KCM restarts.

Proposed fix: Track an absolute mapper age independently of idle usage, for example `createdAt` plus `restMapperRefreshInterval`. Rebuild the entry when that deadline passes even if `lastUsed` is recent; use `lastUsed` only for idle eviction. Add a clock-driven test that repeatedly hits an entry, advances beyond the refresh interval, and verifies that the mapper is replaced

internal/util/kube/restmapper_cache.go
```diff
@@ -0,0 +1,243 @@
...
+)
+
+var sharedRESTMapperCache = newRESTMapperCache(restMapperTTL, restMapperSweepInterval)
+
+func normalizeHost(host string) string { return strings.TrimRight(host, "/") }
+
+func (c *restMapperCache) get(cfg *rest.Config, kubeconfig []byte) (meta.RESTMapper, error) {
+	now := c.nowFunc()
+	c.maybeSweep(now)
+	host := normalizeHost(cfg.Host)
```
zerospiel: One cache slot per host thrashes for valid multi-identity access

Here `internal/util/kube/restmapper_cache.go:101-127` and `internal/util/kube/restmapper_cache.go:158-177`

The map is keyed only by API server URL, but the mapper's discovery transport is credential-specific. Multiple kubeconfigs can legitimately address the same API server through different service accounts, impersonation settings, or cluster aliases. Alternating those kubeconfigs continuously replaces the host's sole entry, so each identity rebuilds a mapper and performs discovery again. Concurrent callers can also receive uncached loser mappers. This defeats the PR's network-call reduction for that topology

Proposed fix: Allow multiple credential variants per normalized host. Key entries by host plus the effective credential/config fingerprint, retain the raw-fingerprint fast path, and let the global size/idle policy remove superseded credentials. Prefer a stable cluster/Secret identity supplied by the caller if the factory API can be extended. Add a test alternating two users on one host and assert that each user reuses its original mapper

internal/util/kube/restmapper_cache.go
```diff
@@ -0,0 +1,243 @@
...
+	// time rather than by cache misses: deleting a cluster produces no miss, so a
+	// miss-driven sweep would never reclaim a fleet that only shrinks.
+	restMapperSweepInterval = 10 * time.Minute
+)
+
+var sharedRESTMapperCache = newRESTMapperCache(restMapperTTL, restMapperSweepInterval)
+
+func normalizeHost(host string) string { return strings.TrimRight(host, "/") }
+
+func (c *restMapperCache) get(cfg *rest.Config, kubeconfig []byte) (meta.RESTMapper, error) {
```
zerospiel: The cache is neither size-bounded nor guaranteed to honor its TTL

Sweeping runs only from `get`. If the last clusters are deleted and no later factory call occurs, their mappers, discovery data, transports, and credential material remain reachable for the process lifetime despite the one-hour TTL. Separately, there is no entry limit, so high host churn within one TTL window can grow the map without bound.

Proposed fix: Add an explicit maximum entry count with LRU eviction, and run TTL cleanup from a lifecycle-owned periodic sweeper rather than only from lookups. If a background sweeper is undesirable, the size bound is still required and the TTL documentation must state that expiry is opportunistic. Add tests for capacity eviction and for expiry without a subsequent cache lookup

internal/util/kube/restmapper_cache.go
```diff
@@ -0,0 +1,243 @@
...
+		return entry.mapper, nil
+	}
+	c.mu.RUnlock()
+
+	var (
+		canonicalFingerprint string
+		observedGeneration   uint64
+		err                  error
+	)
+	if ok {
```
zerospiel: Canonically equivalent kubeconfigs stay on the allocation-heavy path

When new bytes have the same canonical fingerprint as the cached kubeconfig, `get` returns the mapper but leaves `entry.fingerprint` unchanged. Every later lookup with the new bytes repeats `clientcmd.Load`, normalization, serialization, allocation, and hashing. A semantically neutral Secret rewrite can therefore put every poll for that cluster permanently on the expensive path

propsed fix: after confirming canonical equality under the write lock, promote the current raw fingerprint into the entry before returning. Add a test with an injected or counted canonicalizer proving that only the first lookup of the alternate byte representation canonicalizes it

internal/util/kube/restmapper_cache.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	c.mu.RLock()
+	raceHook := c.raceHook
+	c.mu.RUnlock()
+	if raceHook != nil {
+		raceHook()
+	}
+
+	httpClient, err := rest.HTTPClientFor(cfg)
```
zerospiel, line 182: the cached discovery HTTP client is not reused by the object client

`restMapperCache.get` creates an HTTP client for the mapper, then `client.New` creates another because `client.Options.HTTPClient` is unset. client-go can share its base TLS transport for cacheable configurations, but the extra client and wrapper construction remains; configurations with uncacheable transports, such as a kubeconfig proxy function, can also create separate connection pools. The cache already has the correctly credentialed HTTP client needed to avoid this work

Proposed fix: store the `*http.Client` with each mapper entry, return both from `get`, and pass `client.Options{Scheme: scheme, Mapper: mapper, HTTPClient: httpClient}`. Return the winning entry's pair after concurrent convergence so a client never combines a mapper with another credential generation's transport

internal/util/kube/restmapper_cache.go
```diff
@@ -0,0 +1,243 @@
...
+
+	ttl           time.Duration
+	sweepInterval time.Duration
+
+	// generation increments on every store, so a build that started earlier can
+	// tell it is about to overwrite state a later caller installed.
+	generation uint64
+
+	lastSweep atomic.Int64
+	mu        sync.RWMutex
```
zerospiel: Nitpicking: Until it is proven that we have read-heavy workloads, I would suggest using an ordinary `Mutex` instead of RWMutex here. Or please provide sound benchmarks where we can justify the usage of the latter here.

zerospiel (commented):
Thank you for the thoughtful PR and for finding the bottleneck! I've left several comments, mostly regarding further optimizations. Please get through them and let me know your thoughts on them

zerospiel:
@srnbckr we could merge the PR in its current state and I can prepare a new PR addressing the open questions, wdyt?

srnbckr:
Hey @zerospiel, thanks for the review and feedback! I was off work / the laptop for a day, but I've got some time over the weekend and could address your comments if that works for you. Otherwise, feel free to merge the PR in it's current state and prepare a follow-up PR, I'm totally fine with that as well.

zerospiel:
> Hey @zerospiel, thanks for the review and feedback! I was off work / the laptop for a day, but I've got some time over the weekend and could address your comments if that works for you. Otherwise, feel free to merge the PR in it's current state and prepare a follow-up PR, I'm totally fine with that as well.

got it, thanks for the elaboration; if you have some spare time, then i'd gladly wait, we have plenty of time

internal/util/kube/restmapper_cache.go
srnbckr: I've addressed this in ae097021: I picked 30min for the interval (one re-discovery per cluster per half hour, vs. ~6/min before the PR), but happy to tune it if you'd prefer a different bound, wdyt?

internal/util/kube/restmapper_cache.go
srnbckr: Addressed this in ea77284e: keyed by host + canonical kubeconfig fingerprint, raw-fingerprint fast path kept via an alias index, rotations add an entry that ages out via the TTL. I deferred the caller-supplied Secret identity (factory API change across several call sites), but I can to follow up on that separately if you think it's worth it.

internal/util/kube/restmapper_cache.go
srnbckr: Addressed in cccefdba: 256-entry cap with LRU eviction on insert, and TTL expiry moved to a manager-registered sweeper.  I sized the cap at 256 assuming thats well above the realistic per-process fleet size - wdyt?

internal/util/kube/restmapper_cache.go
srnbckr: Addressed in ea77284e

internal/util/kube/restmapper_cache.go
srnbckr: Addressed in feda5aab. I also found that a prebuilt transport skips controller-runtime user-agent defaulting, so the cache now sets the default UA itself (with a regression test), sounds good?

internal/util/kube/restmapper_cache.go
srnbckr: fair suggestion, I've changed it to a Mutex.

srnbckr:
> > Hey @zerospiel, thanks for the review and feedback! I was off work / the laptop for a day, but I've got some time over the weekend and could address your comments if that works for you. Otherwise, feel free to merge the PR in it's current state and prepare a follow-up PR, I'm totally fine with that as well.
> 
> got it, thanks for the elaboration; if you have some spare time, then i'd gladly wait, we have plenty of time

Thanks again for the thorough review @zerospiel. I've tried to address all your comments, please let me know what you think and if further changes are needed.

internal/util/kube/restmapper_cache.go
zerospiel: Sounds good

internal/util/kube/restmapper_cache.go
zerospiel: Yep, thanks for noticing! The default UA also looks fine

internal/util/kube/restmapper_cache.go
```diff
@@ -0,0 +1,334 @@
...
+	// another credential generation's transport.
+	httpClient *http.Client
+	// createdAt is set once at store time and never refreshed by hits, unlike
+	// lastUsed, so an entry's absolute age keeps growing while it is in use.
+	createdAt time.Time
+	// rawFingerprint is the entry's current byte representation — the one the
+	// aliases index resolves. Mutated only under the cache's lock, when a
+	// promotion swaps it for a newer representation.
+	rawFingerprint string
+	lastUsed       atomic.Int64
```
zerospiel: After changing `RWMutex` to `Mutex`, every read and write of `lastUsed` occurs while `c.mu` is held. Atomic here no longer provides safety or concurrency and adds sequentially consistent atomic operations to hits while obscuring the locking invariant

Would it make sense to change `lastUsed` to `int64` and use direct reads and writes under `c.mu`?

But this is a nitpicking comment, thus may be addressed in a follow-up PR by either you or any of the maintainers

zerospiel: But this is a nitpicking comment, thus may be addressed in a follow-up PR by either you or any of the maintainers

internal/util/kube/restmapper_cache.go
```diff
@@ -0,0 +1,334 @@
...
+// lookups hash the bytes and skip parsing; an equivalent but byte-different
+// kubeconfig (e.g. a reordered Secret rewrite) is canonicalized once, promoted
+// into the index, and rides the fast path thereafter. Promotion swaps the
+// entry's single alias rather than accumulating every representation seen, so
+// len(aliases) == len(entries) always and the index cannot outgrow the cache.
+type restMapperCache struct {
+	entries map[string]*restMapperEntry
+	// aliases maps an entry's current raw kubeconfig fingerprint to its
+	// entries key.
+	aliases map[string]string
```
zerospiel, line 58: The hot path computes sha256 and then `hex.EncodeToString` allocates a 64-byte string for every factory call. Canonical keys also concatenate host, separator, and another encoded digest. It'd suggest to represent fingerprints as `[sha256.Size]byte`, use that type directly in `aliases`, and use a struct such as `{host string; fingerprint [sha256.Size]byte}` as the `entries` key. 

But this is also a nitpicking comment, since kubeconfig parsing still dominates clietn construction. So this comment is for a follow-up (if any) PR

internal/util/kube/restmapper_cache.go
zerospiel: So this is a really good approach! I was thinking about what would happen if >256 are in the fleet: every 10secs the poller fetches all of the objects, and then lazily evicts the first stale identity, thus effectively eliminating the hit rate.

I would suggest introducing a configurable cache size (probably via an env to avoid propagating the key through a bunch of structs). But as usual: this could be done as part of a follow-up to avoid blocking this PR

zerospiel (dismissed):
LGTM taking the following comments into the account as potential fixes in a follow up PR

https://github.com/k0rdent/kcm/pull/3008#discussion_r3794707680
https://github.com/k0rdent/kcm/pull/3008#discussion_r3794645472
https://github.com/k0rdent/kcm/pull/3008#discussion_r3794682065

zerospiel:
@srnbckr feel free to comment/resolve any outstanding convos. If you're okay, then I'll proceed with merging and start preparing a follow-up (or you could do this, just let me know)

internal/util/kube/restmapper_cache.go
srnbckr: Good catch, I've changed it to in64 in 41be4ac now

internal/util/kube/restmapper_cache.go
srnbckr: yes, I agree that an env var for the cache size sounds right to not turn the poller into an enviction loop. I would do that in a follow up PR as you suggested.

internal/util/kube/restmapper_cache.go
srnbckr: Agreed, your suggested approach sounds cleaner. Since you flagged it as non-blocking, I'll include it in a follow-up PR.

zerospiel (approved):
LGTM

srnbckr:
> @srnbckr feel free to comment/resolve any outstanding convos. If you're okay, then I'll proceed with merging and start preparing a follow-up (or you could do this, just let me know)

Sounds good @zerospiel, from my side this is ready to merge. I've pulled the lastUsed atomic fix into this PR because I figured it was small enough. 
For the follow-up PR(s) I'll probably have some time at the end of the week and can take it. If you'd like it sooner, feel free to go ahead yourself. 

Thanks again for the great review, the comments were spot on.

zerospiel:
Merged, thanks for the PR and quick communication! We will wait then for the follow-up this week; otherwise, I'll try to pick it up next week.

PR #3010, opened 2026-08-14
feat: access management redesign

eromanova:
## What this PR does / why we need it

`AccessManagement` currently hardcodes one field per distributable Kind (`ClusterTemplateChains`, `ServiceTemplateChains`, `Credentials`, `ClusterAuthentications`, `DataSources`, `ClusterAuditPolicies`), with no way to reference a Kind the API author didn't anticipate and no selector support outside `TargetNamespaces`. This PR replaces that shape with a single generic `resources` list per `AccessRule` and reworks the controller around it.

## API changes

* New `AccessRule.Resources []ResourceRule`, each entry `{apiVersion, kind, names, selector, stringSelector}` selects objects of *any* namespaced Kind (built-in or custom CRD) instead of one hardcoded field per Kind.
* `TargetNamespaces` is now a type alias of the new, generalized `ResourceSelector` type (pure Go-side rename, no wire-format change).
* The six old fields are kept, marked `Deprecated`, to be removed in a future API version.
* New `AccessManagementStatus.Resources []ResourceKindStatus` per-Kind status, since a generic mechanism can fail per-Kind independently, in addition to the existing aggregate `Status.Error`.

## Migration path

* A mutating webhook (`AccessManagementValidator.Default`) migrates the deprecated fields into equivalent `Resources` entries on every create/update, clearing the deprecated fields once migrated (idempotent).
* The controller does **not** depend on the webhook having run: each `AccessRule` is resolved via `EffectiveResources()`, which falls back to synthesizing entries from the deprecated fields, live, on every reconcile, whenever `Resources` is empty. So an object with only deprecated fields set (e.g. because the admission webhook is disabled) keeps being fully reconciled (created/kept/cleaned up) instead of freezing or losing previously-distributed objects.
* When both are populated on the same rule, `Resources` wins outright; the deprecated field is ignored, never merged.

## RBAC changes

* **Dynamic RBAC:** the controller no longer relies solely on a static per-Kind manifest. `ensureDynamicRBAC` maintains a `<name>-resources` `ClusterRole`, label-aggregated into the existing manager `ClusterRole`, granting exactly `get/list/watch/create/delete` on the GVRs currently referenced (minus dangerous/cluster-scoped ones) access grows and shrinks with the spec instead of being pre-provisioned or wildcard.


BROngineer:
The only question - why mutating webhook and not the new API version with two-way conversion? As far as I understand, the migration will be one-way only:

```go
func migrateAccessRule(rule AccessRule) (AccessRule, bool) {
	changed := false

	if synthesized := synthesizeResourceRules(rule); len(synthesized) > 0 {
		rule.Resources = append(rule.Resources, synthesized...)
		changed = true
	}

	if rule.HasDeprecatedFields() {
		rule.ClusterTemplateChains = nil
		rule.ServiceTemplateChains = nil
		rule.Credentials = nil
		rule.ClusterAuthentications = nil
		rule.DataSources = nil
		rule.ClusterAuditPolicies = nil
	}

	for i, res := range rule.Resources {
		if res.APIVersion == "" {
			rule.Resources[i].APIVersion = GroupVersion.String()
			changed = true
		}
	}

	return rule, changed
}
```

thus it will be impossible to rollback after upgrade?

eromanova:
@BROngineer Thanks for the question. I personally think we don't need a new API version for this.
A "convert both ways" approach only actually protects you if the conversion never loses information, but here it can. The whole point of this change is letting people point at any kind of object, not just the original six. If someone uses that new freedom, there's no old-style field to convert it back into (when using a custom kind). So this approach would only help people who never used the new feature, while costing us a lot of new, permanent complexity to build and maintain (we don't have any of that today).

And since there's only ever one AccessManagement per cluster, a simple "back it up before you upgrade" step covers us just fine, without needing to build a whole new versioning system for it.

Happy to discuss it.

internal/webhook/accessmanagement_webhook.go
```diff
@@ -82,3 +101,55 @@ func (v *AccessManagementValidator) ValidateDelete(ctx context.Context, _ *kcmv1
...
+	if err != nil {
+		return fmt.Errorf("failed to resolve %s (ensure the CRD is installed): %w", gvk, err)
+	}
+
+	if mapping.Scope.Name() != apimeta.RESTScopeNameNamespace {
+		return fmt.Errorf("%s is cluster-scoped and cannot be distributed into target namespaces by AccessManagement", gvk)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
eromanova: It was decided not to introduce a list of dangerous resources allowing to copy any resource for the person who has access to edit AccessManagement object. The PR description was outdated.

internal/controller/accessmanagement_controller.go
```diff
@@ -174,374 +241,309 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
...
 	}
 
-	if err := r.processDataSources(ctx, accessMgmt, rule.DataSources, targetNamespace, resources.dataSources, keeper.dataSources); err != nil {
-		errs = errors.Join(errs, fmt.Errorf("failed to process DataSources: %w", err))
+	if mapping.Scope.Name() != apimeta.RESTScopeNameNamespace {
+		l := ctrl.LoggerFrom(ctx)
+		l.Info("skipping cluster-scoped Kind: AccessManagement can only distribute namespaced Kinds", "gvk", gvk.String())
+		r.warnf(accessMgmt, "ClusterScopedKindSkipped", "Skipping %s: cluster-scoped Kinds cannot be distributed by AccessManagement", gvk)
+		return nil, errClusterScopedKindSkipped
 	}
```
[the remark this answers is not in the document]
eromanova: It was decided not to introduce a list of dangerous resources allowing to copy any resource for the person who has access to edit AccessManagement object. The PR description was outdated.

templates/provider/kcm/templates/rbac/controller/roles.yaml
```diff
@@ -32,6 +32,18 @@ rules:
   - rbac.authorization.k8s.io
   resources:
   - clusterroles
+  verbs:
+  - create
+  - watch
+  - get
+  - list
```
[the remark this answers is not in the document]
eromanova: `create/update/watch/get/list/bind/escalate` on clusterroles all predate this PR. This PR only adds `delete`, needed to tear down the `kcm-resources` ClusterRole when no Kinds are referenced anymore.

internal/controller/accessmanagement_controller.go
```diff
@@ -137,35 +140,99 @@ func (r *AccessManagementReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	pollInterval := r.pollInterval
+	if pollInterval == 0 {
+		pollInterval = defaultPollInterval
+	}
+
+	return ctrl.Result{RequeueAfter: pollInterval}, nil
```
zerospiel: why couldn't the existing poller wrapper be used instead of endless re-queuing?

internal/controller/accessmanagement_controller.go
```diff
@@ -174,374 +241,315 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
...
-	return systemTemplateChains, managedTemplateChains, nil
+	_ = unstructured.SetNestedField(obj.Object, newNamespace, fields...)
 }
 
-func (r *AccessManagementReconciler) getCredentials(ctx context.Context) (map[string]*kcmv1.CredentialSpec, []client.Object, error) {
-	credentialList := &kcmv1.CredentialList{}
-	err := r.List(ctx, credentialList)
-	if err != nil {
-		return nil, nil, err
+func rewriteNamespaceIfEmpty(obj *unstructured.Unstructured, sourceNamespace string, fields ...string) {
```
zerospiel: nit: try to keep new funcs as methods as long as they are not being utilized by other controllers within the package to avoid the possibility of misusing them

applicable elsewhere

internal/controller/accessmanagement_controller.go
```diff
@@ -174,374 +241,315 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
...
+		return false, fmt.Errorf("failed to resolve %s: %w", gvk, err)
+	}
 
-		if cap.GetLabels()[kcmv1.KCMManagedLabelKey] == kcmv1.KCMManagedLabelValue {
-			managedClusterAuditPolicies = append(managedClusterAuditPolicies, &cap)
+	if err := r.DynamicClient.Resource(mapping.Resource).Namespace(obj.GetNamespace()).Delete(ctx, obj.GetName(), metav1.DeleteOptions{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
 		}
+		return false, err
```
zerospiel, line 598:
```suggestion
	if err := r.DynamicClient.Resource(mapping.Resource).Namespace(obj.GetNamespace()).Delete(ctx, obj.GetName(), metav1.DeleteOptions{}); client.IgnoreNotFound(err) != nil {
		return false, err
```

nit-picking and applicable elsewhere, up to the implementer whether to apply or not

internal/controller/accessmanagement_controller.go
```diff
@@ -1013,6 +844,91 @@ func (r *AccessManagementReconciler) accessManagementAffectedByNamespaceLabelUpd
...
+		return client.IgnoreNotFound(r.Delete(ctx, clusterRole))
+	case equality.Semantic.DeepEqual(clusterRole.Rules, rules) && clusterRole.Labels[aggregateToManagerLabelKey] == aggregateToManagerLabelValue:
+		return nil
+	default:
+		clusterRole.Rules = rules
+		if clusterRole.Labels == nil {
+			clusterRole.Labels = make(map[string]string)
+		}
+		clusterRole.Labels[aggregateToManagerLabelKey] = aggregateToManagerLabelValue
+		return r.Update(ctx, clusterRole)
```
zerospiel, line 951: can a `clusterrole` object be updated? i mean, in terms of the `.rules` field

internal/controller/accessmanagement_controller.go
```diff
@@ -174,374 +241,315 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
...
-			managedClusterAuditPolicies = append(managedClusterAuditPolicies, &cap)
+	if err := r.DynamicClient.Resource(mapping.Resource).Namespace(obj.GetNamespace()).Delete(ctx, obj.GetName(), metav1.DeleteOptions{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
 		}
+		return false, err
 	}
 
-	return systemClusterAuditPolicies, managedClusterAuditPolicies, nil
+	l.Info(gvk.Kind+" was successfully deleted", "namespace", obj.GetNamespace(), "name", obj.GetName())
```
zerospiel:
```suggestion
	ctrl.LoggerFrom(ctx).Info(gvk.Kind+" was successfully deleted", "namespace", obj.GetNamespace(), "name", obj.GetName())
```

nit to avoid extra bits in case of an error

internal/controller/accessmanagement_controller.go
```diff
@@ -174,374 +241,315 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
...
-	for _, chain := range templateChains {
-		if chain.GetNamespace() == r.SystemNamespace {
-			systemTemplateChains[chain.GetName()] = chain
-			continue
-		}
-
-		if chain.GetLabels()[kcmv1.KCMManagedLabelKey] == kcmv1.KCMManagedLabelValue {
-			managedTemplateChains = append(managedTemplateChains, chain)
-		}
+func rewriteNamespaceIfSet(obj *unstructured.Unstructured, newNamespace string, fields ...string) {
```
zerospiel: what if the rewrite has failed? in this case, we expect that everything is in place but it may not be. might it be a good idea to signal whether it was successful or not?

internal/controller/accessmanagement_controller.go
```diff
@@ -174,374 +241,315 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
...
+			return false, nil
 		}
+		return false, err
 	}
-	return errs
-}
 
-func getNamespacedName(namespace, name string) string {
-	return namespace + "/" + name
+	l.Info(gvk.Kind+" was successfully created", "target namespace", targetNamespace, "source name", source.GetName())
```
zerospiel:
```suggestion
	ctrl.FromContext(ctx).Info(gvk.Kind+" was successfully created", "target namespace", targetNamespace, "source name", source.GetName())
```

same nit-picking comment

internal/controller/accessmanagement_controller.go
```diff
@@ -174,374 +241,315 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
...
+		return false, fmt.Errorf("failed to resolve %s: %w", gvk, err)
+	}
 
-		if deleted {
-			r.eventf(accessMgmt, kind+"Deleted", "Successfully deleted %s %s", kind, namespacedName)
+	if _, err := r.DynamicClient.Resource(mapping.Resource).Namespace(targetNamespace).Create(ctx, target, metav1.CreateOptions{}); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return false, nil
 		}
+		return false, err
```
zerospiel, line 501:
```suggestion
	if _, err := r.DynamicClient.Resource(mapping.Resource).Namespace(targetNamespace).Create(ctx, target, metav1.CreateOptions{}); client.IgnoreAlreadyExists(err) != nil {
		return false, err
```

the same nit-picking: application is up to the implementer

internal/controller/accessmanagement_controller.go
```diff
@@ -174,374 +241,315 @@ func (r *AccessManagementReconciler) reconcileObj(ctx context.Context, accessMgm
...
+	selector, selectorNonEmpty, err := buildLabelSelector(rule.Selector, rule.StringSelector)
+	if err != nil {
+		return nil, fmt.Errorf("failed to construct selector: %w", err)
+	}
 
-		if created {
-			r.eventf(accessMgmt, "DataSourceCreated", "Successfully created DataSource %s/%s", targetNamespace, dsName)
+	names := make([]string, 0, len(system))
+	for name, obj := range system {
+		if !selectorNonEmpty || selector.Matches(labels.Set(obj.GetLabels())) {
```
zerospiel, line 458: empty selector means `match everything`, doesn't it? so, this is expected, that we match here as well?

api/v1beta1/accessmanagement_types.go
```diff
@@ -33,55 +36,267 @@ type AccessManagementSpec struct {
...
+// anything, so callers (the mutating webhook) can skip a write when the rules are already
+// fully migrated and defaulted.
+//
+// Migration is additive and idempotent: deprecated fields are cleared once translated, and
+// re-running against an already-migrated AccessRule is a no-op. If both a deprecated field and
+// a Resources entry for the same Kind are present (e.g. mid-migration hand-edited YAML), the
+// translated entry is appended rather than merged with the existing one; duplicate names in
+// the resolved set are harmless since the controller de-duplicates by namespaced name when
+// distributing objects.
+func MigrateAccessRules(rules []AccessRule) ([]AccessRule, bool) {
```
zerospiel: shouldn't it be a method of `AccessManagement`?

zerospiel (commented):
Looks good, left some questions/nit-picking comments

internal/controller/accessmanagement_controller.go
eromanova: I didn't get the question. Do you mean whether the `clusterrole` `rules` field is immutable? Or, in terms of this PR, can the clusterrole `rules` be updated?

internal/controller/accessmanagement_controller.go
eromanova: Thanks, Michael. I'd keep the previous approach. `client.IgnoreNotFound(err) != nil` is only true for a real error. Both `deleted successfully` (err == nil) and `already gone` (err is NotFound) fall through to the same code path (logger that logs that the object was deleted, and `return deleted == true`). We need to log that the object was deleted and return deleted == true only when the object was actually deleted, so we need to differentiate the error.

internal/controller/accessmanagement_controller.go
eromanova: Same as previously. See: https://github.com/k0rdent/kcm/pull/3010#discussion_r3811618258

internal/controller/accessmanagement_controller.go
```diff
@@ -1013,6 +844,91 @@ func (r *AccessManagementReconciler) accessManagementAffectedByNamespaceLabelUpd
...
+		// only noticing on the next poll.
+		if err := controllerutil.SetControllerReference(accessMgmt, desired, r.Scheme()); err != nil {
+			return fmt.Errorf("failed to set owner reference on ClusterRole %s: %w", name, err)
+		}
+		return r.Create(ctx, desired)
+	case err != nil:
+		return fmt.Errorf("failed to get ClusterRole %s: %w", name, err)
+	case len(rules) == 0:
+		return client.IgnoreNotFound(r.Delete(ctx, clusterRole))
+	case equality.Semantic.DeepEqual(clusterRole.Rules, rules) && clusterRole.Labels[aggregateToManagerLabelKey] == aggregateToManagerLabelValue:
```
zerospiel:
```suggestion
	case clusterRole.Labels[aggregateToManagerLabelKey] == aggregateToManagerLabelValue && equality.Semantic.DeepEqual(clusterRole.Rules, rules):
```

short-circuit evaluation; please check additionally whether there might be extra places in the code that could benefit from this

internal/controller/accessmanagement_controller.go
zerospiel: The former, yes. IIRC, the `.rules` field is immutable, and it seemed that this path still can be reached. Now I realized there was a dedicated case checking the old `.rules` against the new `.rules`, but still, if the rules are not equal — we are trying to update them, and that would lead to an error, again, IIRC

internal/controller/accessmanagement_controller.go
zerospiel: Correct, I've overlooked it thanks for the elaboration

internal/controller/accessmanagement_controller.go
eromanova: The `clusterrole` `.rules` is not immutable, so, it should be fine then :)

zerospiel: Ah, okay, then I mixed it with either another field or simply with another resource. Thanks for the clarification

internal/controller/accessmanagement_controller.go
eromanova: Thanks for this finding. I think an empty label selector here shouldn't match all objects, to prevent accidental copying of everything (unlike with namespaces).

zerospiel (dismissed):
LGTM

internal/webhook/accessmanagement_webhook.go
```diff
@@ -29,22 +30,40 @@ import (
...
+var (
+	_ admission.Validator[*kcmv1.AccessManagement] = &AccessManagementValidator{}
+	_ admission.Defaulter[*kcmv1.AccessManagement] = &AccessManagementValidator{}
+)
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
-func (v *AccessManagementValidator) ValidateCreate(ctx context.Context, _ *kcmv1.AccessManagement) (admission.Warnings, error) {
+func (v *AccessManagementValidator) ValidateCreate(ctx context.Context, am *kcmv1.AccessManagement) (admission.Warnings, error) {
 	itemsList := &metav1.PartialObjectMetadataList{}
 	itemsList.SetGroupVersionKind(kcmv1.GroupVersion.WithKind(kcmv1.AccessManagementKind))
```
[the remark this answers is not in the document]
eromanova: We'll address it separately since it's not related to the current PR.

api/v1beta1/accessmanagement_types.go
```diff
@@ -294,8 +307,10 @@ func migrateAccessRule(rule AccessRule) (AccessRule, bool) {
 	}
 
 	for i, res := range rule.Resources {
-		if res.APIVersion == "" {
-			rule.Resources[i].APIVersion = GroupVersion.String()
+		// Only a built-in Kind can be safely defaulted here: for any other Kind, an omitted
+		// APIGroup means the core group (see ResourceRule.GroupKind), so it must be left alone.
```
zerospiel: > see ResourceRule.GroupKind

the comment seems off

zerospiel (approved):

PR #3022, opened 2026-08-19
refactor: key the RESTMapper cache by digests and make its size configurable

srnbckr:
Follow-up to #3008, addressing the two review suggestions deferred there.

Changes:

**1. Key the cache by raw sha256 digests**
Every factory call hashed the kubeconfig and then hex-encoded the digest, allocating a 64-byte string per lookup, and entry keys concatenated host, separator, and a second encoded digest. Fingerprints are now `[sha256.Size]byte` used directly as the alias keys, and the entries key is a `{host, fingerprint}` struct, as [suggested here](https://github.com/k0rdent/kcm/pull/3008#discussion_r3794682065).

**2. Make the cache size configurable**
A fleet with more simultaneously live cluster identities than the fixed 256-entry cap would turn every poll tick into an eviction cycle, rebuilding mappers and re-running discovery against the same apiservers the cache exists to protect.

The cap can now be overridden with the `RESTMAPPER_CACHE_MAX_ENTRIES` env var (falling back to 256 when unset or invalid), exposed as `controller.restMapperCacheMaxEntries` in the KCM chart and `telemetry.controller.restMapperCacheMaxEntries` in the regional chart. This approach was [suggested here](https://github.com/k0rdent/kcm/pull/3008#discussion_r3794707680).

Follow-up to #3008

### DIFF
--- a/internal/util/kube/restmapper_cache.go
+++ b/internal/util/kube/restmapper_cache.go
@@ -17,8 +17,11 @@ package kube
 import (
 	"context"
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"net/http"
+	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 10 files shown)


zerospiel (approved):
LGTM, thanks for the extensive tests, including helm ones!

PR #3037, opened 2026-08-27
fix: stuck serviceset reconcile

BROngineer:
This PR fixes how the serviceset controller handles the serviceset reconciliation and re-queuing. So that the serviceset will be reconciled until it enters the state confirmed by corresponding sveltos objects - clustersummaries, clusterconfigurations, etc.
Aside from that this PR fixes the stale service status which was not properly updated in case there were no applicable health rules found.

### DIFF
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -372,16 +372,17 @@ func (r *ServiceSetReconciler) verifyServiceStates(ctx context.Context, rgnClien
 	specVersions := make(map[client.ObjectKey]*string, len(serviceSet.Spec.Services))
 	for i := range serviceSet.Spec.Services {
 		svc := &serviceSet.Spec.Services[i]
-		specVersions[client.ObjectKey{Namespace: svc.Namespace, Name: svc.Name}] = svc.Version
+		specVersions[serviceset.ServiceKey(svc.Namespace, svc.Name)] = svc.Version
 	}
 
-	// Only the health pass reaches the target cluster; with no rules to
-	// evaluate, building the client would be pure cost.
-	var childClient client.Client
+	var assessServiceState func(serviceName, serviceNamespace string) (string, []metav1.Condition, error)
 	if len(rules) > 0 {
-		childClient, err = getChildClient(ctx, r.Client, rgnClient, serviceSet)
+		childClient, err := getChildClient(ctx, r.Client, rgnClient, serviceSet)
 		if err != nil {
 			return false, fmt.Errorf("get child cluster client: %w", err)
+		}
+		assessServiceState = func(serviceName, serviceNamespace string) (string, []metav1.Condition, error) {
+			return verifyHelmServiceOnCluster(ctx, childClient, rules, serviceName, serviceNamespace)
 		}
 	}
 
@@ -402,7 +403,7 @@ func (r *ServiceSetReconciler) verifyServiceStates(ctx context.Context, rgnClien
 			continue
 		}
 
-		serviceKey := client.ObjectKey{Namespace: s.Namespace, Name: s.Name}
+		serviceKey := serviceset.ServiceKey(s.Namespace, s.Name)
 		currentHash := serviceHashes[serviceKey]
 
 		newState := s.State
@@ -432,7 +433,7 @@ func (r *ServiceSetReconciler) verifyServiceStates(ctx context.Context, rgnClien
 		// runs. Gating the stamp on health rules is what let a lagging
 		// Status.Version freeze upgrades indefinitely.
 		if len(rules) > 0 {
-			verifierState, conds, err := verifyHelmServiceOnCluster(ctx, childClient, rules, s.Name, s.Namespace)
+			verifierState, conds, err := assessServiceState(s.Name, s.Namespace)
 			if err != nil {
 				errs = errors.Join(errs, fmt.Errorf("verify service %s/%s: %w", s.Namespace, s.Name, err))
 				continue
--- a/internal/controller/adapters/sveltos/serviceset_controller.go
+++ b/internal/controller/adapters/sveltos/serviceset_controller.go
@@ -287,10 +287,17 @@ func (r *ServiceSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 // rules, and (c) decides the final state by combining sveltos's verdict
 // with the on-cluster verdict and the hash gate.
 //
-// Decision table (sveltos verdict, verifier verdict, hash) → final state.
-// State moves to Deployed only when BOTH signals agree; the stamp block
-// below then advances LastDeployedHash+Version only when they agree AND
-// the hash has actually moved:
+// The two passes are independent. Step (b) is skipped entirely when no rules
+// are loaded — sveltos's verdict then stands untouched — but the stamp still
+// runs, because "the spec version reached the cluster" is a fact about the
+// deploy, not about health, and withholding it strands FilterServiceDependencies
+// on a stale Status.Version.
+//
+// Decision table (sveltos verdict, verifier verdict, hash) → final state,
+// applicable only when rules are loaded. State moves to Deployed only when
+// BOTH signals agree; the stamp block below then advances
+// LastDeployedHash+Version only when they agree AND the hash has actually
+// moved:
 //
 //   - sveltos=*,            verifier!=Deployed            → Provisioning
 //     (downgrade —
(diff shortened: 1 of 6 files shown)


internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -373,9 +375,14 @@ func (r *ServiceSetReconciler) verifyServiceStates(ctx context.Context, rgnClien
 		specVersions[client.ObjectKey{Namespace: svc.Namespace, Name: svc.Name}] = svc.Version
 	}
 
-	childClient, err := getChildClient(ctx, r.Client, rgnClient, serviceSet)
-	if err != nil {
-		return false, fmt.Errorf("get child cluster client: %w", err)
+	// Only the health pass reaches the target cluster; with no rules to
+	// evaluate, building the client would be pure cost.
+	var childClient client.Client
```
zerospiel: potential runtime error

BROngineer: yeah. fixed.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -416,25 +416,39 @@ func (r *ServiceSetReconciler) verifyServiceStates(ctx context.Context, rgnClien
...
-			errs = errors.Join(errs, fmt.Errorf("verify service %s/%s: %w", s.Namespace, s.Name, err))
-			continue
-		}
+		//
+		// With no rules loaded none of that applies — we form no health opinion
+		// and sveltos's verdict stands untouched, while the stamp below still
+		// runs. Gating the stamp on health rules is what let a lagging
+		// Status.Version freeze upgrades indefinitely.
+		if len(rules) > 0 {
+			childClient, err := getChildClient(ctx, r.Client, rgnClient, serviceSet)
```
wahabmk: **1. `getChildClient` moved inside the per-service loop — serviceset_controller.go:425**

It used to be called once per reconcile, before the loop. It now runs once per Helm service. Each call is a Secret GET plus `clientcmd.RESTConfigFromKubeConfig` plus client.New (`internal/util/kube/client.go:34`) — the shared RESTMapper cache saves the discovery round-trip, but not the rest. A ServiceSet with 10 Helm services now builds 10 child clients per reconcile. Hoist it above the loop under if `len(rules) > 0`.

That also fixes a secondary issue: an error there now returns `false`, `err` from mid-loop, discarding accumulated `errs`, any `pendingStamp` already set for earlier services, and the `updateServicesInReadyStateCondition` call — after some services have already had `State`/`Conditions` mutated.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -370,12 +372,7 @@ func (r *ServiceSetReconciler) verifyServiceStates(ctx context.Context, rgnClien
 	specVersions := make(map[client.ObjectKey]*string, len(serviceSet.Spec.Services))
 	for i := range serviceSet.Spec.Services {
 		svc := &serviceSet.Spec.Services[i]
-		specVersions[client.ObjectKey{Namespace: svc.Namespace, Name: svc.Name}] = svc.Version
-	}
-
-	childClient, err := getChildClient(ctx, r.Client, rgnClient, serviceSet)
-	if err != nil {
-		return false, fmt.Errorf("get child cluster client: %w", err)
+		specVersions[serviceset.ServiceKey(svc.Namespace, svc.Name)] = svc.Version
```
wahabmk, line 375: **2. `specVersions` is written with a defaulted key and read with a raw one — `serviceset_controller.go:375` vs `:478`**

```go
specVersions[serviceset.ServiceKey(svc.Namespace, svc.Name)] = svc.Version  // ""  → "default"
...
serviceKey := client.ObjectKey{Namespace: s.Namespace, Name: s.Name}        // ""  stays ""
s.Version = specVersions[serviceKey]
```

The PR changed line 375 to `ServiceKey` but left the lookup raw. For a service with an empty `Namespace` the write lands under `default/<name>` and the read misses, so the stamp writes `nil`. `versionsEqual(specVersion, nil)` is then false forever, `stampConverged` never returns true, and the poller never quiesces — the exact failure this PR fixes.

Note `serviceHashes` (line 396) is keyed raw from `rs.ReleaseNamespace`, so the two keys genuinely need to differ; the clean fix is a separate `specKey := serviceset.ServiceKey(s.Namespace, s.Name)` for line 478 only.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -416,25 +416,39 @@ func (r *ServiceSetReconciler) verifyServiceStates(ctx context.Context, rgnClien
...
+			if newState == kcmv1.ServiceStateDeployed && currentHash == "" {
+				newState = kcmv1.ServiceStateProvisioning
+				pendingStamp = true
+			}
+		} else if s.State == kcmv1.ServiceStateDeployed && currentHash == "" {
+			// Requeue to stamp later, but leave State alone — downgrading here
+			// would move Status.Deployed on the strength of an opinion we did
+			// not form.
 			pendingStamp = true
 		}
```
wahabmk, line 460: **3. New unbounded 10s requeue loop in the default configuration — `serviceset_controller.go:450-453`**

```go
} else if s.State == kcmv1.ServiceStateDeployed && currentHash == "" {
    pendingStamp = true
}
```

`pendingStamp` drives `RequeueAfter: r.requeueInterval` (10s, fixed, no back-off) at line 275. Where the hash can never be computed, that requeues forever. The concrete case is again empty-namespace services: `getHelmCharts` defaults `ReleaseNamespace` to `svc.Name` (line 1230) while `Status.Services[i].Namespace` stays `""`, so the `serviceHashes` lookup can never match. Pre-PR, no-rules clusters returned early and never reached this; now every such ServiceSet spins at 10s indefinitely. Worth either bounding the requeue or logging once when the hash is unresolvable.

wahabmk (commented):
**4. The stamp binds `spec.Version` to a hash computed from possibly-stale sveltos artifacts**

`Reconcile` runs `ensureProfile` (writes the new chart version into the Profile) → `collectServiceStatuses` → `verifyServiceStates`, which reads a ClusterSummary sveltos hasn't refreshed yet. The stamp guard is `currentHash != s.LastDeployedHash`, so when `LastDeployedHash` is empty that's trivially true against the old chart — and `s.Version` is set to the new `spec.Version`.

That matters more after this PR than before it: `LastDeployedHash` is `""` for every service on a no-rules cluster today, so the first reconcile after this ships is exactly that window. If a version bump is in flight, the stamp records the new version, `stampConverged` goes true, the poller quiesces, and `FilterServiceDependencies` releases dependents on a version not yet on cluster. It self-heals on the next CS RV bump, but the dependents are already unlocked.

`computeServiceHash` already has `chart.ChartVersion` in hand (`verify.go:1309`). Stamping that, or gating the stamp on `chart.ChartVersion == *spec.Version`, would close this properly.

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: Services added to ServiceSet are produced by the `makeService` function:

```go
func makeService(s kcmv1.Service, version, template string) kcmv1.ServiceWithValues {
	return kcmv1.ServiceWithValues{
		Name: s.Name,
		// We should always use effective namespace, because service namespace in
		// serviceSet's service definition is never empty while service namespace
		// in clusterDeployment's/multiClusterService's service definition can be empty.
		// This will lead to persistent discrepancy between service definitions and
		// lead to continuous serviceSet updates.
		Namespace:   effectiveNamespace(s.Namespace),
		Version:     new(version),
		Template:    template,
		Values:      s.Values,
		ValuesFrom:  s.ValuesFrom,
		HelmOptions: s.HelmOptions,
		HelmAction:  s.HelmAction,
	}
}
```

which in turn sets the effective namespace instead of just copying the whatever namespace is defined in CLD/MCS spec:

```go
func effectiveNamespace(serviceNamespace string) string {
	if serviceNamespace == "" {
		return metav1.NamespaceDefault
	}
	return serviceNamespace
}
```

Thus the state where service namespace in spec is empty is not reachable

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: Same as in comment below - empty service namespace is not reachable for serviceset, because services added to serviceset passes through the `makeService` + `effectiveNamespace` functions.
Will update for consistency tho

internal/controller/adapters/sveltos/serviceset_controller.go
BROngineer: To avoid possible runtime errs, since we do not check the client interface, and to revert the invalid fix with moving getting client under the for loop, now it's under the closure: it's safe in runtime because we build the closure and use it under the same condition `if len(rules) > 0`

wahabmk (approved):

josef-hak (changes requested):
@BROngineer, I tested the PR using ksm test and it failed: https://github.com/josef-hak/k0rdent-ksm-tests/actions/runs/34253626643/job/102154143834 . It's probably because the feature branch is not rebased with main. Could you please rebase so we can ensure it's not a regression?

PR #3014, opened 2026-08-17
feat: introduce KAL

zerospiel:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** _(optional, `Fixes #123`)_:
Closes #3013

PR #3045, opened 2026-09-03
feat: RBAC policy operator

eromanova:
Jira: KCM-1189

Adds a new namespaced RBACPolicy CRD (k0rdent.mirantis.com/v1beta1, shortname rbacpol) that lets a ClusterDeployment declare a role catalog for its child cluster:

* A ClusterDeployment opts in via the new `spec.rbacPolicy` field, which names an RBACPolicy in the same namespace.
* Each `spec.bindings[]` entry maps one ClusterRole to a list of User/Group subjects. The reconciler creates a k0rdent-<slug> ClusterRoleBinding in the child cluster (via the CAPI-generated kubeconfig). If rules are supplied, the ClusterRole is created/updated too; without rules the binding targets a role expected to already exist (admin, edit, view, …). A pre-existing ClusterRole that KCM didn't create is never overwritten.
* Everything created is labeled `k0rdent.mirantis.com/managed-by: rbac-operator`; anything managed but no longer desired is pruned.
* The sync is a step of ClusterDeploymentReconciler, gated on `spec.rbacPolicy` being set, and re-syncs every 5m to correct drift. State is reported as the new RBACPolicyReady condition on the ClusterDeployment.
* A validating webhook (plus inline validation when webhooks are disabled) rejects duplicate binding names and the same ClusterRole carrying rules in more than one binding, and blocks deletion while a ClusterDeployment still references the policy. Ships with rbacpolicies-creator / rbacpolicies-viewer user-facing roles.

## Example

```
kind: RBACPolicy
metadata:
  name: project-2-rbac
  namespace: kcm-system
spec:
  bindings:
    # Bind a built-in ClusterRole that already exists in the child cluster.
    - name: project-viewer
      clusterRole: view
      subjects:
        - kind: User
          name: k0rdent:user:a36bda60-922a-49de-9742-da0db13c3a6b
        - kind: User
          name: k0rdent:sa:sa-482aa77e2fe84760a82633acf65af1a6

    # Define a custom ClusterRole inline and bind it.
    - name: compute-admin
      clusterRole: k0rdent-compute-admin
      rules:
        - apiGroups: [""]
          resources: ["pods", "pods/log"]
          verbs: ["get", "list", "watch", "delete"]
        - apiGroups: ["apps"]
          resources: ["deployments", "statefulsets"]
          verbs: ["*"]
      subjects:
        - kind: Group
          name: k0rdent:project:project-2:compute-admin
        - kind: User
          name: alice@example.com
```


internal/controller/rbac/rbacpolicy.go
```diff
@@ -0,0 +1,212 @@
...
+	rbacSubjectAPIGroup = rbacv1.GroupName
+
+	// ManagedByLabelKey / ManagedByLabelValue mark every ClusterRole/ClusterRoleBinding this package
+	// creates in a child cluster, and are what Prune selects on. Deliberately distinct from the
+	// generic kcmv1.KCMManagedLabelKey so a prune here can never touch some other k0rdent-managed
+	// object in the child cluster that isn't part of this RBACPolicy sync. It is also what tells
+	// applyClusterRole a ClusterRole is safe to overwrite.
+	ManagedByLabelKey   = "k0rdent.mirantis.com/managed-by"
+	ManagedByLabelValue = "rbac-operator"
+)
```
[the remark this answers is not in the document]
eromanova: The description was outdated. Updated.

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,109 @@
...
+// operator reconciles into the child cluster of the [ClusterDeployment] that references this
+// RBACPolicy via spec.rbacPolicy and has spec.clusterAuth set.
+type RBACPolicySpec struct {
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=200
+
+	// Bindings is the role catalog: each entry maps one ClusterRole to a set of subjects,
+	// materialized as a ClusterRoleBinding in the child cluster of the referencing ClusterDeployment.
+	Bindings []RBACPolicyBinding `json:"bindings"`
+}
```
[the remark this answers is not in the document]
eromanova: The description was outdated. `suspend` flag has been removed.

api/v1beta1/clusterdeployment_types.go
```diff
@@ -122,6 +125,9 @@ type ClusterDeploymentSpec struct {
 	// AuditPolicy is the name reference to the related [ClusterAuditPolicy] object located in the same namespace
 	// containing audit policy configuration.
 	AuditPolicy string `json:"auditPolicy,omitempty"`
+	// RBACPolicy string is the name reference to the related [RBACPolicy] object located in the same namespace
+	// containing RBAC policy configuration.
+	RBACPolicy string `json:"rbacPolicy,omitempty"`
```
zerospiel:
```suggestion
    // +optional
    // +kubebuilder:validation:MinLength=1
    // +kubebuilder:validation:MaxLength=256

	// rbacPolicy string is the name reference to the related [RBACPolicy] object located in the same namespace
	// containing RBAC policy configuration.
	RBACPolicy string `json:"rbacPolicy,omitempty"`
```

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+//
+// RBACPolicy makes no assumption about how (or whether) the child cluster's authentication is
+// configured: Name is written verbatim by the author to match whatever that specific setup
+// actually asserts (a ClusterAuthentication's JWT claim mapping, some other identity provider, or
+// none at all) — for example "k0rdent:project:project-2:compute-admin".
+type RBACPolicySubject struct {
+	// +kubebuilder:validation:Enum=User;Group
+
+	// Kind is the Kubernetes RBAC subject kind: "User" or "Group".
+	Kind string `json:"kind"`
```
zerospiel:
```suggestion
    // +required
	// +kubebuilder:validation:Enum=User;Group
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=1024

	// kind is the Kubernetes RBAC subject kind: "User" or "Group".
	Kind string `json:"kind,omitempty"`
```

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// +kubebuilder:validation:MinLength=1
+
+	// Name is the exact subject name (see the RBACPolicySubject doc comment).
+	Name string `json:"name"`
```
zerospiel:
```suggestion
    // +required
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=256

	// Name is the exact subject name (see the RBACPolicySubject doc comment).
	Name string `json:"name,omitempty"`
```

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+//
+// It is the role catalog for exactly one [ClusterDeployment]: a ClusterDeployment opts in by
+// setting spec.rbacPolicy to the name of an RBACPolicy in its own namespace. For each entry in
+// spec.bindings, the RBAC operator creates/updates a "k0rdent-<name>" ClusterRoleBinding in the
+// child cluster, using the kubeconfig generated by Cluster API.
+type RBACPolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec RBACPolicySpec `json:"spec,omitempty"`
```
zerospiel:
```suggestion
    // +required
    
	Spec RBACPolicySpec `json:"spec,omitempty,omitzero"`
```

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+// +kubebuilder:resource:scope=Namespaced,shortName=rbacpol
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Age"
+
+// RBACPolicy is the Schema for the rbacpolicies API.
+//
+// It is the role catalog for exactly one [ClusterDeployment]: a ClusterDeployment opts in by
+// setting spec.rbacPolicy to the name of an RBACPolicy in its own namespace. For each entry in
+// spec.bindings, the RBAC operator creates/updates a "k0rdent-<name>" ClusterRoleBinding in the
+// child cluster, using the kubeconfig generated by Cluster API.
+type RBACPolicy struct {
```
zerospiel, line 111: for some reason, the new CRD is missing from the `PROJECT` file along with the corresponding attributes, like admission webhooks

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+	// +kubebuilder:validation:MaxItems=200
+
+	// Bindings is the role catalog: each entry maps one ClusterRole to a set of subjects,
+	// materialized as a ClusterRoleBinding in the child cluster of the referencing ClusterDeployment.
+	Bindings []RBACPolicyBinding `json:"bindings"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=rbacpol
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Age"
```
zerospiel: it's the default column, isn't it? if so i reckon we could get rid of it

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// RBACPolicyBinding maps a single Kubernetes ClusterRole to the exact set of ClusterRoleBinding
+// subjects that should be bound to it in the targeted child cluster.
+type RBACPolicyBinding struct {
+	// +kubebuilder:validation:MinLength=1
+
+	// Name identifies this binding within the policy (for example "compute-admin"). It is used to
+	// name the generated ClusterRoleBinding ("k0rdent-<name>").
+	Name string `json:"name"`
```
zerospiel:
```suggestion
    // +required
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=256

	// name identifies this binding within the policy (for example "compute-admin"). It is used to
	// name the generated ClusterRoleBinding ("k0rdent-<name>").
	Name string `json:"name"`
```

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+	// where that ClusterRole does not already exist in the child cluster (e.g. a custom role). When
+	// set, the operator creates/updates that ClusterRole (with these rules) alongside the
+	// ClusterRoleBinding. Leave unset to bind to a ClusterRole that is expected to already exist
+	// (e.g. a built-in role) — the operator never modifies a ClusterRole it did not create.
+	Rules []rbacv1.PolicyRule `json:"rules,omitempty"`
+
+	// +kubebuilder:validation:MinItems=1
+
+	// Subjects is the ClusterRoleBinding subject list for this binding. See RBACPolicySubject.
+	Subjects []RBACPolicySubject `json:"subjects"`
```
zerospiel:
```suggestion
	// +optional
	// +listType=atomic
	// +kubebuilder:validation:MinItems=1

	// subjects is the ClusterRoleBinding subject list for this binding. See RBACPolicySubject.
	Subjects []RBACPolicySubject `json:"subjects"`
```

zerospiel (commented):
That's an initial API linter check, could you please stick to #3014 so we won't change the API once again. Basically, most of the fields require different markers

I will continue the review of the logic itself as a separate review

api/v1beta1/rbacpolicy_types.go
eromanova: Subjects is a required field, I updated it accordingly.

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,133 @@
...
+// setting spec.rbacPolicy to the name of an RBACPolicy in its own namespace. For each entry in
+// spec.bindings, the RBAC operator creates/updates a "k0rdent-<name>" ClusterRoleBinding in the
+// child cluster, using the kubeconfig generated by Cluster API.
+type RBACPolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// +required
+
+	Spec RBACPolicySpec `json:"spec,omitempty,omitzero"`
```
zerospiel:
```suggestion
	metav1.TypeMeta   `json:",inline"`
	// +optional

	// metadata contains the object metadata
	metav1.ObjectMeta `json:"metadata,omitempty"`

	// +required

	// spec defines the desired state of the RBACPolicy
	Spec RBACPolicySpec `json:"spec,omitempty,omitzero"`
```

templates/provider/kcm/templates/crds/k0rdent.mirantis.com_rbacpolicies.yaml
```diff
@@ -0,0 +1,180 @@
...
+                          ClusterRoleBinding. Leave unset to bind to a ClusterRole that is expected to already exist
+                          (e.g. a built-in role) — the operator never modifies a ClusterRole it did not create.
+                        items:
+                          description: |-
+                            PolicyRule holds information that describes a policy rule, but does not contain information
+                            about who the rule applies to or which namespace the rule applies to.
+                          properties:
+                            apiGroups:
+                              description: |-
+                                apiGroups is the name of the APIGroup that contains the resources.  If multiple API groups are specified, any action requested against one of
```
zerospiel, line 89: For some reason, the `manifests` target produces diffs here and there. I assume the manifest was amended manually

templates/provider/kcm/templates/rbac/user-facing/rbacpolicies-creator.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    k0rdent.mirantis.com/aggregate-to-namespace-admin: "true"
+rules:
+  - apiGroups:
+      - k0rdent.mirantis.com
+    resources:
+      - rbacpolicies
+    verbs: {{ include "rbac.viewerVerbs" . | nindent 6 }}
+      - create
+      - delete
```
zerospiel: why not just the `{{ include "rbac.editorVerbs" . | nindent 6 }}`?

internal/controller/clusterdeployment_controller.go
```diff
@@ -444,7 +480,15 @@ func (r *ClusterDeploymentReconciler) updateCluster(
...
-	return r.reconcileHelmRelease(ctx, clusterTpl, scope)
+	helmResult, err := r.reconcileHelmRelease(ctx, clusterTpl, scope)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if !helmResult.IsZero() {
+		return helmResult, nil
+	}
+
+	return r.ensureRBACPolicy(ctx, scope)
```
zerospiel: **blocking** — RBAC sync is gated behind full cluster readiness, so changes *and revocations* stall indefinitely.

`ensureRBACPolicy` is only reached when `reconcileHelmRelease` returns a zero `Result`. That function returns `{RequeueAfter}` whenever `!fluxconditions.IsReady(hr)` **or** `aggregateCapiConditions` requeues — and `aggregateCapiConditions` requeues on `capiCondition.Status != metav1.ConditionTrue`.

So any cluster whose CAPI summary condition isn't `True` — mid-upgrade, one machine `NotReady`, a transient provider blip — never syncs RBAC, even though its API server and kubeconfig are perfectly usable. Three consequences:

- new and changed bindings are not applied;
- **clearing `spec.rbacPolicy` revokes nothing**, because the prune branch lives inside the same gated function;
- `RBACPolicyReady=True` stays on the ClusterDeployment, so status asserts the policy is live when it isn't.

The only real precondition for distributing RBAC is the kubeconfig Secret, and `childClientFor` already checks for that. I'd call `ensureRBACPolicy` unconditionally after `reconcileHelmRelease` and join the results/errors, rather than gating it on the helm result:

```go
helmResult, helmErr := r.reconcileHelmRelease(ctx, clusterTpl, scope)
rbacResult, rbacErr := r.ensureRBACPolicy(ctx, scope)
if err := errors.Join(helmErr, rbacErr); err != nil {
    return ctrl.Result{}, err
}
// ... pick the sooner of the two RequeueAfters
```

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1120,113 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+	cd := scope.cd
+
+	active := cd.Spec.RBACPolicy != "" && scope.rbacPolicy != nil
+
+	if active {
+		capiCluster, err := r.getPartialCapiCluster(ctx, scope.rgnClient, cd)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to check for CAPI Cluster: %w", err)
+		}
+		active = capiCluster != nil
```
zerospiel: **blocking** — a missing CAPI `Cluster` silently wipes RBAC from a live child cluster.

When `getPartialCapiCluster` returns nil, `active` goes false and the `!active` branch below runs `rbac.Prune(ctx, childCl, nil, nil)`, deleting every managed ClusterRole and ClusterRoleBinding, then removes the condition.

But `aggregateCapiConditions` explicitly handles "Cluster deleted out-of-band while the ClusterDeployment is alive" (`CAPIClusterMissingReason`) as an anomaly to *report*, not as a teardown signal. If that happens while the kubeconfig Secret still exists — Secret GC is asynchronous, and not every provider owner-references it — this tears down all delegated access in a still-reachable cluster, with no event, no error and no log line.

There's a second, quieter failure here too: `getPartialCapiCluster` looks the Cluster up by `name == cd.Name`, whereas `aggregateCapiConditions` finds it via the `FluxHelmChartNameKey` label. For a template that names the Cluster differently, `active` is permanently false, so RBAC is silently never applied — no condition, no event, not even a `V(1)` line. `ensureAuthConfigSecret`/`ensureAuditPolicyConfigMap` use the same lookup, but only best-effort for an owner reference, so the mismatch is harmless there and load-bearing here.

Suggestion: use the CAPI-Cluster check *only* to decide "this will never get a kubeconfig, so don't start and don't requeue", and make the prune trigger `cd.Spec.RBACPolicy == ""` on its own. And log at `V(1)` when opting out, so an operator who sets `spec.rbacPolicy` on an adopted cluster gets some feedback instead of silence.

internal/controller/clusterdeployment_controller.go
```diff
@@ -248,6 +257,33 @@ func (r *ClusterDeploymentReconciler) getClusterScope(ctx context.Context, cd *k
 		}
 	}
 
+	if cd.Spec.RBACPolicy != "" {
```
zerospiel, line 266: **blocking** — deleting the `RBACPolicy` CR doesn't revoke access, and wedges the ClusterDeployment.

There's no finalizer on `RBACPolicy` and no cleanup in `reconcileDelete`. `RBACPolicyDeletionAllowed` in the delete webhook is the only protection, and it's bypassed in exactly the mode this controller explicitly supports (`IsDisabledValidationWH`), as well as on CRD removal.

Once the policy is gone, this `Get` fails `NotFound` → `RBACPolicyReady=False` → `getClusterScope` returns an error → `reconcileUpdate` never runs. The child cluster keeps every granted ClusterRoleBinding indefinitely, while the controller hot-loops on an error that can never clear without a user editing `spec.rbacPolicy`.

For an API whose whole job is granting access, "delete the policy" is the intuitive way to revoke, and right now it's the one path that provably doesn't. A finalizer on `RBACPolicy` that prunes from each referencing ClusterDeployment's child cluster before allowing removal would close it; at minimum, treating `NotFound` here as "revoke and clear" rather than "error out" would stop the wedge.

internal/controller/rbac/rbacpolicy.go
```diff
@@ -0,0 +1,219 @@
...
+// that name already exists and wasn't created by this package (i.e. it's missing
+// ManagedByLabelKey) — in which case it refuses, so a binding can never overwrite an
+// already-existing ClusterRole such as a built-in "admin"/"edit"/"view", matching
+// RBACPolicyBinding.Rules' doc comment.
+func applyClusterRole(ctx context.Context, childCl client.Client, name string, rules []rbacv1.PolicyRule) (bool, error) {
+	role := &rbacv1.ClusterRole{ObjectMeta: metav1.ObjectMeta{Name: name}}
+	err := childCl.Get(ctx, client.ObjectKeyFromObject(role), role)
+	switch {
+	case err == nil && role.Labels[ManagedByLabelKey] != ManagedByLabelValue:
+		return false, fmt.Errorf("ClusterRole %s already exists and was not created by this RBACPolicy; refusing to overwrite it", name)
```
zerospiel: This is a permanent error that gets retried forever, and it takes the rest of the policy down with it.

`clusterRole: admin` plus a `rules:` block is an easy mistake to make — `admin` exists in every cluster. When it happens, `Sync` returns on the *first* binding, so none of the other bindings in the policy are applied, and the reconcile keeps failing under the rate limiter indefinitely. Nothing about the cluster can change to make it succeed; only a spec edit can.

Two suggestions:

1. Make it terminal — condition `False` with a clear message and no retrigger, in the spirit of the `errNoRetrigger` handling elsewhere in the reconciler.
2. Don't abort the loop. Collect per-binding errors with `errors.Join` and keep going, so one bad binding doesn't hold the other 199 hostage.

Worth noting the webhook can't catch this one, since it depends on child-cluster state — the runtime path is the only place it can be handled well.

internal/controller/rbac/rbacpolicy.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	for _, binding := range policy.Spec.Bindings {
+		if len(binding.Rules) > 0 {
+			roleChanged, err := applyClusterRole(ctx, childCl, binding.ClusterRole, binding.Rules)
+			if err != nil {
+				return nil, nil, false, fmt.Errorf("applying ClusterRole %s: %w", binding.ClusterRole, err)
+			}
+			changed = changed || roleChanged
+			desiredRoles[binding.ClusterRole] = struct{}{}
```
zerospiel: Dropping `rules` from a binding silently deletes the ClusterRole out from under a live ClusterRoleBinding.

`desiredRoles` is only populated when `len(binding.Rules) > 0`, so if a user removes the `rules` block from a binding that previously had one (intending "now bind to a role that already exists"), `Prune` deletes the managed ClusterRole while the ClusterRoleBinding survives pointing at a name that no longer resolves.

Net effect: the subjects lose all access, with no error, no event, and `RBACPolicyReady=True`.

Either keep managed ClusterRoles that are still referenced by a desired binding, or reject the transition in validation so the user has to be explicit about it.

zerospiel (commented):
## Review: RBAC policy operator

Reviewed `5f61ee48`, `c26d9f41` and `915e1f1d`. `go build ./...`, `go vet` and the new unit tests all pass, and `golangci-lint` over the touched packages reports only pre-existing findings in unrelated files. The package split is clean and the doc comments in `internal/controller/rbac` are genuinely good — they made this much easier to review.

Most of my comments are inline. The three I'd treat as blocking are all in the same area, so I'll summarise the theme here: **the sync's activation and deactivation conditions are entangled with cluster-readiness signals that aren't actually preconditions for distributing RBAC, and revocation has no path that is guaranteed to run.**

1. `ensureRBACPolicy` only runs when `reconcileHelmRelease` returns a zero `Result`, which means any cluster whose CAPI summary condition isn't `True` never syncs — and never *revokes* — while still reporting `RBACPolicyReady=True`.
2. A missing CAPI `Cluster` object flips `active` to false, and the `!active` branch prunes everything in a child cluster that may still be perfectly reachable.
3. Deleting the `RBACPolicy` CR leaves every granted binding in place and wedges the `ClusterDeployment` on a permanent error.

### Two design points that don't have a natural inline anchor

**ClusterRole/ClusterRoleBinding only.** There's no namespaced `Role`/`RoleBinding`, no way to bind a subject to a `Role`, and no ServiceAccount subjects. For the multi-tenant story ("project X gets `edit` in namespace Y") the namespaced shape is arguably the more common one. Worth deciding explicitly now, because retrofitting it later means either reshaping `bindings` or adding a second list alongside it. The current `bindings[].clusterRole` naming at least leaves room.

**Privilege escalation has no guard.** `rbacpolicies-creator` aggregates to `namespace-admin`, and `rules` accepts arbitrary `PolicyRule`s including `*/*/*`. So a management-cluster namespace admin can mint themselves `cluster-admin` in every child cluster in that namespace. In-cluster RBAC prevents exactly this with the `escalate`/`bind` verbs; here the controller holds CAPI's admin kubeconfig, so there is no equivalent check. This may well be intended — a namespace admin already owns the cluster — but it's worth being a documented, deliberate decision rather than an emergent one.

### RBAC rules — no issues

The controller's cluster role is complete: `rbacpolicies` added with `editorVerbs` (broader than the `get;list;watch` actually used, but consistent with its siblings), `secrets` already covered for the kubeconfig read, creator/viewer user-facing roles following the existing aggregation pattern exactly. Nothing extra is needed child-side since it goes through CAPI's admin kubeconfig.

### Generated manifests

`templates/provider/kcm/templates/crds/k0rdent.mirantis.com_rbacpolicies.yaml` as committed doesn't match `make manifests` for the pinned `k8s.io/api v0.36.4` — the `rbacv1.PolicyRule` field descriptions differ (`apiGroups is the name...` vs the upstream `APIGroups is the name...`). Regenerating produces a small diff. Might be worth a `make generate manifests && git diff --exit-code` gate in CI.

### Smaller things

- `childClientFactory` is a production struct field that exists only as a test seam ("Left nil in production"), with a `factory == nil` fallback evaluated on every call. Setting it once in `cmd/main.go` would read better.
- `managedSelector()` rebuilds and re-sorts a `labels.Selector` on each of the two `pruneManaged` calls — `client.MatchingLabels{ManagedByLabelKey: ManagedByLabelValue}` is cheaper and more idiomatic.
- `pruneBindings` is a one-line wrapper with a single real caller; consider inlining it.
- The new `clusterdeployment_controller_test.go` subtests use `context.Background()` while `internal/controller/rbac` tests use `t.Context()` — worth being consistent (that's what `usetesting` is for).
- `rbacpolicy_webhook_test.go` shares one `g := NewWithT(t)` across subtests rather than creating one per `t.Run`, so a failure gets attributed to the parent test.
- `Watches(&kcmv1.RBACPolicy{}, ...)` has no predicate, unlike the neighbouring `ClusterAuthentication` watch. RBACPolicy has no status so only spec/metadata changes fire, but `predicate.GenerationChangedPredicate` would still drop label/annotation churn.
- Consider an `RBACPolicyKind` const for symmetry with `ClusterAuditPolicyKind`/`ClusterAuthenticationKind`, and adding `&kcmv1.RBACPolicy{}` to `AccessManagementReconciler.builtinKinds()` so a shared policy in the system namespace is watched rather than only polled.
- `deleteChildResources` lost the `"key", secretKey` log field in the refactor and now recomputes `GetKubeconfigSecretKey` purely for the log line.

internal/controller/rbac/rbacpolicy.go
```diff
@@ -0,0 +1,219 @@
...
+	role := &rbacv1.ClusterRole{ObjectMeta: metav1.ObjectMeta{Name: name}}
+	err := childCl.Get(ctx, client.ObjectKeyFromObject(role), role)
+	switch {
+	case err == nil && role.Labels[ManagedByLabelKey] != ManagedByLabelValue:
+		return false, fmt.Errorf("ClusterRole %s already exists and was not created by this RBACPolicy; refusing to overwrite it", name)
+	case client.IgnoreNotFound(err) != nil:
+		return false, err
+	}
+
+	op, err := controllerutil.CreateOrUpdate(ctx, childCl, role, func() error {
```
zerospiel: Every object costs two round-trips before any write: the `Get` above, then `CreateOrUpdate`'s own internal `Get`. The child client is non-caching, so at the schema's `maxItems: 200` a single resync is up to 800 GETs plus 2 LISTs per cluster — every 5 minutes, per cluster.

Server-side apply collapses each to one call and is idempotent by construction:

```go
role := &rbacv1.ClusterRole{
    TypeMeta:   metav1.TypeMeta{APIVersion: rbacv1.SchemeGroupVersion.String(), Kind: "ClusterRole"},
    ObjectMeta: metav1.ObjectMeta{Name: name, Labels: mergeManagedLabels(nil)},
    Rules:      rules,
}
err := childCl.Patch(ctx, role, client.Apply, client.ForceOwnership, client.FieldOwner("kcm-rbac-operator"))
```

The "refuse to overwrite an unmanaged ClusterRole" guard still needs its own `Get`, but only for bindings that actually carry rules. If SSA is too big a change for this PR, `controllerutil.CreateOrPatch` seeded with the already-fetched object at least avoids the duplicate read.

internal/controller/rbac/rbacpolicy.go
```diff
@@ -0,0 +1,219 @@
...
+// whether anything was actually deleted. Passing a nil/empty map removes every
+// ClusterRoleBinding this package manages there.
+func pruneBindings(ctx context.Context, childCl client.Client, desiredBindings map[string]struct{}) (bool, error) {
+	return pruneManaged(ctx, childCl, &rbacv1.ClusterRoleBindingList{}, desiredBindings, "ClusterRoleBinding")
+}
+
+// pruneManaged deletes every ManagedByLabelKey-selected item in list not present in desired,
+// reporting whether anything was deleted. kind is used only for error messages.
+func pruneManaged(ctx context.Context, childCl client.Client, list client.ObjectList, desired map[string]struct{}, kind string) (bool, error) {
+	if err := childCl.List(ctx, list, &client.ListOptions{LabelSelector: managedSelector()}); err != nil {
```
zerospiel: This lists **full** ClusterRole / ClusterRoleBinding objects — decoding every `Rules` and `Subjects` slice — and then runs them through `apimeta.ExtractList` (reflection, plus a `[]runtime.Object` allocation) only to read `.GetName()`.

`metav1.PartialObjectMetadataList` with the GVK set fetches names only. It's cheaper on both the apiserver and the heap, and it lets you drop the `client.ObjectList` parameter, the `item.(client.Object)` type assertion, and the `kind` string:

```go
list := &metav1.PartialObjectMetadataList{}
list.SetGroupVersionKind(gvk)
if err := childCl.List(ctx, list, client.MatchingLabels{ManagedByLabelKey: ManagedByLabelValue}); err != nil { ... }
for i := range list.Items { ... }
```

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1120,113 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+	pruneChanged, err := rbac.Prune(ctx, childCl, desiredRoles, desiredBindings)
+	if err != nil {
+		err = fmt.Errorf("failed to prune stale RBAC objects: %w", err)
+		if r.setCondition(cd, kcmv1.RBACPolicyReadyCondition, kcmv1.FailedReason, metav1.ConditionFalse, err) {
+			r.warnf(cd, "RBACSyncFailed", err.Error())
+		}
+		return ctrl.Result{}, err
+	}
+
+	r.setCondition(cd, kcmv1.RBACPolicyReadyCondition, kcmv1.SucceededReason, metav1.ConditionTrue, nil)
```
zerospiel: Nothing records *which* revision of the RBACPolicy is actually live.

`setCondition` stamps `ObservedGeneration: cd.Generation`. When the RBACPolicy changes but the ClusterDeployment doesn't, a successful re-sync produces a byte-identical condition — and a re-sync that never happens (see the gating comment in `updateCluster`) leaves the stale `True` in place. Either way there's no way for a user, or for `kubectl wait`, to tell whether the policy revision they just applied has reached the child cluster.

Recording the applied RBACPolicy generation — or a short hash of `spec.bindings`, similar to what `ensureAuthConfigSecret` already does with `authConfigHash` — in the condition message or a dedicated status field, and going `Progressing` when the observed generation is behind, would make this observable.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1120,113 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
 	return nil
 }
 
+// rbacResyncInterval is how often ensureRBACPolicy re-syncs a ClusterDeployment's child-cluster
+// RBAC objects even without any triggering event, since nothing else watches the child cluster
+// for drift on these objects.
+const rbacResyncInterval = 5 * time.Minute
```
zerospiel, line 1155: Two things about this resync.

**No jitter.** `RequeueAfter: rbacResyncInterval` propagates out through `updateCluster` → `reconcileUpdate`, so after a controller restart every ClusterDeployment lands in the same 5-minute phase and then stays locked in step. A little jitter (say ±10%) would spread the load.

**It re-runs the whole ClusterDeployment loop.** Not just the RBAC sync — the ClusterTemplate `Get`, `reconcileHelmRelease`, `updateServices`, the CAPI aggregation, `getPartialCapiCluster`, the kubeconfig Secret `Get` and a fresh child-client construction, every 5 minutes for every CD with a policy. If drift detection on child-cluster RBAC objects is the actual goal, a dedicated runnable keyed off the child clients would be a lot cheaper than driving it from the main reconciler.

Also worth checking against fleet size: the shared REST-mapper cache defaults to 256 entries (`restMapperMaxEntries` in `internal/util/kube/restmapper_cache.go`, overridable via `RESTMAPPER_CACHE_MAX_ENTRIES`). Above that, this resync will thrash it.

internal/util/validation/rbacpolicy.go
```diff
@@ -0,0 +1,67 @@
...
+		names[binding.Name] = struct{}{}
+
+		// binding.Name is concatenated into the generated ClusterRoleBinding's metadata.name
+		// (kcmv1.ClusterRoleBindingNamePrefix + binding.Name), which the child cluster's API
+		// server requires to be a valid DNS-1123 subdomain (also bounding its length). Checking
+		// the combined form here, rather than binding.Name in isolation, fails fast with a clear
+		// error instead of the binding only failing much later when the operator tries to apply
+		// it to the child cluster.
+		objectName := kcmv1.ClusterRoleBindingNamePrefix + binding.Name
+		if errs := apivalidation.IsDNS1123Subdomain(objectName); len(errs) > 0 {
```
zerospiel, line 43: `binding.Name` is validated here as a DNS-1123 subdomain, which is great — but `binding.ClusterRole` never is, even though it becomes a ClusterRole object name in the child cluster whenever `rules` is set.

So `clusterRole: "My Custom Role"` passes admission and then fails forever in `applyClusterRole`, as the same permanent-retry loop described over in `internal/controller/rbac/rbacpolicy.go`. Worth applying `IsDNS1123Subdomain` to `binding.ClusterRole` too when `len(binding.Rules) > 0`.

Two smaller gaps while you're here: duplicate `(kind, name)` subjects within a binding aren't rejected, and neither is the same subject appearing in two bindings for the same ClusterRole. Both are harmless at apply time, just noisy.

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,133 @@
...
+	// materialized as a ClusterRoleBinding in the child cluster of the referencing ClusterDeployment.
+	Bindings []RBACPolicyBinding `json:"bindings,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=rbacpol
+
+// RBACPolicy is the Schema for the rbacpolicies API.
+//
+// It is the role catalog for exactly one [ClusterDeployment]: a ClusterDeployment opts in by
```
zerospiel: Two things about the type itself.

**This 1:1 claim isn't enforced, and isn't true.** Nothing prevents N ClusterDeployments from referencing one RBACPolicy — `deletionAllowedIfUnreferenced` says "one or more ClusterDeployments", and `mapObjectsToClusterDeployments` enqueues all of them. The N:1 case works correctly and is arguably the useful one (one catalog, many clusters). This text lands verbatim in the CRD's `openAPIV3Schema.description`, so it's what users see in `kubectl explain rbacpolicy`.

**No status subresource at all.** The object tells you nothing about itself: not which ClusterDeployments consume it, not whether any sync has succeeded, no `observedGeneration`. All the feedback lives on the ClusterDeployment side. For a resource whose entire job is granting access, `kubectl get rbacpolicy` showing only NAME/AGE is a real gap. I'd suggest at minimum `status.conditions` + `status.observedGeneration` + the set (or count) of referencing ClusterDeployments — plus a couple of printer columns, since none are defined today.

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,133 @@
...
+	// name is the exact subject name (see the RBACPolicySubject doc comment).
+	Name string `json:"name,omitempty"`
+}
+
+// RBACPolicyBinding maps a single Kubernetes ClusterRole to the exact set of ClusterRoleBinding
+// subjects that should be bound to it in the targeted child cluster.
+type RBACPolicyBinding struct {
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
```
zerospiel: `MaxLength=256` is unreachable in practice: `ClusterRoleBindingNamePrefix + Name` has to be a DNS-1123 subdomain, so anything over 245 characters always fails `ValidateRBACPolicy` regardless. `MaxLength=245` would make the schema tell the truth and let the failure surface at the field rather than in a cross-field check.

Minor companion: `RBACPolicySubject.Kind` carries `MaxLength=1024` next to `Enum=User;Group`. Harmless, just noise.

api/v1beta1/rbacpolicy_types.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// RBACPolicySpec defines the role catalog: a set of ClusterRole -> subjects bindings that the RBAC
+// operator reconciles into the child cluster of the [ClusterDeployment] that references this
+// RBACPolicy via spec.rbacPolicy.
+type RBACPolicySpec struct {
+	// +required
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=200
```
zerospiel, line 95: Two schema thoughts.

**`listType=atomic` on `bindings`.** Consider `listType=map` with `listMapKey: name` instead. `ValidateRBACPolicy` already enforces that `bindings[].name` is unique — as a map list the API server would enforce it for free, and you could drop the `duplicate binding name` check entirely. It also makes server-side apply co-ownership possible, should anything ever want to manage a subset of bindings.

**`MaxItems` bounds.** `maxItems: 32768` on both `rules` and `subjects` (inside each of up to 200 bindings) means the theoretical object blows well past etcd's ~1.5 MiB limit long before the schema limits bite, so the failure mode is a confusing apiserver rejection rather than a clear validation error. Something like 1024 on each would still be far more than any real policy needs.

api/v1beta1/rbacpolicy_types.go
eromanova: Right, this is a stale comment. Will be updated. I suggest keeping the RBACPolicy status implementation and whether we need a separate RBACPolicy controller as a follow-up activity.

internal/controller/clusterdeployment_controller.go
eromanova: Let's handle RBACPolicy controller and the finalizer separately. I added the `NotFound` handling, which closes the wedge and revokes access.

internal/controller/rbac/rbacpolicy.go
eromanova: @zerospiel If you don't mind, I'll address this comment separately.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1146,131 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+	}
+
+	desiredRoles, desiredBindings, syncChanged, syncErr := rbac.Sync(ctx, childCl, scope.rbacPolicy)
+	pruneChanged, pruneErr := rbac.Prune(ctx, childCl, desiredRoles, desiredBindings)
+	if err := errors.Join(syncErr, pruneErr); err != nil {
+		err = fmt.Errorf("failed to sync RBAC objects: %w", err)
+		if r.setCondition(cd, kcmv1.RBACPolicyReadyCondition, kcmv1.FailedReason, metav1.ConditionFalse, err) {
+			r.warnf(cd, "RBACSyncFailed", err.Error())
+		}
+		if !rbac.Retriable(err) {
```
zerospiel: **bug** — `Retriable` is called on the *wrapped* error here, which makes it misclassify mixed failures as terminal.

Four lines up, `err` was rewrapped with `fmt.Errorf("failed to sync RBAC objects: %w", err)`, so by the time it reaches `Retriable` it's a `*fmt.wrapError`. `Retriable` starts with `err.(interface{ Unwrap() []error })` — `*fmt.wrapError` implements `Unwrap() error`, not `Unwrap() []error`, so that assertion fails and it falls through to `!errors.Is(err, ErrTerminal)`, which walks the whole join tree and reports terminal if *any* leaf is.

I pulled the function out and ran it standalone to be sure:

```
unwrapped mixed        -> Retriable=true   (want true)
fmt.Errorf-wrapped mix -> Retriable=false  (want true)   <-- this call site
wrapped terminal-only  -> Retriable=false  (want false)
wrapped transient-only -> Retriable=true   (want true)
```

So one binding hitting `ErrTerminal` alongside another hitting a transient apiserver error classifies the whole batch as terminal: the transient failure is never retried, and the resync is dropped too (see the next comment).

Two ways to fix — classify before wrapping:

```go
joined := errors.Join(syncErr, pruneErr)
if joined != nil {
    retriable := rbac.Retriable(joined)
    err := fmt.Errorf("failed to sync RBAC objects: %w", joined)
    ...
}
```

or make `Retriable` unwrap single-error wrappers before the `[]error` assertion, which is more robust against a future caller doing the same thing.

Worth noting all three existing assertions (`rbacpolicy_test.go:152`, `:199`, `:276`) test the raw `Sync` error and never the wrapped form the controller actually produces — which is why this passes CI. A case covering the wrapped mixed error would have caught it.

zerospiel (commented):
## Second pass — `3ba3987f`

Thanks, this addressed almost everything and the shape is much better. `go build ./...`, the RBAC/validation/webhook unit tests and `golangci-lint` over the changed packages are all clean (the one `wastedassign` hit in `api/v1beta1/indexers.go:348` is pre-existing and unrelated to this PR).

**The CRD drift I flagged last round is genuinely resolved.** I regenerated with `controller-gen-v0.21.0` + `yq` into a temp dir and diffed against the committed file — byte-identical. `k8s.io/api` moving to `v0.37.0` is what closed it: that version lowercased the `rbacv1.PolicyRule` field docs, so the committed CRD now matches the pinned deps.

### What checks out

`Sync` pre-populating both desired sets before applying anything is exactly the right shape — a partial failure can no longer make `Prune` revoke a still-wanted binding, and dropping `rules` now keeps the managed ClusterRole alive for the ClusterRoleBinding still pointing at it. Also verified: per-binding error accumulation, the `ErrTerminal` sentinel, the `NotFound` → revoke path that keeps the False condition visible for a dangling reference, decoupling revocation from the CAPI Cluster (with `V(1)` feedback for the adopted-cluster case), `patchIfChanged` replacing the double-Get, `PartialObjectMetadataList` + `MatchingLabels` in `pruneManaged`, resync jitter, the applied generation in the condition message, `listType=map` + `listMapKey`, the tightened `maxItems`/`maxLength`, `clusterRole` DNS-1123 validation, and the corrected N:1 doc comment.

One thing I double-checked because it looked risky: `pruneManaged` passing the singular kind (`WithKind("ClusterRole")`) to a `PartialObjectMetadataList` is fine — controller-runtime's metadata client trims a `List` suffix if present rather than requiring one, and the explicit `SetGroupVersionKind` before `Delete` is both required and present.

### Five new findings, inline

The first is a confirmed logic bug in the new terminal/retriable classification; the rest are smaller. Nothing here is structural — it's all local to the code added in this commit.

### Still open from round one

Listing these only so we can confirm they're the intended follow-up set and nothing fell through the cracks: no status subresource or printer columns on RBACPolicy; no namespaced Role/RoleBinding support; no privilege-escalation guard on `rules`; no `RBACPolicyKind` const or `builtinKinds()` entry; no predicate on the RBACPolicy watch; `childClientFactory` still a nil-defaulted test seam; the new controller tests still use `context.Background()` rather than `t.Context()` (14 occurrences) and `rbacpolicy_webhook_test.go` still shares one `g` across subtests; `deleteChildResources` still recomputes the secret key for its log line without the `key` field.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1146,131 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+			// Only an RBACPolicy edit can clear this; retrying just burns the rate limiter.
+			l.Error(err, "RBACPolicy cannot be applied, will not retrigger this error")
+			return ctrl.Result{}, nil
```
zerospiel: Returning a fully zero `Result` here also drops the periodic drift resync, not just the rate-limited retry.

Not requeueing on the rate limiter for a terminal error is right. But this ClusterDeployment now stops re-syncing RBAC entirely until some unrelated event fires — so if a policy contains one unfixable binding and somebody then deletes a ClusterRoleBinding in the child cluster by hand, nothing restores it. Given the child cluster isn't watched, the 5-minute resync is the only drift detection there is, and one bad binding shouldn't switch it off for the whole policy.

```go
return ctrl.Result{RequeueAfter: jitteredRBACResync()}, nil
```

keeps the cadence without burning the limiter.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1146,131 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+
+	// The applied generation goes in the message so the condition changes when the policy does,
+	// letting a user (or kubectl wait) tell which revision actually reached the child cluster.
+	if apimeta.SetStatusCondition(cd.GetConditions(), metav1.Condition{
+		Type:               kcmv1.RBACPolicyReadyCondition,
+		Status:             metav1.ConditionTrue,
+		Reason:             kcmv1.SucceededReason,
+		Message:            fmt.Sprintf("RBACPolicy %s generation %d applied", scope.rbacPolicy.Name, scope.rbacPolicy.Generation),
+		ObservedGeneration: cd.Generation,
+	}) && (syncChanged || pruneChanged) {
```
zerospiel: **event loss** — chaining the event onto `SetStatusCondition`'s return means drift repairs no longer emit `RBACSynced`.

The condition message only changes when the RBACPolicy generation does. So in the common drift case — policy unchanged, someone deleted a ClusterRoleBinding in the child cluster, `Sync` recreates it — `syncChanged` is true but the condition is byte-identical, `SetStatusCondition` returns false, and no event fires. The repair is invisible in `kubectl describe`, which is precisely where you'd look for it.

These two signals are independent: the condition tracks *which revision is applied*, the event records *that something changed in the child cluster*. The previous form was correct:

```go
apimeta.SetStatusCondition(cd.GetConditions(), metav1.Condition{...})
if syncChanged || pruneChanged {
    r.eventf(cd, "RBACSynced", "ClusterRoles and ClusterRoleBindings synced to the child cluster")
}
```

internal/util/validation/rbacpolicy.go
```diff
@@ -0,0 +1,82 @@
...
+
+		// binding.Name is concatenated into the generated ClusterRoleBinding's metadata.name, which
+		// the child cluster's API server requires to be a valid DNS-1123 subdomain. Checking the
+		// combined form here fails fast instead of only when the operator applies it.
+		objectName := kcmv1.ClusterRoleBindingNamePrefix + binding.Name
+		if errs := apivalidation.IsDNS1123Subdomain(objectName); len(errs) > 0 {
+			return fmt.Errorf("binding name %q produces an invalid ClusterRoleBinding name %q: %s", binding.Name, objectName, strings.Join(errs, "; "))
+		}
+
+		seen := roleSubjects[binding.ClusterRole]
```
zerospiel: This duplicate-subject check spans bindings rather than living inside one, and that rejects legitimate policies.

`seen` is keyed by `binding.ClusterRole` and accumulates across the whole loop, so any subject appearing in two bindings that share a ClusterRole is an error:

```yaml
- name: team-a-view
  clusterRole: view
  subjects: [{kind: User, name: alice}, {kind: User, name: bob}]
- name: team-b-view
  clusterRole: view
  subjects: [{kind: User, name: bob}, {kind: User, name: carol}]   # rejected
```

Two teams sharing a role with one overlapping member is a perfectly normal org structure, and the two resulting ClusterRoleBindings are valid Kubernetes. My original note was about duplicates *within* a single binding — that's the case that's unambiguously a typo. Scoping `seen` to the binding and dropping the `roleSubjects` map would cover it:

```go
seen := make(map[kcmv1.RBACPolicySubject]struct{}, len(binding.Subjects))
for _, subject := range binding.Subjects { ... }
```

Minor either way: the error message names only the second binding, so a user has to go hunting for the first one.

internal/controller/clusterdeployment_controller.go
```diff
@@ -444,7 +491,30 @@ func (r *ClusterDeploymentReconciler) updateCluster(
 		return ctrl.Result{RequeueAfter: r.defaultRequeueTime}, nil
 	}
 
-	return r.reconcileHelmRelease(ctx, clusterTpl, scope)
+	// RBAC distribution only needs the child kubeconfig, not a fully ready cluster, so it runs
+	// regardless of the helm result: gating it would stall grants and, worse, revocations while a
+	// cluster is mid-upgrade or has a single NotReady machine.
+	helmResult, helmErr := r.reconcileHelmRelease(ctx, clusterTpl, scope)
+	rbacResult, rbacErr := r.ensureRBACPolicy(ctx, scope)
```
zerospiel, line 500: Flagging the follow-on cost of ungating this, since it was my request last round rather than anything you got wrong.

`ensureRBACPolicy` now runs on essentially every ClusterDeployment reconcile, and each run does one Get per binding (plus one per role carrying rules) and two Lists, all against the non-caching child client. At the schema's `maxItems: 200` that's roughly 200-400 GETs plus 2 LISTs per reconcile — and reconciles are most frequent exactly when a cluster is busy: helm status churn, CAPI condition updates, service reconciles during an upgrade.

`patchIfChanged` means those are reads rather than writes now, which is a genuine improvement over the previous `CreateOrUpdate` path. But the read volume is new, and it used to be limited by the helm gate.

A cheap guard recovers it: skip the sync when the RBACPolicy generation is unchanged **and** the last successful sync is younger than `rbacResyncInterval`. The applied generation is already in the condition message, though parsing it back out of prose is unpleasant — a small status field (or even an in-memory map keyed by CD UID) would read much better.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1146,131 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+		r.eventf(cd, "RBACSynced", "ClusterRoles and ClusterRoleBindings synced to the child cluster")
+	}
+
+	return ctrl.Result{RequeueAfter: jitteredRBACResync()}, nil
+}
+
+// revokeRBACPolicy removes everything [rbac.Sync] previously created in cd's child cluster.
+func (r *ClusterDeploymentReconciler) revokeRBACPolicy(ctx context.Context, scope *clusterScope) error {
+	cd := scope.cd
+	if apimeta.FindStatusCondition(*cd.GetConditions(), kcmv1.RBACPolicyReadyCondition) == nil {
```
zerospiel: Same read-amplification shape as the sync path, in the misconfiguration case.

A ClusterDeployment with a typo'd `spec.rbacPolicy` gets its False condition set in `getClusterScope`, so this guard always passes, and every subsequent reconcile builds a child client and runs two Lists that delete nothing — indefinitely, for what is purely a bad reference.

Keeping the False condition visible is right; doing the prune every time isn't. Distinguishing "the reference is broken and nothing was ever synced" from "we had a successful sync and now need to revoke" would let this return early. The condition's `Reason` could carry that, or the same applied-generation state suggested on the sync path.

internal/controller/clusterdeployment_controller.go
```diff
@@ -444,7 +491,30 @@ func (r *ClusterDeploymentReconciler) updateCluster(
...
+	if err := errors.Join(helmErr, rbacErr); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return soonerResult(helmResult, rbacResult), nil
+}
+
+// soonerResult merges two reconcile results, keeping the earliest requeue either asks for.
+func soonerResult(a, b ctrl.Result) ctrl.Result {
+	res := ctrl.Result{Requeue: a.Requeue || b.Requeue} //nolint:staticcheck // matches ctrl.Result
```
zerospiel: Neither `reconcileHelmRelease` nor `ensureRBACPolicy` ever sets `Requeue`, so this line is dead weight that also costs a `//nolint:staticcheck` on a deprecated field. Dropping it removes both and leaves the `RequeueAfter` merge, which is the part that does the work.

internal/controller/rbac/rbacpolicy.go
```diff
@@ -0,0 +1,251 @@
...
+// already-existing ClusterRole such as a built-in "admin"/"edit"/"view", matching
+// RBACPolicyBinding.Rules' doc comment.
+func applyClusterRole(ctx context.Context, childCl client.Client, name string, rules []rbacv1.PolicyRule) (bool, error) {
+	role := &rbacv1.ClusterRole{ObjectMeta: metav1.ObjectMeta{Name: name}}
+	err := childCl.Get(ctx, client.ObjectKeyFromObject(role), role)
+	switch {
+	case apierrors.IsNotFound(err):
+		role.Labels = mergeManagedLabels(nil)
+		role.Rules = rules
+		return true, childCl.Create(ctx, role)
```
zerospiel: `return true, childCl.Create(ctx, role)` reports `changed=true` even when the Create fails. Same on line 161 for the binding.

Harmless today — `Sync` discards `changed` on the error path — but it's a claim the code can't actually make, and it'd quietly become an incorrect `RBACSynced` event if the caller ever stopped discarding it. `if err := childCl.Create(...); err != nil { return false, err }; return true, nil` says what's true.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1152,170 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+
+// revokeRBACPolicy removes everything [rbac.Sync] previously created in cd's child cluster.
+func (r *ClusterDeploymentReconciler) revokeRBACPolicy(ctx context.Context, scope *clusterScope) error {
+	cd := scope.cd
+	if apimeta.FindStatusCondition(*cd.GetConditions(), kcmv1.RBACPolicyReadyCondition) == nil {
+		return nil
+	}
+	// A reference that never synced has nothing to revoke, so don't list the child cluster on
+	// every reconcile just because spec.rbacPolicy holds a typo.
+	if cd.Spec.RBACPolicy != "" && !scope.rbacSynced {
```
zerospiel: **A failed revoke is never retried — the bindings stay live forever.**

This guard reads a signal that the same reconcile destroys. `getClusterScope` sets `scope.rbacSynced` from `IsStatusConditionTrue(RBACPolicyReadyCondition)` and then immediately overwrites that condition with `False`, which the deferred `updateStatus` persists. So:

- reconcile 1: condition is `True` → guard passes → prune runs → child cluster unreachable → error, and `False` gets written
- reconcile 2: condition is `False` → `rbacSynced=false` → `return nil`, prune skipped
- forever after: same

```
reconcile 1: revoke attempted and failed as expected (failed to get child cluster client for RBAC cleanup: ...)
reconcile 2: scope.rbacSynced re-derived = false
LEAK: ClusterRoleBinding "k0rdent-compute-admin" still present after the retry — revoke was skipped
```

A controller restart between the `False` write and a successful prune does the same thing. `r.rbacSynced.Delete(cd.UID)` on the line above runs *before* the prune, so the in-memory map can't serve as a fallback either.

Worth saying the cleared-spec path is correct — `cd.Spec.RBACPolicy == ""` short-circuits the guard, the condition is only removed after a successful prune, and failures retry. It's specifically the deleted-CR path that's one-shot. Anything that survives the failure works as a fix: delete the `rbacSynced` entry only after the prune succeeds and guard on that, use a distinct condition `Reason` for "revoked" vs "revoke pending", or add a finalizer.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1152,170 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+
+// rbacSyncFresh reports how long is left before policy needs re-syncing to cd, and whether the
+// sync can be skipped until then.
+func (r *ClusterDeploymentReconciler) rbacSyncFresh(cd *kcmv1.ClusterDeployment, policy *kcmv1.RBACPolicy) (time.Duration, bool) {
+	v, ok := r.rbacSynced.Load(cd.UID)
+	if !ok {
+		return 0, false
+	}
+	state, ok := v.(rbacSyncState)
+	if !ok || state.generation != policy.Generation {
```
zerospiel: **Switching `spec.rbacPolicy` between two policies at the same generation skips the sync entirely.**

Freshness is keyed on the generation number and nothing else, and two different RBACPolicy objects are both at generation 1 almost all the time. So `policy-a` → `policy-b` looks fresh:

```
after switching to policy-b: result={false 4m59.999985417s <nil>}
NOT APPLIED: policy-b's binding k0rdent-b is absent
NOT REVOKED: policy-a's binding k0rdent-a is still live
```

For up to five minutes the child cluster holds the old policy's grants and none of the new one's, with `RBACPolicyReady=True` on the ClusterDeployment. The cached state needs the policy identity as well — UID is safest, name would do.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1152,170 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+	// Cluster deleted out-of-band is an anomaly to report, not a signal to strip access from a
+	// still-reachable child cluster.
+	if scope.rbacPolicy == nil {
+		return ctrl.Result{}, r.revokeRBACPolicy(ctx, scope)
+	}
+
+	// Reconciles are frequent (helm status churn, CAPI updates); the child client is non-caching,
+	// so re-reading every binding on each of them is wasted. Nothing but this controller writes
+	// these objects, so an unchanged policy only needs the periodic drift resync.
+	if left, ok := r.rbacSyncFresh(cd, scope.rbacPolicy); ok {
```
zerospiel, line 1180: Small one: the early return happens before any condition handling, so the condition goes unmaintained while fresh.

Two ways that shows up. If the status write from the successful sync didn't land (conflict → `updateStatus` errors → reconcile retries), the retry takes this path and never re-sets the condition, so `RBACPolicyReady` is missing from the persisted status until the interval elapses. And the condition's `ObservedGeneration` is stamped with `cd.Generation` at sync time, so editing the ClusterDeployment spec without touching the policy leaves it lagging `metadata.generation` for five minutes — which is the staleness check tooling looks at.

Re-asserting the condition with the same message before returning costs nothing and covers both.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1152,170 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+func (r *ClusterDeploymentReconciler) rbacSyncFresh(cd *kcmv1.ClusterDeployment, policy *kcmv1.RBACPolicy) (time.Duration, bool) {
+	v, ok := r.rbacSynced.Load(cd.UID)
+	if !ok {
+		return 0, false
+	}
+	state, ok := v.(rbacSyncState)
+	if !ok || state.generation != policy.Generation {
+		return 0, false
+	}
+	left := rbacResyncInterval - time.Since(state.syncedAt)
```
zerospiel, line 1265: The jitter doesn't survive this. `jitteredRBACResync()` returns something in [4.5m, 5.4m), but freshness is measured against the un-jittered constant — so when jitter lands below 5m the wakeup finds `left > 0`, skips, and requeues for the remainder. The sync always lands at exactly 5m, every ClusterDeployment settles onto a precise 5-minute grid, and the below-5m half of the jitter is discarded at the cost of an extra no-op reconcile per cycle.

Storing the computed next-sync deadline in `rbacSyncState` instead of recomputing `syncedAt + rbacResyncInterval` makes the two agree.

internal/controller/clusterdeployment_controller.go
```diff
@@ -2219,6 +2457,17 @@ func (*ClusterDeploymentReconciler) warnf(cd *kcmv1.ClusterDeployment, reason, m
 }
 
 // SetupWithManager sets up the controller with the Manager.
+// ignoreDeletePredicate passes create and update events only: deleting a referenced object must
```
zerospiel, line 2468: This landed between `// SetupWithManager sets up the controller with the Manager.` and the function that comment belongs to, so godoc now shows that line above `ignoreDeletePredicate` and `SetupWithManager` is undocumented. Just needs the comment moved down.

zerospiel (commented):
## Third pass — `72a83952`

Everything from the last round is fixed and verified, including the `Retriable` unwrap (I re-ran the wrapped-mixed case). Build, vet, tests and lint are all clean, with no lint hits in any changed file.

The new freshness cache brought two bugs with it, though, and both are access-control correctness rather than style — I'd want them fixed before merge. I confirmed both with throwaway tests against the real package rather than by reading, so the output in the comments is real.

Everything else below is small.

Two things I checked because they looked suspicious and turned out fine: `GenerationChangedPredicate` on the RBACPolicy watch does still pass Create and Delete (only `Update` is overridden), so deletion enqueues and revokes as your comment says; and RBACPolicy having no status subresource doesn't break it either, since the apiextensions strategy bumps `generation` on any non-metadata change regardless.

One small thing with no natural anchor: `rbacSynced` entries are only removed in `reconcileDelete`'s success tail and in `revokeRBACPolicy`, so a ClusterDeployment whose finalizer is force-removed, or that's deleted while the controller is down, leaves its entry for the life of the process. Tiny each, but unbounded — low priority, and `restMapperCache` already has a TTL-sweeper pattern if you ever want one.

Still open from round one, unchanged and presumably deferred: no status subresource or printer columns, no namespaced Role/RoleBinding, no privilege-escalation guard on `rules`, no `RBACPolicyKind` const or `builtinKinds()` entry, `childClientFactory` still a nil-defaulted test seam, and `deleteChildResources` still recomputes the secret key for its log line without the `key` field.

zerospiel (commented):
## Fourth pass — `708757f6`

Both major fixes are correct. The UID+generation key really does fix the policy-switch case, and moving ownership of `RBACPolicyReadyCondition` into `revokeRBACPolicy` — keeping it `True` until the prune succeeds, and deleting the cache entry after rather than before — fixes the one-shot revoke properly, including across a restart. The two new tests cover both, and the displaced `SetupWithManager` comment is sorted. Build, vet, tests and lint all clean, nothing in the changed files.

One new finding inline, same family as the one it replaced but much narrower, plus a cosmetic nit. I reproduced the first one against the real package.

---

## Suggested follow-up PR

Pulling together everything deferred across the four rounds, roughly in the order I'd tackle it.

**API surface — the biggest gap.** RBACPolicy has no status subresource at all, so the object tells you nothing about itself: not which ClusterDeployments consume it, not whether any sync succeeded, no `observedGeneration`. Every signal lives on the ClusterDeployment side. For a resource whose entire job is granting cluster access, `kubectl get rbacpolicy` showing only NAME/AGE is a real gap. Worth `status.conditions` + `status.observedGeneration` + the referencing ClusterDeployments (or a count), and a couple of printer columns while you're in there. This would also give the controller somewhere durable to record "something is granted", which is the state both revoke bugs have been dancing around.

**Scope of what can be granted.** Only ClusterRole/ClusterRoleBinding today — no namespaced `Role`/`RoleBinding`, no binding a subject to a `Role`, no ServiceAccount subjects. For the multi-tenant story ("project X gets `edit` in namespace Y") the namespaced shape is arguably the more common one, and retrofitting it later means either reshaping `bindings` or adding a second list beside it.

**Privilege escalation.** `rbacpolicies-creator` aggregates to `namespace-admin` and `rules` accepts arbitrary PolicyRules including `*/*/*`, so a management-cluster namespace admin can mint themselves `cluster-admin` in every child cluster in that namespace. In-cluster RBAC prevents exactly this with the `escalate`/`bind` verbs; here the controller holds CAPI's admin kubeconfig, so there's no equivalent. Very possibly intended — a namespace admin already owns the cluster — but it should be a documented decision rather than an emergent one.

**Sync freshness leftovers** (from the third round, all small):
- The freshness skip returns before any condition handling. A lost status write (conflict → `updateStatus` errors → retry) isn't re-asserted, and the condition's `ObservedGeneration` lags `metadata.generation` for up to five minutes after an unrelated ClusterDeployment spec edit. Re-asserting the condition with the same message before returning covers both.
- `rbacSyncFresh` measures against the un-jittered `rbacResyncInterval`, so a sub-5m jittered wakeup finds `left > 0`, skips, and requeues for the remainder. The sync always lands at exactly 5m and every ClusterDeployment settles onto a precise grid — the jitter is discarded, at the cost of an extra no-op reconcile per cycle. Storing the computed next-sync deadline in `rbacSyncState` makes the two agree.
- `rbacSynced` entries are only removed in `reconcileDelete`'s success tail and in `revokeRBACPolicy`, so a ClusterDeployment whose finalizer is force-removed, or that's deleted while the controller is down, leaves its entry for the life of the process. Tiny each but unbounded; `restMapperCache` already has a TTL-sweeper pattern if you want one.

**Housekeeping:**
- Add an `RBACPolicyKind` const for symmetry with `ClusterAuditPolicyKind` / `ClusterAuthenticationKind`, and `&kcmv1.RBACPolicy{}` to `AccessManagementReconciler.builtinKinds()` so a shared policy in the system namespace is watched rather than only polled.
- `childClientFactory` is a production struct field documented as "Left nil in production", with a `factory == nil` fallback evaluated on every call. Setting it once in `cmd/main.go` would read better.
- `deleteChildResources` lost its `"key", secretKey` log field in the original refactor and now recomputes `GetKubeconfigSecretKey` purely for the log line.

**Leftovers:**
https://github.com/k0rdent/kcm/pull/3045#discussion_r3978580185
https://github.com/k0rdent/kcm/pull/3045#discussion_r3978580190

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1148,183 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+}
+
+// revokeRBACPolicy removes everything [rbac.Sync] previously created in cd's child cluster, once
+// spec.rbacPolicy is cleared or the RBACPolicy it names is gone. RBACPolicyReadyCondition stays
+// True for as long as anything is still granted, so a failed revoke is retried.
+func (r *ClusterDeploymentReconciler) revokeRBACPolicy(ctx context.Context, scope *clusterScope) error {
+	cd := scope.cd
+	conds := cd.GetConditions()
+
+	if cd.Spec.RBACPolicy != "" && !apimeta.IsStatusConditionTrue(*conds, kcmv1.RBACPolicyReadyCondition) {
```
zerospiel, line 1276: **A partially-successful sync grants bindings but leaves the condition `False`, and this guard reads `False` as "nothing was ever granted".**

`Sync` deliberately continues past per-binding failures — that was the fix two rounds ago — so a policy with nine good bindings and one bad one creates nine ClusterRoleBindings and *still* ends with the condition `False`. This guard then skips the prune entirely:

```
after partial sync: RBACPolicyReady=False (failed to sync RBAC objects: applying ClusterRole admin:
  terminal error: ClusterRole admin already exists and was not created by this RBACPolicy...)
k0rdent-good IS live in the child cluster
LEAK: k0rdent-good survived the revoke - the grant is still live
```

Much narrower than the bug it replaced: it only bites on the deleted-CR path, and `RBACPolicyDeletionAllowed` normally blocks deleting a referenced policy — so it needs the webhook out of the picture, which is exactly the `IsDisabledValidationWH` mode the controller supports. The cleared-spec path is fine, since it skips this guard and reaches the prune via the `FindStatusCondition != nil` check below.

Root cause is that condition-`True` is standing in for "something is granted", and a partial sync breaks that equivalence. `Sync` already returns `changed`, so recording the grant whenever `changed` is true — error or not — would close it. A persisted marker would be better still, and is the same thing the status subresource in the follow-up would give you.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1076,6 +1148,183 @@ func (r *ClusterDeploymentReconciler) ensureAuditPolicyConfigMap(ctx context.Con
...
+			return fmt.Errorf("failed to revoke RBAC objects: %w", err)
+		}
+	}
+	r.rbacSynced.Delete(cd.UID)
+
+	if cd.Spec.RBACPolicy == "" {
+		apimeta.RemoveStatusCondition(conds, kcmv1.RBACPolicyReadyCondition)
+		return nil
+	}
+	err = fmt.Errorf("RBACPolicy %s/%s not found, the RBAC objects it granted have been revoked", cd.Namespace, cd.Spec.RBACPolicy)
```
zerospiel, line 1304: Cosmetic: this message survives exactly one reconcile.

The next pass takes the dangling-reference branch at the top (spec still names the policy, condition is now `False`), which rewrites the message to the shorter `"RBACPolicy %s/%s not found"`. So the condition flips once, fires a second `RBACPolicyError` warning event, and ends up with the *less* informative of the two messages.

Sharing one message between the branches, or having the dangling branch leave an existing `False` condition alone, avoids it.

internal/controller/clusterdeployment_controller.go
zerospiel: Part of the follow-up

internal/controller/clusterdeployment_controller.go
zerospiel: Part of the follow-up

internal/controller/clusterdeployment_controller.go
zerospiel: Part of the follow-up

internal/controller/clusterdeployment_controller.go
zerospiel: Part of the follow-up

internal/controller/clusterdeployment_controller.go
zerospiel: Part of the follow-up

zerospiel (approved):

PR #3065, opened 2026-09-14
fix: RBAC policy operator follow-ups

eromanova:
Jira: KCM-1238

Follow-up to #3045.

## Revocation correctness

**A partially applied policy is no longer left granted after a revoke.** `rbac.Sync` continues past a failing binding, so nine good bindings plus one bad one grants nine ClusterRoleBindings and still reports `RBACPolicyReady=False`. The prune guard read that False as "nothing was ever granted" and skipped the prune, leaking those grants — reachable on the deleted-CR path with the webhook disabled. The condition now doubles as the grant record via the new `RBACPolicyPartiallyAppliedReason`, which `rbacGranted` (backed by the `rbacSynced` entry, so a lost status write doesn't lose the marker) checks instead.

**Condition message churn.** The deleted-CR message survived one reconcile before the dangling-reference branch rewrote it and fired a second `RBACPolicyError`. Both branches now share one message and `RBACPolicyNotFoundReason`.

## Sync freshness

- The freshness skip returned before any condition handling, so a lost status write was never re-asserted and `ObservedGeneration` lagged for up to 5m. It now re-asserts the condition first.
- Freshness was measured against the un-jittered interval, so sub-5m wakeups skipped and requeued for the remainder — half the jitter discarded, every ClusterDeployment on a 5m grid. `rbacSyncState` now stores the computed `nextSyncAt`.
- `rbacSynced` entries leaked for the life of the process on a force-removed finalizer or a delete during downtime. Added a TTL sweeper following the `restMapperCache` pattern.

## Housekeeping

- `RBACPolicyKind` const, `&kcmv1.RBACPolicy{}` in `builtinKinds()` so shared system-namespace policies are watched rather than polled, and `RBACPolicyKind` in `builtinResourceKinds` for `apiGroup` defaulting.
- `childClientFactory` defaulted once in `SetupWithManager` instead of per call.
- `childClientFor` returns the kubeconfig Secret key, so `deleteChildResources` logs it without recomputing.
- `rbac.Prune` reports what it deleted before failing.
- Documented in the `RBACPolicy` godoc (and so in the CRD description) that grants are cluster-scoped only and that `rules` accepts arbitrary PolicyRules by design.


internal/controller/clusterdeployment_controller.go
```diff
@@ -1214,7 +1237,15 @@ func (r *ClusterDeploymentReconciler) ensureRBACPolicy(ctx context.Context, scop
...
+		// real access while reporting False. The Reason carries that, since a False condition on
+		// its own reads as "nothing was ever granted" — see rbacGranted. Unconditional rather than
+		// keyed on syncChanged/pruneChanged: whether this pass happened to write anything says
+		// nothing about what an earlier one left in the child cluster, and re-applying an
+		// already-applied binding reports no change, so the very state this marker exists for is
+		// the one where both flags are false. Overstating a grant only costs revokeRBACPolicy one
+		// List of managed objects; understating one leaks access that is never revoked.
+		if r.setCondition(cd, kcmv1.RBACPolicyReadyCondition, kcmv1.RBACPolicyPartiallyAppliedReason, metav1.ConditionFalse, err) {
 			r.warnf(cd, "RBACSyncFailed", err.Error())
 		}
```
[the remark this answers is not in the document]
eromanova: Fixed in 079e5eb6. `ensureRBACPolicy` now stores an `rbacSynced` entry on the partial-failure path too, right after the condition is set — unconditionally rather than keyed on `syncChanged`/`pruneChanged`, for the same reason the condition is: re-applying an already-applied binding reports no change, so the state this marker exists for is exactly the one where both flags are false. `nextSyncAt` is `time.Now()`, so the entry records the grant without making the next reconcile treat the failed sync as fresh.

The other paths you list are covered separately:
- 1310: real, and fixed — `getClusterScope` set `FailedReason` unconditionally on the fetch and inline-validation failures, erasing a `RBACPolicyPartiallyApplied` marker before `ensureRBACPolicy` ever ran. Both now go through `rbacFailureReason`.
- 1385: real, and fixed — `sweepRBACSynced` now uses `CompareAndDelete` on the value read by `Range`.
- 1425: real, and fixed — a nil child client with the CAPI Cluster still present now returns an error instead of clearing the marker and the condition. With no CAPI Cluster (an adopted cluster, or one already gone) there is nothing left to prune, so the revoke still completes.
- 1303: not changed. The persisted condition is the durable record and survives a restart; the `sync.Map` only covers the window where `rbac.Sync` has written to the child but the status update has not landed. Closing that remaining crash window would mean listing the child cluster on reconciles that have no reason to build a child client at all, which is the cost this design is avoiding — noted in `rbacGranted`'s doc comment.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1226,27 +1271,85 @@ func (r *ClusterDeploymentReconciler) ensureRBACPolicy(ctx context.Context, scop
...
+
+// rbacGranted reports whether ClusterRoles/ClusterRoleBindings this operator created may still be
+// live in the ClusterDeployment's child cluster, which is what decides whether revokeRBACPolicy
+// has anything to prune. RBACPolicyReadyCondition doubles as that record: True means the whole
+// policy is applied, kcmv1.RBACPolicyPartiallyAppliedReason that some of its bindings are.
+// [rbac.Sync] deliberately continues past a failing binding, so a partial sync grants real access
+// while still reporting False — reading the status alone would call that "nothing was ever
+// granted" and skip the prune. It has to be recorded on the ClusterDeployment rather than
+// anywhere on the RBACPolicy, since the case it exists for is the RBACPolicy being deleted.
+func (r *ClusterDeploymentReconciler) rbacGranted(cd *kcmv1.ClusterDeployment) bool {
```
zerospiel: There's still a window where a grant exists and nothing records it.

The marker is written after `rbac.Sync` has already written to the child cluster. For the first-ever sync, if the status write never lands (crash, or a conflict that's retried into a different path) *and* the process restarts, taking `rbacSynced` with it, *and* the RBACPolicy is deleted before the next successful sync, then `rbacGranted` returns false and the revoke is skipped:

```
condition: status=False reason=RBACPolicyNotFound
RESIDUAL LEAK: k0rdent-good still live; no record said anything was granted
```

It needs all three to line up, and any earlier successful sync closes it, because the persisted `True` then carries the marker. Your own test covers the same-process version of this, which the in-memory map handles.

The only thing that closes it fully is a marker written *before* the sync rather than after — a finalizer, an annotation on the ClusterDeployment, or a status field. Fine to leave as a known gap, but worth writing down somewhere rather than leaving it implied.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1226,27 +1271,85 @@ func (r *ClusterDeploymentReconciler) ensureRBACPolicy(ctx context.Context, scop
...
+	if cond == nil {
+		return false
+	}
+	return cond.Status == metav1.ConditionTrue || cond.Reason == kcmv1.RBACPolicyPartiallyAppliedReason
+}
+
+// rbacFailureReason picks the Reason for a False RBACPolicyReadyCondition, carrying forward any
+// grant an earlier reconcile recorded: the marker must survive every failure until a revoke
+// actually clears it.
+func (r *ClusterDeploymentReconciler) rbacFailureReason(cd *kcmv1.ClusterDeployment) string {
```
zerospiel: Using the Reason as the grant marker costs you the failure detail it used to carry.

Once anything is granted, this and `rbacProgressingReason` replace the real reason at every failure site. So after the first successful sync, "failed to fetch the RBACPolicy", "RBACPolicy is invalid" and "child client error" all report `RBACPolicyPartiallyApplied`, and `FailedReason` never appears again for this condition. The Message still tells them apart, but Reason is the machine-readable half, so anything filtering on it can no longer distinguish them.

It also produces combinations like `Unknown` / `RBACPolicyPartiallyApplied` / `"child cluster kubeconfig not ready yet"`, where the reason doesn't describe the status.

Same fix as the other comment: a dedicated field for the marker (a `status.rbacPolicyGranted` on ClusterDeployment, or a finalizer) would free Reason to mean what it normally means, and would also close the window above if it's written before the sync.

internal/controller/rbac/rbacpolicy.go
```diff
@@ -217,9 +217,11 @@ func mergeManagedLabels(existing map[string]string) map[string]string {
 // there — used both for normal drift cleanup after a [Sync], and to tear down everything a
 // ClusterDeployment's child cluster once had once it stops referencing an RBACPolicy at all.
 func Prune(ctx context.Context, childCl client.Client, desiredRoles, desiredBindings map[string]struct{}) (bool, error) {
+	// rolesChanged is carried out even on failure: pruneManaged reports what it deleted before it
```
zerospiel: Nothing reads this. `ensureRBACPolicy` stores the `rbacSynced` entry unconditionally on the error path, the `RBACSynced` event only fires when `err == nil`, and `revokeRBACPolicy` discards the bool entirely — so no caller looks at `rolesChanged` when `Prune` returns an error.

Harmless as a change, but the comment claims a caller behaviour that doesn't exist yet. Either wire it up or trim the comment.

Unrelated nit while I'm here: `childClientFor` now assumes `childClientFactory` is non-nil. That's true via `SetupWithManager`, but a reconciler built without it panics rather than degrading. Worth a line in the field comment.

zerospiel (commented):
## Review of the follow-ups

Checked both commits. Build, vet, the full test run and lint are clean, with no lint hits in any touched file. I regenerated the three CRDs with `controller-gen-v0.21.0` and they match what's committed. I also re-ran the round-4 partial-sync case against the real package to confirm the fix, and it passes:

```
after partial sync: status=False reason=RBACPolicyPartiallyApplied
FIXED: k0rdent-good was revoked
```

### Fixed

- partial sync grants access but the revoke skips it (round 4)
- condition message flipping once after a revoke (round 4)
- freshness skip leaving the condition unmaintained (round 3)
- jitter discarded by the freshness check (round 3)
- `rbacSynced` map leak — the sweeper plus `CompareAndDelete` handles the Range race properly (round 3)
- `RBACPolicyKind` const, `builtinKinds()`, `builtinResourceKinds` and the AccessManagement doc (round 1)
- `childClientFactory` defaulted once in `SetupWithManager` (round 1)
- `deleteChildResources` log field, both `secret` and `key` (round 1)

The twelve new subtests cover the marker surviving each failure site, the sweeper race, the jittered deadline and the condition settling. Good coverage.

### Closed as documented decisions

Namespaced Role/RoleBinding, ServiceAccount subjects and the privilege-escalation question are now spelled out in the RBACPolicy godoc, so they show up in `kubectl explain`. That works for me as a way to close them — the escalation note especially.

### Not addressed

The RBACPolicy status subresource and printer columns. No `Status`, no `+kubebuilder:subresource:status`, no print columns, so the object still says nothing about itself: no `observedGeneration`, no record of whether a sync succeeded, no list of the ClusterDeployments using it. Worth confirming that's deliberately deferred rather than dropped, since it's also where the grant marker in the two comments below would naturally live.

### New

Three comments inline. Two of them come from putting the grant marker in the condition Reason rather than in a field of its own; neither is a regression on what's already merged.

eromanova:
@zerospiel All comments are addressed in the latest commit: the grant marker moved to a dedicated `ClusterDeployment.status.rbacPolicyGranted` written before `rbac.Sync`, which closes the leak window and frees the condition Reason to report real failures again; the sweeper now drops entries whose ClusterDeployment is gone instead of expiring them on age; and the `rolesChanged` comment, the `childClientFactory` note and the ordering-dependent prune test are fixed. The one thing not done is a status on RBACPolicy itself — a policy is shared by many ClusterDeployments, so "grants may be live" is a fact about a (policy, child cluster) pair rather than a property of the policy.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1262,67 +1334,147 @@ func (r *ClusterDeploymentReconciler) rbacSyncFresh(cd *kcmv1.ClusterDeployment,
...
+}
+
+// sweepRBACSynced drops every entry whose ClusterDeployment is gone, and every malformed one.
+// Keyed on the ClusterDeployment still existing rather than on the entry's age, so that a
+// ClusterDeployment which goes a long time without reconciling — a paused Region, say — keeps
+// its entry and doesn't re-read its child cluster for nothing on the way back. The list is served
+// by the manager's cache, so this costs no API calls.
+func (r *ClusterDeploymentReconciler) sweepRBACSynced(ctx context.Context) error {
+	cds := new(kcmv1.ClusterDeploymentList)
+	if err := r.MgmtClient.List(ctx, cds); err != nil {
```
zerospiel: The comment is right that a cache-served list costs no API calls, but controller-runtime deep-copies every item on a cache `List` by default. So this allocates and discards a full copy of every ClusterDeployment status — conditions, services, available upgrades — every 10 minutes, to build a `map[types.UID]struct{}`.

Nothing here mutates the result, so:

```go
if err := r.MgmtClient.List(ctx, cds, client.UnsafeDisableDeepCopy); err != nil {
```

Available in the pinned controller-runtime and not used anywhere in the repo yet, so worth a short comment saying the result is read-only.

internal/controller/clusterdeployment_controller.go
```diff
@@ -1226,27 +1259,66 @@ func (r *ClusterDeploymentReconciler) ensureRBACPolicy(ctx context.Context, scop
...
+func (r *ClusterDeploymentReconciler) markRBACGranted(ctx context.Context, cd *kcmv1.ClusterDeployment) error {
+	if cd.Status.RBACPolicyGranted {
+		return nil
 	}
-	r.rbacSynced.Store(cd.UID, rbacSyncState{syncedAt: time.Now(), uid: scope.rbacPolicy.UID, generation: scope.rbacPolicy.Generation})
 
-	return ctrl.Result{RequeueAfter: jitteredRBACResync()}, nil
+	patch := client.MergeFrom(cd.DeepCopy())
+	cd.Status.RBACPolicyGranted = true
+	if err := r.MgmtClient.Status().Patch(ctx, cd, patch); err != nil {
```
zerospiel, line 1302: This is the only failure path in `ensureRBACPolicy` that returns an error without touching `RBACPolicyReadyCondition`, so a persistent management-API problem here shows up as a reconcile error with the condition frozen at whatever it was:

```
err = failed to record the RBAC grant on ClusterDeployment test-ns/test-cd: Operation cannot be fulfilled...
NOTE: no RBACPolicyReadyCondition set on this failure path
```

The important half is right — it fails closed, nothing reaches the child cluster before the marker is durable. Just observability.

zerospiel (commented):
## Review of 9a54f942

Both points from the last round are fixed. I checked the leak window against the real package: the eager patch lands before `rbac.Sync` and writes only the one field, and a fresh reconciler revokes correctly after a simulated restart.

```
persisted after the eager patch: rbacPolicyGranted=true, conditions=0
FIXED: revoked after restart
```

`rbacFailureReason`/`rbacProgressingReason` are gone and the three sites report `FailedReason`/`ProgressingReason` again. `Prune`'s comment, the `childClientFactory` doc and the List-ordering dependency in `TestPruneReportsDeletesMadeBeforeAFailure` are all sorted.

Switching the sweeper from a TTL to "does the ClusterDeployment still exist" also fixes something I hadn't raised: under the TTL a CD in a paused Region would have lost its freshness entry and re-read its child cluster for nothing on the way back.

Build, vet, tests clean. Both CRDs and the deepcopy regenerate with no diff.

Two minor comments inline, neither a correctness problem.

zerospiel (approved):

Issue #3066, opened 2026-09-14
[bug] MCS with dependsOn services cannot be deleted


josef-hak:
## Summary

When a `MultiClusterService` whose services form a `dependsOn` chain is deleted, the
services are uninstalled without respecting the reverse of the install order. A service
that owns CRDs is uninstalled before the service whose release contains custom resources
of those CRDs. The dependent release then fails to uninstall with

```
ensure CRDs are installed first
```

which the addon-controller retries forever. The helm release stays `deployed`, the
`ServiceSet` stays at `Deleting`, and the `MultiClusterService` never loses its
`k0rdent.mirantis.com/multicluster-service` finalizer — so the object cannot be deleted
without editing the finalizer by hand.

Install honours the `dependsOn` order; uninstall does not appear to reverse it.

## Environment

| | |
|---|---|
| KCM | `main` @ `fc4201b4` (built from source) — **and** release `1.12.0-rc.3` |
| projectsveltos | `1.14.0` |
| Provider | `ksm-projectsveltos`, `selfManagement: true` |
| Cluster | single-node k0s in Docker, `docker.io/k0sproject/k0s:v1.36.3-k0s.0` |
| Charts | `oci://ghcr.io/k0rdent/catalog/charts` (public) |

Reproduced on freshly built clusters every time: twice locally on `1.12.0-rc.3`, and in
CI on both legs of the matrix — `main` @ `fc4201b4` and `1.12.0-rc.3` — while all ten
other scenarios in the same run passed on both.

## Reproduction

Install `ServiceTemplate`s for `cert-manager` 1.20.2, `kserve-crd` v0.18.0 and
`kserve-resources` v0.18.0 from `oci://ghcr.io/k0rdent/catalog/charts`, then apply:

```yaml
apiVersion: k0rdent.mirantis.com/v1beta1
kind: MultiClusterService
metadata:
  name: mcs-dependson-teardown
spec:
  serviceSpec:
    provider:
      name: ksm-projectsveltos
      selfManagement: true
    services:
      - template: cert-manager-1-20-2
        name: cert-manager
        namespace: cert-manager
        values: |
          cert-manager:
            crds:
              enabled: false
            fullnameOverride: scenario-cert-manager
      - template: kserve-crd-0-18-0
        name: kserve-crd
        namespace: kserve
        dependsOn:
          - name: cert-manager
            namespace: cert-manager
      - template: kserve-resources-0-18-0
        name: kserve-resources
        namespace: kserve
[...]

PR #3073, opened 2026-09-15
fix: remove dependent services in order to prevent deadlock during deletion

kylewuolle:
This PR deletes the dependent services in order to prevent a deadlock when deleting a MultiClusterService

#3066


josef-hak (commented):
Hi @kylewuolle , it doesn't seem to work:
- failing tests: https://github.com/josef-hak/k0rdent-ksm-tests/actions/runs/35063076539

I think I managed to fix it here: https://github.com/k0rdent/kcm/pull/3075 - could you review, please?
- passing all tests: https://github.com/josef-hak/k0rdent-ksm-tests/actions/runs/35090274789/job/104774725674

PR #3074, opened 2026-09-16
fix(am): skip distribution in the same and system namespaces

### LINKED ISSUE #3064: AccessManagement distributes into its own system namespace and can replace the source object with a managed copy it never cleans up
**Describe the bug**

`AccessManagementReconciler` distributes objects from `SystemNamespace` into every target namespace a rule resolves to, with no guard excluding `SystemNamespace` itself. When a rule's targets include the system namespace, the controller creates a "managed copy" of each source object back into the namespace it just read it from.

Normally that is a no-op — `createManagedObject` gets `AlreadyExists` from the source object and returns `created=false`. But the list and the create are not atomic. If the source object is deleted between `collectGroupKindResources`'s `List` and `processResourceRule`'s `Create`, the copy is created **under the source object's own name**. It is no longer a copy; it is the only object with that name.

Two things make that state permanent and lossy:

1. **It is never cleaned up.** `collectGroupKindResources` excludes managed objects in the system namespace from the `managed` list that drives cleanup (`internal/controller/accessmanagement_controller.go:388`):

   ```go
   if managedList.Items[i].GetNamespace() == r.SystemNamespace {
       continue
   }
   ```

   The orphan is invisible to `cleanupManagedResources` forever.

2. **It is not a faithful copy.** `createManagedObject` strips metadata wholesale (`internal/controller/accessmanagement_controller.go:482-485`):

   ```go
   target.SetOwnerReferences(nil)
   target.SetFinalizers(nil)
   target.SetAnnotations(nil)
   target.SetLabels(map[string]string{kcmv1.KCMManagedLabelKey: kcmv1.KCMManagedLabelValue})
   ```

   The object now holding the source name has lost the original's labels, annotations, owner references and finalizers.

Every subsequent attempt to recreate the source object then fails with `409 AlreadyExists`, indefinitely.

This is not limited to rules that explicitly name the system namespace. `getTargetNamespaces` returns `targetNamespaces.List` verbatim, and when neither a list nor a selector is set it lists **every** namespace — which includes `SystemNamespace`.

**To Reproduce**

1. Create a `ClusterTemplateChain` in the KCM system namespace.
2. Add an `AccessManagement` rule distributing it with `targetNamespaces.list: [kcm-system]` (or with no list and no selector, which resolves to all namespaces).
3. Delete and immediately recreate the chain under the same name. The chain `spec` is immutable (`self == oldSelf`), so delete + create is the only way to change it.
4. If the delete lands between the reconcile's `List` and `Create`, the recreate returns `409 AlreadyExists`. `kubectl get clustertemplatechain -n kcm-system <name> -o yaml` then shows an object labelled `k0rdent.mirantis.com/managed: "true"` occupying the name and carrying the pre-delete spec. Every later recreate keeps failing.

Observed on KCM 1.12.0-rc2. Controller log at the moment it happens:

```
{"level":"info","msg":"ClusterTemplateChain was successfully created",
 "controller":"accessmanagement","target namespace":"kcm-system","source name":"<chain-name>"}
```

**Expected behavior**

`processResourceRule` should skip `targetNamespace == r.SystemNamespace`. Distributing an object into the namespace it was read from has no purpose — the only two outcomes are a no-op or this corruption. Equivalently, `createManagedObject` could refuse when `targetNamespace == sourceObj.GetNamespace()`.

**Additional context**

Found from an integration test that deletes and recreates a `ClusterTemplateChain` to work around the immutable spec, and lost this race; the chain name was then unusable for the remainder of the run.

zerospiel:
Skips distribution and marking objects as managed in the system namespace.
Skips distribution into the same namespace from which the object had been read.

Fixes #3064


eromanova (approved):
LGTM. Thanks for the fix @zerospiel

Issue #1836, opened 2025-08-06
[enhancement] allow to generate multiple cluster profiles/profiles per cluster


aglarendil:
**Is your enhancement / feature request related to a problem? Please describe.**
When we deploy a cluster, we attach a bunch of things to it, although those things might require different sveltos management policies, for instance, we want to sync specific configmaps continuously, but we actually do not want to sync some of the services as they might be more fragile or we can adopt them into a different management tool or pipeline.
Additionally, sometimes we need **strict** order of how services are being deployed, so we need "dependsOn".

**Describe the solution you'd like**
This means that in terms of Sveltos we need to have them in different cluster profiles, i.e. when I specify services in the list, I should be able to attach them to a different clsuter profile than the default one being generated.

**Describe alternatives you've considered**
Supposedly, we can use MCS, but this makes not too much sense as we will have a handful of MultiClusterServices with a single cluster attached to them

**Additional context**

zerospiel:
@wahabmk please triage it, it is 100% ksm part, and prioritization and determining demand for the feature is up to you folks.

@aglarendil for the future, if you clearly understand an issue is related to `ksm`, please attach the corresponding label to it. Thanks

wahabmk:
> This means that in terms of Sveltos we need to have them in different cluster profiles, i.e. when I specify services in the list, I should be able to attach them to a different clsuter profile than the default one being generated.

A `ClusterDeployment` has been designed to represent a single cluster. So currently our API has also been designed in a way that `.spec.serviceSpec` is supposed to apply to that single cluster. It would break our current design if we target multiple clusters from a cluster deployment's spec. We have a larger backlog task to separate out KSM into it's own repo. This would likely require de-coupling service spec from `ClusterDeployment`, so this suggestion could be considered then.

As already mentioned in the description, we can currently use MCS to achieve this.

> Additionally, sometimes we need strict order of how services are being deployed, so we need "dependsOn".

This will be implemented as part of a significant API resign currently in progress: https://github.com/k0rdent/kcm/issues/1462.

PR #3075, opened 2026-09-16
fix(ksm): uninstall services in reverse dependency order

### LINKED ISSUE #3021: respect dependencies during services deletion
Deletion has to run in reverse dependency order for the same reason install runs in forward order: a resource must never be torn down while something that still needs it is also being torn down.

Concrete case — kserve (cert-manager → kserve-crd → kserve-resources):
- kserve-crd installs the CRDs (incl. ClusterStorageContainer).
- kserve-resources creates a ClusterStorageContainer CR of that CRD.

Because delete isn’t reversed, kserve-crd is removed before kserve-resources. The CRD is now gone, so Helm can’t uninstall kserve-resources: resource mapping not found ... no matches for kind "ClusterStorageContainer" → uninstall loops forever → release never deletes → MCS finalizer never clears → removal hangs. (Works on addon-controller v1.9.0, hangs on v1.12.0+.)

### LINKED ISSUE #3066: [bug] MCS with dependsOn services cannot be deleted

josef-hak:
Sveltos uninstalls the Helm charts of a Profile in the order they are listed, which is the order KCM installed them in. A dependsOn chain was therefore torn down front to back, helm refused to uninstall a dependent whose CRDs were already gone, and the aborted undeploy left the Profile and the ServiceSet finalizer stuck.

Rewrite the charts in reverse dependency order before deleting the Profile, and wait for the ClusterSummary to carry that order.

- fix #3066
- fix #3021

### Scope

This is a targeted fix for the teardown path: KCM collapses all services of one MCS
into a single sveltos Profile, so the only ordering signal sveltos has is the order of
`spec.helmCharts`, which it walks forward on both deploy and undeploy. Reversing that
list before the Profile is deleted is the smallest change that makes a `dependsOn` chain
uninstallable.

The structural answer is one Profile per service wired together with sveltos'
profile-level `spec.dependsOn`, which enforces the ordering in both directions on its own
(`areDependenciesDeployed` on deploy, `areDependentsRemoved` on delete). That is tracked
in #1836 and would remove the handshake added here entirely — it is out of scope for this
PR, which fixes the deadlock on the current model.


kylewuolle (dismissed):

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -519,6 +524,14 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
 	// to initiate the deletion of the Profile or wait for it to be deleted.
 	if err == nil {
 		if profile.GetDeletionTimestamp().IsZero() {
+			requeue, orderErr := r.ensureTeardownOrder(ctx, rgnClient, serviceSet, profile)
+			if orderErr != nil {
+				return ctrl.Result{}, orderErr
+			}
+			if requeue {
+				return ctrl.Result{RequeueAfter: teardownOrderRequeueInterval}, nil
```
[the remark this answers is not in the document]
josef-hak: `ServiceSetFinalizer` is added and persisted before any Profile/ClusterProfile is
created, and with the default background propagation GC deletes dependents only
after the owner is gone from etcd — i.e. after `reconcileDelete` has deleted the
Profile and cleared that finalizer. On self-management the dependent is a
cluster-scoped ClusterProfile owned by a namespaced ServiceSet, which GC never
deletes at all. Only an explicit foreground cascade would reach the handshake
with the Profile already terminating, and KCM never sets a `PropagationPolicy`.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +552,118 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	charts, ok := profileHelmCharts(profile)
+	if !ok || len(charts) < 2 {
+		return false, nil // a single release cannot be torn down out of order
+	}
+
+	if deletedAt := serviceSet.DeletionTimestamp; deletedAt != nil &&
+		r.timeFunc().Sub(deletedAt.Time) > teardownOrderTimeout {
+		l.Info("Giving up on ordering the teardown, deleting the Profile as is",
+			"profile", profile.GetName(), "waited", teardownOrderTimeout)
+		return false, nil
```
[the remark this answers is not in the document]
josef-hak: Good catch on the ordering of the check: the deadline was evaluated before the
reorder, so a Profile deleted past it went into undeploy in install order —
exactly the state the issue wedges on. The cap now ends only the waiting; the
charts are rewritten on every pass, and the "stalled handshake" test asserts the
stored Profile still carries the teardown order.

The cap itself stays deliberately: a ClusterSummary that never converges would
otherwise leave the ServiceSet finalizer stuck forever, which is the same
unrecoverable state from the user's point of view.

josef-hak:
@zerospiel I updated according to copilot comments. Can we merge now?
- it works for all ksm tests now: https://github.com/josef-hak/k0rdent-ksm-tests/actions/runs/35123491904

... while the main is failing for one of them: https://github.com/josef-hak/k0rdent-ksm-tests/actions/runs/34942459926

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +552,122 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	}
+
+	return false, nil
+}
+
+// orderHelmChartsByRank sorts charts by the rank of their release. Unranked
+// releases sink to the end: no known dependency constrains them.
+func orderHelmChartsByRank(charts []addoncontrollerv1beta1.HelmChart, rank map[client.ObjectKey]int) []addoncontrollerv1beta1.HelmChart {
+	rankOf := func(chart addoncontrollerv1beta1.HelmChart) int {
+		if i, ok := rank[serviceset.ServiceKey(chart.ReleaseNamespace, chart.ReleaseName)]; ok {
```
zerospiel: The chart key and the service key can disagree. `ServiceKey` maps an empty namespace to `default` (`internal/serviceset/util.go:1010`), but `helmChartFromSpecOrRef` sets `ReleaseNamespace` to `svc.Name` when the service namespace is empty (`serviceset_controller.go:1344-1349`). `ServiceWithValues.Namespace` is `MinLength=0` and documented as "Empty value means default namespace", so this is reachable.

Probe on this branch:

```
svcKey=default/kserve-resources  chartKey=kserve-resources/kserve-resources  equal=false
ordered=[cert-manager kserve-resources]
```

Every such chart lands in the `len(rank)` bucket, the stable sort keeps the install order, and the reorder becomes a no-op. Could `helmChartFromSpecOrRef` use `effectiveNamespace` for `ReleaseNamespace`, or `rank` be built with the same fallback?

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +552,122 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	charts, ok := profileHelmCharts(profile)
+	if !ok || len(charts) < 2 {
+		return false, nil // a single release cannot be torn down out of order
+	}
+
+	// The cap only ends the waiting, never the reordering: a Profile deleted past
+	// it still carries the order sveltos needs, which is the best a deletion that
+	// must not hang forever can do.
+	deletedAt := serviceSet.DeletionTimestamp
+	timedOut := deletedAt != nil && r.timeFunc().Sub(deletedAt.Time) > teardownOrderTimeout
```
zerospiel: `timedOut` is measured from the ServiceSet deletion, not from when the handshake started. If this is first reached more than 30s after the ServiceSet was marked for deletion (controller restart, regional-client errors, the status->Deleting pass plus 10s requeues), the first call writes the new order and then returns `!timedOut == false`, so `reconcileDelete` deletes the Profile in the same reconcile and sveltos never copies the order into the ClusterSummary.

Requeueing at least once after a successful write would avoid that, with the deadline stamped when the handshake starts rather than taken from `DeletionTimestamp`.

Minor: `deletedAt != nil` is dead - `Reconcile` only calls `reconcileDelete` when `DeletionTimestamp` is non-zero (line 144).

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +552,122 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	profile client.Object,
+) (requeue bool, _ error) {
+	l := ctrl.LoggerFrom(ctx)
+
+	charts, ok := profileHelmCharts(profile)
+	if !ok || len(charts) < 2 {
+		return false, nil // a single release cannot be torn down out of order
+	}
+
+	// The cap only ends the waiting, never the reordering: a Profile deleted past
```
zerospiel: This says a Profile deleted past the cap still carries the order sveltos needs, but the doc at line 558 says the undeploy walks the ClusterSummary. Writing only the Profile does not get it there.

With `syncMode: OneTime` it never will - `updateClusterSummary` returns early (addon-controller `controllers/profile_utils.go:468`), so the wait always burns the full 30s and the order never lands. Worth rewording, and possibly skipping the handshake in that mode.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +552,122 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+		}
+		l.V(1).Info("Waiting for the ClusterSummary to carry the teardown order",
+			"clusterSummary", client.ObjectKeyFromObject(summary))
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// orderHelmChartsByRank sorts charts by the rank of their release. Unranked
```
zerospiel, line 735: `rank` is built from all of `serviceSet.Spec.Services` (line 594), so a chart can only be unranked when its key fails to match its own service - the mismatch flagged below. Reads as a design choice but it is a symptom.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +552,122 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	}
+
+	ordered := serviceset.TeardownOrder(serviceSet.Spec.Services, dependencies)
+	rank := make(map[client.ObjectKey]int, len(ordered))
+	for i, svc := range ordered {
+		rank[serviceset.ServiceKey(svc.Namespace, svc.Name)] = i
+	}
+
+	orderedCharts := orderHelmChartsByRank(charts, rank)
+	if !sameReleaseOrder(charts, orderedCharts) {
```
zerospiel, line 663: This block reruns on every 2s poll: `TeardownOrder` clones the slice and allocates four maps, `orderHelmChartsByRank` clones `[]HelmChart`, and line 606 deep-copies the whole ServiceSet - up to ~15 times per deletion, almost always only to find nothing changed. A `slices.IsSortedFunc(charts, byRank)` check before cloning would cover the common case.

The `DeepCopy` is only needed because `getClusterSummaryForServiceSet` writes `serviceSet.Status.Deployed` (line 1139). Dropping that write looks cleaner than copying around it.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +552,122 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+		setProfileHelmCharts(profile, orderedCharts)
+		l.Info("Reordering Profile Helm charts for teardown", "profile", profile.GetName())
+		if err := rgnClient.Update(ctx, profile); err != nil {
```
zerospiel: Full `Update` of the object read at the top of `reconcileDelete`. A concurrent write turns into a conflict, an error-level log and a backoff requeue, inside a budget that is already 30s. `client.MergeFrom` patching `spec.helmCharts` only would be smaller and would not conflict.

internal/serviceset/util.go
```diff
@@ -418,6 +418,109 @@ func FilterServiceDependencies(
...
+			if depDepth := depthOf(dep) + 1; depDepth > d {
+				d = depDepth
+			}
+		}
+		delete(visiting, key)
+		depth[key] = d
+		return d
+	}
+	for key := range present {
+		if depthOf(key); cyclic {
```
zerospiel: A cycle silently returns the install order - the exact state #3066 wedges on - and nothing logs it. There is no webhook rejecting `dependsOn` cycles either, so it is reachable. A line from `ensureTeardownOrder` saying why the order was left alone would help whoever debugs this next.

Minor: `if depthOf(key); cyclic {` uses the init statement purely for a side effect; splitting it reads easier.

internal/serviceset/util.go
```diff
@@ -418,6 +418,109 @@ func FilterServiceDependencies(
...
+// ServiceSet.Spec.Services holds [kcmv1.ServiceWithValues], which carries no
+// DependsOn, so the dependency graph has to be read back from the owner. A
+// missing owner is not an error, only an absence of dependency information.
+func ResolveOwnerServices(ctx context.Context, c client.Client, serviceSet *kcmv1.ServiceSet) ([]kcmv1.Service, error) {
+	if serviceSet.Spec.MultiClusterService != "" {
+		mcs := new(kcmv1.MultiClusterService)
+		key := client.ObjectKey{Name: serviceSet.Spec.MultiClusterService}
+		if err := c.Get(ctx, key, mcs); err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil, nil
```
zerospiel: `(nil, nil)` here means "owner gone", and the caller reads it as "no dependencies" and tears down in install order. Probably safe given the owner finalizers, but it is silent - worth a log at the call site.

zerospiel (commented):
Read this against the sveltos source. The mechanism checks out: `uninstallHelmCharts` walks `clusterSummary.Spec.ClusterProfileSpec.HelmCharts` forward (addon-controller v1.14.0, `controllers/handlers_helm.go:966`), and reordering does not change the helm feature hash (`getSortedHelmCharts` sorts before hashing), so no redeploy is triggered by the rewrite. Comments inline - one of them looks like it can silently disable the fix.

**Open questions**

- Has the #3066 repro been run against this branch? v1.11.0 and v1.14.0 walk the charts forward identically (`handlers_helm.go:937` vs `:966`), so chart order does not explain the 1.11.1-good / 1.12.0-bad bisect in the issue, and the issue itself argues a sort is not the fix. The reorder looks like the right answer for #3021; less certain about #3066 without a repro run.
- Should the 30s cap be a field on the reconciler like `requeueInterval`, and should expiry emit an Event? It expires silently today.
- Should the handshake be skipped under `syncMode: OneTime`, where it can only ever time out?

**Residual risks**

- If the cap expires, behaviour returns to the pre-fix order with no record that the guarantee was dropped.
- `effectiveNamespace` and the `svc.Namespace`-or-`svc.Name` fallback are pre-existing with other callers, so fixing the key mismatch there needs a wider look.

internal/controller/adapters/sveltos/teardown_order_test.go
```diff
@@ -0,0 +1,305 @@
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	kcmv1 "github.com/K0rdent/kcm/api/v1beta1"
+)
+
+// Test_ensureTeardownOrder walks the handshake that keeps sveltos from
+// uninstalling a dependsOn chain front to back (#3066).
```
zerospiel, line 40: Every service here has a non-empty namespace equal to its own name, so the key mismatch flagged in the controller, a chart with no matching service, and the unranked bucket are all unexercised.

The timeout case at line 268 also only covers "past the cap on a later pass", not "past the cap on the first pass", which is the racy one.

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: Fixed in `a93cf689`. `orderHelmChartsByRank` returns the input untouched when it is already
sorted, so the steady-state poll allocates nothing, and `getClusterSummaryForServiceSet` no
longer writes `Status.Deployed` — that moved to the status collector that owns it, which
let the `DeepCopy` go.

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: Agreed, it was a symptom. With `ed30d1d5` the only way to land there is a chart with no
service at all, and there is now a test for that case.

internal/serviceset/util.go
josef-hak: Fixed in `a93cf689`. `TeardownOrder` returns a `cyclic` flag and `ensureTeardownOrder` logs
why the order was left alone. The init-statement `if` is split.

internal/serviceset/util.go
josef-hak: Fixed in `a93cf689`. `ResolveOwnerServices` returns `found`, and the call site logs that the
owner is gone before falling back to the listed order.

internal/controller/adapters/sveltos/teardown_order_test.go
josef-hak: Both covered in `ed30d1d5` and `2442f742`: a service with an empty namespace, a chart with
no matching service, and "past the cap on the first pass" — which now asserts the handshake
still reorders and waits rather than deleting.

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: Comment reworded in `a93cf689`: the Profile is only where the order is written, it counts
once sveltos has copied it into the ClusterSummary.

`OneTime` is the part I'd like you to decide. `updateClusterSummary` returns early there, so
the ClusterSummary never converges, the wait always burns the full cap and the order never
lands. Either I skip the handshake in that mode — honest, but the uninstall stays broken for
`OneTime` profiles — or KCM writes `orderedCharts` into
`ClusterSummary.Spec.ClusterProfileSpec.HelmCharts` as well, which makes it work and removes
the timing question, at the cost of writing to an object sveltos derives from our Profile.
Which do you prefer?

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: Right on both counts, fixed in `2442f742`. The deadline is stamped in an annotation on the
Profile when the wait starts, so a restart or a backlog does not spend it before the
handshake ever runs, and it survives the restart instead of resetting. `deletedAt != nil`
is gone — it was dead, as you say.

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: Right on both counts, fixed in `2442f742`. The deadline is stamped in an annotation on the
Profile when the wait starts, so a restart or a backlog does not spend it before the
handshake ever runs, and it survives the restart instead of resetting. `deletedAt != nil`
is gone — it was dead, as you say.

josef-hak:
@zerospiel Thanks — all of it is in, in three commits: `ed30d1d5` (the key mismatch), `2442f742` (the
timing of the wait and the deadline) and `a93cf689` (observability, the poll, the knobs).

On your open questions:

The cap is now a field on the reconciler next to `requeueInterval` (`a93cf689`). I did not
add an Event on expiry: there is no `EventRecorder` anywhere in `internal/controller`, so it
would be a new pattern to introduce in this PR — expiry logs at Info instead. Happy to add
the recorder if you want it.

`OneTime` — left as a question in the thread on line 571.

On the repro: the ksm e2e scenario `02dep01_valid` is the kserve chain from the issue and it
passes on this branch while it fails on main (runs linked above), with
`Reordering Profile Helm charts for teardown` in the controller log. But you are right that
this does not explain the 1.11.1-good / 1.12.0-bad bisect: both versions walk the charts
forward, so the ordering defect was always there and the version difference is something
else. I'd keep that as a separate investigation rather than block this fix on it.

internal/controller/adapters/sveltos/teardown_order_test.go
```diff
@@ -0,0 +1,399 @@
...
+			Status:     addoncontrollerv1beta1.Status{MatchingClusterRefs: []corev1.ObjectReference{clusterRef}},
+		}
+	}
+	summary := func(charts []addoncontrollerv1beta1.HelmChart) *addoncontrollerv1beta1.ClusterSummary {
+		return &addoncontrollerv1beta1.ClusterSummary{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: namespace,
+				Name: clusterops.GetClusterSummaryName(
+					addoncontrollerv1beta1.ProfileKind, "test-serviceset", clusterRef.Name, true),
+			},
```
zerospiel, line 121:
```suggestion
				Name: clusterops.GetClusterSummaryName(
					addoncontrollerv1beta1.ProfileKind, "test-serviceset", clusterRef.Name, true,
				),
			},
```

internal/controller/adapters/sveltos/teardown_order_test.go
```diff
@@ -0,0 +1,399 @@
...
+			Status:     addoncontrollerv1beta1.Status{MatchingClusterRefs: []corev1.ObjectReference{clusterRef}},
+		}
+	}
+	clusterSummary := func(charts []addoncontrollerv1beta1.HelmChart) *addoncontrollerv1beta1.ClusterSummary {
+		return &addoncontrollerv1beta1.ClusterSummary{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: namespace,
+				Name: clusterops.GetClusterSummaryName(
+					addoncontrollerv1beta1.ClusterProfileKind, "test-serviceset", clusterRef.Name, true),
+			},
```
zerospiel, line 150:
```suggestion
				Name: clusterops.GetClusterSummaryName(
					addoncontrollerv1beta1.ClusterProfileKind, "test-serviceset", clusterRef.Name, true,
				),
			},
```

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,13 +567,191 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	if !sameReleaseOrder(charts, orderedCharts) {
+		patch := client.MergeFrom(profileCopy(profile))
+		setProfileHelmCharts(profile, orderedCharts)
+		l.Info("Reordering Profile Helm charts for teardown", "profile", profile.GetName())
+		if err := rgnClient.Patch(ctx, profile, patch); err != nil {
+			return false, fmt.Errorf("failed to reorder Profile Helm charts for teardown: %w", err)
+		}
+		// Always requeue: deleting in the same pass would never give sveltos the
+		// chance to copy the new order into the ClusterSummary.
+		return true, nil
```
zerospiel: The deadline is only stamped and checked past the `sameReleaseOrder(charts, orderedCharts)` gate, so this branch has no cap. If the write never takes effect - a lagging regional cache, anything putting the order back - it patches and requeues forever. `return !timedOut, nil` used to bound it.

Probe with the Patch intercepted as a no-op and the Get still returning the install order:

```
still requeueing after 100 passes / 3m18s elapsed; 100 patches issued
```

That is a ServiceSet that never finishes deleting, plus a write every 2s. Stamping and checking the deadline before this branch, once the graph is known non-empty, would put both paths on one budget.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -65,6 +65,15 @@ const (
...
 	managementSveltosCluster = "mgmt"
+
+	// The teardown-order handshake holds the Profile deletion back, so it polls
+	// fast and gives up rather than wedging a ServiceSet sveltos never answers.
+	teardownOrderRequeueInterval = 2 * time.Second
+	defaultTeardownOrderTimeout  = 30 * time.Second
+
+	// Stamped on the Profile when the wait for the ClusterSummary begins, so the
+	// deadline survives a restart instead of starting over.
+	teardownOrderWaitAnnotation = "k0rdent.mirantis.com/teardown-order-waiting-since"
```
zerospiel: `ServiceSetPausedAnnotation`, `RegionPauseAnnotation` and `KCMRegionalTemplateAnnotation` are all in `api/v1beta1`. This one is written onto a cluster object too, so it would be easier to find there.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,13 +567,191 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *ServiceSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.Client = mgr.GetClient()
 	if r.timeFunc == nil {
 		r.timeFunc = time.Now
 	}
 	r.requeueInterval = 10 * time.Second
+	r.teardownOrderTimeout = defaultTeardownOrderTimeout
```
zerospiel: This makes the zero-fallback in `teardownOrderDeadline()` dead outside tests. Either the assignment or the getter can go.

internal/controller/adapters/sveltos/serviceset_controller.go
zerospiel: I'm not the right person to settle this one; it should go to the KSM team, since the answer depends on whether anyone is actually running `OneTime` with `dependsOn` chains, and I don't have that.

For what it's worth, my read leans to skipping the handshake under `OneTime`:

- It isn't a regression — every sync mode tears down in install order today, so skipping leaves `OneTime` where it already is while the default `Continuous` case gets fixed.
- Writing the `ClusterSummary` only sticks under `OneTime` anyway. Under `Continuous`, `updateClusterSummary` overwrites `Spec.ClusterProfileSpec` from the `Profile` whenever the two differ, so that write gets reverted. The coherent version is "write it only under `OneTime`", which is a special case rather than a simplification.
- KCM (KSM part) reading `ClusterSummary` is established; writing `Spec` on an object sveltos derives from our `Profile` is a different class of coupling, and it would break in a teardown path that is already awkward to debug.

If that's the way it goes, worth making it visible rather than silent: log the mode and that teardown order is not honoured, document it next to `syncMode`, and open a follow-up so the gap is tracked. Possibly upstream too — `updateClusterSummary` skipping OneTime is reasonable for a sync, less so when the Profile is being deleted.

But please take the decision to the KSM team rather than from me

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: You're right, and it is what `4bd7d497` broke: the stamp only went down past the
`sameReleaseOrder` gate, so a write that never takes effect had nothing bounding it.

Fixed in `678ffb5a`. The budget opens as soon as there is an order to enforce, and both
the write and the wait spend the same one, so your probe now ends at the cap. There is a
test for it: `Patch` intercepted and dropped, charts still reading back in install order.

The budget is in memory on the reconciler, like `loadOrCreateEnqueueState` in
`enqueue.go`. On the objects it would be as unreliable as the order itself — in your
probe an annotation would not have persisted either — and `DeletionTimestamp` is what you
objected to last round. Cost: a restart grants a fresh budget, i.e. one more idempotent
reorder.

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: Moot: the annotation is gone with `678ffb5a`, nothing of ours is written onto the sveltos
object any more.

internal/controller/adapters/sveltos/teardown_order_test.go
josef-hak: Applied in `15698eeb`.

internal/controller/adapters/sveltos/teardown_order_test.go
josef-hak: Applied in `15698eeb`.

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: Went with your read: `9a9a8fb2` skips the handshake under `syncMode: OneTime` and logs the
mode and that teardown order is not enforced. Your second argument decided it — a
ClusterSummary write only sticks under OneTime, so it would be a special case, not a
simplification.

@kylewuolle — ack? The gap: a `dependsOn` chain on `syncMode: OneTime`
keeps tearing down in install order, exactly as every mode does today; `Continuous`, the
default, gets the fix. If anyone runs OneTime with dependsOn chains, say so and we do it
differently. Follow-up issue either way, and probably one upstream too — the early return
in `updateClusterSummary` is reasonable for a sync, less so while the Profile is deleted.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +568,225 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	// patch and requeue forever.
+	key := client.ObjectKeyFromObject(serviceSet)
+	startedAt := r.teardownOrder.begin(key, r.timeFunc())
+	waited := r.timeFunc().Sub(startedAt)
+	timedOut := waited > r.teardownOrderDeadline()
+	defer func() {
+		if !requeue {
+			r.teardownOrder.end(key)
+		}
+	}()
```
zerospiel, line 660: `requeue` is false on every error return too, so any error - a transient ClusterSummary read, a failed patch - clears the budget and the next pass opens a fresh one. Persistent errors are loud and back off, so those are fine; intermittent ones mean the cap never accumulates.

Probe, erroring the ClusterSummary Get on every other pass with the ClusterSummary stuck in install order:

```
still holding the deletion after 2m0s, cap is 30s
```

It carries on indefinitely. Naming the error return and ending only on `err == nil` closes it.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +568,225 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	if !ok || len(charts) < 2 {
+		return false, nil // a single release cannot be torn down out of order
+	}
+
+	// Under OneTime sveltos stops copying the Profile into the ClusterSummary
+	// (updateClusterSummary returns early), so the order could never reach the
+	// undeploy and the handshake would only ever burn its budget. Not a
+	// regression - every sync mode tears down in install order today - but logged
+	// rather than left silent.
+	if mode := profileSyncMode(profile); mode == addoncontrollerv1beta1.SyncModeOneTime {
```
zerospiel: `LeavePolicies` has the same shape. `isLeavePolicies` (addon-controller `controllers/handlers_utils.go:1213`) skips the uninstall entirely when the ClusterSummary carries a deletion timestamp and `StopMatchingBehavior` is `LeavePolicies` - exactly this path - so the handshake spends its budget and a patch ordering an uninstall that never runs. Opt-in like `OneTime`: `applyProfileSpecDefaults` defaults to `WithdrawPolicies` (line 1870), `LeavePolicies` is reachable at line 1828. Reads like one more condition in this guard.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +568,225 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+
+func sameReleaseOrder(a, b []addoncontrollerv1beta1.HelmChart) bool {
+	return slices.EqualFunc(a, b, func(x, y addoncontrollerv1beta1.HelmChart) bool {
+		return x.ReleaseNamespace == y.ReleaseNamespace && x.ReleaseName == y.ReleaseName
+	})
+}
+
+func profileSyncMode(profile client.Object) addoncontrollerv1beta1.SyncMode {
+	switch p := profile.(type) {
+	case *addoncontrollerv1beta1.Profile:
```
zerospiel, line 770: Four type switches over the same two kinds now: this one, `setProfileHelmCharts`, `profileCopy`, `profileSyncMode`. A single `profileSpec(profile) (*addoncontrollerv1beta1.Spec, bool)` returning a pointer would cover charts read, charts write and sync mode.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +568,225 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+// teardownOrderBudget remembers when the handshake for a ServiceSet started.
+//
+// In memory on purpose: a deadline read back from the objects under handshake
+// would be as unreliable as the order itself, and taking it from the ServiceSet
+// deletion would spend it on everything that ran before the first pass. A restart
+// grants a fresh budget, which costs one more round of an idempotent reorder.
+type teardownOrderBudget struct {
+	startedAt map[client.ObjectKey]time.Time
+	sync.Mutex
+}
```
zerospiel, line 713: Embedding puts `Lock`/`Unlock` on the type's surface, and the mutex sits below what it guards. `mu sync.Mutex` above `startedAt` is the usual shape.

Separately: entries only leave via the defer in `ensureTeardownOrder` or the finalizer path, so a ServiceSet that stops being reconciled mid-handshake (finalizer stripped by hand, `getRegionalClient` failing before `reconcileDelete`) leaves one behind. Tiny, but there is no other eviction.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +568,225 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+		}
+	}
+
+	// One budget for the whole handshake, opened as soon as there is an order to
+	// enforce. Both the write and the wait spend it, so a write that never takes
+	// effect - a lagging regional cache, anything putting the order back - cannot
+	// patch and requeue forever.
+	key := client.ObjectKeyFromObject(serviceSet)
+	startedAt := r.teardownOrder.begin(key, r.timeFunc())
+	waited := r.timeFunc().Sub(startedAt)
```
zerospiel: `r.timeFunc()` twice; one `now :=` makes the first pass exactly zero.

internal/controller/adapters/sveltos/serviceset_controller.go
josef-hak: Agreed, `268ea612` guards it next to OneTime, with a test. One correction: under
`Continuous` the budget would not have burned — `updateClusterSummary` does carry the new
order over, so the handshake converged and merely did pointless work. The guard is still
right, the cost was a patch and a requeue rather than the full cap.

wahabmk:
@josef-hak If I understood this PR correctly, the service order is re-arranged on the Profile object according to their reversed dependency-order once the ServiceSet is being deleted.

Have you considered the other approach, in which while creating the Profile object, we make sure the services are added to the Profile object in their reversed dependency-order?

For example if we have an MCS deploying `C dependsOn B dependsOn A`, on the 1st run the ServiceSet's and Profile's spec will only have A. Once A has been Deployed, then B can be added to the ServiceSet and Profile object in reverse dependency-order. Similarly C can be added in reverse dependency order once B's status is Deployed.

This way, we probably won't have to create the teardown mechanism that this PR creates. That would enable us to:
1. Save processing time in the ServiceSet reconciler by not having to make Kube API calls to get the owner of the ServiceSet for the dependency-order. As we already have the owner object in memory while creating the ServiceSet, so we can have the service added to ServiceSet's spec in reverse dependency order, and then the ServiceSet controller can create the Profile with the same order as the services appear in the ServiceSet's spec.

3. Less code to maintain.

josef-hak:
@wahabmk no, I haven't considered that. BUT are you sure it would be less code? You would need to update Profile.spec multiple times (after each service deployment). In my approach it will only be updated once - before the ServiceSet removal.

wahabmk:
> @wahabmk no, I haven't considered that. BUT are you sure it would be less code? You would need to update Profile.spec multiple times (after each service deployment). In my approach it will only be updated once - before the ServiceSet removal.

The Profile.spec is already updated each time a new service is added. Taking the same example of `C dependsOn B dependsOn A`. This is the summary of how the flow currently is:
1. MCS controller updates ServiceSet to add A.
2. ServiceSet controller updates Profile to add A.
3. Once A is Deployed, the MCS controller updates ServiceSet to add B.
_(This happens bcz the poller setup by ServiceSet controller watches ClusterSummary for status change, once it detects change it triggers the ServiceSet controller, which updates the ServiceSet's status, which then in-turn triggers the MCS or CD controller)_
4. ServiceSet controller updates Profile to add B.
5. Once B is Deployed, the MCS controller updates ServiceSet to add C.
6. ServiceSet controller updates Profile to add C.

So we already update the ServiceSet and Profile object each time a new service is added.

One drawback of the approach I mentioned in https://github.com/k0rdent/kcm/pull/3075#issuecomment-5718190071 is that we want the MCS controller or CD controller to be completely agnostic of the KSM provider (Sveltos in this case). Therefore, it is not ideal to make the decision to add services in reverse-dependency order to the ServiceSet (as this decision would happen in the MCS and CD controllers).

To avoid that, we could keep the services order in the ServiceSet as it is (so that we don't have to make the reverse dependency-order decision in the MCS and CD controllers). Then, within the ServiceSet's controller we could fetch its owner and then make the decision to add the services in reverse-dependency order when creating the Profile object. This way we will incur the cost of 1 extra API call to fetch the owner (as is done in this PR). But we can avoid creating the teardown order mehcanism.

josef-hak:
> The Profile.spec is already updated each time a new service is added. Taking the same example of C dependsOn B dependsOn A. This is the summary of how the flow currently is:

@wahabmk Keeping the Profile permanently reversed would trade the teardown deadlock for an install
one. `walkChartsAndDeploy` walks the list forward and `deploySingleChart` aborts the pass on
the first failing chart (`ContinueOnError` is false by default). The incremental flow you
describe is safe — one new chart per wave, dependency already in the cluster. But the
ServiceSet keeps deployed services in its spec, so when several charts install from scratch
at once (child cluster rebuilt, releases deleted, ClusterSummary recreated), a reversed list
installs C before B, fails on the missing CRDs, aborts, and the next pass repeats it —
#3066 mirrored onto the install path.

So the order in the Profile has to match the deploy direction, and the flip belongs right
before the uninstall. It only happens when `dependsOn` is actually used — without an edge
the handshake returns immediately and makes no extra call.

Your point about keeping MCS/CD provider-agnostic is exactly why this PR puts the decision
in the ServiceSet controller. And the graph is only needed there while deleting; in your
variant it would be needed on every reconcile, so the owner lookup moves from deletion-only
to always. Cheapest either way would be carrying `DependsOn` in `ServiceSet.Spec.Services`.

Happy to spike your shape separately — the first thing to check there is the fresh
multi-chart install.

wahabmk:
> one new chart per wave, dependency already in the cluster.

The AI correctly identified that putting the services in reversed order would not affect installation because we add 1 service at a time anyway.

> But the ServiceSet keeps deployed services in its spec, so when several charts install from scratch
at once (child cluster rebuilt, releases deleted, ClusterSummary recreated), a reversed list
installs C before B, fails on the missing CRDs, aborts, and the next pass repeats it

We do not consider dependencies after the initial install. So if B depends on A and they are both deployed, and then A becomes unavailable, we do not block/uninstall B after it has already been deployed.

However, I do not have the entire picture of what unforeseen circumstances could occur that could break our control of deploying one service at a time. So perhaps, it's safer to reverse the dependency order only for the uninstall.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +568,227 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+		return false, nil // a single release cannot be torn down out of order
+	}
+	charts := spec.HelmCharts
+
+	// Under OneTime sveltos stops copying the Profile into the ClusterSummary
+	// (updateClusterSummary returns early), so the order could never reach the
+	// undeploy and the handshake would only ever burn its budget. Not a
+	// regression - every sync mode tears down in install order today - but logged
+	// rather than left silent.
+	if mode := spec.SyncMode; mode == addoncontrollerv1beta1.SyncModeOneTime {
```
wahabmk: internal/controller/adapters/sveltos/serviceset_controller.go:597 — paused Profiles are a third dead end the guards don't cover. `handlePauseAnnotations` (same file, line 1042) stamps `addoncontrollerv1beta1.ProfilePausedAnnotation` onto the Profile whenever the ServiceSet carries `ServiceSetPausedAnnotation`. Upstream `profile_controller.go:147` then bails out of reconciliation entirely ("profile is paused. Skip reconciliation"), so the reordered `spec.helmCharts` is never copied into the ClusterSummary — exactly the situation the OneTime and LeavePolicies branches were added to avoid. Concretely: delete a paused ServiceSet with a 3-service `dependsOn` chain and `reconcileDelete` patches the Profile, requeues at 2 s, and repeats for the full 30 s budget (~15 Profile patches) before giving up and deleting in install order anyway. Adding `annotations.HasPaused`-equivalent check next to the SyncMode/StopMatchingBehavior guards would short-circuit it.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +568,227 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	key := client.ObjectKeyFromObject(serviceSet)
+	now := r.timeFunc()
+	startedAt := r.teardownOrder.begin(key, now)
+	waited := now.Sub(startedAt)
+	timedOut := waited > r.teardownOrderDeadline()
+	// Errors keep the budget open: they end this pass without ending the
+	// handshake, and closing on them would restart the clock every time an
+	// intermittent one lands - the cap would never be reached and the deletion
+	// would be held for good.
+	defer func() {
```
wahabmk: internal/controller/adapters/sveltos/serviceset_controller.go:656 — the give-up path closes the budget, so a failing Profile delete re-arms a full 30 s hold. On the timeout branches (lines 670-674, 689-693) the function returns `(false, nil)`, so the deferred `r.teardownOrder.end(key)` fires. Control returns to `reconcileDelete:552`, and if `rgnClient.Delete(ctx, profile)` fails (transient API error, webhook rejection), the error propagates and the next pass calls `ensureTeardownOrder` again — `begin` finds no entry and stamps a fresh start time. With the ClusterSummary still not converging (the reason we gave up), the handshake re-blocks for another full 30 s before the delete is retried, indefinitely. Keeping the budget open until the Profile is actually gone (i.e. only ending it in the not-found branch at line 558) would make the cap mean what the comment says it means.

internal/controller/adapters/sveltos/serviceset_controller.go
```diff
@@ -539,6 +568,227 @@ func (r *ServiceSetReconciler) reconcileDelete(ctx context.Context, rgnClient cl
...
+	dependencies, ownerFound, err := serviceset.ResolveOwnerServices(ctx, r.Client, serviceSet)
+	if err != nil {
+		return false, fmt.Errorf("failed to resolve service dependencies for teardown: %w", err)
+	}
+	if !ownerFound {
+		l.Info("Owner of the ServiceSet is gone, tearing down in the order the services are listed")
+		return false, nil
+	}
+
+	if !slices.ContainsFunc(dependencies, func(s kcmv1.Service) bool { return len(s.DependsOn) > 0 }) {
```
wahabmk: internal/controller/adapters/sveltos/serviceset_controller.go:622 — the "no edges" fast path tests the owner's services, not this ServiceSet's, so a no-op handshake can still block deletion. `dependencies` is the owner's full list, but `TeardownOrder` drops every edge whose endpoints aren't in `serviceSet.Spec.Services`. If the MCS declares a `dependsOn` between two services that `FilterServiceDependencies` kept out of this ServiceSet, the guard passes, `TeardownOrder` returns the input order unchanged, and the code still falls through to the ClusterSummary comparison — where a ClusterSummary lagging for any unrelated reason (e.g. a service was removed from the MCS moments before it was deleted, so chart counts differ and `sameReleaseOrder` is false on length) holds the deletion for the full 30 s budget with no reordering to wait for. Checking `sameReleaseOrder(charts, orderedCharts) && ordered-equals-input` (i.e. nothing actually moved) before entering the wait would skip it.

wahabmk:
We might not need the teardown mechanism in this PR since https://github.com/projectsveltos/addon-controller/pull/1984 was merged. Although, we do still need to make some other changes after bumping to sveltos version which contains this change.

josef-hak:
@wahabmk ok, closing this PR. BTW any info about when the fix in sveltos will be available - new version of addon-controller release?

wahabmk:
I think the dependency order fix in https://github.com/k0rdent/kcm/pull/3081 is what we needed anyway after sveltos is upgraded.

PR #3081, opened 2026-09-22
fix(ksm): an unchanged service must not unlock what depends on it

josef-hak:

On upgrade, a service could be upgraded before the service it transitively depends on, whenever an intermediate service in the chain has no version change in that release.

`FilterServiceDependencies` lets a service satisfy its dependents when it is `Deployed`, `Status.Version == Spec.Version` and `Spec.Version == desired`. All three test the service on its own account and the result was never propagated down the chain, so an intermediate whose version does not change passes them all while the service it depends on is still mid-upgrade — a transparent node that unlocks everything behind it. The doc comment already promised "directly **or transitively**"; the code did not.

In `cfg → keycloak → bootstrap`, with `cfg` and `bootstrap` bumped and `keycloak` unchanged, `bootstrap` goes first. In the reported workload that deadlocked: `bootstrap`'s Helm hook waits for a Secret only the new `cfg` creates, so its `helm upgrade` never returns, the Helm feature stays held and `cfg` can never upgrade — with every layer above reporting healthy.

Three commits, so the behaviour change and the ordering can be judged separately:

- `test(ksm)`: the chain from the report, as two table cases. Fails on `main`.
- `fix(ksm): lock a dependent behind its whole ancestry`: `deployedServices` is reduced to a fixpoint — satisfying your dependents means your own dependencies are satisfied too.
- `fix(ksm): apply a batch in dependency order`: one reconcile's services are written as one list, and that list is the order sveltos applies the charts in; it was ordered by name. `sortByDependency` places a dependency first and leaves the name order to break ties, so the result stays stable across reconciles.

**Why the ordering is still needed**: the gate compares versions, but a service also carries its values, `valuesFrom`, helm options and action. A dependency whose values change while its version stays put is fully synced as far as the gate can see, so it unlocks its dependents in the very reconcile carrying its own new values, and both land in one batch — the last case of `Test_FilterServiceDependencies_VersionGate`, whose expectations are now the order the services come out in rather than just the set. Gating on deployed configuration instead would close that properly (`ServiceState.LastDeployedHash` already fingerprints values), but KSM has no matching hash for what it wants to deploy and `valuesFrom` changes without the spec moving, so that belongs in its own issue.

**Behaviour change to look at first**: three existing cases in `Test_FilterServiceDependencies` asserted that a deployed-but-locked `b` satisfies `c`. That is the same transparent node with a different cause — a failed or absent dependency rather than an unchanged version — and they now expect `c` to be locked with `b`. Nothing is removed from the cluster: `BuildServicesList` carries a locked service over at its stored version.

These are unit tests on `FilterServiceDependencies`. The path `ServiceSet → sveltos Profile → helm` is not covered here, and since a buggy and a fixed run end in the same state, an e2e scenario for it would have to assert ordering rather than final versions.


internal/serviceset/util.go
```diff
@@ -415,7 +435,55 @@ func FilterServiceDependencies(
 		return cmp.Compare(aKey.Name, bKey.Name)
 	})
```
wahabmk: The original intent of sorting by Namespace/Name was to make the list of filtered services stable so that services B,C,A are always sorted as A,B,C so that a un-necessary reconcile is not triggered on the ServiceSet just because the services are in a different order now in its spec.


So, keeping this in mind, do we need to sort by Namespace/Name if we are then sorting by dependency below?

internal/serviceset/util.go
```diff
@@ -415,7 +435,55 @@ func FilterServiceDependencies(
...
 
-	return filtered, nil
+	return sortByDependency(filtered), nil
```
wahabmk, line 440: **internal/serviceset/util.go:438 —** the new ordering gets discarded. `ResolveServicesToApply` feeds the sorted `filtered` into `ServicesToDeploy`, whose first loop appends every in-flight service before the second loop appends the rest in sorted order. A dependency can never be in-flight, but a dependent can — so the dependent lands first. Verified empirically: `a-leaf→z-root`, with `z-root` carrying new values and `a-leaf` Provisioning, yields `[a-leaf, z-root]`, which `getHelmCharts` passes straight into `Profile.Spec.HelmCharts`. That's the exact "dependency whose values change while its version stays put" case the PR cites. The unit test misses it because neither fixture service is in-flight.

internal/serviceset/util.go
```diff
@@ -415,7 +435,55 @@ func FilterServiceDependencies(
...
+}
+
+// sortByDependency puts a dependency ahead of its dependents within one batch.
+// Services eligible in the same reconcile are applied in the order they are
+// listed, so name order would upgrade a dependent first whenever it happens to
+// sort earlier.
+//
+// Input is expected sorted by key, which is what makes the result stable: of the
```
wahabmk: **internal/serviceset/util.go:446 —** comment doesn't match the code. It describes Kahn's with a name tie-break; the implementation is DFS post-order (different, still deterministic, permutation). The claim about out-of-batch edges is also wrong — every dependency of a batch member is in the batch.

internal/serviceset/util.go
```diff
@@ -415,7 +435,55 @@ func FilterServiceDependencies(
...
+// services free to go at each step, the first by name is taken. Edges to
+// services outside the batch are not constraints - those are either already
+// deployed or locked, and neither is applied here. A cycle would leave services
+// unplaced; they keep their name order rather than being dropped, and the
+// webhook is where a cycle is reported.
+func sortByDependency(services []kcmv1.Service) []kcmv1.Service {
```
wahabmk, line 456: Also I think this sort function (or any upgraded one) needs unit tests to test its correctness.

wahabmk:
@kylewuolle This PR and https://github.com/k0rdent/kcm/pull/3083 are fixing the same issue.

internal/serviceset/util.go
josef-hak: Yes, still needed. `filtered` is built by ranging over the `dependsOnCount` map, and `sortByDependency` only refines the order it is given — DFS walks the slice in order, so independent services keep their input order. It inherits determinism rather than producing it. Made that coupling explicit in the comment, and `Test_sortByDependency` now asserts the input order is kept.

internal/serviceset/util.go
josef-hak: You're right, confirmed with your exact case. The first loop appends in-flight services before the second walks the sorted batch, so the result is [in-flight in stored order] ++ [the rest in dependency order].

Fixed: the first loop now only records the in-flight entry, and the single pass over `filteredServices` emits everything. Added `Test_ServicesToDeploy_DependencyOrder` — your scenario, asserting the emitted order. It fails on the previous commit with exactly `[a-leaf, z-root]`.

internal/serviceset/util.go
josef-hak: Correct on both counts. Rewrote the comment to describe the depth-first post-order the code actually implements, and dropped the out-of-batch claim. Kept the `index` lookup guard — it costs nothing and doesn't rest on that reasoning.

internal/serviceset/util.go
josef-hak: Agreed. Added `Test_sortByDependency`: chain against name order, diamond, independent services keeping input order, edge outside the batch, cycle, single service.

